PR #3, opened 2015-11-15
Add ex mode

guillermooo:
@jpoon The `w` command is sort of working, so I think we can merge this.


jpoon:
After merging this, I'm not able to run the extension anymore? Still tracking down the error.

Issue #13, opened 2015-11-18
request "true" command/normal mode for VS Code


guillermooo:
Tracking this here. I suspect handling key strokes they way we're doing now doesn't cut it. There must be an explicit "command/normal" mode where key presses are not inserted as characters in the buffer.

Also, we need a block caret.

I don't think these things will block us for now, but I'm adding this issue to track this.

- [x] request command mode for VS Code
- [ ] request block caret

jpoon:
Duping this against #19

PR #16, opened 2015-11-19
Command/Insert Modes

jpoon:
Update: Ready to merge.

With this PR, implements basic key bindings in the various modes and ability to toggle between `esc` and insert keys

Supported Commands:
* Insert Mode - i, I, a, A, o, O
* Command Mode - h, j, k, l

The current mode is displayed at the bottom left status bar.

Somewhat addresses #13


guillermooo:
One thing to keep in mind if that Vim uses key names like `<CR>`, `<C-Home>`, etc. We may want to translate to Vim-speak on input and back to a sensible output when needed.

adriaanp:
I would suggest that we register the commands as "vsc_vim.a" instead of "extension.vimMode_a"

What do you guys think?

jpoon:
@adriaanp Absolutely. I'll do that.

PR #20, opened 2015-11-19
Ex mode

guillermooo:
@jpoon I think this can be merged.


jpoon:
Nice! One thing, squash commits please :)

guillermooo:
@jpoon When you say squash, you mean just one commit per PR? Sometimes I find it clearer to have separate commits if they are unrelated (not that I'm very disciplined with this, though). For example, above we have commits that change whitespace only. To me, it makes sense to keep those separate. But I'm just trying to figure out the style we have to follow. I'm ok either way.

jpoon:
I'd like to keep master branch as clean as possible. People (maybe just me) don't really care about how the sausage is made -- they just want the sausage. I don't need to see the various commits that were made to get a feature or bug fixed, I just want to know that this one particular commit did it all. In my mind, that means squashing commits down to a few, or one, discreet commit. In turn, this will make it easier to navigate through the commit history.

http://programmers.stackexchange.com/questions/263164/why-squash-git-commits-for-pull-requests

PR #26, opened 2015-11-21
Command base idea

adriaanp:
:bulb: This is an idea I have for separating the different operations.

What I would like to achieve is that to add a new vim feature should be as simple as mapping a command to an implementation class.

What do you guys think?


guillermooo:
Looks really tidy to me. I like it.

src/mode/mode_command.ts
```diff
@@ -18,22 +22,22 @@ export default class CommandMode extends Mode {
...
-                vscode.commands.executeCommand("cursorDown");
-                break;
-            case 'k':
-                vscode.commands.executeCommand("cursorUp");
-                break;
-            case 'l':
-                vscode.commands.executeCommand("cursorRight");
-                break;
+            default:
+                var map = this.motionMappings[key];
```
guillermooo: I think we need to implement one action soon to see how this fits in the bigger picture. For now it should be fine.

src/motion/cursor.ts
```diff
@@ -0,0 +1,33 @@
+import * as vscode from 'vscode';
+
+export default class Cursor {
```
guillermooo: I'd call this class `Caret`. What's `default` for? Couldn't find it in the TS spec.

src/motion/cursor.ts
```diff
@@ -0,0 +1,33 @@
...
+	editor: vscode.TextEditor;
+
+	constructor(editor: vscode.TextEditor) {
+		this.editor = editor;
+	}
+	
+	moveLeft(times = 1) {
+		for (var index = 0; index < times; index++) {
+			vscode.commands.executeCommand('cursorLeft');
```
guillermooo: This loop was super slow in Sublime. We may need to calc the x/y coords of the whole motion and move once. That mean we'd have to handle multiple carets our selves too, which isn't great.

src/motion/cursor.ts
jpoon: A nicety introduced in ES6 to make it easier to import modules --  http://www.2ality.com/2014/09/es6-modules-final.html

src/motion/cursor.ts
jpoon: +1

`mode_insert.ts` contains an example of how to calculate and set a new position based on line, char (roughly translates to x/y).

jpoon:
:+1:

adriaanp:
Added proof of concept how pending operations could work.

Like @guillermooo explained that we will need a way of switching between modes without user input.

Let me know what you think?

jpoon:
What are same examples of switching modes without user input?

guillermooo:
> What are same examples of switching modes without user input?

It's rather implicit switching---without the user requesting it. For example `:s/foo/bar/g` will leave you in normal mode regardless in which mode you started.

But I think operator pending is the main case. `daw` will switch from normal to operator pending even though `d` isn't really a mode switching command in the same sense as `esc` or `v`. I may be wrong, but I think in master we can't do that now.

jpoon:
Is there a way to execute `:/s/foo/bar/g` without first being in normal mode? For normal VIM, `daw` remains in normal mode? 

My initial thinking with the pending operations was that all key press history is kept in `keyHistory`. Then each subsequent key press can determine what to do.

Maybe I'm missing something :confounded:

guillermooo:
> Is there a way to execute  :/s/foo/bar/g  without first being in normal mode? For normal VIM,  daw  remains in normal mode? 

You could start in visual mode.

BTW, it's possible that I'm too focused on switching modes explicitly. It may not be necessary after all, as you say below.

> My initial thinking with the pending operations was that all key press history is kept in  keyHistory . Then each subsequent key press can determine what to do.

Yeah, it's possible that I'm missing the bigger picture. So from this perspective, each key doubles as a command name too? I think that could work.

> Maybe I'm missing something

It's very possible that we're trying to implement the same behavior in three different ways. :)

drewcassidy:
Couldnt you just make Visual mode just a special state of normal mode? most of the commands are still available between the two modes

guillermooo:
@adriaanp You may want to join the Slack thingie so we can better coordinate?

adriaanp:
> My initial thinking with the pending operations was that all key press history is kept in keyHistory. Then > each subsequent key press can determine what to do.

I do like this idea better than switching between modes.

This will allow to chain Operators and Motions together and then process the chain. The trick would be composing these together.

adriaanp:
Cleaned up the code, and add motions for w,b

rashwell:
Just a quick note related to this thread <c-o> is one that seems like a challenge as it shifts to normal mode from insert mode but only for the next command, and then end result is you are back in insert mode.  So the key history logic is complicated, (doable) but complicated, because the state of how you got into mode is important, for the mode outcome as well.

jpoon:

PR #30, opened 2015-11-23
Mappings

guillermooo:
:warning: **Not ready** :warning:

### DIFF
--- a/src/mode/mode_handler.ts
+++ b/src/mode/mode_handler.ts
@@ -6,10 +6,13 @@ import {Mode, ModeName} from './mode';
 import CommandMode from './mode_command';
 import InsertMode from './mode_insert';
 import VisualMode from './mode_visual';
+import {Mappings} from '../mapping/main';
+import * as util from '../util';
 
 export default class ModeHandler {
     private modes : Mode[];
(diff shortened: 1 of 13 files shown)


src/mode/mode_handler.ts
```diff
@@ -6,16 +6,25 @@ import {Mode, ModeName} from './mode';
...
 
     constructor() {
+        this.mappings = new Mappings();
+                
         this.modes = [
             new CommandMode(),
             new InsertMode(),
             new VisualMode(),
+            new CommandLineMode(this.mappings),
+            // new OperatorPendingMode();
```
adriaanp: I was putting some logic together for Operator commands, but have not thought about in this "mode" way. Was trying to do everything in command mode, this might be a better idea.

guillermooo: I think we need a way to allow transitioning from one mode to the next, as requested by the current mode (different to switching modes explicitly at user's request, I think). For example, after a command line action, we'll have to switch back to normal mode. Same for operator pending, I believe. I don't think this transitioning is possible with the current ModeHandler.

jpoon:

PR #32, opened 2015-11-25
Support `<<` and `>>`

jpoon:
Small PR that adds support for `>>` and `<<`.

### DIFF
--- a/src/mode/mode_normal.ts
+++ b/src/mode/mode_normal.ts
@@ -1,39 +1,50 @@
+import * as _ from 'lodash';
+import * as vscode from 'vscode';
 import {ModeName, Mode} from './mode';
 import {showCmdLine} from './../cmd_line/main';
-import * as vscode from 'vscode';
 
 export default class CommandMode extends Mode {
+    private keyHandler : { [key : string] : () => void; } = {};
+    
     constructor() {
         super(ModeName.Normal);
+        
+        this.keyHandler = {
+            ":" : () => { showCmdLine(); },   
+            "h" : () => { vscode.commands.executeCommand("cursorLeft"); },        
+            "j" : () => { vscode.commands.executeCommand("cursorDown"); },        
+            "k" : () => { vscode.commands.executeCommand("cursorUp"); },      
+            "l" : () => { vscode.commands.executeCommand("cursorRight"); },
+            ">>" : () => { vscode.commands.executeCommand("editor.action.indentLines"); },
+            "<<" : () => { vscode.commands.executeCommand("editor.action.outdentLines"); },
+            "dd" : () => { vscode.commands.executeCommand("editor.action.deleteLines"); }
+        };
     }
 
     ShouldBeActivated(key : string, currentMode : ModeName) : boolean {
-        return (key === 'esc');
+        return (key === 'esc' || key === 'ctrl+[');
     }
 
     HandleActivation(key : string) : void {
         // do nothing
-    }    
+    }
 
     HandleKeyEvent(key : string) : void {
-        this.keyHistory.push(key);
-
-        switch (key) {
-            case ':':
-                showCmdLine();
+        let keyHandled = false;
+        for (let window = 0; window <= this.keyHistory.length; window++) {
+            var keysPressed = key + _.takeRight(this.keyHistory, window).join('');
+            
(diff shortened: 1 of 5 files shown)


src/mode/mode_command.ts
```diff
@@ -17,23 +35,28 @@ export default class CommandMode extends Mode {
...
-            case 'l':
-                vscode.commands.executeCommand("cursorRight");
-                break;
+        var keyHandled = false;
+        for (var window = 1; window <= 2 && window <= this.keyHistory.length && keyHandled === false; window++) {
+            var keys : string  = _.takeRight(this.keyHistory, window).join('').toString();
+            
+            console.log(keys);
+            if (this.keyHandler[key] !== undefined) {
+                // Bug: 'two characters does not work'
```
jpoon: There's a bug right now where for some reason, I can't build a key with more than one chars.

src/mode/mode_command.ts
```diff
@@ -1,14 +1,27 @@
...
+            "k" : (position) => { vscode.commands.executeCommand("cursorUp"); },      
+            "l" : (position) => { vscode.commands.executeCommand("cursorRight"); },
+            ">>" : (position) => { vscode.commands.executeCommand("editor.action.indentLines"); },
+            "<<" : (position) => { vscode.commands.executeCommand("editor.action.outdentLines"); }, 
+        };
     }
 
     ShouldBeActivated(key : string, currentMode : ModeName) : boolean {
-        return (key === 'esc');
+        return (key === 'esc' || key == 'ctrl+[');
```
guillermooo: When we have mappings, we can make `ctrl+[` one and expand it to `esc` somewhere upstream. So `esc` is the only "switch to normal mode" command. Just an idea for the future, though.

jpoon: Absolutely! That is a much better way of doing that.

jpoon:
@guillermooo: All the comments/commits from before has been lost.

I hadn't realized you were working out of this branch as well. I just `git push --force` it :(. In any case, I fixed the `key` vs `keys` typo.

src/mode/mode_normal.ts
```diff
@@ -16,24 +29,24 @@ export default class CommandMode extends Mode {
...
-            case 'j':
-                vscode.commands.executeCommand("cursorDown");
-                break;
-            case 'k':
-                vscode.commands.executeCommand("cursorUp");
-                break;
-            case 'l':
-                vscode.commands.executeCommand("cursorRight");
+        var keyHandled = false;
+        for (var window = 0; window <= this.keyHistory.length; window++) {
```
guillermooo: Use `let` here because that's the intent?

src/mode/mode_normal.ts
```diff
@@ -16,24 +29,24 @@ export default class CommandMode extends Mode {
...
+            var keysPressed = key + _.takeRight(this.keyHistory, window).join('');
```
guillermooo: var ==> let ?

src/mode/mode_normal.ts
```diff
@@ -16,24 +29,24 @@ export default class CommandMode extends Mode {
...
+            
+            if (this.keyHandler[keysPressed] !== undefined) {
```
guillermooo: Would this be enough? `if (this.keyHandler[keysPressed]) ...`

src/mode/mode_normal.ts
```diff
@@ -1,14 +1,27 @@
+import * as _ from 'lodash';
+import * as vscode from 'vscode';
 import {ModeName, Mode} from './mode';
 import {showCmdLine} from './../cmd_line/main';
-import * as vscode from 'vscode';
 
 export default class CommandMode extends Mode {
+    private keyHandler : { [key : string] : (position : vscode.Position) => void; } = {};
```
guillermooo: I'm unsure we'll ever need to pass the `position` explicitly. If we need it, we can retrieve it from the current document. Also, why pass just one position when there may be multiple carets?

jpoon: True, I'll get rid of that.

src/mode/mode_normal.ts
jpoon: Yeah, habit. Should be adopting `let`.

PR #41, opened 2015-11-26
ensure user can dismiss global messages with esc

guillermooo:


guillermooo:
@jpoon PTAL Without this, users cannot dismiss global messages by pressing esc.

PR #44, opened 2015-11-28
'$' EoL Motion

josephliccini:
Hey guys

Wrote the $ motion, but I wasn't sure where to put logic for registering a command that isn't in VSCode Built in.

Also, I what I've written isn't particularly reusable for combining with another motion (i.e. 'd$', or 'y$')..

Feedback / thoughts on where to put the 'cursorEndOfLine' command?

Thanks
Joe

### DIFF
--- a/extension.ts
+++ b/extension.ts
@@ -98,7 +98,6 @@ export function activate(context: vscode.ExtensionContext) {
     
     vscode.commands.registerCommand('extension.vim_$', () => handleKeyEvent("$"));
     
-    registerCustomCommands();
     context.subscriptions.push(cmdLineDisposable);
 }
 
--- a/extension.ts
+++ b/extension.ts
@@ -105,23 +105,3 @@ export function activate(context: vscode.ExtensionContext) {
 function handleKeyEvent(key:string) {
     modeHandler.handleKeyEvent(key);
 }
-
-function registerCustomCommands() {
-    // TODO: Not sure where to put this..
-    // Will move cursor to EOL
-    vscode.commands.registerCommand("cursorEndOfLine", () => {
-        const cursorCharIndex = vscode.window.activeTextEditor.selection.active.character;
-        const lineNumber = vscode.window.activeTextEditor.selection.active.line;
-        
-        const lineLength = vscode.window.activeTextEditor.document.lineAt(lineNumber).text.length;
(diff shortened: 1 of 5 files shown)


jpoon:
Thanks @josephliccini!  As we don't really have a visual mode yet, would you mind adding this to a new `$` handler to `/src/mode/modeNormal.ts`? 

We need to come up with a better strategy of combining motions but current focus is to get a quick initial release out the door.

jpoon:
(Argh. Accidentally pressed the `x` button to delete my previous message).

Thanks @josephliccini! As we haven't put any effort in the visual mode yet, would you mind putting your `cursorEndOfLine` logic to the `$` delegate that is in `src/mode/modeNormal.ts`? 

We don't yet have a good strategy for combining with other motions; want to push out a small release soon!

extension.ts
```diff
@@ -96,9 +96,32 @@ export function activate(context: vscode.ExtensionContext) {
...
+    
+    registerCustomCommands();
     context.subscriptions.push(cmdLineDisposable);
 }
 
 function handleKeyEvent(key:string) {
     modeHandler.handleKeyEvent(key);
 }
+
+function registerCustomCommands() {
```
jpoon: Move this inside `"$": () => { <Here> }`? instead of `modeNormal.ts` instead?

extension.ts
```diff
@@ -96,9 +96,32 @@ export function activate(context: vscode.ExtensionContext) {
...
+        const lineLength = vscode.window.activeTextEditor.document.lineAt(lineNumber).text.length;
+        const charDelta = lineLength - cursorCharIndex;
+        
+        const position = new vscode.Position(lineNumber, cursorCharIndex);
+        const updatedPosition = position.translate(0, charDelta);
+        
+        const selection = new vscode.Selection(updatedPosition, updatedPosition);
+        
+        vscode.window.activeTextEditor.selection = selection;
+        vscode.window.activeTextEditor.revealRange(selection, vscode.TextEditorRevealType.Default);
```
jpoon: I haven't tested this out but would this be equivalent to one of the built-in VS commands: `cursorEnd` (e.g. `vscode.commands.executeCommand("cursorEnd")`?

josephliccini:
Came up with a way that matches what you guys have put together with the Cursor class coming in soon

PR #50, opened 2015-11-29
move cursor position after getting normal mode

kimitake:
After getting normal mode (e.g. from insert mode to normal mode), the cursor position should be moved to 1 left.


jpoon:
Thanks @kimitake!

PR #52, opened 2015-11-29
fix cursor behavior for h,j,k,l, and ESC

kimitake:
default cursor moving function such as vscode.commands.executeCommand("cursorLeft") is a bit different from original vim's behavior, so just added similar functions in TextEditor class and call them instead.


jpoon:
Manually merged with c018f165de2d6970b0ea3b96617561b5649d17e3

PR #61, opened 2015-11-30
map keys from US keyboard to other layouts

guillermooo:
Uses a `vim.keyboardLayout` setting to determine which keyboart layout to use; defaults to letting the key trhough as is.

### DIFF
--- a/extension.ts
+++ b/extension.ts
@@ -95,6 +95,8 @@ export function activate(context: vscode.ExtensionContext) {
     
     registerCommand(context, 'extension.vim_<', () => handleKeyEvent("<"));
     registerCommand(context, 'extension.vim_>', () => handleKeyEvent(">"));
+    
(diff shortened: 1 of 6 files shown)


jpoon:
:+1: x :100: 

How does one configure what keyboard to use?

guillermooo:
`"vim.keyboardLayout" = "es-ES"` in you user settings, but that code is a bad name for a keyboard, because it could be DVORAK, QWERTY or a different variant. I think we'll have to use other names.

In any case, I believe VSC cannot really forward all needed characters from non-US QWERTY keyboards. For example, I cannot capture `<` for Spanish Spain QWERTY because VSC believes it's `\`, and the actual US QWERTY `<` is `;` in Spanish Spain QWERTY.

TD;DR Not ready for merging. :)

guillermooo:
Basic functionality ready.

PR #63, opened 2015-11-30
Update README.md

markrendle:
Added step to run `tsd install` to Contributing/GettingStarted section. Without this the `gulp` step errors.


jpoon:
FYI. There's also a gulp command that does this.

PR #66, opened 2015-12-01
'$' and '^' for Moving to beginning and end of line

josephliccini:
This one works better than #44 that I had earlier since it uses the Cursor you guys have in there.

Let me know if any issues
Thanks,
Joe


jpoon:
FYI, you could have continued pushing commits and it would have updated the PR. No need to create a new PR.

PR #67, opened 2015-12-01
support repeat for h,j,k,l,x

kimitake:
support repeat for h,j,k,l,x

### DIFF
--- a/src/cursor.ts
+++ b/src/cursor.ts
@@ -13,66 +13,79 @@ export default class Cursor {
 		return vscode.window.activeTextEditor.selection.active;
 	}	
 	
-	static left() : vscode.Position {
+	static left(n = 1) : vscode.Position {
 		let pos = this.currentPosition();
 		let column = pos.character;
-		
-		if (column > 0) {
-			column--;
+
+		column -= n;
+		if (column < 0) {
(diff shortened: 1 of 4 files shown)


jpoon:
The current method in which we handle keys is not scalable. @adriaanp and I were discussing (via Slack) a better pattern for handling these kinds of commands. We were considering having different command types and then handling composition between the different types -- movement, editing, repeat.

kimitake:
I see. I fixed line end determination bug, too. So I will create a new PR.

kimitake:
Can I add '0' command, too? I just wanted it but I don't know why but I implemented repeat function...

jpoon:
First off, thanks soooo much for the PR. We might need to approach repeatable commands in a different manner considering that the same needs to be applied to both commands (e.g. deletes) and motions (e.g. cursor moves). 

With the recent [refactor](https://github.com/VSCodeVim/Vim/commit/d04b0d75aa6f7d5e51f56e9dfa3e073e22459c03) this is vastly out of date. As such, I'm going to close this PR. Please submit a new one against the latest `master` and/or file an issue regarding implementing repeatable commands such that we can discuss what the best way of approaching that may be.

PR #68, opened 2015-12-01
fix line end determination for a, A, $

kimitake:
fix line end determination for a, A, $

### DIFF
--- a/test/cursor.test.ts
+++ b/test/cursor.test.ts
@@ -1,7 +1,7 @@
 import * as assert from 'assert';
 import * as vscode from 'vscode';
 import TextEditor from './../src/textEditor';
-import Cursor from './../src/cursor';
+import Cursor from './../src/cursor/cursor';
 
 suite("cursor", () => {
 	let text: Array<string> = [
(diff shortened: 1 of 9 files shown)


jpoon:
I think it might be more proper to separate this into two classes. A cursor class (in which insert mode will use) and a caret class (in which command mode will use). 

As the same bug will appear for `lineEnd` and `documentEnd`

kimitake:
just created Caret class which is derived from Cursor class, and defined "maxLineLength" method for override between Caret and Cursor class.
And normal mode uses Caret class instead of Cursor.
How about this?

test/cursor.test.ts
```diff
@@ -56,13 +57,13 @@ suite("cursor", () => {
...
+		Caret.move(new vscode.Position(0, 7));
 
-		let current = Cursor.currentPosition();
+		let current = Caret.currentPosition();
 		assert.equal(current.line, 0);
 		assert.equal(current.character, 7);
 
-		let right = Cursor.right();
+		let right = Caret.right();
 		assert.equal(right.line, 0);
```
jpoon: Can we keep this test suite focused on the `Cursor` class? Keep using the Cursor functions but change the expected values on the asserts?

jpoon:
:+1: Just one note. After that, we can mergeeeeee!

kimitake:
Is this correct?

Issue #19, opened 2015-11-19
Block cursor in command mode


snowiow:
Hi, I'm interested in implementing this feature and studied the API. I didn't find anything and asked on the official issue tracker of [vscode](https://github.com/Microsoft/vscode/issues/200), but no answer so far. Does someone of you know, how I can get it done, to manipulate the cursor shape?

jpoon:
Related: https://github.com/VSCodeVim/Vim/issues/13

Thanks so much for helping look into this! I couldn't find anything in the vscode documentation either; but you could potentially *fake* it but selecting the current character to mimic the block cursor.

snowiow:
Ok, I was already afraid of something like that. After reading [their approach](https://code.visualstudio.com/docs/extensions/our-approach) it looked like it isn't allowed at all for the moment, because they are going a very conservative way of extension possibilities. Hope they allow more in the future though. So long this fake selection, seems to be the only possibility.

adriaanp:
Would the [DecorationRenderOptions](https://code.visualstudio.com/docs/extensionAPI/vscode-api#DecorationRenderOptions) with cursor property not style the cursor when setting a TextEditorDecorationType on the editor?

adriaanp:
Never mind after a quick test, that is the actual mouse cursor styling

drewcassidy:
Excuse my lack of experience with javascript and web-based stuff, but didnt someone say in #18 it was possible to render things directly to the canvas? would that work to draw the block cursor?

markrendle:
Looks like the best way to do this right now is to define a single-character range in the textEditor with the backgroundColor style set, and move it with the cursor.

jpoon:
Also just tried setting mimicking the block as a selection:

```
                let position = new vscode.Position(0, 0);
		let selection = new vscode.Selection(position, position.translate(0, 1));
		vscode.window.activeTextEditor.selection = selection ;
```

It sorta works but you still see the cursor, so....suboptimal -- css styling may be the way to go.

jpoon:
https://github.com/VSCodeVim/Vim/pull/102

PR #69, opened 2015-12-01
Fake a Block cursor when in Normal mode

markrendle:
Solution for #19 until the Code team provide a better way of styling the cursor.

Slightly hacky solution using textEditor.setDecorations, but it does give a clear visual indicator when you're in Normal mode.

At present you can't set decorations past the end of text on a line, so when the cursor is on an empty line it doesn't show the block.

### DIFF
--- a/extension.ts
+++ b/extension.ts
@@ -6,6 +6,7 @@ import {showCmdLine} from './src/cmd_line/main';
 import * as cc from './src/cmd_line/lexer';
 import ModeHandler from "./src/mode/modeHandler";
 import {ModeName} from "./src/mode/mode";
+import Cursor from "./src/cursor/cursor";
 
 var modeHandler : ModeHandler;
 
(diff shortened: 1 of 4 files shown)


jpoon:
Nice! How would this work in Visual mode?

markrendle:
@jpoon This doesn't do anything with Visual Mode, that's a whole other thing which should probably just work with setting actual selections. I don't mind having a go at Visual Mode if you like, seems like this SDK is pretty easy to work with.

markrendle:
Squashed. I can squash now! :grinning:

kimitake:
I opened long file and moved cursor by j or k, but window is not scrolled.
Up/Down allow keys work, mouse wheel and scroll bar are working, too.

In addition, the block area becomes big at tab space.

markrendle:
I'll take a look at the scrolling issue. Not sure how much can be done about the big tab space thing, since I've basically used a hack to simulate the block cursor. Question is, is the big cursor a bigger issue than having no visual indicator that you're in Normal mode?

jpoon:
@markrendle filed a separate issue for the scrolling issue https://github.com/VSCodeVim/Vim/issues/89 and I already have a fix in https://github.com/VSCodeVim/Vim/pull/87

jpoon:
@andersonj has resolved the conflicts and submitted a new PR. Closing this in favor of https://github.com/VSCodeVim/Vim/pull/102

PR #72, opened 2015-12-01
Add commands support for 'yy' and 'p'

liushuping:
Add commands support for 'yy' and 'p' and 'P'
 - 'yy': copy current line from cursor to the unnamed register
 - 'p': paster content from unnamed register to line below current cursor
 - 'P': paste content from unnamed register to line above current cursor

test file register.test.ts added.
The cursor.test.ts file is also updated to remove the failed code where the async code is not working in the test.

### DIFF
--- a/src/mode/modeNormal.ts
+++ b/src/mode/modeNormal.ts
@@ -2,7 +2,9 @@ import * as _ from 'lodash';
 import * as vscode from 'vscode';
 import {ModeName, Mode} from './mode';
 import {showCmdLine} from './../cmd_line/main';
+import {TextAction, Register} from './../register';
 import {Caret} from './../motion/motion';
+import TextEditor from "./../textEditor";
 
 export default class CommandMode extends Mode {
 	private keyHandler : { [key : string] : () => void; } = {};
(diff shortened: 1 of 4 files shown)


src/mode/modeNormal.ts
```diff
@@ -7,6 +7,7 @@ import TextEditor from './../textEditor';
 
 export default class CommandMode extends Mode {
 	private keyHandler : { [key : string] : () => void; } = {};
+	private registers: string[] = [];
```
jpoon: Registers are a global concept that exist outside of `CommandMode`.

src/textEditor.ts
```diff
@@ -34,5 +34,23 @@ export default class TextEditor {
...
+
+    	return register;
+    }
+    
+    static PasteBelow(text: string): Thenable<boolean> {
+    	let pos = vscode.window.activeTextEditor.selection.active;
+    	var insertPos = new vscode.Position(pos.line + 1, 0);
+    	
+    	return vscode.commands.executeCommand("editor.action.insertLineAfter").then(
+    		_ => TextEditor.Insert(text, insertPos)
```
jpoon: For fat arrow functions without any arguments, you can use `() => { my cool function }`

src/mode/modeNormal.ts
liushuping: @jpoon could you please propose a suggestion where to put the registers for copy and paste commands? Thanks!

src/textEditor.ts
liushuping: Thanks for pointing out this!

src/mode/modeNormal.ts
jpoon: Consider moving it out to it's own class?

jpoon:
Thanks @liushuping. I'm doing some clean-up and am going to close this PR.

Please reopen when:
1) conflicts with `master` fixed
2) fix the commit message to say something more useful :)

PR #76, opened 2015-12-01
implement correct w,b motion behaviour

adriaanp:
the vs code behaviour of moving words left and right does not follow the way vim does it.


jpoon:
Yay for tests! I'll merge this in.

PR #71, opened 2015-12-01
Add commands support for 'gg' and 'G'

liushuping:
Add navigation commands support:
 - 'gg': go to fist line, on first non-blank character
 - 'G': go to last line, on first non-blank character

PR #77, opened 2015-12-02
Implemented command mode, $, 0, ^, gg, G, plus inline search fx and Fx

rashwell:
Looking over the changes I realize I might have to watch the formatting in things like the package.json.  I don't understand how mine was formatted so differently.  But I am also very new to git so I am not really sure how to selectively do stuff.  But it was suggested to me to make this pull request so that the logic for 0 vs ^ could be in there.  To keep this to a single commit, the fx and Fx are in here as well.  I am hoping to chat with the major directors here though on the approach for thing like f and t commands, as you will see in my commit, I created sort of a duo key handler since it is different that two key commands.  But there are a lot of ways to approach this.  I have ideas, but right now this extension is so early, it might need a lot of refactoring, as we puzzle through this.

### DIFF
--- a/src/mode/modeNormal.ts
+++ b/src/mode/modeNormal.ts
@@ -6,6 +6,7 @@ import Cursor from './../cursor';
 
 export default class CommandMode extends Mode {
 	private keyHandler : { [key : string] : () => void; } = {};
+	private duoKeyHandler : { [key : string] : (string) => void; } = {};
 
 	constructor() {
 		super(ModeName.Normal);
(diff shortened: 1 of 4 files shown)


jpoon:
> .. package.json. I don't understand how mine was formatted so differently.

Looks like you may have added lodash as a dev dependency. I have found that after doing an `npm install --save/save-dev`, it re-formats the package.json. You can commit certain parts of a file using `git commit -i`

src/mode/modeNormal.ts
```diff
@@ -13,20 +14,30 @@ export default class CommandMode extends Mode {
...
 			"w" : () => { vscode.commands.executeCommand("cursorWordRight"); },
-			"b" : () => { vscode.commands.executeCommand("cursorWordLeft"); },
+			"b" : () => { vscode.commands.executeCommand("cursorWordLeft"); },			
 			">>" : () => { vscode.commands.executeCommand("editor.action.indentLines"); },
 			"<<" : () => { vscode.commands.executeCommand("editor.action.outdentLines"); },
 			"dd" : () => { vscode.commands.executeCommand("editor.action.deleteLines"); },
 			"dw" : () => { vscode.commands.executeCommand("deleteWordRight"); },
 			"db" : () => { vscode.commands.executeCommand("deleteWordLeft"); },
+			"gg" : () => { vscode.commands.executeCommand("cursorTop"); },
+			"G" : () => { vscode.commands.executeCommand("cursorBottom"); },
```
jpoon: This will conflict with https://github.com/VSCodeVim/Vim/pull/71

src/mode/modeNormal.ts
```diff
@@ -13,20 +14,30 @@ export default class CommandMode extends Mode {
 		this.keyHandler = {
 			":" : () => { showCmdLine(); },
 			"u" : () => { vscode.commands.executeCommand("undo"); },
-			"ctrl+r" : () => { vscode.commands.executeCommand("redo"); },
+			"ctrl\+r" : () => { vscode.commands.executeCommand("redo"); },
```
jpoon: This key is passed down through `extension.ts` and should remain `ctrl+r`

rashwell: Ugh sorry that was actually a left over, when I was experimenting with how to hand fx and Fx I at one point turned the whole table into an array of regex's instead.  The thought at the the time was then f. and F. could be matched, but I didn't like that you for each handle you would have to loop through the array of matches, rather than like now where it is a direct lookup via the associative array.

PR #85, opened 2015-12-02
reduce suggestion

kimitake:
When inputting, the suggestion window is displayed, but sometimes it is unnecessary.

### DIFF
--- a/extension.ts
+++ b/extension.ts
@@ -21,6 +21,7 @@ export function activate(context: vscode.ExtensionContext) {
     registerCommand(context, 'extension.vim_esc', () => handleKeyEvent("esc"));
     registerCommand(context, 'extension.vim_colon', () => handleKeyEvent(":"));
     registerCommand(context, 'extension.vim_space', () => handleKeyEvent("space"));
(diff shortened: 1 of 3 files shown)


jpoon:
This still doesn't fix the actual issue. I have filed -- https://github.com/Microsoft/vscode/issues/997 against vscode for this.

kimitake:
Yes, but much better than current one. At least, editor.action.triggerSuggest should be called in "then" method, otherwise any suggest window won't be displayed during inputting (it is displayed when hit backspace).

jpoon:
> At least, editor.action.triggerSuggest should be called in "then" method, otherwise any suggest window won't be displayed during inputting (it is displayed when hit backspace).

If you want to update the PR to only address this, I can merge that.

I am hesitant to take this PR as a whole as it's a hack that only serves as a band-aid fix. I'd like to see VSCode's response to https://github.com/Microsoft/vscode/issues/997 before we decide how to address this issue.

kimitake:
As you said, the best solution is the fix you request above, but if the fix is merged, we will see a lots of suggest window and can't write code easily using the vim plugin...so I also wait for the response Microsoft/vscode#997 and your decision.

PR #87, opened 2015-12-03
Refactor motion.

jpoon:
Create a base motion class in which cursor and caret extend. The difference between the`caret` vs `cursor` and why I thought it necessitated separate classes:

* Caret's can never have a position that at the very end of the string. Therefore, some of the logic differs when calculating the end of the line.
* Whenever we figure out styling of the caret block, we can place that in their respective class

Another reason for the refactor was to move towards a model of OOP rather than static functions. Additionally, motions are now composable. This makes repeatable actions super easy:

`new Cursor().left().left().left().left()`

or mix and match:

`new Cursor().left().lineEnd().down()`

The UI is only updated on the `move()` command => `new Cursor(0, 2).lineBegin().down().move()`


jpoon:
As this is a major refactor, let me know what you think @guillermooo or @adriaanp or @kimitake (as you had done the original work).

adriaanp:
:+1:  Love it!

kimitake:
I strongly agree with to instantiate the cursor/caret in mode class, actually I tried to do when I created caret class, but I needed to modify a lots, so I gave it up :)

btw, do you have a plan to add default argument as 1 for left, right, up, down etc?

jpoon:
> do you have a plan to add default argument as 1 for left, right, up, down etc?

No, you can achieve the same composing it => new cursor().right().right().right().

kimitake:
e.g. the user inputs "10h", so 

for (var i=0; i<10; i++) {
	this.caret.left();
}
this.caret.move();

Is this correct? how about 1000G etc?

jpoon:
Well, `10G` vs `100G` is the same. I think that logic should be somewhere else as I'm trying to follow *separation of concerns*.

benjaminRomano:
:+1:

PR #90, opened 2015-12-03
another POC to handling normal mode/keys

adriaanp:
:warning: Proof of Concept

This is kind of the idea I had from the start with some help from atom/vim-mode.

Basically as keys are pressed we push a motion or operator onto a stack, and then when ready the stack is processed.

Operators use the motion behaviour of selecting and performing its operation on that selection. This is quite nice because with an operator like d, any motion after words just works. For this POC, dl, dh, dj, dk, dw, db is implemented without duplicating calculations of selections, and any new operator will be able to work with all motions already in place.

This also allows for composing an operator and motion together for one key, like the x, X commands in the POC.

dd is also implemented here, its not obvious but the d operator with a "special Motion.select" will delete the current line.

By adding a Repeater operation on the stack we could implement the count for each command as well.

### DIFF
--- a/src/cursor.ts
+++ b/src/cursor.ts
@@ -176,7 +176,7 @@ export default class Cursor {
 
 	private static _nonWordCharacters = "/\\()\"':,.;<>~!@#$%^&*|+=[]{}`?-";
 
-	private static getNextWordPosition(): vscode.Position {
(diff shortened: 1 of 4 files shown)


jpoon:

PR #91, opened 2015-12-03
fix cursor position after entering command mode (again)

kimitake:
after merged Refactor motion. #87, this process was discarded.

### DIFF
--- a/src/mode/modeNormal.ts
+++ b/src/mode/modeNormal.ts
@@ -42,7 +42,7 @@ export default class CommandMode extends Mode {
 	}
 
 	HandleActivation(key : string) : void {
-		this.caret.reset();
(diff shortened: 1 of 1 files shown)


src/mode/modeNormal.ts
```diff
@@ -37,6 +37,7 @@ export default class CommandMode extends Mode {
 
 	ShouldBeActivated(key : string, currentMode : ModeName) : boolean {
 		if (key === 'esc' || key === 'ctrl+[') {
+			this.caret.reset().left().move();
```
jpoon: Ah! Thanks! Can you actually move this to `HandleActivation()`?

jpoon:
Thanks!

PR #94, opened 2015-12-05
Experimental key handling 2

guillermooo:
Revised experimental branch for handling keys. This POC prints command names to the debug panel; it does not run any commands. It only works in normal and insert modes. You can switch between the two using `i` or `esc` only.

The code is able to process all types of commands, like `w`, `dw`, `daw`, `dd`, `D`, `dfa`, `d/foo`.

screenshot 2015-12-05 00 55 20


jpoon:

PR #95, opened 2015-12-05
De keyboard layout

snowiow:
Started some mappings for German keyboards, but there are still some keys missing for {,},[ and ]. Because they are combinded trigger of alt gr + 7/8/9/0.


jpoon:
@guillermooo, mind taking a look as you've been working in this area?

Can you squash your commits?

snowiow:
I don't know what it is yet, but I will have a look later and squash them.

guillermooo:
@snowiow Try `git rebase -i <ref>` where \<ref\> is the parent ref to your first commit in this branch. There are other ways, but that's easy to start with.

snowiow:
Ok seems that they are squashed together now, but don't know why there is an empty merge commit

guillermooo:
Merged by 12ae0ba102cadc61661914a441984cb94f8c7df7

Issue #103, opened 2015-12-20
Implement `e`


pottereric:
I'd like to have the "e" motion implemented.I've submitted a pull request that imlements the feature.

PR #109, opened 2015-12-27
Fix character position persistence on up/down commands, add : "e", "0", and fix "^" 

corymickelson:
Note: The property this.prevColumn was renamed after finding that the property did not in fact store the previous position but the position after the new position was set. To make this a little more clear I renamed it to activeColumn. This was also the correct property to use to fix the up/down column position persistence problem.


jpoon:
Fixes https://github.com/VSCodeVim/Vim/issues/103.

Thanks!

PR #121, opened 2016-01-03
Small change, 0 is equivalent to ^

sharpoverride:
Only invoked the cursorHome command

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,14 +9,14 @@
     "color": "#a5c9a2",
     "theme": "light"
   },
-  "license" : "MIT",
+  "license": "MIT",
   "keywords": [
     "vim",
     "vi",
     "vscodevim",
     "vsc-vim"
   ],
- "repository": {
+  "repository": {
     "type": "git",
     "url": "https://github.com/VSCodeVim/Vim.git"
   },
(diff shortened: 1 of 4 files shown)


jpoon:
0 and ^ are different. According to vim `:help`: 

```
0			To the first character of the line.  |exclusive|
			motion.
^			To the first non-blank character of the line.
			|exclusive| motion.
```

As far as I can tell, the current implementation is correct.

sharpoverride:
To be honest I always used 0, because I wanted to get to the start of the line. When I started the change I had a version that did not support the motion. 

After I rebased latest I saw it was changed to what I expected it to be.

I've also commited multiple other changes since :P, I see it gets them as part of the same pull request. 
I'll change it back to your version that uses motion.

package.json
```diff
@@ -140,14 +429,14 @@
   "devDependencies": {
```
jpoon: I don't see any version revs?

src/motion/motion.ts
```diff
@@ -81,8 +81,10 @@ export class Motion implements vscode.Disposable {
     }
 
     public move(line : number = null, character : number = null) : Motion {
-        if (line !== null && character !== null) {
+        if (line !== null) {
             this._position = this.position.setLocation(line, character);
+        }
+        if (character !== null && character > this._desiredColumn) {
```
jpoon: I believe this check is still required as if you move left, you want to adjust the desiredColumn.

test/motion.test.ts
```diff
@@ -120,13 +120,12 @@ suite("motion", () => {
     suite("line up", () => {
         motionModes.forEach(o => {
             test("should move cursor one line up", () => {
+
                 let motion = new Motion(o).move(1, 0);
                 assert.equal(motion.position.line, 1);
-                assert.equal(motion.position.character, 0);
```
jpoon: Why removing the assertion? It's still an essential assertion that the 'up' move does not change the character/column.

package.json
```diff
@@ -36,88 +36,377 @@
...
-
-        { "key": "Shift+4", "command": "extension.vim_$", "when": "editorTextFocus" },
-        { "key": "Shift+6", "command": "extension.vim_^", "when": "editorTextFocus" },
-
-        { "key": "Ctrl+[", "command": "extension.vim_ctrl_[", "when": "editorTextFocus" },
-        { "key": "Ctrl+r", "command": "extension.vim_ctrl_r", "when": "editorTextFocus" },
-        
-        { "key": "Shift+,", "command": "extension.vim_<", "when": "editorTextFocus" },
-        { "key": "Shift+.", "command": "extension.vim_>", "when": "editorTextFocus" }
+      {
```
jpoon: I'd prefer the single line for the keybindings as it's easier to read.

jpoon:
In the future, if you are reverting a commit, use `git revert` as its cleaner. As per [Contributing.md](https://github.com/VSCodeVim/Vim/blob/master/CONTRIBUTING.md), please squash your commits.

package.json
sharpoverride: I think the fix whitespaces changed that I did not notice it

test/motion.test.ts
sharpoverride: The test only mentions the line. Plus the _desiredColumn value is cached and the test failed. It probably needs to have the _desiredColumn reset beforeEach

src/motion/motion.ts
sharpoverride: The thing is, that for me when I moved up/down it reset the _desiredColumn to 0 for empty lines.

When you move left or right by one character it already resets the _desiredColumn as far as I could tell

PR #122, opened 2016-01-05
Delete Right when user presses x

sharpoverride:
Small change to invoke deleteRight command when you press x.

BTW, are there any tests around the normal mode keys handling. I admit I'm still looking around and trying to add the small things that I feel productive with.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "displayName": "Vim",
   "description": "Vim emulation for Visual Studio Code",
   "icon": "images/icon.png",
-  "version": "0.0.8",
+  "version": "0.0.9",
   "publisher": "vscodevim",
   "galleryBanner": {
     "color": "#a5c9a2",
--- a/src/mode/modeNormal.ts
+++ b/src/mode/modeNormal.ts
@@ -26,7 +26,7 @@ export default class NormalMode extends Mode {
         "dd" : () => { return vscode.commands.executeCommand("editor.action.deleteLines"); },
         "dw" : () => { return vscode.commands.executeCommand("deleteWordRight"); },
         "db" : () => { return vscode.commands.executeCommand("deleteWordLeft"); },
-        // "x" : () => { this.CommandDelete(1); }
+        "x" : () => { return vscode.commands.executeCommand("deleteRight"); },
         "esc": () => { return vscode.commands.executeCommand("workbench.action.closeMessages"); }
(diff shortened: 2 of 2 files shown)


jpoon:
All are tests are here: https://github.com/VSCodeVim/Vim/tree/master/test. We currently don't have any that test normal mode commands as we are in the process of refactoring it all to better handle commands.

src/mode/modeNormal.ts
```diff
@@ -26,7 +26,7 @@ export default class NormalMode extends Mode {
         "dd" : () => { return vscode.commands.executeCommand("editor.action.deleteLines"); },
         "dw" : () => { return vscode.commands.executeCommand("deleteWordRight"); },
         "db" : () => { return vscode.commands.executeCommand("deleteWordLeft"); },
-        // "x" : () => { this.CommandDelete(1); }
+        "x" : () => { return vscode.commands.executeCommand("deleteRight"); },
```
jpoon: Tbh, this will likely get refactored in the future to better handle things like `4x` hence the CommandDelete implementation.

sharpoverride: Yeah, I expect that to happen, but I need to have small things like x, or cw available in the meantime. Also I'll probably try to add some dumb visual selection. I don't use all of what vim offers, mostly stuff I got used to.

I can deploy manually for myself if you think it's better to wait for the proper handling implementation.

package.json
```diff
@@ -3,7 +3,7 @@
   "displayName": "Vim",
   "description": "Vim emulation for Visual Studio Code",
   "icon": "images/icon.png",
-  "version": "0.0.8",
+  "version": "0.0.9",
```
jpoon: Can you revert this change? I'd like to rev the version on a difference schedule (ie. grouping several changes into one version).

src/mode/modeNormal.ts
jpoon: Or...you can help with the proper handling implementation ;)

package.json
jpoon: Actually this is fine, it so happens that I'm about to push out a release.

sharpoverride:
Ok, next time I'll make sure I don't change the version. thx

As per your comment with multi-motion I need to find some time to think about how to go about it. Or is there a document that has some details about the approach you or somebody else expects to have?

PR #133, opened 2016-01-30
Add paragraph motions

johnfn:
Add `{` and `}` motions, plus a couple of tests.

### DIFF
--- a/src/motion/position.ts
+++ b/src/motion/position.ts
@@ -48,6 +48,9 @@ export class Position extends vscode.Position {
         return this;
     }
 
+    /**
+     * Get the position of the line directly below the current line.
+     */
     public getDown(desiredColumn: number) : Position {
         if (this.getDocumentEnd().line !== this.line) {
             let nextLine = this.line + 1;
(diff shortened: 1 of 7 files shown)


src/motion/position.ts
```diff
@@ -156,10 +156,56 @@ export class Position extends vscode.Position {
...
+     * 2) the end of the document.
+     *
+     * If the current line is empty, go to the first nonempty line or end of document, then
+     * do the above.
+     */
+    public getCurrentParagraphEnd(): Position {
+      let pos = this.translate(0);
+
+      while (TextEditor.getLineAt(pos).text === "" && !TextEditor.isLastLine(pos)) {
+        pos = pos.translate(1);
```
jpoon: Perhaps you can reuse `pos.getRight()`?

jpoon: Perhaps you can re-use `getDown()` here?

src/motion/position.ts
```diff
@@ -156,10 +156,56 @@ export class Position extends vscode.Position {
...
+      }
+
+      for (pos; pos.line < TextEditor.getLineCount() - 1; pos = pos.translate(1)) {
```
jpoon: Similarly, `getRight()` here?

PR #134, opened 2016-02-03
Implement % to jump to matching brace

tma-isbx:
Implement % in normal mode to jump to matching brace.

Basically, binds % to "editor.action.jumpToBracket".

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "displayName": "Vim",
   "description": "Vim emulation for Visual Studio Code",
   "icon": "images/icon.png",
-  "version": "0.0.9",
+  "version": "0.0.10",
   "publisher": "vscodevim",
   "galleryBanner": {
     "color": "#a5c9a2",


jpoon:
Thanks for the PR!

From my testing, it seems like a partial implementation when relying on `jumpToBracket` as it only seems to jump to the matching brace when the brace is selected. 

* With VIM, `%`, it should jump to the next item in the line with a match. 
* and as we're relying on `jumpToBracket` no support C-style comments, C preprocessor matching

Judging from the caveats of this implementation, I'm OK with merging this PR unless any other contributors had objections?

tma-isbx:
Ah, yes. Sorry, I forgot to list the caveats.

Like all vim users using different subsets of functionality, this implementation met the subset that I used for %.

I was hoping this would be a good starter PR to get familiar with Extensions in general, and this extension in particular.

tma-isbx:
Oops. I was moving my master to track your master not realizing that would affect the pull request, and I guess 0 commits go merged into your master.

I can make a new pull request with the feature branch I had moved the commit to.

PR #137, opened 2016-02-08
Convert to ES6, Promises, async and await.

johnfn:
Converts the entire project to use ES6 Typescript, and converts all promise based code (that I could find) to use async/await.

This makes all of the Promise-based code we had before look *a lot* nicer. Compare e.g. test/mode/modeInsert.test.ts, which used to be an awful mess of `.then()`, with how nice it looks now! Since a good chunk of the VSC API is promise-based, I think this will help beautify a lot of our code in the future.

There is a drawback to this approach, which is that until https://github.com/Microsoft/vscode/issues/2778 gets resolved, I had to disable some of the ES6 features we were using that are currently hidden behind node flags. These are:

* Default parameters - `function foo(x: number = 6)`
* Rest arguments - `function foo(...x: number[])`
* The default export - `export default function foo()`

I don't see these things as a huge loss compared to the gain of async/await. The frustrating thing is that if you do e.g. use default arguments, you get a fairly obscure error message during runtime that, if you've never seen it before, could be pretty confusing. If the VSC team gets back to us quickly on https://github.com/Microsoft/vscode/issues/2778 it won't be a big issue, but if not we could catch it by wrapping import statements in a try catch and giving a more helpful error message.

---------

I also took the opportunity to make a few minor stylistic changes:

* For modes, we were naming functions in TitleCase, which was inconsistent with the rest of our naming conventions. I changed it to camelCase.
* Use `for ... of` in some places. `for ... of` is objectively better than `for (let i = 0; i < arr.length; i++) { let obj = arr[i]...` and subjectively better than `_.forEach` because it is prettier. :smile: 

Hopefully the above isn't contentious and I can sneak it in this PR :wink:


jpoon:
Yay! Thank you. I've been always meaning to do this! Although, looks to be a merge conflict, likely caused I just merged https://github.com/VSCodeVim/Vim/pull/138

johnfn:
@jpoon :+1: Just fixed up the merge conflicts, should be good to go!

jpoon:
@johnfn dammit. I should really merge this PR first. Conflicts again after merging your other one :(

johnfn:
@jpoon Totally unacceptable.

johnfn:
...But I fixed it up anyways. :+1:

PR #138, opened 2016-02-09
Fix word and back-word motions, and fix tests.

johnfn:
The word motion has a weird edge case where it doesn't go to the final character of a word on any line except the final line. Fixed it and fixed the tests as well.


jpoon:
Thanks!

Issue #127, opened 2016-01-07
`x` at end of line doesn't work


jpoon:

jpoon:
Fixed with https://github.com/VSCodeVim/Vim/issues/127

chevex:
Is this fix in the live plugin? I'm experiencing this issue but maybe it's just not released yet?

jpoon:
Looks like a regression with the latest release :(

jpoon:
Ok, pushed an update. This should be fixed in 0.0.15.

I noticed a couple of other regressions with the delete operators -- sorry -- I'll try a fix as soon as I can.

chevex:
Confirmed to be working as expected now. Thanks!

tunz:
I think it does not work in 0.0.16 version of OS X vscode. Is it fixed?

PR #141, opened 2016-02-11
[#127] Fix 'x' behavior at EOL 

tma-isbx:
Determine if we're at the EOL before we modify the text.

Also resolve() when we handle keys in Normal mode (probably a bug). This allows us to test keys.


src/mode/modeNormal.ts
```diff
@@ -64,7 +64,7 @@ export default class NormalMode extends Mode {
 
             if (keyHandled) {
                 this.keyHistory = [];
-                return this.keyHandler[keysPressed](this.Motion);
```
jpoon: I think this goes away with the async/await stuff that was just merged a couple of days ago.

jpoon:
Merge conflicts will need to be fixed before I can merge the PR

tma-isbx:
Resolved merge conflicts.

jpoon:
Fixes https://github.com/VSCodeVim/Vim/issues/127

PR #144, opened 2016-02-13
Visual Mode + Rudimentary Operators

johnfn:
First cut at Visual Mode!


johnfn:
I now think this is good enough to get reviewed and merged in. There's still lots of work to do, obviously, but this works more or less like visual mode does. Next I want to revisit what @guillermooo was looking at in #94 to make better infrastructure for operators.

jpoon:
Beautiful. I'll take a closer look tomorrow :)

src/mode/modeHandler.ts
```diff
@@ -22,23 +23,20 @@ export class ModeHandler implements vscode.Disposable {
...
-            return mode.isActive;
-        });
-
-        return currentMode;
+        return this._modes.find(mode => mode.isActive);
     }
 
     setCurrentModeByName(modeName : ModeName) {
-        this._modes.forEach(mode => {
+        for (let mode of this._modes) {
```
jpoon: Nit: I'd prefer `forEach` :)

src/mode/modeVisual.ts
```diff
@@ -0,0 +1,140 @@
...
+    constructor(motion: Motion, modeHandler: ModeHandler) {
+        super(ModeName.Visual, motion);
+
+        this._modeHandler = modeHandler;
+        this._keysToOperators = {
+            // TODO: use DeleteOperator.key()
+
+            // TODO: Don't pass in mode handler to DeleteOperators,
+            // simply allow the operators to say what mode they transition into.
+            'd': new DeleteOperator(modeHandler),
```
jpoon: :+1: Awesome, this is the pattern in which I was envisioning as well but never got around to it.

src/mode/modeVisual.ts
```diff
@@ -0,0 +1,140 @@
...
+            'x': new DeleteOperator(modeHandler),
+            'c': new ChangeOperator(modeHandler)
+        };
+    }
+
+    shouldBeActivated(key: string, currentMode: ModeName): boolean {
+        return key === "v";
```
jpoon: capital V?

src/mode/modeVisual.ts
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    async handleActivation(key: string): Promise<void> {
+        this._selectionStart = this.motion.position;
```
jpoon: This is great first cut. Would it be worthwhile to move selection to it's own class? That is what I had envisioned for the `motion` class.

src/motion/motion.ts
```diff
@@ -52,8 +56,10 @@ export class Motion implements vscode.Disposable {
             let selection = e.selections[0];
 
             if (selection) {
-                let line = selection.active.line;
-                let char = selection.active.character;
+                const whosFirst = selection.anchor.compareTo(selection.active);
```
jpoon: Whta does this do?

src/operator/delete.ts
```diff
@@ -0,0 +1,43 @@
+"use strict";
+
+import { Position } from './../motion/position';
```
jpoon: This would replace `/action/deleteAction.ts`, can we delete that?

jpoon:
Sickkkk! :+1: :heart: 

Dropped a couple of comments.

src/mode/modeHandler.ts
johnfn: Why? IMO, `for of` is much better than `forEach` - it reads way nicer, and you can do `for (const ... of)` which is not possible with `forEach`. (Are you sure you haven't confused it with `for in`?)

src/motion/motion.ts
johnfn: Um... actually, not totally sure what I was doing here o_O

Gonna revert this.

src/operator/delete.ts
johnfn: Yeah :+1:

src/mode/modeVisual.ts
johnfn: Hmm... Well, to start, I still don't fully grasp the concept of `Motion`. It seems to me like it's just a wrapper around Position to update VSCode, but I think those are actually 2 separate responsibilities that don't need to be in the same class. 

This seems like a good place to discuss some of my thoughts about structuring this code. My idea is to pass around a state object which could look like this (very rough)

```{
    cursorStart: Position(0, 5),
    cursorEnd: Position(0, 7),
    selecting: true,
    mode: Mode.Visual,
    textInEditor: "hello world" // to facilitate undo - maybe stored as diffs to save memory, idk?
}```

All motions would just take a state and return a new state object. When you pressed a key the object would be passed through some transformations and eventually you'd get a new one. When the process is done you would then update VSCode to show the new selections and cursor positions.

The idea is to make stuff like macros/number commands easier - if, say, the user goes into visual mode, makes 2w into a macro and does 100@@, we don't redraw the selection 200 times. 

This would also make proper undo (e.g., stepping back through full actions rather than what we have now) way more straightforward, since actions are nicely discretized. 

Anywho, I still have to dig more into the source of Vintageous and VSVim before I feel confident about making sweeping changes like this, but those are my general thoughts. What do you think?

(I also am a fan of separating out the motions into classes like VSVim and Vintageous do - otherwise the Motion and Position classes would eventually become monstrous!)

src/mode/modeVisual.ts
johnfn: And then later we turn on `--experimentalDecorators` and this list autopopulates and its magic and unicorns everywhere! :star: :fireworks: :dancers:

johnfn:
@jpoon I think I got it all. Except the `for of` thing, which I'll defend until I die :wink: Or until someone gives me a good reason to use it, hehe.

src/mode/modeHandler.ts
jpoon: :ok_woman: Not that strongly attached to it so let's just go with `for of`

src/mode/modeVisual.ts
jpoon: Yeah, I think that's what I had initially intended for `Motion`. Essentially what you described, some sort of state that would be passed around as this is super simliar to how VsVim does it (Vim for Visual Studio).

jpoon:
Weird, all the builds are failing. I'll take a look in a couple of days -- work's got me busy :(

src/mode/modeVisual.ts
johnfn: Ah... that makes more sense now!

src/mode/modeHandler.ts
johnfn: Haha. The closer I get to bike shedding, the more argumentative I get, or something. :rainbow:

PR #147, opened 2016-02-16
Capital W/B word movement

tma-isbx:
Implement W/B WORD movements
- update regex so as to treat “) {“ as two words instead of one
- update word movements to use indexes returned by regex directly
- fix b/B behavior when moving up a line (should move to beginning of last word/WORD)
- fix w/W behavior when moving down a line (should move to beginning of first word/WORD)


src/mode/modeNormal.ts
```diff
@@ -23,8 +23,10 @@ export class NormalMode extends Mode {
         "gg" : async (c) => { return c.firstLineNonBlankChar().move(); },
         "G" : async (c) => { return c.lastLineNonBlankChar().move(); },
         "w" : async (c) => { return c.wordRight().move(); },
+        "W" : async (c) => { return c.WORDRight().move(); },
```
jpoon: Not a big fan of differentiating functions using casing. I understand that vim differentiates using word vs. WORD, but I'm wondering if there are alternatives?

tma-isbx: Yeah, I'm not usually a fan either. Though I figured in this case it would be the most obvious vocabulary for someone familiar with vim.

I took a look at various other projects to see what they use, and I think I like "Big Word" the best. Still short, and the connection can be made to WORD pretty easily and recalled easily. I'll make an update to the pull request.

tma-isbx:
Updated.

src/motion/position.ts
```diff
@@ -11,22 +11,30 @@ export enum PositionOptions {
...
-        segments.push(`[\\s${_.escapeRegExp(Position.NonWordCharacters) }]+`);
-
         this.positionOptions = options;
-        this._nonWordCharRegex = new RegExp(segments.join("|"), "g");
+
+        this._nonWordCharRegex = this.makeWordRegex(Position.NonWordCharacters);
+        this._nonBigWordCharRegex = this.makeWordRegex(Position.NonBigWordCharacters);
+    }
+
+    private makeWordRegex(characterSet: string) : RegExp {
```
jpoon: Nit: mind moving the f(x) around such that public comes first then privates?

jpoon:
Awesome Thanks! Mind also updating the Readme?

I'll push an update to the extension following this PR merge. Sidenote: I should setup a continuous delivery system. Such that every time we push to a stable branch, it auto publishes it to VSCE. I'll file an issue to track this actually :)

jpoon:
Meh, we'll update the readme later -- want to push a new version out.

PR #154, opened 2016-02-18
Fix for block cursor in insert mode

### LINKED ISSUE #149: When you go into insert mode, the block cursor should disappear
Really trivial bug, just leaving it here so I (or someone else) remembers.

sWW26:
Fixes #149


johnfn:
Would you consider adding a test for this as well? I realize it is fairly trivial, but you can never have too many tests.

jpoon:
:+1:

sWW26:
There don't appear to be any tests around the cursor decoration and I can't seem to find a way to get a list of current decorations for a given a text editor that would allow me to test this.

PR #160, opened 2016-02-23
Add E (end of WORD), and fix up e (end of word).

tma-isbx:


johnfn:
Hey @tma-isbx! This is looking pretty good, but one thing that I noticed is that in a case like this:

```
word word word|

word
```

With | being the cursor, if I press `e`, the cursor doesn't make it to the next work and actually the extension crashes! I left you a comment where I think this happens. Could you fix that case and add a test for it? Thanks!

src/motion/position.ts
```diff
@@ -311,4 +295,39 @@ export class Position extends vscode.Position {
...
+            currentLine = TextEditor.getLineAt(workingPosition);
+            currentCharacter = 0;
+        }
+
+        let positions = [];
+
+        regex.lastIndex = 0;
+        while (true) {
+            let result = regex.exec(currentLine.text);
+            if (result === null) {
```
johnfn: Here we get into an infinite loop.

tma-isbx:
Fixed infinite loop occurring with w/b/e when trying to traverse empty lines.

Also made sure to match vim's behavior.

w/W stops on empty lines, but skips over whitespace only lines.
b/B stops on empty lines, but skips over whitespace only lines.
e/E skips over both empty lines and whitespace only lines.

johnfn:
Nice! Those are some bizarre edge cases...

You've got me wondering though. Your code seems a little complicated. The way I'd write this for word ends might look something like this (this is pseudocode, btw)

```
for (let line = currentLine(); line < maxLines; line++) {
    for (position of wordRegex.exec(currentLine.getText()).allPositions()) {
        if (position > currentPosition) { return position; }
    }
}

return endOfDocument();
```

It seems this could easily be generalized to all of wWeEbB. Is there any reason why such an approach wouldn't work here?

tma-isbx:
I've cleaned it up a little bit. It is fundamentally along the same lines as the psuedo code.

Fixed a bug in the regex that required the previous guards against empty lines (so was able to remove those). But still have the special case logic to stop on empty lines.

johnfn:
Hey @tma-isbx , check this out: https://github.com/VSCodeVim/Vim/commit/131bf4f5b0d227ba874f5f17e981292a563e4f99

I was able to clean up your code a little bit by incorporating the empty line case in the regex that you use for matching. I also tightened up the outer loop by making it loop directly over current line rather than keeping track of that separately. Finally, I decomposed out the bit of code that turns regexes into positions that match because you use that in a couple different places.

Could you finish off the work I've done (I only did it for the 'w' motion so far) and then commit that? Then I think we can merge!

tma-isbx:
Looks good. I think it could be directly applied to wordLeft, but I think the regex change breaks e/E which doesn't stop on empty lines. And another version of allPositions will be needed for e/E (though it may be possible to use it with just different regexes -- something based off of \S\s+, though there are probably some corner cases there).

I'll take a look.

johnfn:
Yeah, the way I'm imagining it we could just use separate regex rules for the different motions - or just build one off the other (since they are very similar)? What's wrong with allPositions in the e/E case?

tma-isbx:
Current implementation of e/E uses the same word/WORD regexes but uses match.position + match.length as the returned position to find the end of word/WORD.

tma-isbx:
Updated.

johnfn:
Ah, beautiful! :)

PR #164, opened 2016-02-26
Handle [count] on Normal Mode

frarees:
Allows combinations such as `6j` or `10dw`.

### DIFF
--- a/src/mode/modeNormal.ts
+++ b/src/mode/modeNormal.ts
@@ -64,10 +64,21 @@ export class NormalMode extends Mode {
 
         let keyHandled = false;
         let keysPressed: string;
+        let count = 1;
 
         for (let window = this.keyHistory.length; window > 0; window--) {
             keysPressed = _.takeRight(this.keyHistory, window).join('');
             if (this.keyHandler[keysPressed] !== undefined) {
+                if (this.keyHistory.length - 1 - window >= 0) {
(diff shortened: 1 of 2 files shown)


frarees:
Fail actually caused by #165, which I also PR'd.

johnfn:
This is actually a really difficult thing to implement right now. For instance, this doesn't really handle motions such as `d3w`, or (as you pointed out) any motion that uses any other mode e.g. `3ihelloESC` `2vw` etc. 

Essentially, the correct way to handle this would be to encapsulate some concept of an 'action' which can span many character presses and mode changes, rewrite all of the code to use actions rather than what it does right now, and then repeat that 'action' many times. 

It's quite an involved change, and I don't really blame you for not doing all that! :stuck_out_tongue:

frarees:
Yeah, motions are hard to handle properly. This is barely a quick implementation to "get something". As you state properly implementing this would require quite a rewrite of how actions are parsed and treated right now 😅

And undo actions get not grouped using this method, so e.g. `6u` would be needed to fully undo.

DanEEStar:
Is the concept of an 'action' on the roadmap for this project?
Is there a roadmap for this project?

jpoon:
I'm going to close this PR due to inactivity. As it stands, it's also wayyyy out of date with master. Feel free to re-open.

PR #166, opened 2016-02-29
WIP - Normal mode input handling

tma-isbx:
Some stuff that I've been working on. I think with a little bit more work, it can handle the majority of my typical normal mode use cases.

Supports
- [count][command][motion count][motion] - eg: c2w, 2dw, 2w
- [count][command][argument] - eg: 2fr
- [count][command][motion count][motion][argument] - eg: d3fe
- . - repeats the last command with motion count and motion
- text objects - in the sense that it recognizes them, but need to build the corresponding range makers for the different text objects (similar to what we have for motions).
- adding new motions and commands should be mostly straightforward.

TODO items
- ~~move mode switching out of modeHandler.handleKeyEvent(). Should probably be handled as a command/key in the individual modes. It currently intercepts i/a/etc and prevents you from doing "fa".~~
- ~~examine "off by one behavior" with "dfr", delete/change needs to be inclusive when used with t/f but not with other motions ("dw") ...~~
- ~~revisit decision to loop a motion count times versus passing the count to the motion. "2tr" is not the same as "tr" "tr".~~
- clean up
- a lot of unit tests


johnfn:
Ooh, interesting, I will have to dig into this later. But just from peeking at it now, don't you think that the handling code will need to be moved out of modeNormal? e.g. consider something like this:

`3i<c-w><c-w><esc>`

The whole `i<c-w><c-w>` motion will need to be repeated, but two of the commands were in insert mode.

tma-isbx:
Yeah, the key handling should be moved into the parent class. Visual mode would also need the key handling for motions (t/f).

tma-isbx:
I think everything pretty much works now. Needs some more clean up.

. (repeat) needs to be changed to support repeating Inserts.

johnfn:
Wow, this is awesome. Going to look into it more this weekend. :)

tma-isbx:
TODO investigate use of PositionOptions.CharacterWiseInclusive in a couple of places.

jpoon:
I'm going to close this PR due to inactivity. As it stands, it's also wayyyy out of date with master. Feel free to re-open.

PR #167, opened 2016-02-29
Fix infinite loop when workspace is empty

frarees:
Repro steps:

1. Open an empty document
2. Use `b`
3. Notice the plugin won't respond to events anymore

### DIFF
--- a/src/motion/position.ts
+++ b/src/motion/position.ts
@@ -254,12 +254,14 @@ export class Position extends vscode.Position {
         let positions = [];
 
         regex.lastIndex = 0;
-        while (true) {
-            let result = regex.exec(currentLine.text);
-            if (result === null) {
-                break;
+        if (currentLine.text) {
+            while (true) {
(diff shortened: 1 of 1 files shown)


frarees:
Found out that `w` also triggers this issue. Included that fix as well.

jpoon:
Can we add tests? And can you squash your commits?

tma-isbx:
This is also fixed in #160 and contains tests for these conditions.

frarees:
@tma-isbx checked out #160. Looking great!
@jpoon how do we deal with infinite loops in tests? Been playing with Mocha's `timeout` but I'm afraid we cannot gracefully catch an infinite loop happening on tests. Any ideas?

jpoon:
Maybe test the affirmative? Test that the wordLeft does what it should when the workspace is empty?

frarees:
#160 has been merged already so I'm closing this one as fixed.

PR #168, opened 2016-02-29
Fix visual mode activated on insert mode

frarees:


src/mode/modeVisual.ts
```diff
@@ -40,6 +40,9 @@ export class VisualMode extends Mode {
     }
 
     shouldBeActivated(key: string, currentMode: ModeName): boolean {
+        if (currentMode !== ModeName.Normal) {
+            return false;
+        }
         return key === "v";
```
jpoon: I think it's simpler if you do this:

`return key === "v" && currentMode === ModeName.Normal`

Can you also add in the `modeNormal.ts` that is you are in Visual Mode and `v` is entered, it should go to Normal Mode?

johnfn: Whoa, been using vim for ages and never knew that `v` in visual mode brings you out of visual mode!

frarees:
@jpoon yeah let me revise it. Later today I will push some changes.

src/mode/modeVisual.ts
jpoon: I'm in the same boat -- learning a lot of little things here and there building out this extension.

jpoon:
Thanks @frarees

PR #175, opened 2016-03-04
Wait for test files to get written

frarees:
Waiting asynchronously for the workspace test files to get written helps to get rid of those unexpected `cannot open file` failures on test runs.


jpoon:
Can we simply use `writeFileSync` (https://nodejs.org/api/fs.html#fs_fs_writefilesync_file_data_options)?

frarees:
Yes. I tried to maintain the async API, but in this case there's not too much we can benefit from. I will push some changes here.

PR #177, opened 2016-03-04
Commands can write to status bar

### LINKED ISSUE #156: How to turn off info bar on saving a file?
I save my code very frequently and info bar appears every time telling me that code is saved. It gets very annoying because I save almost ten times/mintue. Is there a way to turn it off on saving code?

frarees:
Fixes #156.

### DIFF
--- a/src/mode/modeHandler.ts
+++ b/src/mode/modeHandler.ts
@@ -49,7 +49,7 @@ export class ModeHandler implements vscode.Disposable {
         }
 
         const statusBarText = (this.currentMode.name === ModeName.Normal) ? '' : ModeName[modeName];
-        this.setupStatusBarItem(statusBarText.toUpperCase());
+        this.setupStatusBarItem(statusBarText ? `-- ${statusBarText.toUpperCase()} --` : '');
     }
 
     handleKeyEvent(key : string) : void {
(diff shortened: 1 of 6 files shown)


jpoon:
I was in the process of removing the whole message actually. Moving it towards the status bar displaying a short message.

frarees:
We could move the message to the status bar. 
`"<filename>" <lines>L, <chars>C written`. 
What do you think?

jpoon:
:+1:

src/mode/modeHandler.ts
```diff
@@ -81,12 +81,16 @@ export class ModeHandler implements vscode.Disposable {
         }
     }
 
-    private setupStatusBarItem(text : string) : void {
+    private setupModeStatusBarItem(text : string) : void {
```
jpoon: Can we consolidate these functions? `setupModeStatusBarItem` and `setupStatusBarItem`?

frarees: :+1:

frarees:
I've passed the mode handler down to the command. Will there be more than one instance of mode handler? Otherwise we could make it a singleton and that could simplify the API.

jpoon:
Nah, this is fine.

Issue #181, opened 2016-03-07
Can't highlight last character of a line.


khisakuni:
I am unable to highlight the last character of a line without having trailing spaces. This is the case both when trying to move to the last character, and trying to select the last character in visual mode.

bradphelan:
When I'm on the last character of a line and hit __x__ it doesn't delete the character. If I'm on the second last character and hit __x__ it does delete the character. I guess it's the same bug.

jpoon:
Closed with https://github.com/VSCodeVim/Vim/pull/182

KennethAshley:
I'm still having this issue.

PR #182, opened 2016-03-07
Fixup/highlight eol char

khisakuni:
Addresses issue of being unable to highlight the last character in a line.
https://github.com/VSCodeVim/Vim/issues/181

### DIFF
--- a/src/motion/motion.ts
+++ b/src/motion/motion.ts
@@ -114,7 +114,7 @@ export class Motion implements vscode.Disposable {
      * space at the provided position in a lighter color.
      */
     private highlightBlock(start: Position): void {
-        this.highlightRange(start, start.getRight());
+        this.highlightRange(start, new Position(start.line, start.character + 1, start.positionOptions));
     }
 
     /**


src/motion/motion.ts
```diff
@@ -75,7 +75,7 @@ export class Motion implements vscode.Disposable {
         switch (this._motionMode) {
             case MotionMode.Caret:
                 // Valid Positions for Caret: [0, eol)
-                this.position.positionOptions = PositionOptions.CharacterWiseExclusive;
+                this.position.positionOptions = PositionOptions.CharacterWiseInclusive;
```
jpoon: With VsCode, they give you indices on where you are in the sentence. For instance, the sentence, `abc` has valid positions of 0, 1, 2, 3. The intention of the character wise inclusive vs exclusive is to denote whether `3` is an appropriate index. In the case of caret, it is not as you should not be able to have a caret starting from index 3 hence it is an exclusive motion.

Maybe the better way of doing it would be below and maybe have it forgo that inclusive/exclusive check in the `getRight` call?

```
    /**
     * Allows us to simulate a block cursor by highlighting a 1 character
     * space at the provided position in a lighter color.
     */
    private highlightBlock(start: Position): void {
        this.highlightRange(start, start.getRight());
    }
```

khisakuni:
Thanks for the comments @jpoon. I made some changes based on your suggestions.

src/motion/motion.ts
```diff
@@ -114,7 +114,9 @@ export class Motion implements vscode.Disposable {
      * space at the provided position in a lighter color.
      */
     private highlightBlock(start: Position): void {
+        start.positionOptions = PositionOptions.CharacterWiseInclusive;
```
jpoon: This still feels hacky.

Can you change line 118 to `this.highlightRange(start, start.translate(0, 1));` and then that should fix the bug.

jpoon: Actually, that may not be viable as it returns a different type.

khisakuni:
@jpoon Updated per your recommendation. I rebased off of master so everything should be up to date. Not sure why CI is failing though.

jpoon:
Thanks :rocket: ! Yeah, I don't think the build failing is your fault.

PR #208, opened 2016-04-26
Support for copy and p command

petegleeson:
Hello, I have added two things:

1. Support for copying to the system clipboard
2. A basic version of the put command that pastes from the system clipboard

When first looking around I found [this issue](https://github.com/Microsoft/vscode/issues/217) on the vscode repository about supporting copy/paste operations for extensions. I have followed the advice in that issue and included the [copy-paste](https://www.npmjs.com/package/copy-paste) npm package to do  copy/paste.

Be great to hear what everyone thinks 👍

PR #211, opened 2016-04-30
Install XClip for Travis CI

edthedev:
This fixes the Travis build failure in #208.

Requests sudo and installs xclip.

Note that this switches the project to 'Standard Infrastructure'
rather than 'container based intrastructure'
http://stackoverflow.com/questions/26299552/travis-sudo-is-disabled

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 notifications:
   email: false
 
-sudo: false
+sudo: required
 
 language: node_js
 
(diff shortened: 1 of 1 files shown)


jpoon:
Thanks @edthedev. :rocket: I'd much prefer if we can merge this PR directly with https://github.com/VSCodeVim/Vim/pull/208 and then into master as the two changes are super related.

edthedev:
#208 includes a better solution to the travis.yml issue. @petegleeson uses addon, which allows continued use of the Travis container system.

#208 build is still currently failing, but it is on a later test in the process.

So this pull is not needed. This pull should be closed without merging.

PR #212, opened 2016-05-03
Trying to figure out why Travis will not approve #208

edthedev:
Sorry for the spam. Want to see if I create a new pull from the same branch and commit if Travis CI will show the test passing.

The diff here looks identical to #208, to me; but #208 shows as not passing in Travis CI. 
Yet when I wired Tavis CI to my fork and pushed a copy of the code in #208, the build passes.

Very odd.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,14 +3,18 @@ notifications:
 
 sudo: false
 
+addons:
+  apt:
+    packages:
+      - xclip
+
 language: node_js
 
 node_js:
   - "4.2"
-
 env:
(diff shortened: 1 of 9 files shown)


petegleeson:
Thanks for investigating this @edthedev. I agree, very strange behaviour with these CI builds. Seems ok now though.

edthedev:
Now that Travis CI passed here, I see that it has started passing in #208. Very odd. I will close this in favor of #208, since this pull seems to have served it's purpose: tricking the build on #208 to retry.

Issue #98, opened 2015-12-08
Add support for motion with jkl;


mloskot:
Many of Vim users prefer the ergonomic `jkl;` over the historic `hjkl` mapping..

Would it be possible to add support for the former too, please?

I also use [VsVim](https://github.com/jaredpar/VsVim) extension with Visual Studio, and it is easy to configure it there with vimrc (or the specific vsvimrc file).

jpoon:
I can think of two ways in which we can implement this.

1) config in the user settings (similar to how we do keyboard mapping) that denotes if it is "ergo mode", then bind the proper keys. Simple and easy, but the better way might be..
2) Support `map`, `nmap` etc

jpoon:
Keybindings can be overridden for a mode by supplying a {string: string} object defining what key or keys should perform what action when pressed.

Example:

{
    "vim.normalModeKeybindings": {
        "d": "DeleteChar",
        "D": "DeleteLastChar"
    },
    "vim.insertModeKeybindings": {
        "e": "InsertAtCursor",
        "E": "InsertAfterCursor"
    }
}

mloskot:
Thanks!

jpoon:
All I did was press the `merge` button, all thanks to @Lindenk! 👍

mloskot:
Hmm, I'm having no luck in binding `jkl;`. Is this mapping incorrect?

```
    "vim.normalModeKeyBindings": {
        "j":"MoveLeft",
        "k":"MoveDown",
        "l":"MoveUp",
        ";":"MoveRight",
        "0":"MoveLineBegin",
        "$":"MoveLineEnd"
    }
```

ant-fx:
I got this working with these settings:

```json
    "vim.normalModeKeyBindings": [
        {
            "before": [";"],
            "after": ["right"]
        },
        {
            "before": ["l"],
            "after": ["down"]
        },
        {
            "before": ["k"],
            "after": ["up"]
        },
        {
            "before": ["j"],
            "after": ["left"]
        }
    ],
    "vim.visualModeKeyBindings": [
        {
            "before": [";"],
            "after": ["right"]
        },
        {
            "before": ["l"],
            "after": ["down"]
        },
        {
            "before": ["k"],
            "after": ["up"]
        },
        {
            "before": ["j"],
            "after": ["left"]
        }
    ],
```

mloskot:
For the records and closure, here is the current way to achieve the modified motion mapping:

[...]

Issue #185, opened 2016-03-14
Remapping keys / vimrc support


AntJanus:
I understand that full vimrc support is impossible but I'd love to be able to remap some of the keys that I use often. For example, it's very common to use `,` or `;` instead of `:`.

patoupatou:
I love remapping jk to <esc>, for me I can't use VSCode until I can remap keys like vim.

BR

johnfn:
Done!!!

Issue #128, opened 2016-01-07
Keybinding discussion (non-US mostly)


AntonAderum:
Hi!

(Note: This became quite long. but the question have quite a lot of history in it so I had a hard time making it shorter...)

We have started using vscode as our developing tool at work, and since im a long time vim user I've been looking for a good vim-extension. As far as I can tell this is the best and most ambitious project.

However the keybindings for non-us keyboard are not optimal. I know this is mostly because of vscode using Electron (Atom has the same problem), and I know you have had some discussion about it already: #49 here discusses it and @guillermooo  even created an issue over at vscodes repository. 

There's also the issue over at vscode repo that discusses this at length: https://github.com/Microsoft/vscode/issues/713

The end result of that issue created the Keybinding widget (which is great), and also fixed key-bindings in general for non-us keybindings. 

When I tried using the Keybinding widget I noticed that the "<"-key was still not working on my (Swedish) keyboard. So I submitted an issue (https://github.com/Microsoft/vscode/issues/1794) and it got fixed to the master branch just the day after. I have built vscode from source and testet that it works. Great!

pressing the ```<``` key in using the Keybinding widget will result in this:
```
{
    "key": "oem_102",
    "command": "commandId",
    "when": "editorTextFocus"
}
```

Now I have noticed that you guys have the ```keyboard.ts``` file that tries to handle keybindings for different layouts, this also works fine - except for when there are keys that you havent declared in ```package.json``` since these wont be handled (and so never gets to be "translated"). 

So if i tried to create a swedish KeyMapper in ```keyboard.ts``` i would do this:
```
this.mappings = {
            'oem_102': '<',
            'shift_oem_102': '>'
        };
```
But that wouldnt execute because ```oem_102``` would never be handled and and sent to that translate function i just created.

This isnt just a problem with ```oem_102``` but really with every key that isnt handled in ```package.json```

Sooooo, I guess my question is: How do you want to handle translating keys that arent in ```package.json```? Should we simply add lines in ```package.json``` like this:
```
{ "key": "oem_102", "command": "extension.vim_oem_102", "when": "editorTextFocus" },
{ "key": "Shift+oem_102", "command": "extension.vim_shift_oem_102", "when": "editorTextFocus" }
```
And then do the mapping I did above (+ actually writing the command as well of course)? 
This seems to work fine (only way to test at the moment is to run from source), and it will allow me to indent using the actual ```<``` button on my keyboard.

AntonAderum:
[...]

PR #219, opened 2016-05-12
Added basic support for rebinding keys.

Lindenk:
Support is limited but nothing should be changed for anything not related to rebinding keys using user configuration. 

Partial fix for #185 and #128.


jpoon:
Closes https://github.com/VSCodeVim/Vim/issues/98

Issue #188, opened 2016-03-16
Switch to new VSCode APIs to better support input processing and autocompletion


tma-isbx:
See sample for usage of new APIs: https://github.com/alexandrudima/vscode-vim

jpoon:
I haven't looked into the new APIs, but I'm imagining the worst and that there's going to be a lot of work involved with upgrading to the new APIs.

tma-isbx:
I think it is work that will be ultimately necessary ;)

But if the work that was needed for amVim ( https://github.com/aioutecism/amVim-for-VSCode/pull/50/files ), it doesn't seem like it should be TOO bad.

Issue #157, opened 2016-02-22
After installing this extension, caps lock stops working.


segmentationfaulter:
I have installed VSCode (version 0.10.8) on Ubuntu 15.10 via umake, after installing this extension, caps lock is not honored in main code editing area. The caps lock starts working again after I uninstall this extension.

jpoon:
Doesn't seem to work on Windows either.

jpoon:
@segmentationfaulter Thanks for filing the issue. Seems like the bug lies within VSCode, issue filed with them to track.

Issue #126, opened 2016-01-06
suggest window is displayed despite not having any suggestions


jpoon:
Tracks: https://github.com/Microsoft/vscode/issues/997

PR #224, opened 2016-05-19
Update to vscode 0.10.12 APIs

jpoon:
. Requires VSCode ^1.0.0

With the new APIs, no longer need to bind every single letter in `package.json`. Fixes several issues with the keybindings including #157, #126, #188. 

For good measure, also updated outdated npm packages and typings.


jpoon:
The copy/paste tests continually fail on travis. They pass locally on my mac. Merging this PR regardless, looking into failures later.

PR #230, opened 2016-05-28
Rewrite Normal Mode tests to use the ModeHandler interface.

johnfn:
This is much nicer because now it's an end-to-end test of the Vim interface. Also, it's much easier to test that commands allow us to enter/leave different modes.


johnfn:
Woot, looks like I fixed the tests intermittently failing, as well! 😀

PR #231, opened 2016-05-29
Rewrite normal mode tests.

johnfn:
Continuing from https://github.com/VSCodeVim/Vim/pull/230

This actually caught a bug with how we were doing tests before. The `i` in `ciw` when running the extension forces you to go into insert mode, but the test incorrectly passed anyways. Commented out those tests for now.

### DIFF
--- a/extension.ts
+++ b/extension.ts
@@ -32,7 +32,7 @@ export function activate(context: vscode.ExtensionContext) {
         }
     });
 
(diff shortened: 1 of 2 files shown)


johnfn:
Ugh, looks like I was wrong about getting those tests to pass. So annoying. 😞  

Works on my machine! 🚢

PR #233, opened 2016-05-29
Don't use ctrl-c to leave insert mode by default.

johnfn:
I'm a solipsistic OSX user, my bad!


johnfn:
The only reason these tests aren't passing is because I made the other PR first.

PR #234, opened 2016-05-29
Refactor commands [WIP]

johnfn:
Stuff I did:

* Add a class for every action
* Make 3 base types of actions: movements (e.g. `h`), operators (e.g. `d`) and commands (e.g. `i`). There may be more? I'm not sure. 
* Completely rewrite the core processing function. For example (it can't do all this yet, but the infrastructure is there) if you type `5"pdw`, it'll gather: 
* * That it's going to happen 5 times
* * It's going into the `p` register
* * It's a delete operation
* * Then it sees the `w` and does the whole thing. 
* Handle commands like `dw`, `y}` - commands that spell out an operation and a movement to do it over -  in a unified manner rather than special casing all of them.
* Instead of special casing commands that make you switch between modes, they're now just command actions and they work like anything else.
* Remove CharacterWiseInclusive/Exclusive. We instead handle that stuff in the mode-specific processing in modeHandler.
* Make `aw` and `iw` movements work as arguments to `d`, `y` and `c`.


src/actions/actions.ts
```diff
@@ -0,0 +1,568 @@
...
+   * Every Vim action will be added here with the @RegisterAction decorator.
+   */
+  public static allActions: BaseAction[] = [];
+}
+
+function RegisterAction(action) {
+    Actions.allActions.push(new action());
+}
+
+@RegisterAction
```
jpoon: Oohh! neat, didn't know TS/JS had this.

johnfn: Yeah, it's really snazzy. Unfortunately I'm running into some annoying bugs where depending on compile order things might not work, so I might have to remove it...

johnfn:
Bah! The tests totally pass. I don't understand why they're timing out on the server. 

I'LL JUST DO IT LIVE!!!

PR #238, opened 2016-06-05
Cleanup cursor drawing and remove Motion class

johnfn:
I took out the special bit of functionality that Motion had to draw the block cursor etc., and then since the rest of Motion was pretty much just a wrapper around Position, I was able to get rid of it. Woo!


johnfn:
Whoa, the tests started passing again. I blame Travis.

Issue #132, opened 2016-01-27
Use VSCode native block cursor setting


markrendle:
Adding this as a place-holder because there should be a new release of VSCode this weekend which provides a block cursor style out of the box.

jpoon:
:dancers:

jpoon:
I saw that the PR got merged but how does one use it and toggle between the different cursor styles?

tma-isbx:
This sample project has the details: https://github.com/alexandrudima/vscode-vim

PR #245, opened 2016-06-06
Use vscode built in support for block cursors

Paxxi:
Added some simple tests for this and moved it out of the hot path as it's something that only happens during mode changes.

### DIFF
--- a/test/mode/modeHandler.test.ts
+++ b/test/mode/modeHandler.test.ts
@@ -1,6 +1,7 @@
 "use strict";
 
 import * as assert from 'assert';
+import * as vscode from 'vscode';
 import {setupWorkspace, cleanUpWorkspace} from './../testUtils';
 import {ModeName} from '../../src/mode/mode';
 import {ModeHandler} from '../../src/mode/modeHandler';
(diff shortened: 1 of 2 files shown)


jpoon:
@Paxxi sweet! there's a minor tslint error that's making the build fail

> [16:16:39] [gulp-tslint] error mode/modeHandler.ts[195, 13]: unused variable: '_caretDecoration'

Paxxi:
Ahh right, forgot to run tslint, will fix tomorrow

Paxxi:
Fixed the tslint error and rewrote it a bit to make it a bit shorter

test/mode/modeHandler.test.ts
```diff
@@ -32,4 +33,14 @@ suite("Mode Handler", () => {
...
+
+    test("Uses correct cursor style depending on mode", async () => {
+        const modeHandler = new ModeHandler();
+
+        modeHandler.setCurrentModeByName(ModeName.Normal);
+        assert.equal(vscode.window.activeTextEditor.options.cursorStyle, vscode.TextEditorCursorStyle.Block);
+
+        modeHandler.setCurrentModeByName(ModeName.Insert);
+        assert.equal(vscode.window.activeTextEditor.options.cursorStyle, vscode.TextEditorCursorStyle.Line);
+    });
```
jpoon: Can you add another case here for Visual mode?

Paxxi:
Ok added the test case.

Note for the future, I can be sloppy sometimes in PR's so don't hesitate to "nitpick", I expect it :smile:

jpoon:
@Paxxi one more tiny thing, can you merge the latest changes from `master` to this branch?

Paxxi:
done

jpoon:
Fixes https://github.com/VSCodeVim/Vim/issues/132! Thanks @Paxxi!

johnfn:
@Paxxi unfortunately this breaks visual line mode :( In visual line mode, the block cursor is often in the middle of the linewise selection (Vim will always select the entire line the cursor is in, no matter if it's at the beginning or the end).

Can you find any way to fix this? Otherwise we'll have to revert, which would be a shame because this cursor is really nice.

Paxxi:
I see the issue, will investigate and see how we can solve it

Paxxi:
Not sure if I will have time to investigate in the next few days so maybe revert this in case you want to do a release and we can give it another try once I figure it out

Issue #256, opened 2016-06-08
better testing infrastructure


johnfn:
I would really love it if we could write tests like this:
```
{
    start: `abc |def ghi`,
    keysPressed: `ww`,
    end: `abc def ghi|`
}
```

johnfn:
Closed thanks to @darrenweston

PR #292, opened 2016-06-14
Add simpler test mechanism and convert some tests

darrenweston:
Hi I've had a first crack at #256.

I'm new to typescript/javascript/git/github so feel free to tell me how to improve or change this to what you have in mind.

I've added an extra optional **endMode**: parameter so we can check the final mode.
```typescript
+    newTest("Can handle 'caw' on blanks",
+        {
+            start: ['text   tex|t'],
+            keysPressed: '^lllllcaw',
+            end: ['text|'],
+            endMode: ModeName.Insert
+        }
+    );
```

Here is a code snippet which will make entering tests easier. Activate with 'newt':
```typescript
	"Add new test": {
		"prefix": "newt",
		"body": [
    	"newTest(\"$1\",",
		"	{",
		"		start: ['$2'],",
		"		keysPressed: '$3',",
		"		end: ['$4'],",
		"		//endMode: ModeName.Insert",
		"	}",
    	");",
		"$5"
		],
		"description": "Create a new test"
	}
```

I also added a few 'P' tests. Note that 3 tests are currently failing.

### DIFF
--- a/test/mode/modeNormal.test.ts
+++ b/test/mode/modeNormal.test.ts
@@ -1,22 +1,47 @@
 "use strict";
 
 import * as assert from 'assert';
-import { setupWorkspace, cleanUpWorkspace, assertEqualLines, assertEqual } from './../testUtils';
+import { setupWorkspace, cleanUpWorkspace, assertEqualLines,assertEqual } from './../testUtils';
 import { ModeName } from '../../src/mode/mode';
 import { TextEditor } from '../../src/textEditor';
 import { ModeHandler } from '../../src/mode/modeHandler';
+import { Position } from '../../src/motion/position';
+import { TestObject, testIt } from '../testSimplifier';
+
 
 suite("Mode Normal", () => {
 
     let modeHandler: ModeHandler;
-
+    let testWithObject: (t: TestObject) => Promise<void>
+    
     setup(async () => {
         await setupWorkspace();
 
         modeHandler = new ModeHandler();
+        testWithObject = testIt.bind(null, modeHandler);
     });
 
     teardown(cleanUpWorkspace);
+    
+    function newTest(title: string, testObj: TestObject) {
+        let niceStack = (new Error).stack.split('\n').splice(2, 1).join('\n');
+        test(title, async () => testWithObject(testObj)
+            .catch(reason => {
+                reason.stack = niceStack;
+                throw reason;
+            })
+        );
+    }
+    function newTestOnly(title: string, testObj: TestObject) {
+        console.log("!!! Running single test !!!");
+        let niceStack = (new Error).stack.split('\n').splice(2, 1).join('\n');
+        test.only(title, async () => testWithObject(testObj)
+            .catch(reason => {
+                reason.stack = niceStack;
+                throw reason;
+            })
+        );
+    }
 
     test("can be activated", async () => {
         let activationKeys = ['<esc>', '<ctrl-[>'];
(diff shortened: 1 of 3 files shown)


johnfn:
omg thankyou thankyou thankyou thankyou!!!! I can't wait to merge this in, it will make writing tests so much easier!!!

Does `newTestOnly` really only run one test because holy smokes that is so useful.

Your Typescript looks pretty good to me for your first try! I'll leave you a few comments to help you learn :)

test/mode/modeNormal.test.ts
```diff
@@ -277,327 +290,218 @@ suite("Mode Normal", () => {
...
-        assertEqual(TextEditor.getSelection().start.character, 6, "F failed");
-    });
-
-    test("Can handle 'T' twice", async () => {
-        await modeHandler.handleMultipleKeyEvents(
-            'itext text'.split('')
-        );
+    newTest("$ always keeps cursor on EOL",
+        {
+            start: ['text text', 'text', 'text tex|t'],
```
johnfn: You can use backtick strings in JavaScript/TypeScrip. They can span multiple lines.

test/testSimplifier.ts
```diff
@@ -0,0 +1,153 @@
...
+    
+    constructor(private _testObject: TestObject) {
+        this._parse(_testObject);
+    }
+
+    public get isValid(): boolean {
+        return this._isValid;
+    }
+    
+    private _setStartCursorPosition(lines: string[]) {
```
johnfn: In TypeScript it's good to mark what types functions return.

test/testSimplifier.ts
```diff
@@ -0,0 +1,153 @@
...
+
+    private _setEndCursorPosition(lines: string[]) {
+        let result = this._getCursorPosition(lines);
+        this.endPosition = result.position;
+        return result.success;
+    }
+    
+    private _getCursorPosition(lines: string[]) {
+        let ret = { success: false, position: new Position(0, 0) };
+        lines.forEach((line, lineIdx) => {
```
johnfn: I generally prefer `for` loops over `forEach` etc because they read better and you can more easily `break` out of them if it becomes necessary.

test/testSimplifier.ts
```diff
@@ -0,0 +1,153 @@
...
+        let ret = this._testObject.end.slice(0);
+        ret[this.endPosition.line] = ret[this.endPosition.line].replace('|', '');
+        return ret;
+    }
+
+    /** Returns a sequence of Vim movement characters 'hjkl' as a string array
+     * which will move the cursor to the start position given in the test. */
+    public getKeyPressesToMoveToStartPosition(): string[] {
+        let ret: string = '';
+        let h: number, j: number, k: number, l: number;
```
johnfn: You can initialize multiple variables all in one go:

`let h = 0, j = 0, k = 0, l = 0;` 

Also if you initialize a variable when you create it, you don't need to declare a type for it, because it will be inferred by TypeScript.

test/testSimplifier.ts
```diff
@@ -0,0 +1,153 @@
...
+
+        let cursorPosAfterEsc =
+            this._testObject.start[this._testObject.start.length - 1].replace('|', '').length - 1;
+        let numCharsInCursorStartLine =
+            this._testObject.start[this.startPosition.line].replace('|', '').length - 1;
+        let columnOnStartLine = Math.min(cursorPosAfterEsc, numCharsInCursorStartLine);
+        let charactersToMove = this.startPosition.character - columnOnStartLine;
+
+        if (linesToMove > 0) {
+            ret += Array(linesToMove + 1).join('j');
```
johnfn: Huh, I haven't seen this trick before.

johnfn:
Ok, I looked a bit closer and the only thing I would like before we merge this in would be the original call stack for the error. Right now, no matter what test fails, the callstack is always `testSimplifier.ts:137:5`. It would be great to be able to click on a line in the callstack and have it bring me to the error.

The easiest way I can think of to do this is have the newTest() function simply return a boolean that indicates whether the test has failed, and then wrap assert around newTest e.g. `assert(newTest(/* stuff */))`. Perhaps newTest could return an object of both the bool and some corresponding text that explains why/how the test failed, and then you could do something like `assert(test.result, test.msg)`. 

Again, thanks a bunch! ❤️

darrenweston:
Hi Grant,

Thanks for the coding tips. I have made the suggested changes.
I have figured out how to hack the call stack when a test fails. Not sure if it is good practice but I think it is better than wrapping all tests in assert().

Yes you can call a single test with newTestOnly it makes testing way faster but you have to remember to change it back to newTest before committing :)

Let me know if there are any other things that need changing.

Have fun!
Darren.

PR #294, opened 2016-06-15
Refactor dot

johnfn:
This makes . work with actions such as `cawasdf<esc>` and even the extra-tricky `d/blah<esc>`. 

. is actually pretty solid now. Now I just need to do repeated actions and then my work here is done. 😉


johnfn:
Looks like Travis got into a bad state somehow...

PR #302, opened 2016-06-16
Select into visual mode

johnfn:
Clicking should move the cursor, and selecting should cause you to go into visual mode.

It's hard to do this perfectly with the current state (and/or my understanding) of the VSCode API, but this solution works 99% of the time.


johnfn:
Travis is a strugglebus.

PR #312, opened 2016-06-17
Add support for 'U' uppercase

rebornix:
Convert selected range to Uppercase is an often-used command.

The code is well designed so it's easy enough to add new commands support. The only catch is all commands are implemented inside a single file, which makes it lack of categorization and readability, any ideas to refactor it?

### DIFF
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -229,6 +229,17 @@ export class BaseOperator extends BaseAction {
      * Run this operator on a range, returning the new location of the cursor.
      */
     run(vimState: VimState, start: Position, stop: Position): Promise<VimState> { return; }
+
+    public transformRange(start: Position, end: Position): [Position, Position] {
+      if (start.compareTo(end) <= 0) {
+        end = new Position(end.line, end.character + 1);
+      } else {
+        [start, end] = [end, start];
+        end = new Position(end.line, end.character + 1);
+      }
+
+      return [start, end];
+    }
 }
 
 export enum KeypressState {
@@ -656,16 +667,7 @@ export class DeleteOperator extends BaseOperator {
      * Deletes from the position of start to 1 past the position of end.
      */
     public async run(vimState: VimState, start: Position, end: Position): Promise<VimState> {
-        if (start.compareTo(end) <= 0) {
-          end = new Position(end.line, end.character + 1);
-        } else {
-          const tmp = start;
-          start = end;
-          end = tmp;
-
-          end = new Position(end.line, end.character + 1);
-        }
-
+        [start, end] = this.transformRange(start, end);
         const isOnLastLine = end.line === TextEditor.getLineCount() - 1;
 
         // Vim does this weird thing where it allows you to select and delete
@@ -767,6 +769,20 @@ export class DeleteOperatorXVisual extends BaseOperator {
 
     public async run(vimState: VimState, start: Position, end: Position): Promise<VimState> {
       return await new DeleteOperator().run(vimState, start, end);
+    }
+}
+
+@RegisterAction
+export class UpperCaseOperator extends BaseOperator {
+    public keys = ["U"];
+    public modes = [ModeName.Visual, ModeName.VisualLine];
+    
+    public async run(vimState: VimState, start: Position, end: Position): Promise<VimState> {
+      [start, end] = this.transformRange(start, end);
+      let text = vscode.window.activeTextEditor.document.getText(new vscode.Range(start, end));
+      await TextEditor.replace(new vscode.Range(start, end), text.toUpperCase());        
+      vimState.currentMode = ModeName.Normal;
+      return vimState;
     }
 }
 
--- a/test/mode/modeVisualLine.test.ts
+++ b/test/mode/modeVisualLine.test.ts
@@ -64,6 +64,18 @@ suite("Mode Visual", () => {
         assertEqual(modeHandler.currentMode.name, ModeName.Normal);
     });
 
+    test("Can handle U", async () => {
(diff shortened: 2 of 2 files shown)


johnfn:
Hey @rebornix, thanks for the awesome PR! I'm glad to hear that the code is easy to understand for the most part.

I hear you about the gigantic file. Honestly, for whatever reason, I've always had a lower sensitivity to this issue than other people, which is why I haven't split it yet. When I tried briefly, I was having difficulty with decorators and splitting actions.ts into multiple files - I think TypeScript decorators may have some implementation bugs (they are classified as experimental, after all). 

TL;DR - I'm totally open to PRs that split it up. It's a task I plan to do at some point, but it's not high priority. If a couple people nag me to do it, I'll bump the priority 😉

src/actions/actions.ts
```diff
@@ -43,6 +43,42 @@ const compareKeypressSequence = function (one: string[], two: string[]): boolean
   return true;
 };
 
+const adjustVimSelectionPosition = function(vimState: VimState, start: Position, end: Position): any {
```
johnfn: Two comments here:

* `adjustVimSelectionPosition` is an ambiguous name. Something like `transformRangeForDeleteOperator` might be better. 
* Instead of `any` I would prefer using a TypeScript tuple type: `[Position, Position]`

johnfn: Also I generally prefer static functions to free-floating ones. When we split this file up into many files, the decision will make our lives easier. 😃

src/actions/actions.ts
```diff
@@ -583,6 +589,21 @@ export class DeleteOperatorXVisual extends BaseOperator {
 }
 
 @RegisterAction
+export class UpperCaseOperatorXVisual extends BaseOperator {
+    public keys = ["U"];
+    public modes = [ModeName.Visual, ModeName.VisualLine];
+    
+    public async run(vimState: VimState, start: Position, end: Position): Promise<VimState> {
+      [start, end] = adjustVimSelectionPosition(vimState, start, end);
```
johnfn: I actually don't understand why you need to transform the range here. Most of the logic in DeleteOperator is to ensure we get exactly the right newlines we need to delete, but we don't really care about capitalizing newlines. 

I think just running from start to end.getRight() should be enough?

src/actions/actions.ts
```diff
@@ -583,6 +589,21 @@ export class DeleteOperatorXVisual extends BaseOperator {
 }
 
 @RegisterAction
+export class UpperCaseOperatorXVisual extends BaseOperator {
```
johnfn: You can just name this UpperCaseOperator.

johnfn:
Last thing - would you mind adding a few tests?

src/actions/actions.ts
rebornix: Thanks for the naming suggestion, it's always pain for non-English speaker like me :) The strict return type suggestion is also good, will change it later.

src/actions/actions.ts
rebornix: When users select ranges in reverse order, we need to swap `start` and `end` first, this is contained in the extracted method so I just reuse it. 

Moreover, this method can be used for any Visual/VisualLine selection and I think it would be perfect to deal every special case if possible, then people don't need to worry about when to `getRight` or `getLeft`, whether to take current character under cursor into account or not, etc.

rebornix:
@johnfn thanks for your detailed review and suggestion. I tried to run test cases on my Windows Machine previously but seems VS Code package will download Linux related package instead. I'll look into it and add tests :smiley:

src/actions/actions.ts
johnfn: Ah; all that logic is contained within executeOperator inside modeHandler.ts.

However, I think that only the swapping of start and stop should be moved in there (that's a good catch, btw!) - the rest of this stuff is DeleteOperator specific.

rebornix: Cool, let me see if I can add this refactor together with this pr.

johnfn: 👍  I don't think that you'll end up needing to refactor out this function at all.

johnfn: (Though if you want you can remove the start/stop switching logic I have in the other operators...)

rebornix: Will do that. Once the code is updated you can know if I fully get your ideas :)

BTW, send you an email about other funny stuff.

rebornix:
I've tried to move the start/end swapping for Visual mode but failed. The reason is that this will breaks deletion.

1. Delete Operator may be triggered by `executeOperator` or `runAction`, while we don't have any position information inside `runAction`, so we have to keep the swapping inside Delete Operator.
2. With above info in mind, we have the swapping inside Delete Op. So putting the swapping inside `executeOperator` will lead to the duplication of swapping for Delete Op as `executeOperator` might trigger Delete Op as well.

So the temporary solution for UpperCase operator is reusing the swapping method. This is caused by my limited knowledge of Action/Operator relationships and the workflow, once I have better understanding of the code, I may come up with a prettier solution.

src/actions/actions.ts
```diff
@@ -229,6 +229,17 @@ export class BaseOperator extends BaseAction {
      * Run this operator on a range, returning the new location of the cursor.
      */
     run(vimState: VimState, start: Position, stop: Position): Promise<VimState> { return; }
+
+    public transformRange(start: Position, end: Position): [Position, Position] {
```
rebornix: This needs to be reconsidered whether to put the method here.

johnfn:
Hmm, I see. I will dig into this tomorrow and see what I can figure out!

PR #319, opened 2016-06-18
Add %.

johnfn:
We do a fairly basic implementation that only tracks the state of the type character you're over and its pair (e.g. "[" and "]"). This is similar to what Vim does.

It can't handle strings very well - something like "|( ')' )" where | is the cursor will cause it to go to the ) in the quotes, even though it should skip over it.

PRs welcomed! I have this case at low priority because I believe it's a lot of effort for very little gain. 

Ideally though, VSC implements https://github.com/Microsoft/vscode/issues/7177.


johnfn:
Oh great... now Travis is telling me the tests pass when they fail. 

Apparently the only way to get them to pass is to cause them to fail so bad they don't even run. 🔥

PR #325, opened 2016-06-20
Tilde key toggles case and moves forwards

markrendle:
I'm just adding things a key at a time until VsCode/Vim matches my muscle-memory. Hope that's OK :smile:


johnfn:
Not a problem at all! 😉

PR #350, opened 2016-06-24
Align TextEditorOptions between test code and workspace

rebornix:
When we format code, change code indentation, the final output relies on TextEditorOptions like `tabSize` and `insertSpaces`. Here we allow changing the options in test cases to keep alignment.


johnfn:
Look at that beautiful green checkmark! 😄

PR #351, opened 2016-06-24
Implement <count>yy

rebornix:
With `getDownByCount`, supporting Nyy is real simple.


johnfn:
As well as it should be! 😀

PR #359, opened 2016-06-26
Update Roadmap Motions section

rebornix:
Align the categories with Vim Docs http://vimdoc.sourceforge.net/htmldoc/motion.html#word-motions


johnfn:
I just took it from here http://vimdoc.sourceforge.net/htmldoc/quickref.html

rebornix:
Oh that would be better, my change is somewhat too detailed.

PR #362, opened 2016-06-27
Sentence

rebornix:
Sentence related movements.

A paragraph boundary is also a sentence boundary, so the regex can handle all sentence endings pattern. Another tricky part is when the cursor is at the end of a paragraph (seems it's also the beginning of next paragraph), next section will jump to the first non-whitespace character whether it's a valid sentence or not.


rebornix:
Add sentence backward and fix numeric prefix bugs.

johnfn:
Awesome!

The only thing I don't really like is the type of tests. In the future, I'd prefer using the `newTest` style and sending keys, because it's much easier to verify the test is correct just by looking at it (vs counting characters) and you're testing the thing you actually care about (vim motions) rather than internals of our code.

rebornix:
@johnfn agree, while writing cases I found `newTest` is the easiest way of validating line counts, cursor position, etc. I'll go with that in following commits.

PR #363, opened 2016-06-27
[WIP]: Switch Window

rebornix:
Leverage VS Code TabGroup to implement Windows management. Mark it as Work in progress as we still have no key mapping and it will still people's Ctrl+w on Windows. Besides, VS Code treats Ctrl as modifiers but not keys, so you must press `ctrl+w` first then `hjkl` so the experience is somewhat not smooth.


rebornix:
Besides, VS Code supports Tabs inside a group. We may want to map it to Vim's tab feature.

johnfn:
Remapping keys is my top priority right now. 

(Unfortunately that requires proper history management, so it could be a ways off.)

LeonB:
Would it also be possible to implement `ctrl+w q` for closing a split?

PR #364, opened 2016-06-27
Bring Ctrl keys back

rebornix:
Since VS Code treats ctrl keys as modifiers, all modifiers related keys need be registered in contribution. Here I just register them then we have all related functionalities back

1. `ctrl-b`
2. `ctrl-f`
3. `ctrl-r`


johnfn:
👍

PR #365, opened 2016-06-27
Add Ncc support and revise cc behavior

rebornix:
Original `cc` behaviour is not the same as Vim. When you trigger `cc` in Vim with below content,

```
one
|two
three
```

You will get
```
one
|
three
```

However, what we get in VSCodeVim is

```
one
|three
```

I suppose it's not a correct way of `cc` and it's caused by treating `cc` as `LineWise`. Now I add numeric support and set it as `CharacterWise` to fix this problem.

### DIFF
--- a/src/register/register.ts
+++ b/src/register/register.ts
@@ -4,7 +4,7 @@ import { VimState } from './../mode/modeHandler';
  * There are two different modes of copy/paste in Vim - copy by character
  * and copy by line. Copy by line typically happens in Visual Line mode, but
  * also shows up in some other actions that work over lines (most noteably dd,
- * yy, and cc).
+ * yy).
  */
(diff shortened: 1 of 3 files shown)


src/register/register.ts
```diff
@@ -4,7 +4,7 @@ import { VimState } from './../mode/modeHandler';
  * There are two different modes of copy/paste in Vim - copy by character
  * and copy by line. Copy by line typically happens in Visual Line mode, but
  * also shows up in some other actions that work over lines (most noteably dd,
- * yy, and cc).
```
johnfn: Haha, nice catch!

johnfn:
Good stuff. This has been bugging me for a while.

PR #366, opened 2016-06-27
Support `C`

rebornix:
Delete from the cursor position to the end of the line and [count]-1 more lines [into register x], and start insert.

### DIFF
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -958,9 +958,15 @@ class CommandDeleteToLineEnd extends BaseCommand {
 class CommandChangeToLineEnd extends BaseCommand {
   modes = [ModeName.Normal];
   keys = ["C"];
+  canBePrefixedWithCount = true; 
 
   public async exec(position: Position, vimState: VimState): Promise<VimState> {
     return new ChangeOperator().run(vimState, position, position.getLineEnd().getLeft());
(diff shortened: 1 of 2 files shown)


src/actions/actions.ts
```diff
@@ -958,10 +958,16 @@ class CommandDeleteToLineEnd extends BaseCommand {
 class CommandChangeToLineEnd extends BaseCommand {
   modes = [ModeName.Normal];
   keys = ["C"];
+  canBePrefixedWithCount = true;
 
   public async exec(position: Position, vimState: VimState): Promise<VimState> {
     return new ChangeOperator().run(vimState, position, position.getLineEnd().getLeft());
   }
+
+  public async execCount(position: Position, vimState: VimState): Promise<VimState> {
```
johnfn: If you override `execCount`, you don't need to keep `exec`. `execCount`'s default behavior is to call `exec`, but if you override it, then `exec` is never called.

johnfn:
Remove the old `exec` function and we should be good here 👍

rebornix:
@johnfn good catch, it's not necessary at all. To allow users not override it every time, I remove the `abstract` keyword of the method definition.

johnfn:
Oh, good idea. Thanks!

PR #368, opened 2016-06-28
Tabs

rebornix:
Mapping vim tab commands to VS Code new tab features. For now only part of the commands can be implemented with existing APIs but we are going to see more APIs in following sprints.


rebornix:
Already add all Tab commands possible to implement with existing VS Code commands https://github.com/Microsoft/vscode/issues/6605 . It's also a good way to help define VS Code Tab features by looking at Vim Tab commands definition.

johnfn:
Awesome. 👍

PR #377, opened 2016-06-30
Add jsdiff typings and fix tslint

rebornix:
Add typings for jsdiff to bypass `vscode:precompile` failure. The typings file is a draft and only contains what we used for now. We can add more detailed if we have more usage later and then extract them out of our repo.

Another fix is tslint :)

### DIFF
--- a/src/mode/modeHandler.ts
+++ b/src/mode/modeHandler.ts
@@ -583,7 +583,7 @@ export class ModeHandler implements vscode.Disposable {
         // have changed it as well. (TODO: do you even decomposition bro)
 
         if (vimState.currentMode !== this.currentModeName) {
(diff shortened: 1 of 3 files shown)


johnfn:
My bad, the typings already exist: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/diff/diff.d.ts

That `newTestOnly` thing is annoying, though. There should be a way to get tslint to ignore test files.

johnfn:
Actually, the typings folder is in .gitignore.

johnfn:
* Added `tsd install` to .travis.yml
* Removed typings folder (that was confusing)
* Implemented your tslint fixes

PR #383, opened 2016-07-01
Arrows

### LINKED ISSUE #382: visual mode extension using arrow keys
Hi, I know this is a new project so thanks very much.

when I do this:
v [arrow]
or V [arrow]

I expect this:
enters visual mode (or visual linewise mode) and extends the selection in the direction of the arrow key

But in fact this happens:
enters visual mode (or visual linewise mode), but then cancels it, moving the cursor in the direction of the arrow key.

I can work around this by doing this:
(vV)(hjkl)

rebornix:
Enable users to leverage arrow keys as `hjkl`. Fixes #382.


rebornix:
Running into `The command "typings install" failed and exited with 127 during .` error in CI. It's related to our diff typings change.

johnfn:
Bah, that's my bad. I'll fix it soon.

PR #385, opened 2016-07-04
Word in visual mode

rebornix:
1. Enrich `aw` with `executeWithOperator` and `executeWithCount`
2. Add `aW`
3. Enrich `iw`
4. Add `iW`
5. Normal/Visual mode test cases.

### DIFF
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -2160,7 +2160,7 @@ class MovementAWordTextObject extends BaseMovement {
               await this.execAction           (position, vimState);
 
           if (firstIteration) {
-            result.start = temporaryResult.start;           
+            result.start = temporaryResult.start;
           }
 
           // Result is always a Movement.
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -2126,7 +2126,7 @@ class MovementAWordTextObject extends BaseMovement {
           start = position.getLastWordEnd().getRight();
         } else {
           start = position.getWordLeft(true);
-    }
+        }
     }
 
     if (vimState.currentMode === ModeName.Visual && !vimState.cursorPosition.isEqual(vimState.cursorStartPosition)) {
@@ -2192,6 +2192,35 @@ class MovementAWordTextObject extends BaseMovement {
     }
 
     return res;
+  }
+}
+
+@RegisterAction
+class MovementABigWordTextObject extends MovementAWordTextObject {
+  keys = ["a", "W"];
+
+  public async execAction(position: Position, vimState: VimState): Promise<IMovement> {
+    let start: Position;
+    let stop: Position;
+
+    const currentChar = TextEditor.getLineAt(position).text[position.character];
+
+    if (/\s/.test(currentChar)) {
+        start = position.getLastBigWordEnd().getRight();
+        stop = position.getCurrentBigWordEnd();
+    } else {
+        start = position.getBigWordLeft();
+        stop = position.getBigWordRight().getLeft();
+    }
+
+    if (vimState.currentMode === ModeName.Visual && !vimState.cursorPosition.isEqual(vimState.cursorStartPosition)) {
+        start = vimState.cursorStartPosition;
+    }
+
+    return {
+      start: start,
+      stop: stop
+    };
   }
 }
 
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -198,6 +198,13 @@ export abstract class BaseMovement extends BaseAction {
             result = temporaryResult;
             position = temporaryResult;
           } else if (isIMovement(temporaryResult)) {
+            if (result instanceof Position) {
+              result = {
+                start : new Position(0, 0),
+                stop : new Position(0, 0)
+              };
+            }
+
             if (firstIteration) {
               (result as IMovement).start = temporaryResult.start;
             }
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -2173,23 +2173,12 @@ class MovementAWordTextObject extends BaseMovement {
 
       return result;
   }
-
   public async execActionForOperator(position: Position, vimState: VimState): Promise<IMovement> {
-    const line = TextEditor.getLineAt(position).text;
-    const currentChar = line[position.character];
     const res = await this.execAction(position, vimState);
-    const wordEnd = await new MoveWordBegin().execActionForOperator(position, vimState);
-
-    // TODO(whitespace)
-    if (currentChar !== ' ' && currentChar !== '\t') {
-      res.stop = wordEnd.getRight();
-    } else {
-      res.stop = wordEnd;
-    }
-
-    if (res.stop.character === line.length + 1) {
-      res.start = (await new MoveLastWordEnd().execAction(res.start, vimState)).getRight();
-    }
+    // Since we need to handle leading spaces, we cannot use MoveWordBegin.execActionForOperator
+    // In normal mode, the character on the stop position will be the first character after the operator executed
+    // and we do left-shifting in operator-pre-execution phase, here we need to right-shift the stop position accordingly.
+    res.stop = new Position(res.stop.line, res.stop.character + 1);
 
     return res;
   }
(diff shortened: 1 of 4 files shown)


test/mode/modeVisual.test.ts
```diff
@@ -5,9 +5,14 @@ import { ModeHandler } from '../../src/mode/modeHandler';
 import { setupWorkspace, cleanUpWorkspace, assertEqualLines, assertEqual } from './../testUtils';
 import { ModeName } from '../../src/mode/mode';
 import { TextEditor } from '../../src/textEditor';
+import { getTestingFunctions } from '../testSimplifier';
```
johnfn: Thanks! I need to change the other tests at some point...

src/actions/actions.ts
```diff
@@ -1812,53 +1812,108 @@ class ActionChangeChar extends BaseCommand {
...
   }
 
+  public async execActionWithCount(position: Position, vimState: VimState, count: number): Promise<Position | IMovement> {
+      let recordedState = vimState.recordedState;
+      let result: IMovement = {
+        start: undefined,
+        stop: undefined
+      };
+
+      if (count < 1) {
```
johnfn: It seems like you copied this from `BaseMovement#execActionWithCount`, which I'm not a big fan of. DRY and all that. Is there any other way to do this?

rebornix: It's slightly different with `BaseMovement#execActionWithCount`, especially the way we calculate `start` and `stop`.

In `BaseMovement#execActionWithCount`, we will assign the return value of each time Movement to `result`, which means both `start` and `stop` of `result` will be updated. This works perfectly when you run `2w`, we just move the `start` property twice forward.

While here, we are not just moving cursor, instead we are expanding the selection. We need to cache the `start` of the first movement and cache the `stop` of the last movement.

The only common part of these two methods are the `count` tweaking. But you are right, we can have a better way of this method to avoid confusing other maintainers.

test/mode/modeVisual.test.ts
rebornix: Oh we can do this step by step. Each time we add new test cases to old ones, we can do the tweaking at the same time.

rebornix:
Add logic for `iw` and `iW`.

Just like I mentioned above, selecting words, sentences, paragraphs, etc all share the same `executeWithCount` that we use in `MovementAWordTextObject`. It's different from `BaseMovement#execActionWithCount`.

In summary, `BaseMovement#execActionWithCount` is returning a movement, which tells us how cursor moves from old position to new position. While `MovementAWordTextObject#executeWithCount` is returning a selection, which tells us which range (start and stop) should we operate on. `IMovement` is not an ideal name for this situation IMHO.

rebornix:
Just realize that `IMovement` is for selection. The funny thing is the `IMovement` part of `BaseMovement#execActionWithCount` is never being called (actually I don't know what kind of movement will use original logic). So I just moved my `execActionWithCount` into it.

rebornix:
Update and fix conflicts

src/actions/actions.ts
```diff
@@ -171,7 +171,15 @@ export abstract class BaseMovement extends BaseAction {
    */
   public async execActionWithCount(position: Position, vimState: VimState, count: number): Promise<Position | IMovement> {
       let recordedState = vimState.recordedState;
+<<<<<<< eaeb9ce3df631aafdd4b909978d9512fbe6aea47
       let result: Position | IMovement = new Position(0, 0); // bogus init to satisfy typechecker
+=======
```
johnfn: There are still some merge conflicts here.

johnfn:
Sorry, I've left this hanging for too long. Can you explain what you did? Just add counts for aw iw?

rebornix:
The first thing I did is fixing `aw` and `iw` with count and then add `aW`, `iW`, the Big Word feature.

src/actions/actions.ts
```diff
@@ -180,21 +185,31 @@ export abstract class BaseMovement extends BaseAction {
...
 
-          result = temporaryResult;
-
-          if (result instanceof Position) {
-            position = result;
-          } else if (isIMovement(result)) {
-            position = result.stop;
+          if (temporaryResult instanceof Position) {
+            resultPosition = temporaryResult;
+            position = temporaryResult;
```
johnfn: Why do we assign to `position` here and elsewhere? As far as I can see, `position` is never again read from.

src/actions/actions.ts
```diff
@@ -171,7 +171,12 @@ export abstract class BaseMovement extends BaseAction {
    */
   public async execActionWithCount(position: Position, vimState: VimState, count: number): Promise<Position | IMovement> {
       let recordedState = vimState.recordedState;
-      let result: Position | IMovement = new Position(0, 0); // bogus init to satisfy typechecker
+      let resultPosition: Position = new Position(0, 0);
```
johnfn: Isn't `resultPosition` the exact same thing as `position`?

rebornix:
Fix the comment by @johnfn , I was doing wrong with Union Type but now turned around.

src/actions/actions.ts
```diff
@@ -180,17 +180,25 @@ export abstract class BaseMovement extends BaseAction {
...
+            position = temporaryResult;
+          } else if (isIMovement(temporaryResult)) {
+            if (firstIteration) {
+              (result as IMovement).start = temporaryResult.start;
+            }
+
+            if (lastIteration) {
+              (result as IMovement).stop = temporaryResult.stop;
+            } else {
+              position = temporaryResult.stop.getRightThroughLineBreaks();
```
johnfn: This getRight concerns me. Why is it necessary?

src/actions/actions.ts
```diff
@@ -1900,94 +1908,168 @@ class ActionChangeChar extends BaseCommand {
 class MovementAWordTextObject extends BaseMovement {
   modes = [ModeName.Normal, ModeName.Visual];
   keys = ["a", "w"];
+  canBePrefixedWithCount = true;
 
   public async execAction(position: Position, vimState: VimState): Promise<IMovement> {
-    if (vimState.currentMode === ModeName.Visual && !vimState.cursorPosition.isEqual(vimState.cursorStartPosition)) {
-      // TODO: This is kind of a bad way to do this, but text objects only work in
-      // visual mode if you JUST entered visual mode
```
johnfn: Did you manage to fix the case this is guarding against?

johnfn:
Just want to hear back on my 2 comments and we should be good to merge!

src/actions/actions.ts
rebornix: yup, it's necessary for IMovment. This is also why the logic of IMovement and Position is totally different. 

When we run a single movement without an operator, like `w`, we want to move cursor to the beginning of next word. While if we run a single movement with an operator, like `aw`, we want to move the end cursor to **the end of current word** (aka, the left-side character of the beginning of next word). Then we sucessfully get a selection of the whole word.

Above is about single movement w/ operator. Then if we run movements with an operator multple times, like `2aw`, after running the first `aw`, the end of seletion or the cursor is at the end of first word. As we need to include next following word into the selection, we need to pass the position of the begining of next word to `aw` command, then we can get the correct end position of the second `aw`. Lastly, we combine the beginning of first word and the end of last word and return the result selection.

Like I mentioned before, Movement (contains a start and a stop) is likely a selection. When executing selection with numeric prefix, I think we always need to move cursor to next character of the end of current selection.

src/actions/actions.ts
rebornix: I moved this piece of code to below https://github.com/VSCodeVim/Vim/pull/385/files#diff-7addcd1a8a241060a73b1fe4abb39960R1932 . If we entered Visual Mode and moved the cursor (selection length > 1), the final result selection depends on the cursor direction. So I checked `vimState.cursorPosition.isBefore(vimState.cursorStartPosition)`. So in my implementation, all cases are covered.

rebornix:
@johnfn rebased code and replied to your comments :)

rebornix:
Rebase again to resolve conflicts.

johnfn:
Woo finally! Sorry for taking so long, and thanks @rebornix!

PR #399, opened 2016-07-06
Fix #358.

### LINKED ISSUE #358: current line scroll to top when begin editing in split editor
occurs very often after the first occurrence, but not 100% reproducible, and sometimes it is more often in one pane than the other:

1. open a file
2. split the editor
3. scroll down the left pane
4. mouse click to switch between the 2 edit panes
5. begin edit in the left pane via `o`, `i` or `p` (and maybe other commands)

the current line scroll to the top of the pane.

rebornix:
Yay! We love PRs! 🎊

- [x] Commits are squashed 

From my observation, if you open the same file twice in two window viewport inside a single tab, these two viewports are sharing the same buffer. The history is shared between these two viewports, selection is shared either. The only thing not shared is current cursor position in normal mode, totally no idea why is that.

VS Code takes almost the same concept, if we open the same file in two Grouped Editors, they are sharing the same buffer. And VS Code doesn't share cursor between these two panes either. So each time when we switch between two panes which open the same file, we just reassign the cursor position to `vimState`. That's all we need to handle as VS Code stores everything about the buffer like Vim does.

The only thing that might break is history. But currently everything works fine.

I added test cases for this change on my local box. But it turned out that VS Code `splitEditor` or `focusPreviousGroup` or `focusNextGroup` doesn't move the active cursor to target Group Editor. So our correct code can never pass test cases and it's false negative. So before VS Code test library has a fix for this thing, I'd like to keep the test cases on my local box. But I've tested this fix manually, it works.

### DIFF
--- a/extension.ts
+++ b/extension.ts
@@ -12,37 +12,72 @@ import { ModeHandler } from './src/mode/modeHandler';
 import { TaskQueue } from './src/taskQueue';
 import { Position } from './src/motion/position';
 
+export class EditorIdentity {
+  private _fileName: string;
+  private _viewColumn: vscode.ViewColumn;
+
+  constructor(textEditor?: vscode.TextEditor) {
+    this._fileName = textEditor && textEditor.document.fileName || "";
+    this._viewColumn = textEditor && textEditor.viewColumn || vscode.ViewColumn.One;
+  }
+
+  get fileName() {
+    return this._fileName;
+  }
+
+  get viewColumn() {
+    return this._viewColumn;
+  }
+
+  public hasSameBuffer(identity: EditorIdentity): boolean {
+    return this.fileName === identity.fileName;
+  }
+
+  public isEqual(identity: EditorIdentity): boolean {
+    return this.fileName === identity.fileName && this.viewColumn === identity.viewColumn;
+  }
+
+  public toString() {
+    return this.fileName + this.viewColumn;
+  }
+}
+
 let extensionContext: vscode.ExtensionContext;
 
 /**
  * Note: We can't initialize modeHandler here, or even inside activate(), because some people
  * see a bug where VSC hasn't fully initialized yet, which pretty much breaks VSCodeVim entirely.
  */
-let modeHandlerToFilename: { [key: string]: ModeHandler } = {};
-let previousActiveFilename: string | undefined = undefined;
+let modeHandlerToEditorIdentity: { [key: string]: ModeHandler } = {};
+let previousActiveEditorId: EditorIdentity = new EditorIdentity();
 
 let taskQueue = new TaskQueue();
 
-function activeFileName(): string {
-  return vscode.window.activeTextEditor.document.fileName;
-}
-
 export async function getAndUpdateModeHandler(): Promise<ModeHandler> {
-  const oldHandler = previousActiveFilename ? modeHandlerToFilename[previousActiveFilename] : undefined;
+  const oldHandler = modeHandlerToEditorIdentity[previousActiveEditorId.toString()];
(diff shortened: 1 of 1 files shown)


johnfn:
I'm holding off on merging this because now we track both activeFileName and activeFileIdentity, which seems confusing. activeFileName was intended to be the identity of the active file anyways. Can you think of a way to merge both of them into one?

rebornix:
@johnfn yeah `activeFileName` is part of `activeFileIdentity` so that's doable, I'll revise the code.

jpoon:
@rebornix, you're still working on one common file identity?

rebornix:
@jpoon sorry for the delay, I'll update this PR later today together with other PRs.

rebornix:
Add a new class for File/Editor Identity. I was trying to put the declaration into `textEditor.ts` but  if I did that, I kept getting `textEditor_1.EditorIdentiy is not a constructor` when running test cases (launch extension is perfect). Not sure what kind of issue I'm running into.

extension.ts
```diff
@@ -11,38 +11,83 @@ import { showCmdLine } from './src/cmd_line/main';
 import { ModeHandler } from './src/mode/modeHandler';
 import { TaskQueue } from './src/taskQueue';
 import { Position } from './src/motion/position';
+<<<<<<< aa8f6051a847c442e374ce22920ca5bd21835aec
```
johnfn: You still have a merge conflict here

extension.ts
```diff
@@ -11,38 +11,83 @@ import { showCmdLine } from './src/cmd_line/main';
...
+=======
+
+export class EditorIdentity {
+    private _fileName: string;
```
johnfn: Do we really need a whole class? Isn't a string sufficient?

rebornix: The catch is we need to store both file name and column number and compare file name or both of them sometimes. If we merge file name and column number into one single string, we need to split it when we only need to compare file name. Here I found a separate class is more intuitive.

rebornix:
Rebase and fix conflicts.

extension.ts
johnfn: Isn't it sufficient to ALWAYS compare filename AND column number? When don't we need to?

johnfn: Oh.. I see

I need to think about this a little more.

rebornix: Yeah, filename is referring to Buffer while filename+column number is actually referring to the real editor (Tab in VS Code). We need to share almost everything (VimState, History, Register) except cursor position.

The most interesting thing is, in Vim two window reading the same buffer even shares Selection.

extension.ts
```diff
@@ -12,37 +12,79 @@ import { ModeHandler } from './src/mode/modeHandler';
...
+
+  public hasSameBuffer(identity: EditorIdentity): boolean {
+    return this.fileName === identity.fileName;
+  }
+
+  public isEqual(identity: EditorIdentity): boolean {
+    return this.fileName === identity.fileName && this.viewColumn === identity.viewColumn;
+  }
+
+  public withColumn(column: vscode.ViewColumn): EditorIdentity {
```
jpoon: Sorry for the nitpick. Where is this function used?

rebornix: Oh it's not used for now, I thought this can be used to update Identity by column but finally it's not yet necessary. If other parts are good, I'll remove it and push a update.

jpoon: Yep, that's the only comment I have :)

rebornix:
@jpoon we are good to go now :)

jpoon:
Thanks @rebornix! 🎈

PR #400, opened 2016-07-07
Fix TS2318 and ignore .vscode-test folder

rebornix:
Yay! We love PRs! 🎊

- [x] Commits are squashed 

Firstly ignore `.vscode-test` folder, otherwise it will slow down the auto-compilation. Secondly as we set `noLib` as `true` and we are using ES6 most of the time (currently we don't have any ES7 syntax yet), so I add es6 lib file to it.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -9,6 +9,10 @@
     "experimentalDecorators": true
   },
   "exclude": [
-    "node_modules"
+    "node_modules",
+    ".vscode-test"
+  ],
+  "files": [
+    "node_modules/typescript/lib/lib.es6.d.ts"
   ]
 }


johnfn:
👍 , though you can actually just delete vscode-test entirely and test within VSCode itself.

tsconfig.json
```diff
@@ -9,6 +9,10 @@
     "experimentalDecorators": true
   },
   "exclude": [
-    "node_modules"
+    "node_modules",
+    ".vscode-test"
+  ],
+  "files": [
```
jpoon: I was looking into this a bit earlier, doesn't the `files` config define what files `tsc` will compile? (ref: https://www.typescriptlang.org/docs/handbook/tsconfig-json.html). 

> If no "files" property is present in a tsconfig.json, the compiler defaults to including all TypeScript (*.ts or *.tsx) files in the containing directory and subdirectories. When a "files" property is present, only the specified files are included.

I think doing this just masks the problem?

rebornix: I was referring to https://github.com/Microsoft/TypeScript/issues/3930 , seems if you set `nolib` as true you must add your own lib file as TypeScript compiler always rely on lib files to generate output. So the main purpose of `nolib` is allowing you to replace the original `lib.d.ts` to the one you want, here it's `lib.es6.d.ts`.

I'm not 100% sure whether it's correct but currently it's the only way to bypass the errors.

rebornix:
@johnfn the reason I always test those cases with command line is it's the same way as Travis does. Currently if you run tests inside VS Code, everything is good while if you run through command line, it will fail immediately. I mentioned this issue in Slack and it's due to HistoryTracker initialization.

johnfn:
The Travis thing _is_ annoying, but it's not really indicative of how real people will use our extension. I put more weight on running the tests within VSCode because those results seem to correlate stronger with actual bugs existing in the code.

tsconfig.json
johnfn: @jpoon you were correct; unfortunately, this change will break typescript compilation. 

Be sure to test `tsc --watch` before making changes to tsconfig.json 😉

rebornix:
@jpoon @johnfn sorry for the break and thanks for @johnfn 's quick turnaround. The only thing that bothers me is, why we put `noLib` here. Per documentation, it's used for `Do not include the default library file`, then which file should it use? Currently we have no lib files in our code base, the result is when you run `compile`, you will always get TS2318 errors

```
> vim@0.0.28 compile /Users/rebornix/Documents/code/Vim
> node ./node_modules/vscode/bin/compile -watch -p ./

error TS2318: Cannot find global type 'Array'.
error TS2318: Cannot find global type 'Boolean'.
error TS2318: Cannot find global type 'Function'.
error TS2318: Cannot find global type 'IArguments'.
error TS2318: Cannot find global type 'Iterable'.
error TS2318: Cannot find global type 'IterableIterator'.
error TS2318: Cannot find global type 'Iterator'.
error TS2318: Cannot find global type 'Number'.
error TS2318: Cannot find global type 'Object'.
error TS2318: Cannot find global type 'RegExp'.
error TS2318: Cannot find global type 'String'.
error TS2318: Cannot find global type 'Symbol'.
error TS2318: Cannot find global type 'TemplateStringsArray'.
error TS2468: Cannot find global value 'Symbol'.
4:41:04 PM - Compilation complete. Watching for file changes.
```

I can ignore these things as it doesn't affect the functionality but the catch is, VS Code is relying on the lib file to do intellisense and since the lib file is missing, we are seeing hundreds of errors highlighting in editor, which makes the whole development hard.

Maybe I did something wrong with my local setup, any comments?

johnfn:
@rebornix My mistake. You're right about TS2318. We have some files in our typings/ directory that have been gitignore'd from an old version (including lib.es6). You don't have those files. So we don't see TS2318, but you do.

**HOWEVER**, we can't just add lib.es6 to the files array, because that _unadds_ all of the files in src/, which breaks `tsc --watch`. 

The correct solution for now is to just add back lib.es6 to the typings directory.

Issue #384, opened 2016-07-03
Ctrl+w h will open up a new tab if there isn't one


johnfn:
`Ctrl+w h` doesn't actually work like that in Vim.

rebornix:
I suppose you are referring to Window, right? I'll take it :)

johnfn:
I'm not quite sure what the terminology is. A new pane perhaps? 😛

rebornix:
Not sure whether I used Vim wrongly but on my Mac and Linux VM, `<ctrl-w>+h` or `<ctrl-w>+l` won't create new tab/window if there isn't one. Is there any configuration setup in your vimrc?

BTW, we still need to implement `vsp {file}` and if the `{file}` is empty, create a new window and edit a new empty file.

johnfn:
Hmm, I'm using VSC Insiders with the latest build and `    "workbench.editor.showTabs": true,` in my user settings file.

rebornix:
This one can be closed now

PR #404, opened 2016-07-07
Open file in new window.

rebornix:
Yay! We love PRs! 🎊

- [x] Commits are squashed 

Support `:vsp`, `:vne` and `:vnew`.

This feature is limited to VS Code's group editors, which only support 3 groups at most. Besides, all horizontal splitting commands wont' work in VS Code

### DIFF
--- a/src/cmd_line/commands/file.ts
+++ b/src/cmd_line/commands/file.ts
@@ -4,8 +4,14 @@ import * as vscode from "vscode";
 import * as path from "path";
 import * as node from "../node";
 
+export enum FilePosition {
+    CurrentWindow,
+    NewWindow
+}
+
 export interface IFileCommandArguments extends node.ICommandArgs {
     name?: string;
+    position?: FilePosition;
 }
 
 
@@ -23,13 +29,48 @@ export class FileCommand extends node.CommandBase {
         return this._arguments;
     }
 
+    getViewColumn(sideBySide: boolean) : vscode.ViewColumn {
+        const active = vscode.window.activeTextEditor;
+        if (!active) {
+            return vscode.ViewColumn.One;
+        }
+
+        if (!sideBySide) {
+            return active.viewColumn;
+        }
+
+        switch (active.viewColumn) {
+            case vscode.ViewColumn.One:
+                return vscode.ViewColumn.Two;
+            case vscode.ViewColumn.Two:
+                return vscode.ViewColumn.Three;
+        }
+
+        return active.viewColumn;
+    }
+
     execute(): void {
+        if (!this._arguments.name) {
+            // Open an empty file
+            if (this._arguments.position === FilePosition.CurrentWindow) {
+                vscode.commands.executeCommand("workbench.action.files.newUntitledFile");
+            }
+
+            vscode.commands.executeCommand("workbench.action.splitEditor").then(() => {
+                vscode.commands.executeCommand("workbench.action.files.newUntitledFile").then(() => {
+                    vscode.commands.executeCommand("workbench.action.closeOtherEditors");
+                });
+            });
+
+            return;
+        }
+
         let currentFilePath = vscode.window.activeTextEditor.document.uri.path;
         let newFilePath = path.join(path.dirname(currentFilePath), this._arguments.name);
 
         if (newFilePath !== currentFilePath) {
             let folder = vscode.Uri.file(newFilePath);
-            vscode.commands.executeCommand("vscode.open", folder);
+            vscode.commands.executeCommand("vscode.open", folder, this.getViewColumn(this._arguments.position === FilePosition.NewWindow));
         }
     }
 }
--- a/src/cmd_line/commands/file.ts
+++ b/src/cmd_line/commands/file.ts
@@ -29,14 +29,21 @@ export class FileCommand extends node.CommandBase {
         return this._arguments;
     }
 
-    getViewColumn(sideBySide: boolean) : vscode.ViewColumn {
+    getActiveViewColumn(): vscode.ViewColumn {
         const active = vscode.window.activeTextEditor;
+
         if (!active) {
             return vscode.ViewColumn.One;
         }
 
-        if (!sideBySide) {
-            return active.viewColumn;
+        return active.viewColumn;
+    }
+
+    getViewColumnToRight() : vscode.ViewColumn {
+        const active = vscode.window.activeTextEditor;
+
+        if (!active) {
+            return vscode.ViewColumn.One;
         }
 
         switch (active.viewColumn) {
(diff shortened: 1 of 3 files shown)


johnfn:
Sorry, would you mind fixing those merge conflicts?

rebornix:
@johnfn I've updated the code to latest :)

rebornix:
BTW, this PR fixes #384 as well.

src/cmd_line/commands/file.ts
```diff
@@ -23,13 +29,48 @@ export class FileCommand extends node.CommandBase {
...
+        if (!active) {
+            return vscode.ViewColumn.One;
+        }
+
+        if (!sideBySide) {
+            return active.viewColumn;
+        }
+
+        switch (active.viewColumn) {
+            case vscode.ViewColumn.One:
```
johnfn: So what is this function doing? What is the sideBySide parameter? This switch appears to get the view column one to the right, but how come it doesn't handle the ViewColumn.Three case? 

It seems to me like you might want two separate functions. getActiveViewColumn() and getViewColumnToRight()

src/cmd_line/commands/file.ts
```diff
@@ -23,13 +29,48 @@ export class FileCommand extends node.CommandBase {
...
+    }
+
     execute(): void {
+        if (!this._arguments.name) {
+            // Open an empty file
+            if (this._arguments.position === FilePosition.CurrentWindow) {
+                vscode.commands.executeCommand("workbench.action.files.newUntitledFile");
+            }
+
+            vscode.commands.executeCommand("workbench.action.splitEditor").then(() => {
```
johnfn: Can you use async/await here?

src/cmd_line/subparsers/file.ts
```diff
@@ -12,6 +12,26 @@ export function parseEditFileCommandArgs(args: string): node.FileCommand {
...
+export function parseEditNewFileInNewWindowCommandArgs(args: string): node.FileCommand {
+    return new node.FileCommand({
+        position: node.FilePosition.NewWindow
+    });
+}
+
+export function parseEditFileInNewWindowCommandArgs(args: string): node.FileCommand {
+    let name = "";
+    if (args) {
+        var scanner = new Scanner(args);
```
johnfn: Can you use var, not let?

src/cmd_line/commands/file.ts
rebornix: This function is used to get the correct column of group editors when we open new files. If `sideBySide` is true, it means, we want to get the column number of next Group Editor instead of current active editor. But since we can only have 3 columns at most, we need to do some validation on the column number of current editor, if it's already 3, we have to open the file in current editor.

If `sideBySide` is false, what we want to do is get the column of current active editor.

This is what we do in VS Code or any file-related extension (DefaultConfig/UserConfig, Markdown Preview, reStructuredText Preview, etc). Since people might want to tweak it a little bit sometimes, we don't make it an API of any VS Code libraries. 

If we separate it to two functions, we may need move logics to the caller so I'd like to keep it this way.

src/cmd_line/commands/file.ts
rebornix: good catch!

src/cmd_line/subparsers/file.ts
rebornix: Already change it to `var`. That's just my personal preference of using `let` all the time as people always get confused with the scoping of JavaScript's `var`, but here it's totally the same so use `var` is perfect either.

rebornix:
Revise code per comments by @johnfn :)

src/cmd_line/commands/file.ts
johnfn: Ok, I understand now. 

> If we separate it to two functions, we may need move logics to the caller so I'd like to keep it this way.

I don't think so. The caller still needs to determine whether to set `sideBySide` to be `true` or `false`. That's the same as determining whether to call getActiveViewColumn() and getViewColumnToRight(). I suggest splitting the function since those feel to me like separate bits of functionality.

src/cmd_line/subparsers/file.ts
```diff
@@ -9,7 +9,7 @@ export function parseEditFileCommandArgs(args: string): node.FileCommand {
     }
 
     var scanner = new Scanner(args);
-    let name = scanner.nextWord();
+    var name = scanner.nextWord();
```
johnfn: Wait, why did you change _more_ `let`s to `var`s? That's the opposite of what I want! 😉

src/cmd_line/commands/file.ts
rebornix: Sure, will do that.

src/cmd_line/subparsers/file.ts
rebornix: Now I'm a little bit confused. Previously I was thinking your way of deciding when to use `let` or `var` is

* If it's inside a block (`{ }`, for loop), then we can use `let`
* If it's just in the very top level of the function, then we use `var`

Since JavaScript does hoisting and moves all declaration to the beginning this `var name` here is almost the same as `var name = ""` below in `parseEditFileInNewWindowCommandArgs` (https://github.com/VSCodeVim/Vim/pull/404/files#diff-16bc6b0f3306c3b8bdc5733a584411ffR27). You must have solid reason for this but I'm sorry that I'm not that clear now, do you mind pointing it out?

johnfn: My thoughts about variable types:

* Use `const` as much as possible
* When `const` is impossible (the variable gets changed), use `let`

Even though `let` and `var` are equivalent in this case, `let` is generally superior to `var` and there is no disadvantage to using `let; therefore we should always use `let`.

I'll whip up a style guide soon. :smile:

rebornix: oh your suggestion is using `let` whenever possible.

so previously when you said `use var, not let` in ` parseEditFileInNewWindowCommandArgs` is kind of typo? :)

johnfn: OOPS! Yes, that was a typo! Sorry!

rebornix: perfect. We are using `let` totally the same way. `let` always :) I'll give it a quick turnaround.

rebornix:
@johnfn and I misunderstood each other about the usage of `let` and `var` but actually we are saying the same thing: use `let` always possible.

Fixed all existing comments, I suppose it's a good time to merge.

johnfn:
@rebornix 2 things left:

* Split that function into 2
* Switch that one `var` to a `let` that I commented on (sorry!)

Then it should be good to go!

rebornix:
I must be drunk last time I checked in code, I thought I already fixed those issues ...

Follow your `Better Naming` suggestion and replace all `var` with `let` in my PR. Maybe we can create some gulp task `fix-var` similar to `fix-whitespace` :)

PR #413, opened 2016-07-08
Fix #397.

### LINKED ISSUE #397: `db` off-by-one
Typing `db` in normal mode behaves differently from vim.

Here I'm surrounding a character with underscores to indicate the location of the cursor. (E.g. `x_y_z` represents the cursor on character `y` in string `xyz`.) These examples use the string `abc 123` and start with the cursor on `1`:

In vscodevim (v0.0.28), typing db does this: 

`abc _1_23` -> `a_2_3`

In vim:

`abc _1_23` -> `_1_23`

rebornix:
Yay! We love PRs! 🎊

- [x] Commits are squashed 

While we run `b` movement, cursor start position is after stop position and we did some stop position adjustment, that's why we are not doing the `db` right.

Besides, the test cases covering `db` happen to be false positive. Run `db` inside `test test` always pass because both the first word and second word start and stop with `t`. It's a really interesting trap.

### DIFF
--- a/src/mode/modeHandler.ts
+++ b/src/mode/modeHandler.ts
@@ -736,6 +736,10 @@ export class ModeHandler implements vscode.Disposable {
         let recordedState = vimState.recordedState;
 
         if (recordedState.operator) {
+            if (start.compareTo(stop) > 0) {
+                [start, stop] = [stop, start];
+            }
+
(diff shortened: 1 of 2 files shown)


jpoon:
Thanks @rebornix

Issue #416, opened 2016-07-08
ctrl+f stopped working in vsCode 1.3.0 due to this extension. Works fine after removing this.


ashishgunjotikar:
Do you need help or have a question? Ask us on Slack (https://vscodevim-slackin.azurewebsites.net/) 👫. 
Found a bug? Please fill out the sections below 👍.

### Issue Summary

A summary of the issue -- what blew up? what major feature are we missing? 

### Steps to Reproduce

1. This is the first step
2. This is the second step, etc.

Any other info e.g. Why do you consider this to be a bug? What did you expect to happen instead?

### Technical details:

* VSCode Version:
* OS:

johnfn:
Kind of the point of Vim is that you now use the Vim way of doing things. 😉 For example, instead of ctrl-f you now use `/` and `?`, which are better in every regard. 😛 

Is there something you want to do with ctrl+f that can't be accomplished with `?` or `/`? If so, we should try to incorporate that functionality.

The actual fix here is not too hard; I just want to know what functionality you're missing from ? or /.

jpoon:
Also note that `ctrl+f` maps to scrolling the window downward in [Vim](http://vimdoc.sourceforge.net/htmldoc/scroll.html#CTRL-F).

Actually, @johnfn, just noticed the extension is doing the opposite `ctrl+f` => upward; `ctrl+b` => downward.

PR #418, opened 2016-07-08
Fix ctrl+b, ctrl+f

jpoon:
As noted in https://github.com/VSCodeVim/Vim/issues/416, ctrl+b should be upward, ctrl+f should be downward.

### DIFF
--- a/src/mode/modeHandler.ts
+++ b/src/mode/modeHandler.ts
@@ -930,5 +930,6 @@ export class ModeHandler implements vscode.Disposable {
     }
 
     dispose() {
+        // do nothing
     }
 }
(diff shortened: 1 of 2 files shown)


src/mode/modeHandler.ts
```diff
@@ -930,5 +930,6 @@ export class ModeHandler implements vscode.Disposable {
     }
 
     dispose() {
+        // do nothing
```
jpoon: Added a comment here so tslint wouldn't complain.

johnfn: My bad. Thanks!

PR #430, opened 2016-07-12
Ctrl+U and Ctrl+D

### LINKED ISSUE #214: I think CTRL-d should be half-page-down
> I think CTRL-d should be half-page-down

Original Vim do have a its binding.
Why not use `CTRL-d`?

rebornix:
Yay! We love PRs! 🎊

- [x] Commits are squashed 

Fix #214. Since currently we don't have VS Code Screen Line support, we simply use `20` as the half window size, but users can configure `vim.scroll` to override this setting the same as Vim.


johnfn:
hype

sachin21:
Thank you so much for implement!

PR #433, opened 2016-07-12
Fix #279.

### LINKED ISSUE #279: case insensitive/smart case search
When you search with / or ?, by default vim does case insensitive search. If you type an upper case letter in the search string, Vim switches to case sensitive. It's a clever system that we should emulate; currently / and ? are permanently case sensitive, which is annoying.

rebornix:
Yay! We love PRs! 🎊

- [x] Commits are squashed 

This bug fix is a good starter for adding Mode. This feature requires user config while now we can't change user config by API so the test cases are missing.


src/mode/modeHandler.ts
```diff
@@ -147,6 +149,7 @@ export class SearchState {
      */
     public searchDirection = 1;
 
+    public vimSettings: VimSettings;
```
johnfn: Can we have vimSettings be a property on VimState instead?

src/mode/modeHandler.ts
```diff
@@ -162,16 +165,28 @@ export class SearchState {
         // Calculate and store all matching ranges
         this.matchRanges = [];
 
+        let ignorecase = this.vimSettings.ignorecase;
+
+        if (ignorecase && this.vimSettings.smartcase && /[A-Z]/.test(search)) {
+            ignorecase = false;
+        }
+
+        const regex = new RegExp(search.replace(/[\-\[\]{}()*+?.,\\\^$|#\s]/g, "\\$&"), ignorecase ? 'gi' : 'g');
```
johnfn: Um, what does this do? Could you leave a comment?

src/mode/modeHandler.ts
```diff
@@ -484,6 +502,8 @@ export class ModeHandler implements vscode.Disposable {
         this._vimState.settings.useSolidBlockCursor = vscode.workspace.getConfiguration("vim")
             .get("useSolidBlockCursor", false);
         this._vimState.settings.scroll = vscode.workspace.getConfiguration("vim").get("scroll", 20) || 20;
+        this._vimState.settings.ignorecase = vscode.workspace.getConfiguration("vim").get("ignorecase", false) || false;
```
johnfn: Can we default smartcase to true? Even though Vim doesn't default it to true, I think we should, because it's better - in the same way that we show incremental search results, even though Vim doesn't by default.

src/mode/modeHandler.ts
```diff
@@ -220,9 +235,12 @@ export class SearchState {
         }
     }
 
-    constructor(direction: number, startPosition: Position, searchString = "") {
+    constructor(direction: number, startPosition: Position, vimSettings: VimSettings, searchString = "") {
         this.searchDirection = direction;
         this.searchCursorStartPosition = startPosition;
+        this.vimSettings = vimSettings;
+        // vimSettings must be put before assigining searchString 
+        // as we will run `_recalculateSearchRanges` immediately after the assignment.
```
johnfn: How do you feel about removing my setter for searchString, and instead use a method that takes both the search string and the settings object? It would eliminate this comment and bug.

src/mode/modeHandler.ts
rebornix: We have `vimSettings` in `VimState` already and it's initialized in `modeHandler`. Besides, we have `searchState` in VimState, the model between VimState and SearchState is Composition. So what we are doing is accessing outter object's property from inner object, right? My current solution is passing `vimSettings` alone to `SearchState` while initalization as I don't want to any other part of `VimState`.

So are you saying we should pass VimState directly to SearchState while initializing SearchState?

Another way to handle this situation is moving `SearchState` definition into `VimState` as we don't use it without the outter VimState. Making it a subclass can provide us the ability to access members of VimState. Any idea?

src/mode/modeHandler.ts
rebornix: oh, this is another good way of decoupling, I'll try.

rebornix:
@johnfn As we are not caching vim setting in SearchState, we have to pass the setting each time we invoke SearchState's member. The smell is still not good now, I suppose extracting VimSettings to a separate singleton will be better and will benefit later when we implementing vimrc loading.

Make `ignorecase` and `smartcase` as true as @johnfn suggested. Add comments for handling `ignorecase` and `smartcase` options.

src/mode/modeHandler.ts
johnfn: Ah, good point, we don't actually have VimState inside of SearchState.

johnfn:
I think the singleton VimSettings is the best idea; it's a giant blob of globals that will hopefully never change. 

Not necessary for this PR, however.

johnfn:
Actually I take back what I said before. 

Do you think you can make VimSettings into a singleton class? It's unfortunate how it ends up getting passed in everywhere. That would be annoying to fix in the future, we may as well nip it in the bud now.

rebornix:
I'll first write a real simple Singleton for VimSetting and unblock this PR.

Speaking of VimSetting, do you think detecting and read `.vimrc` is a valid feature for Vim users?

johnfn:
Thanks! 

I absolutely do think that .vimrc is something we should eventually support. Not a full vimrc, but some basic commands.

rebornix:
Since this feature relies on #508 and it's just 20 lines of code change with #508 . So I'd like to close this one and move features there.

PR #434, opened 2016-07-12
Fix spec for otherModesKeyBindings to match insert

sectioneight:
No existing issue that I can find, happy to open one if that helps.

### DIFF
--- a/package.json
+++ b/package.json
@@ -113,7 +113,7 @@
                     "description": "Keyboard layout to use to translated key presses."
                 },
                 "vim.otherModesKeyBindings": {
-                    "type": "object",
+                    "type": "array",
                     "description": "Keybinding overrides to use for normal mode."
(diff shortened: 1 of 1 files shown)


sectioneight:
image

sectioneight:
Note: This is my first time editing a VSCode plugin and I don't think I have my node environment set up properly so I haven't actually tested this, but looking at the code the "other" and "insert" shares the same class with only a different config key.

sectioneight:
Ok, I've tested this now, and set up remaps using the array syntax in the development plugin environment.

sectioneight:
I have no idea what to do about the test failures. They seem unrelated to this change.

johnfn:
Thanks for the PR, @sectioneight! Apologies for the broken tests; we're working on it... 😅

PR #444, opened 2016-07-14
Read TextEditor options from active editor

rebornix:
Yay! We love PRs! 🎊

- [x] Commits are squashed 

We were reading TextEditor options from configuration but there are some drawbacks:

1. We didn't monitor the configuration change (not sure if there is any event exposed)
2. VS Code puts a copy of the text editor related configuration in each Text Editor object and there is no bidirectional between them. If people change the configuration file, every copy of text editor option will be reloaded but if we change `vscode.window.activeTextEditor.options` through API on the fly, it's not synced back to `vscode.workspace.getConfiguration("editor")`. It does make sense because `vscode.workspace.getConfiguration("editor")` is likely representing the state of the configuration file.

Here I changed our logic to read from `vscode.window.activeTextEditor.options` always because we can't make sure we are the only one that modifies it. People might install other extensions that modify the options at runtime as well. `vscode.window.activeTextEditor.options` is our best choice now.

BTW, I'll raise the issue to @code team about the configuration sync issue because we are adding API support for other configuration part.

### DIFF
--- a/src/textEditor.ts
+++ b/src/textEditor.ts
@@ -134,8 +134,8 @@ export class TextEditor {
     }
 
     static setIndentationLevel(line: string, screenCharacters: number): string {
-        let tabSize = vscode.workspace.getConfiguration("editor").get<number>("tabSize");
-        let insertTabAsSpaces = vscode.workspace.getConfiguration("editor").get<boolean>("insertSpaces");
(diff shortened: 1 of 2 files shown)


johnfn:
LGTM :+1:

PR #445, opened 2016-07-14
Fix #426

### LINKED ISSUE #426: Ctrl-{ isn't binded to switch to normal mode
`Ctrl-{` in vim is another keybinding to get you into normal mode but it doesn't appear to work right now.

* VSCode Version: 1.3.0
* VsCodeVim Version: 0.1.1
* OS: Windows 10 Professional

arussellk:
This binds ctrl+[ to the escape command.

I removed an existing CommandOpenBracket. This was only used by the test and couldn't be accessed when actually using the editor.
I also had to add a "hack" alongside the `"<c-r>"` remap in modeHandler.ts.

We need something like this in the tests:
```
editor.pressKeys("ctrl+[");
```
so that we can ensure the `ctrl+` mappings from package.json are working properly.


arussellk:
I'll look into why this failed and fix it. I knew my first pull request would be a bit rocky...

rebornix:
LGTM. The test failure has nothing to do with your code, thanks for your contribution, it's really neat!

src/mode/modeHandler.ts
```diff
@@ -505,6 +505,7 @@ export class ModeHandler implements vscode.Disposable {
 
     async handleKeyEvent(key: string): Promise<Boolean> {
         if (key === "<c-r>") { key = "ctrl+r"; } // TODO - temporary hack for tests only!
+        if (key === "<ctrl-[>") { key = "<esc>"; } // TODO - temporary hack because package.json maps ctrl+[ to the escape command
```
rebornix: Actually I think we don't need this line and the corresponding test as well because `<ctrl-[>` will only be passed in by test code, right? VS Code will ensure it will translate `ctrl-[` to `vim_esc` command and we don't need test this feature in our code, what do you think?

arussellk: Correct. `<ctrl-[>` is only in test code.

I would _like_ to have a test to verify that pressing `ctrl+[` results in the right action, but perhaps a comment in the test to reference package.json is adequate. What do you think?

Something like this:
```
test("can be activated", async () => {
        let activationKeys = ['<esc>'];
        // ctrl+[ is mapped to the vim_esc command by package.json
...
    });
```

package.json
```diff
@@ -48,6 +48,11 @@
                 "when": "editorTextFocus"
             },
             {
+                "key": "ctrl+[",
+                "command": "extension.vim_esc",
```
johnfn: I would prefer to send the actual ctrl+[ key sequence here, so that you can test it later on. (It's also important for remapping!)

src/mode/modeHandler.ts
johnfn: Be sure to remove this line too.

package.json
arussellk: So, something like this?

package.json:
```
{
  "key": "ctrl+[",
  "command": "extension.ctrl+[",
  "when": "editorTextFocus"
}
```

actions.ts (I think that's the file name):
```
CommandControlOpenBracket extends CommandEscape {
  keys = ["ctrl+["]
}
```

I don't know Typescript, but it seems like I should be able to leverage the existing escape command.

johnfn: Yup, that looks good to me.

arussellk:
I have `ctrl+[` working the way I think you want it. Let me know if something else needs to change.

Sorry about all the commits - I don't know how to squash them with a remote involved. This seems to indicate that, as admins, you can fix it: https://help.github.com/articles/about-pull-request-merge-squashing/

jpoon:
Nope, no option to auto-squash, likely cause it's out of date with master. You should be able to squash them using `git rebase -i HEAD~5`.

johnfn:
This PR looks excellent. Let's squash and merge this bad boy!

johnfn:
Actually, I was able to squash it myself. How awesome is that? I guess I'm just cooler than you, @jpoon. 😉 

Thanks for the contribution, @arussellk!

jpoon:
Hmm. Did GH give you the option to squash after you push that big red merge button?

johnfn:
Yeah, I clicked the drop down and there was an option to squash.

PR #446, opened 2016-07-14
Fix #341 CJK Problem.

### LINKED ISSUE #341: issue with Chinese IME
in insert mode, I want to input `## 查询`, to input the 2 Chinese characters, I need to switch to Chinese IME and type `iaxp `, but after I type all of these, the actual result becomes `#ch查询`: `#` and ` ` were removed and `ch` (IME expands `i` to `ch`) was added.

I think the issue is similar with an issue of amVim (another vim plug-in of VSCode), and maybe a bug of  VSCode instead of the plugin: 
https://github.com/aioutecism/amVim-for-VSCode/issues/101

OS: Windows 10 Professional
VSCode: 1.2.1
IME: Microsoft Pinyin (Double Pinyin Mode)

rebornix:
Yay! We love PRs! 🎊

- [x] Commits are squashed 

Two impoartant things to note here:

1. We are handing `type` events **asynchronously**, in which case, `default:replacePreviousChar` might be executed before our customized `type` event. When we type in CJK characters, we are typing English characters firstly then replace them with the desired CJK character/s. If the `replacePreviousChar` happens before its corresponding `type` event, we are running into chaos.
2. We are maintaining `cursorPosition` in `vimState` and we will leverage this state to implement dot repeat somehow. However, when we perform `type` then `replacePreviousChar`, it will ruin our `cursorPosition` maintaining. So we should always update the cusor info after triggering `default:replacePreviousChar`.

After this fix, the CJK typing experience in Insert Mode is likely perfect (I'm not 100% sure whether there would be other issues as I just write Chinese). CJK handling in Normal Mode is still blocked by VS Code right now but the priority is really low, why would you bother tying CJK while using Vim commands?

### DIFF
--- a/extension.ts
+++ b/extension.ts
@@ -10,6 +10,7 @@ import * as vscode from 'vscode';
 import { showCmdLine } from './src/cmd_line/main';
 import { ModeHandler } from './src/mode/modeHandler';
 import { TaskQueue } from './src/taskQueue';
+import { Position } from './src/motion/position';
 
 let extensionContext: vscode.ExtensionContext;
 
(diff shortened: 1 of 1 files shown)


johnfn:
Nice!

PR #450, opened 2016-07-15
Implement additional text object commands

### LINKED ISSUE #449: Implement i< and i(, and i{ text objects
I am going to send a PR for these shortly.

sectioneight:
This allows for e.g. 'ci(' and 'da<' commands.

Fixes #449

### DIFF
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -2208,16 +2208,15 @@ abstract class MoveInsideCharacter extends BaseMovement {
   modes = [ModeName.Normal, ModeName.Visual, ModeName.VisualLine];
   protected charToMatch: string;
   protected includeSurrounding = false;
-  protected lineMatchOnly = false;
 
   public async execAction(position: Position, vimState: VimState): Promise<Position | IMovement> {
     const text = TextEditor.getLineAt(position).text;
 
     // First, search backwards for the opening character of the sequence
-    let startPos = position.lastIndexOf(this.charToMatch, this.lineMatchOnly);
+    let startPos = position.lastIndexOf(this.charToMatch);
     const startPlusOne = new Position(startPos.line, startPos.character + 1);
 
-    let endPos = PairMatcher.nextPairedChar(startPlusOne, this.charToMatch, false, this.lineMatchOnly);
+    let endPos = PairMatcher.nextPairedChar(startPlusOne, this.charToMatch, false);
 
     // Poor man's check for whether we found an opening character
     if (startPos === position && text[position.character] !== this.charToMatch) {
@@ -2320,36 +2319,6 @@ class MoveAClosingCaret extends MoveInsideCharacter {
   keys = ["a", ">"];
   charToMatch = "<";
   includeSurrounding = true;
-}
-
-@RegisterAction
-class MoveIDoubleQuote extends MoveInsideCharacter {
-  keys = ["i", "\""];
-  charToMatch = "\"";
-  lineMatchOnly = true;
-}
-
-@RegisterAction
-class MoveISingleQuote extends MoveInsideCharacter {
-  keys = ["i", "'"];
-  charToMatch = "'";
-  lineMatchOnly = true;
-}
-
-@RegisterAction
-class MoveADoubleQuote extends MoveInsideCharacter {
-  keys = ["a", "\""];
-  charToMatch = "\"";
-  includeSurrounding = true;
-  lineMatchOnly = true;
-}
-
-@RegisterAction
-class MoveASingleQuote extends MoveInsideCharacter {
-  keys = ["a", "'"];
-  charToMatch = "'";
-  includeSurrounding = true;
-  lineMatchOnly = true;
 }
 
 @RegisterAction
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -3,6 +3,7 @@ import { ModeName } from './../mode/mode';
 import { TextEditor } from './../textEditor';
 import { Register, RegisterMode } from './../register/register';
 import { Position } from './../motion/position';
+import { PairMatcher } from './../matching/matcher';
 import * as vscode from 'vscode';
 
 const controlKeys: string[] = [
@@ -2170,68 +2171,17 @@ class MoveToMatchingBracket extends BaseMovement {
   modes = [ModeName.Normal, ModeName.Visual, ModeName.VisualLine];
   keys = ["%"];
 
-  pairings: { [key: string]: { match: string, nextMatchIsForward: boolean }} = {
-    "(" : { match: ")",  nextMatchIsForward: true  },
-    "{" : { match: "}",  nextMatchIsForward: true  },
-    "[" : { match: "]",  nextMatchIsForward: true  },
-    ")" : { match: "(",  nextMatchIsForward: false },
-    "}" : { match: "{",  nextMatchIsForward: false },
-    "]" : { match: "[",  nextMatchIsForward: false },
-    // "'" : { match: "'",  direction:  0 },
-    // "\"": { match: "\"", direction:  0 },
-  };
-
-  nextBracket(position: Position, charToMatch: string, toFind: { match: string, nextMatchIsForward: boolean }, closed: boolean = true) {
-      /**
-       * We do a fairly basic implementation that only tracks the state of the type of
-       * character you're over and its pair (e.g. "[" and "]"). This is similar to
-       * what Vim does.
-       *
-       * It can't handle strings very well - something like "|( ')' )" where | is the
-       * cursor will cause it to go to the ) in the quotes, even though it should skip over it.
-       *
-       * PRs welcomed! (TODO)
-       * Though ideally VSC implements https://github.com/Microsoft/vscode/issues/7177
-       */
-
-      let stackHeight = closed ? 0 : 1;
-      let matchedPosition: Position | undefined = undefined;
-
-      for (const { char, pos } of Position.IterateDocument(position, toFind.nextMatchIsForward)) {
-        if (char === charToMatch) {
-          stackHeight++;
-        }
-
-        if (char === this.pairings[charToMatch].match) {
-          stackHeight--;
-        }
-
-        if (stackHeight === 0) {
-          matchedPosition = pos;
-
-          break;
-        }
-      }
-
-      if (matchedPosition) {
-        return matchedPosition;
-      }
-
-      // TODO(bell)
-      return position;
-  }
-
   public async execAction(position: Position, vimState: VimState): Promise<Position> {
     const text = TextEditor.getLineAt(position).text;
     const charToMatch = text[position.character];
-    const toFind = this.pairings[charToMatch];
+    const toFind = PairMatcher.pairings[charToMatch];
 
-    if (!toFind) {
+    if (!toFind || !toFind.matchesWithPercentageMotion) {
       // If we're not on a match, go right until we find a
       // pairable character or hit the end of line.
 
       for (let i = position.character; i < text.length; i++) {
-        if (this.pairings[text[i]]) {
+        if (PairMatcher.pairings[text[i]]) {
           return new Position(position.line, i);
         }
       }
@@ -2240,8 +2190,8 @@ class MoveToMatchingBracket extends BaseMovement {
       return position;
     }
 
-    return this.nextBracket(position, charToMatch, toFind, true);
-}
+    return PairMatcher.nextPairedChar(position, charToMatch, true);
+  }
 
   public async execActionForOperator(position: Position, vimState: VimState): Promise<Position> {
     const result = await this.execAction(position, vimState);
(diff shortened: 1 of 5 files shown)


sectioneight:
To be clear, any and all feedback, no matter how minor or how major, is appreciated. I'd like to continue to contribute to the project, and I'd like to write idiomatic TypeScript, but I'm nearly certain this code is not optimal as-is.

src/actions/actions.ts
```diff
@@ -2190,11 +2188,11 @@ class MoveToMatchingBracket extends BaseMovement {
       let matchedPosition: Position | undefined = undefined;
 
       for (const { char, pos } of Position.IterateDocument(position, toFind.nextMatchIsForward)) {
-        if (char === charToMatch) {
+        if (char === charToMatch && charToMatch !== toFind.match) {
           stackHeight++;
         }
 
-        if (char === this.pairings[charToMatch].match) {
+        if (char === toFind.match) {
```
sectioneight: All of the callers of this function (`this.nextBracket`) are already passing an object from `this.pairings[char]` as the `toFind`, so I think this is more correct. This also allows callers of this function to use characters that don't exist in `this.pairings`, such as `"` and `'`

src/actions/actions.ts
```diff
@@ -2233,7 +2236,7 @@ class MoveToMatchingBracket extends BaseMovement {
     }
 
     return this.nextBracket(position, charToMatch, toFind, true);
-}
+  }
```
sectioneight: Unrelated style cleanup, not sure if this belongs in this PR or not.

src/actions/actions.ts
```diff
@@ -2246,6 +2249,48 @@ class MoveToMatchingBracket extends BaseMovement {
...
+  modes = [ModeName.Normal, ModeName.Visual, ModeName.VisualLine];
+  protected charToMatch: string;
+
+  public async execAction(position: Position, vimState: VimState): Promise<Position | IMovement> {
+    const text = TextEditor.getLineAt(position).text;
+    const pairedCharacter = this.pairings[this.charToMatch]
+    const charToClose = pairedCharacter ? pairedCharacter.match : this.charToMatch;
+
+    // First, search backwards for the opening character of the sequence
+    let start = -1;
```
sectioneight: Is there a more idiomatic way to accomplish this without looping backwards? It feels weird using these -1 sentinel values, and using a plain for loop, but String.indexOf can only search forwards.

src/actions/actions.ts
```diff
@@ -2246,6 +2249,48 @@ class MoveToMatchingBracket extends BaseMovement {
...
+      }
+    }
+    const startPos = new Position(position.line, start + 1);
+    const endPos = this.nextBracket(startPos, this.charToMatch, { match: charToClose, nextMatchIsForward: true }, false);
+    if (start == -1 || endPos.character == start + 1 || start > position.character || endPos.character < position.character ) {
+      return position;
+    }
+
+    return {
+      start : startPos,
```
sectioneight: Not sure about style here. Was following some surrounding style, but VSCode seems to want to format this as `start: startPos` (no space before opening colon).

src/actions/actions.ts
```diff
@@ -2246,6 +2249,48 @@ class MoveToMatchingBracket extends BaseMovement {
...
+
+@RegisterAction
+class MoveIParentheses extends MoveInsideCharacter {
+  keys = ["i", "("];
+  charToMatch = "(";
+}
+
+@RegisterAction
+class MoveIDoubleQuote extends MoveInsideCharacter {
+  keys = ["i", "\""];
```
sectioneight: This feels kind of silly. Is there a way to set the `keys` property programatically and just register a single action/class that can handle all of these? (since there are more characters to implement, but wanted to start small)

test/mode/modeNormal.test.ts
```diff
@@ -220,6 +220,27 @@ suite("Mode Normal", () => {
     });
 
     newTest({
+      title: "Can handle 'ci(' on first parentheses",
```
sectioneight: Is there a way to run a single test case?

src/actions/actions.ts
```diff
@@ -2246,6 +2249,48 @@ class MoveToMatchingBracket extends BaseMovement {
...
+    // First, search backwards for the opening character of the sequence
+    let start = -1;
+    for (let i = position.character; i >= 0; i--) {
+      if (text[i] == this.charToMatch) {
+        start = i;
+        break;
+      }
+    }
+    const startPos = new Position(position.line, start + 1);
+    const endPos = this.nextBracket(startPos, this.charToMatch, { match: charToClose, nextMatchIsForward: true }, false);
```
sectioneight: This is where the behavior differs from vim. This should be calling some variant of `nextBracket` that only iterates positions up to the end of the current line, not the whole document.

src/actions/actions.ts
johnfn: No worries; I'm totally fine with you sneaking in stuff like this.

src/actions/actions.ts
johnfn: I am actually okay with this. I'd prefer to explicitly list out every key combination that exists, even though it's slightly less DRY than we'd otherwise want.

test/mode/modeNormal.test.ts
jpoon: I haven't found a way :(

src/actions/actions.ts
johnfn: Good catch! That must have been vestigial.

I would also suggest adding ' and " back to our `pairings` list, perhaps adding a flag so that we don't match them with %.

test/mode/modeNormal.test.ts
johnfn: There is a way!

newTestOnly

src/actions/actions.ts
johnfn: No hard rules; I just like to line up colons sometimes.

johnfn:
This is looking good so far, and I'm pretty hyped to have this functionality! 😀 

The only thing I'd suggest to change would be how you use inheritance to share the matching functionality with `MoveToMatchingBase`. You could just extract the function out into a new file and our class hierarchy would become 1 less deep, which would be nice.

src/actions/actions.ts
rebornix: This `pairings` bothers me a lot when I implement [MoveToUnclosedRoundBracketBackward](https://github.com/sectioneight/Vim/blob/5a50f2105119bd8cb7ba4a105560a76b2369bb59/src/actions/actions.ts#L2294). And the first hack I did is moving the bracket matching to a separate method called `nextBracket` in https://github.com/vscodevim/Vim/commit/b0a203830c7cda40100c678bd79c2e45c495af94 but now it seems `nextBracket` is no longer a good name as we might match quotes. Something called `nextPairedChar` looks better to me.

Besides, your PR is good enough for this fix but you may want to take a look at [all text objects commands](https://github.com/VSCodeVim/Vim/blob/master/ROADMAP.md#text-objects-only-in-visual-mode-or-after-an-operator), maybe we can make this `abstract class MoveToMatchingBase` more generic?

src/actions/actions.ts
sectioneight: Sounds good to me. Just didn't want to leave you with some massive copy-pasta mess if there was a better way.

src/actions/actions.ts
sectioneight: Yes, I plan on making the class generic enough so that it can handle `a<`, `i)` etc (and may add them to this PR now that I have some better confidence I'm going down the right track).

Will come up with some better naming, and add `'`/`"` into `pairings`

sectioneight:
Ok, I think I have most of what I need to get this PR to a merge-able state.

One open question that I left at the top: is there a way to return something to indicate that this **didn't** match, so that for example an unsuccessful `ci"` won't switch to insert mode?

sectioneight:
Needs a few more tests, but I think this is almost there.

johnfn:
Ahh. Your question about a failed ci" is actually something I haven't quite got around to yet. For now could you just mark it with a TODO and return a range of length 0? It's not perfect, I admit.

src/actions/actions.ts
```diff
@@ -2154,68 +2155,19 @@ class MoveToMatchingBracket extends BaseMovement {
...
-      // TODO(bell)
-      return position;
-  }
+  matcher = new PairMatcher();
 
   public async execAction(position: Position, vimState: VimState): Promise<Position> {
     const text = TextEditor.getLineAt(position).text;
     const charToMatch = text[position.character];
-    const toFind = this.pairings[charToMatch];
+    const toFind = PairMatcher.pairings[charToMatch];
```
johnfn: What do you think about making these static methods rather than instance methods? There isn't really any associated state we need to carry along, is there?

src/matching/matcher.ts
```diff
@@ -0,0 +1,78 @@
...
+export class PairMatcher {
+  static pairings: { [key: string]: { match: string, nextMatchIsForward: boolean, limited?: boolean }} = {
+    "(" : { match: ")",  nextMatchIsForward: true  },
+    "{" : { match: "}",  nextMatchIsForward: true  },
+    "[" : { match: "]",  nextMatchIsForward: true  },
+    ")" : { match: "(",  nextMatchIsForward: false },
+    "}" : { match: "{",  nextMatchIsForward: false },
+    "]" : { match: "[",  nextMatchIsForward: false },
+    // These characters can't be used for "%"-based matching, but are still
+    // useful for text objects. TODO limited is a terrible name.
```
johnfn: Haha, nice, I was going to suggest you change limited to something else also. 

matchesWithPercentageMotion might be ok.

src/actions/actions.ts
sectioneight: Yea, I originally had it being stateful (where you initialized a PairMatcher with the character), but that seemed wasteful as well. Wasn't sure about preferences for static/non-static. Will change to static.

src/matching/matcher.ts
```diff
@@ -0,0 +1,78 @@
...
+    "'" : { match: "'",  nextMatchIsForward: true, limited: true },
+    "\"": { match: "\"", nextMatchIsForward: true, limited: true },
+    "<" : { match: ">",  nextMatchIsForward: true, limited: true },
+  };
+
+  previousChar(position: Position, charToMatch: string, currentLineOnly: boolean = false): Position {
```
johnfn: This strikes me as something that belongs on Position. Maybe cal it lastIndexOf to indicate similarity with the string method of the same name.

src/actions/actions.ts
johnfn: Yeah. I prefer static to indicate stateless, though unfortunately there are places where I don't adhere to that.

src/matching/matcher.ts
sectioneight: Indeed. Done.

sectioneight:
By "return a range of length 0" do you mean an IMovement with the same start/stop? I tried that, but it still didn't "cancel" the insert/change.

johnfn:
Yeah, there's no way to cancel, so we just return an empty selection. It's not great- we'll have to fix it later.

sectioneight:
Yea, I understand that there isn't currently a cancel. Maybe there's a bug in my code, but when I return a `{start: pos, stop: pos}` (which I think is what you meant), a `ci"` still deletes the first character.

rebornix:
@sectioneight these commands are on my list before I worked on Window/Tab and CJK stuff. Good to see you can help on this, they are quite important.

sectioneight:
Disregard my previous comment, `ci"` without a match is not deleting any characters (only changing to insert mode, which is sub-par but not a dealbreaker)

sectioneight:
NOTE: There is one limitation to this that I just discovered.

If you have:

```
"foo"
```

and you have your cursor on the closing `"` mark, and type e.g. `ci"`, it won't recognize the match while vim will.

I don't consider it a blocker, but if there's an obvious way to fix this I'd be open to doing it now.

johnfn:
Ahh, yes. When I initially said that i" and i' would be harder, that was a bit of a concern of mine. I'm not convinced there's an easier way than parsing the entire line.

rebornix:
We are using `inclusive` parameter quite often in [Position](https://github.com/VSCodeVim/Vim/blob/master/src/motion/position.ts#L184), I suppose that's also what we need for this generic function.

I'd like to see this PR merged as I don't want to see it Open for too long. But just make sure we don't release this feature without a fix, people are likely to run `i"` while the cursor is on a `"`.

sectioneight:
To be clear, the limitation is only on the closing quote. ci" on an opening quote behaves as expected, as does `ci(` on the opening **or** closing parens, as well as inside. Because of the observed behavior, I'm pretty certain it's due to `nextPairedChar` not correctly handling the closing and opening chars being the same.

sectioneight:
My current thinking: handling quotes is going to take some more work, and I want to balance making forward progress with not introducing buggy code. I'm going to remove the single- and double-quote functionality from this PR and work out the proper algorithm to handle them (since it's really a different problem than handling the rest).

That way we can (hopefully) get this PR merged this weekend and I'll spend some more time working through the edge cases of properly handling quotes.

sectioneight:
Ok, PR updated, I think the functionality contained is complete (although I marked it as "partially supported" on the roadmap until at least one other person tries it out and says it works).

johnfn:
Awesome work! This looks good to me.

PR #453, opened 2016-07-16
Add support for ctrl+w in insert mode

### LINKED ISSUE #452: Add control-w insert mode delete backwards
I am implementing this currently since I use it so much for some reason.

sectioneight:
This deletes backwards, and is for some reason one of my most commonly
used keyboard shortcuts.

Fixes #452

### DIFF
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -361,6 +361,21 @@ class CommandCtrlOpenBracket extends CommandEsc {
 }
 
 @RegisterAction
+class CommandCtrlW extends BaseCommand {
+  modes = [ModeName.Insert];
+  keys = ["ctrl+w"];
+
+  public async exec(position: Position, vimState: VimState): Promise<VimState> {
+    const wordBegin = position.getWordLeft();
+    await TextEditor.delete(new vscode.Range(wordBegin, position));
+
+    vimState.cursorPosition = wordBegin;
+
+    return vimState;
+  }
+}
+
+@RegisterAction
 class CommandCtrlC extends CommandEsc {
   modes = [ModeName.Insert, ModeName.Visual, ModeName.VisualLine];
   keys = ["ctrl+c"];
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -353,7 +353,7 @@ Status | Command | Description
     | CTRL-X ...	                 |  complete the word before the cursor in various ways
     | BS or CTRL-H	                 |  delete the character before the cursor
 :white_check_mark:    | Del		                 |  delete the character under the cursor
-    | CTRL-W		                 |  delete word before the cursor
+:white_check_mark:    | CTRL-W		                 |  delete word before the cursor
     | CTRL-U		                 |  delete all entered characters in the current line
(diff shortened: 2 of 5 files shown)


src/actions/actions.ts
```diff
@@ -357,7 +357,22 @@ class CommandEsc extends BaseCommand {
 
 @RegisterAction
 class CommandCtrlOpenBracket extends CommandEsc {
-  keys = ["ctrl+["]
+  keys = ["ctrl+["];
```
sectioneight: Unrelated, but necessary to get `gulp tslint` passing locally.

sectioneight:
Note: I see that for some other control- commands, there are references in `extension.ts` and `package.json`, e.g.:

```json
"command": "extension.vim_ctrl+b",
```

I don't fully understand what these do. I tested this change interactively in the debug editor, so I know it works, I guess these give people some knobs to twiddle?

rebornix:
Here is the thing, we are listening to vscode's `type` [event](https://github.com/VSCodeVim/Vim/blob/master/extension.ts#L61) but it will not pass any to this event, which means, we can't catch `ctrl` keystroke anyway. The only suggested way to handle those keys are using key bindings.

So in our case, we translate `ctrl-w` to `switch_window`. But since I didn't know we can use `ctrl-w` to perform delete backwards so I made this key binding apply to all Vim Modes. While what you need to is simply add restriction to `switch_window` by checking `vim.mode`, see [samples here](https://github.com/VSCodeVim/Vim/blob/master/package.json#L68). Then finally add a new contribution in this package.json file for `ctrl+w` in Insert Mode.

sectioneight:
Makes sense. I'm a little confused by https://github.com/VSCodeVim/Vim/blob/master/extension.ts#L102

```
'rfbduc['.split('').forEach(key => {
  registerCommand(context, `extension.vim_ctrl+${key}`, () => handleKeyEvent(`ctrl+${
});
```

Which appears to register the various control keys, but notably `w` is not one of them. I'll update `package.json` now.

rebornix:
@sectioneight good catch man. Just remove `w` to the list. Both lines are added by me but I don't see any reason to keep `ctrl+w` specific.

rebornix:
@sectioneight miss an important link in previous comments, take a look at VS Code [key binding doc](https://code.visualstudio.com/docs/customization/keybindings#_accepted-keys) then you can have an idea which keys will be passed to `type` or not and which keys should be handled in contribution always.

sectioneight:
Thanks for the pointers! Can you take a look at the most recent diff? I **think** we're on the same page (making ctrl+w look like the rest of the ctrl+* keys).

rebornix:
LGTM 💯

jpoon:
One last thing, would you mind resolving the conflicts with `master`?

sectioneight:
@jpoon yup, updated, I think we both fixed the tslint error in different branches :)

jpoon:
@sectioneight :tada:

sectioneight:
Dope. Really happy with how easy it is to implement features. Looking
forward to helping getting the rest of my favorites in :)

PR #455, opened 2016-07-16
Add back missing control-c registration

sectioneight:
Looks like this was accidentally removed in #445

### DIFF
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -360,6 +360,7 @@ class CommandCtrlOpenBracket extends CommandEsc {
   keys = ["ctrl+["]
 }
 
+@RegisterAction
 class CommandCtrlC extends CommandEsc {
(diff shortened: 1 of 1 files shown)


rebornix:
LGTM. This might happen as Git Diff looks always weird when adding new actions.

sectioneight:
Yea, this is what the diff looks like for that PR:

image

You can see ControlC is now orphaned without a decorator (is that what they're called in TS?)

rebornix:
Yup, that's Decorator and @johnfn said we had some problems with it and that's why we have to put every command in one single file.

image

See image above, if we add a new command starting with `@RegisterAction`, Git will not treat the first line as `add`, instead it will make the `@RegisterAction` after your change as `add`, which is weird to me and sometimes I revert the last line by mistake.

jpoon:
Thanks @sectioneight for the fix

Issue #448, opened 2016-07-15
Tab Stops Ignored When Using Backspace 


cswartzvi:
### What did you do?

While editing a file with `editor.useTabStops` enabled the extension seems to cause `backspace` to ignore the tab stops.

### What did you expect to happen?

With the extension disabled, vscode follows the tab stops upon deleting whitespace with `backspace`:

![vscode-backspace](https://cloud.githubusercontent.com/assets/1111348/16860782/259c7308-4a09-11e6-94b5-0b9726f9f124.gif)

### What happened instead?

However with the extension enabled the tab stops seem to be ignored:

![vim-ext-backspace](https://cloud.githubusercontent.com/assets/1111348/16860785/2d2ce35a-4a09-11e6-9995-0c60b0183800.gif)

I have been bypassing this issue by reverting the `backspace` keybinding to `deleteLeft` (you could also use `shift+backspace` which is set to `deleteLeft` as well). Sadly, I recently discovered that this does not play nice with the vim search feature as you cannot delete characters in your search string with `deleteLeft`.

### Technical details:

* VSCode Version: 1.3.0
* VsCodeVim Version: 0.1.2
* OS: Fedora 23

PS. I've been VERY impressed with the speed of improvement these last few weeks. You're doing a fantastic job!

jpoon:
Whoa, what are you using the get those keystrokes outputted on the screen like that?

cswartzvi:
It's a nice little program called [screenkey](https://www.thregr.org/~wavexx/software/screenkey/).

rebornix:
Fixed and you can see the turnround in our new release.

cswartzvi:
Again I say, fantastic job!

PR #461, opened 2016-07-16
FixTabStops

rebornix:
Yay! We love PRs! 🎊

- [x] Commits are squashed 

Since this configuration is not accessible in `vscode 1.0.0`, we can't write test cases for it currently. So I just tested manually for now. A proper workaround is adding a test folder for testing (mentioned this before with @johnfn ) or waiting for VS Code's July release.

### DIFF
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -538,10 +538,20 @@ class CommandInsertInInsertMode extends BaseCommand {
           vimState.cursorStartPosition = position.getPreviousLineBegin().getLineEnd();
         }
       } else {
-        await TextEditor.delete(new vscode.Range(position, position.getLeft()));
+        let leftPosition = position.getLeft();
 
-        vimState.cursorPosition      = position.getLeft();
-        vimState.cursorStartPosition = position.getLeft();
+        if (position.getFirstLineNonBlankChar().character >= position.character) {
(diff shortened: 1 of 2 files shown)


rebornix:
Fixes #448.

johnfn:
Fix the merge conflicts and we should be good to go.

The more of the VSC insert algo that we replicate inside InsertInInsertMode, the sadder I get. Those API extensions can't come soon enough. 😛

rebornix:
Fix conflicts. It's almost the first thing I need to do after I wake up every day :)

I have a bad feeling about VSC stuff as well. While fixing bugs for CJK and a few other bugs, I found these bugs only exist in our code base but not am-vim or Alex's Vim prototype, and that's because we are handling the cursor position ourselves while they are delegating Insert/Delete in Insert Mode directly to VS Code. Not sure if there will be other hidden problems about the cursor problem any more but if it happens again, I'd like to further investigate.

johnfn:
Unfortunately the cursor tracking is necessary for a number of things we do that amVim and the other Vim prototypes don't do - the most notable of which is history tracking.

johnfn:
Sorry about making you fix all the conflicts though haha. If only this project wasn't so active... _sigh_... 😉

PR #466, opened 2016-07-17
Add support for failed motions

johnfn:
It's always a pleasant thing when a change that seems tricky ends up being very simple. 😃 

Fixes a bug I promised @sectioneight I'd get to where ci{ when you're not inside {} would cause you to switch into insert mode. Also fixes an old bug where dtx when x couldn't be found would delete a character.

### DIFF
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -67,7 +67,7 @@ export interface IMovement {
    * will not move the cursor. Furthermore, dfx won't delete *anything*, even though
    * deleting to the current character would generally delete 1 character.
    */
-  failed       : boolean;
+  failed?      : boolean;
 
   // It /so/ annoys me that I have to put this here.
   registerMode?: RegisterMode;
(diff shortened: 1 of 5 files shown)


src/actions/actions.ts
```diff
@@ -2219,10 +2242,18 @@ abstract class MoveInsideCharacter extends BaseMovement {
...
+
+      return {
+        start : position,
+        stop  : position,
+        failed: true
+      };
     }
-    // Make sure the start position is inside the selection
+
     if (startPos.isAfter(position) || endPos.isBefore(position)) {
```
sectioneight: We can remove this now, right?

johnfn: Oh yeah. Good catch.

johnfn: Fixed on master, thanks!

src/mode/modeHandler.ts
```diff
@@ -707,6 +707,10 @@ export class ModeHandler implements vscode.Disposable {
     if (result instanceof Position) {
       vimState.cursorPosition = result;
     } else if (isIMovement(result)) {
+      if (result.failed) {
+        vimState.recordedState = new RecordedState();
```
sectioneight: I believe this is the line that is causing #474, since the stacktrace we're getting is coming from `modeHandler.ts`:

```
    if (!recordedState.operator) {
      throw new Error("what in god's name");
    }
```

But more guessing than actually understanding what "recorded state" is.

johnfn: Yeah, sorry, RecorededState is an awful name! It used to be called ActionState, which was a little better, but still bad.

Issue #287, opened 2016-06-11
All text objects


johnfn:
- [x] `a(` and `i(`
- [x] `a"` and `i"`
- [x] `a'` and `i'`
- [x] `a<` and `i<`
- [x] `a[` and `i[`
- [x] `a{` and `i{`
- [x] `aw` and `iw`

Most of these are tedious but easy. The quoted ones are going to be a little harder, however.

johnfn:
Thanks to @sectioneight for allowing me to tick off a number of these boxes! 😃

sectioneight:
no prob! Iirc I didn't get square brackets yet. It should be a trivial change though.

rebornix:
With https://github.com/VSCodeVim/Vim/pull/385, we have `aW`, `iW` as well.

johnfn:
Already checked off 😉

johnfn:
Thanks to @sectioneight for nailing all of these!

hew:
Any plans to support `at`/`it` (tag)?

johnfn:
We actually do support at/it. We do have an open bug about not supporting multiline tags, however. #1108

PR #467, opened 2016-07-17
Implement square-bracket text object

sectioneight:
References #287

### DIFF
--- a/test/mode/modeNormal.test.ts
+++ b/test/mode/modeNormal.test.ts
@@ -268,6 +268,14 @@ suite("Mode Normal", () => {
     });
 
     newTest({
+      title: "Can handle 'ci[' spanning multiple lines",
+      start: ['one [', '|', ']'],
+      keysPressed: 'ci[',
+      end: ['one [|]'],
+      endMode: ModeName.Insert
+    });
+
+    newTest({
       title: "Can handle 'df'",
       start: ['aext tex|t'],
       keysPressed: '^dft',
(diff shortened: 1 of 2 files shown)


test/mode/modeNormal.test.ts
```diff
@@ -268,6 +268,14 @@ suite("Mode Normal", () => {
     });
 
     newTest({
+      title: "Can handle 'ci[' spanning multiple lines",
```
rebornix: consider adding cases for `a[`, `a]`, `i]`?

sectioneight: Yes, I wasn't sure about the level of testing desired, given that the underlying functionality (character matching) is already covered. Happy to add more tests (and that's generally my default), but wanted to be sensitive to test run times if that's a concern for this project.

rebornix: I agree, the underlying functionality is coverd and no idea to test it again. I'll suggest adding tests for corner cases then we can make sure that we don't break this feature if we want to modify the architecture somehow one day.

For example, per [Vim doc](http://vimdoc.sourceforge.net/htmldoc/motion.html#i[) for `a[`, `When used in Visual mode it is made characterwise.` while this case is linewise. Another case is `i"`, we need take escaped `"` into consideration.

Not sure did I cover this kind of corner cases perfectly when I implement `aw/iw`, but we can together push it to the limit, what do you think? I'm not saying your test cases really miss some corner cases(I'm just not sure).

rebornix:
LGTM

johnfn:
Thanks, @sectioneight!

PR #472, opened 2016-07-18
Fix handling of opener for nested text objects

sectioneight:
Prior to this commit, we would naively search for the previous e.g. "{"
character when you wanted to "ci{".

However, if you were inside two levels of nesting:

```
  one {
    two {}
    // cursor here
  }
```

Then the backwards search would take you to the wrong place.

Switching to the same logic for forward searching, we now handle this
case properly.


jpoon:
Awesome! Thanks @sectioneight :rocket:

PR #473, opened 2016-07-19
Fix regressions in text objects

sectioneight:
This PR fixes a regression in text objects when positioned over the final character.

It also fixes a regression with `<` objects, which is unfortunately not easy to test right now due to the way test cases are tokenized. I'd like to follow up and make it testable and add the appropriate tests, but I also want to leave master in a stable state.

I am starting to appreciate just how much vim does under the hood :)


johnfn:
> I am starting to appreciate just how much vim does under the hood :)

Welcome to the Bram Moolenaar fan club 😉

johnfn:
I understand the thing about not being easy to test < or >. Eventually I want to tokenize all characters just how Vim does, so there's no confusion. One day!

PR #475, opened 2016-07-19
CJK in all modes

rebornix:
Yay! We love PRs! 🎊

- [x] Commits are squashed 

With the newly API introduced in https://github.com/Microsoft/vscode/commit/70a083d54b2dd39220cd4d999dc1ee3c361d8936 , now we can handle CJK in all modes.

### DIFF
--- a/extension.ts
+++ b/extension.ts
@@ -55,8 +55,14 @@ export async function getAndUpdateModeHandler(): Promise<ModeHandler> {
   return handler;
 }
 
+class CompositionState {
+  public isInComposition: boolean = false;
+  public composingText: string = "";
+}
+
 export async function activate(context: vscode.ExtensionContext) {
   extensionContext = context;
+  let compositionState = new CompositionState();
 
   registerCommand(context, 'type', async (args) => {
     if (!vscode.window.activeTextEditor) {
@@ -66,7 +72,12 @@ export async function activate(context: vscode.ExtensionContext) {
     taskQueue.enqueueTask({
       promise: async () => {
         const mh = await getAndUpdateModeHandler();
-        await mh.handleKeyEvent(args.text);
+
+        if (compositionState.isInComposition) {
+          compositionState.composingText += args.text;
+        } else {
+          await mh.handleKeyEvent(args.text);
+        }
       },
       isRunning: false
     });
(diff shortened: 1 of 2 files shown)


src/mode/modeHandler.ts
```diff
@@ -129,6 +131,40 @@ export class VimSettings {
   useCtrlKeys     = false;
 }
 
+export class ComposeState {
```
johnfn: 1. I don't think this should be on VimState, it doesn't really have anything to do with it.
2. This seems like a rather heavyweight wrapper around what is really just a boolean and a string. Just define `class CompositionState { content = ""; isComposing = false; }` and then just use it.

rebornix: Then I think you might be okay with just using some closure to handle the sharing of `content` and `isComposing`, right?

rebornix:
Since we just need a bool and a string, we don't need a class to wrap them as well.

johnfn:
The reason I suggested class is because

1. They are linked enough to belong in a single structure. A class indicates that relation.
2. A `class` allows you to reset just by going `new MyThingy()`.

rebornix:
@johnfn yup now I got your idea.

jpoon:
I assume the next release of VSCode will have https://github.com/Microsoft/vscode/commit/70a083d54b2dd39220cd4d999dc1ee3c361d8936? Is there any concern for merging this prior to the API existing?

extension.ts
```diff
@@ -55,8 +55,19 @@ export async function getAndUpdateModeHandler(): Promise<ModeHandler> {
   return handler;
 }
 
+class CompositionState {
```
johnfn: You can just do

```
class CompositionState {
    public isInComposition = false;
    public composingText = "";
}
```

johnfn:
Yeah, I don't want to merge stuff that involves new features until those features land on VSCode proper (not insiders). 

LGTM to me however.

rebornix:
Did a tiny update per @johnfn 's comment.

@jpoon I didn't remove the logic in `replacePreviousChar`, which makes all these functions backwards compatible. So in this case, once we have this PR merged and released, people using Insider can benefit from the new feature and test in prod, if they find issues with this new feature, whether it's a VS Code bug or ours, we have chance to turnaround. This is also why we provide users Insiders.

But you guys do raise concerns about backwards compatibility, when we leverage new APIs from VS Code, we should make sure they won't break people who are using old version of VS Code. I'll see if I can get telemetry data of our Vim extension and see how people are using our extension and what's the percentile of each version.

jpoon:
@rebornix is the extension's package.json taken into account when doing updates?

```
    "engines": {
        "vscode": "^1.0.0"
    },
```

i wonder if we can use this such that people running old versions won't see the vscodevim extension updates.

rebornix:
@jpoon with commit https://github.com/Microsoft/vscode/commit/c06d57865dcb149c53d29ee24a19768458a69088 , Code will only install compitable version of extensions, but it's after and including Code 1.3.1. It means people using any version prior to 1.3.1 will always install the latest version of extensions. So it's doable, but not for all users.

However if we use `engine` to control what users will use, from the code level, `master` branch is no longer enough, right? If we add a new feature which is not related to Code's version, we should push the change to both old version users and latest version users. This will make the development and release complex.

But it's a common extension management issue, not just ours and I don't know if Marketplace guys have a plan for this or not. It's a good topic to sync with the team.

For now the easy way is making sure there is no regression when adding new features, what do you think?

jpoon:
I'm ok with leaving people behind and that `master` would refer to the current released version of VSCode. I'd like to avoid logic checking for specific versions of VSCode (ie. what happens with CSS + IE).

In any case, I'll merge this PR.

PR #482, opened 2016-07-20
Fix #338 - add gt, gT support

### LINKED ISSUE #338: gT gt
VSCode added tabs!!!

arussellk:
I did not write tests for this since the existing, closely-related actions `MoveToRightPane` and `MoveToLeftPane` did not have tests.

Let me know if anything looks wrong.

### DIFF
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -4,6 +4,7 @@ import { TextEditor } from './../textEditor';
 import { Register, RegisterMode } from './../register/register';
 import { Position } from './../motion/position';
 import { PairMatcher } from './../matching/matcher';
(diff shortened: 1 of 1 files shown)


rebornix:
I suppose this is about tabs, not editor groups, right? We are mapping editor groups to Vim's Window concept. For moving Tabs, you need to use other API, take a look at https://github.com/VSCodeVim/Vim/blob/master/src/cmd_line/commands/tab.ts . We've already had commands for them, and the logic is similar.

BTW, no worries about test cases of Tab, it's not testable with current test infra.

arussellk:
@rebornix I rewrote my code to use the existing tab stuff. Check it out.

johnfn:
LGTM! Thanks for the PR, @arussellk!

PR #483, opened 2016-07-20
Implement quoted text objects

### LINKED ISSUE #471: ci' does not delete between single quotes and enter insert mode
### What did you do?

1. Placed the cursor on some text inside two single quotes (ie. 'testing test')
2. Pressed: ci'

1. Delete what's inside the single quotes
2. Enter insert mode

### What happened instead?

Nothing

### Technical details:

* VSCode Version: 1.3.1
* VsCodeVim Version: 0.1.2
* OS: OSX

sectioneight:
Fixes #471, #287

### DIFF
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -4,6 +4,7 @@ import { TextEditor } from './../textEditor';
 import { Register, RegisterMode } from './../register/register';
 import { Position } from './../motion/position';
 import { PairMatcher } from './../matching/matcher';
+import { QuoteMatcher } from './../matching/quoteMatcher';
 import { Tab, TabCommand } from './../cmd_line/commands/tab';
 import * as vscode from 'vscode';
 
@@ -2518,6 +2519,89 @@ class MoveASquareBracket extends MoveInsideCharacter {
 class MoveAClosingSquareBracket extends MoveInsideCharacter {
   keys = ["a", "]"];
   charToMatch = "[";
+  includeSurrounding = true;
+}
+
+abstract class MoveQuoteMatch extends BaseMovement {
+  modes = [ModeName.Normal, ModeName.Visual, ModeName.VisualLine];
+  protected charToMatch: string;
+  protected includeSurrounding = false;
+
+  public async execAction(position: Position, vimState: VimState): Promise<IMovement> {
+    const text = TextEditor.getLineAt(position).text;
+    const quoteMatcher = new QuoteMatcher(this.charToMatch, text);
+    const start = quoteMatcher.findOpening(position.character);
+    const end = quoteMatcher.findClosing(start + 1);
+
+    if (start === -1 || end === -1 || end === start || end < position.character) {
+      return {
+        start: position,
+        stop: position,
+        failed: true
+      };
+    }
+
+    let startPos = new Position(position.line, start);
+    let endPos = new Position(position.line, end);
+    if (!this.includeSurrounding) {
+      startPos = startPos.getRight();
+      endPos = endPos.getLeft();
+    }
+
+    return {
+      start: startPos,
+      stop: endPos
+    };
+  }
+
+  public async execActionForOperator(position: Position, vimState: VimState): Promise<IMovement> {
+    const res = await this.execAction(position, vimState);
+
+    res.stop = res.stop.getRight();
+
+    return res;
+  }
+}
+
+@RegisterAction
+class MoveInsideSingleQuotes extends MoveQuoteMatch {
+  keys = ["i", "'"];
+  charToMatch = "'";
+  includeSurrounding = false;
+}
+
+@RegisterAction
+class MoveASingleQuotes extends MoveQuoteMatch {
+  keys = ["a", "'"];
+  charToMatch = "'";
+  includeSurrounding = true;
+}
+
+@RegisterAction
+class MoveInsideDoubleQuotes extends MoveQuoteMatch {
+  keys = ["i", "\""];
+  charToMatch = "\"";
+  includeSurrounding = false;
+}
+
+@RegisterAction
+class MoveADoubleQuotes extends MoveQuoteMatch {
+  keys = ["a", "\""];
+  charToMatch = "\"";
+  includeSurrounding = true;
+}
+
+@RegisterAction
+class MoveInsideBacktick extends MoveQuoteMatch {
+  keys = ["i", "`"];
+  charToMatch = "`";
+  includeSurrounding = false;
+}
+
+@RegisterAction
+class MoveABacktick extends MoveQuoteMatch {
+  keys = ["a", "`"];
+  charToMatch = "`";
   includeSurrounding = true;
 }
 
--- a/src/matching/quoteMatcher.ts
+++ b/src/matching/quoteMatcher.ts
@@ -1,0 +1,53 @@
+enum QuoteMatch {
+  None,
+  Opening,
+  Closing
+}
+
+/**
+ * QuoteMatcher matches quoted strings, respecting escaped quotes (\") and friends
+ */
+export class QuoteMatcher {
+  static escapeChar = "\\";
+
+  private quoteMap: QuoteMatch[] = [];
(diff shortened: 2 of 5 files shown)


src/matching/quoteMatcher.ts
```diff
@@ -0,0 +1,36 @@
...
+    for (let i = start - 1; i >= 0; i--) {
+      const currentChar = corpus[i];
+      if (currentChar === char && corpus[i - 1] !== this.escapeChar) {
+        return i;
+      }
+    }
+
+    // Didn't find one behind us, the string may start ahead of us
+    for (let i = start; i < corpus.length; i++) {
+      const currentChar = corpus[i];
```
sectioneight: I feel a bit bad (but not too bad) about copy-pasting the body of these loops. Really what I want is to have some thing that takes an iterator and returns Either<position **or** notfound> and then pass that iterator to the loops. But, I'm a total TypeScript noob and I didn't see an obvious way to do it.

src/matching/quoteMatcher.ts
```diff
@@ -0,0 +1,36 @@
+/**
+ * QuoteMatcher matches quoted strings, respecting escaped quotes (\") and frieds
+ */
+export class QuoteMatcher {
+  static escapeChar = "\\";
+
+  static findOpening(char: string, corpus: string, start: number): number {
+    // first, search backwards to see if we could be inside a quote
```
rebornix: This comment is somewhat inconsistent with the code below. You are going to check whether we are inside a quote, the code is seaching for a quote backwards. But what if there are two paired quotes on the left side? Finding the closing one doesn't mean we are inside a quote, right?

Did I miss something or is it a Vim design?

sectioneight: It's a bit tricky, and I probably didn't comment enough. From my observed behavior of Vim, this algorithm works as expected. First, you check if the opening quote is behind you (the first iteration of this loop). If you find a valid "opening" quote, stop processing and look for the closing quote after the opening position.

If this loop **fails**, then search forwards to see if the quoted string starts **after** the current position.

sectioneight: And yes, it doesn't handle the case of:

```
"foo" <cursor> "bar"
```

where a `ci"` will leave you with

```
"foo"<cursor>"bar"
```

But, this is also how vim treats this.

rebornix: Great! This doesn't surprise me as Vim is not always consistent with its documentation. As it's kind of `incomplete`(Vim's behavior), do you mind adding this `"foo"‹cursor›"bar"` case to test suite then other people won't see it as our code issue?

sectioneight: Good call! Doing so now.

src/matching/quoteMatcher.ts
```diff
@@ -0,0 +1,36 @@
...
+      const currentChar = corpus[i];
+      if (currentChar === char && corpus[i - 1] !== this.escapeChar) {
+        return i;
+      }
+    }
+
+    // Didn't find one behind us, the string may start ahead of us
+    for (let i = start; i < corpus.length; i++) {
+      const currentChar = corpus[i];
+      if (currentChar === char && corpus[i - 1] !== this.escapeChar) {
```
rebornix: I'm a little worried about the escape char, this can't handle `\\"`, right?

sectioneight: Correct, that is a limitation.

I looked at the algorithm used by other vim emulation stacks and this seemed to be the best one. I think we could handle that case by having another look-behind for another escape character? Or are there other pathological cases we'd still be missing?

rebornix: We have handled escape char while parsing subsitute commands https://github.com/VSCodeVim/Vim/blob/master/src/cmd_line/subparsers/substitute.ts . The logic is reading char one by one and handling every `\`. I suppose we can do the same thing here?

For now I don't see any other case but I'll investigate.

rebornix:
Everything looks good to me except the `\` escaping, it won't be too complex to handle as we already did that in substitute parsing :)

sectioneight:
@rebornix updated with handling of escaped backslashes

rebornix:
@sectioneight I'm really sorry that I didn't think carefully about the escaping before I made suggestions. Checking double `\` before `"` is somewhat not perfect, for example `"abc\\\"def"` will pass the validation

```
currentChar === char &&
(corpus[i - 1] !== this.escapeChar || corpus[i - 2] === this.escapeChar)
```

but it's not a true closing `"`. Now it turns out to be a text lexing problem, remind me of old school days...

sectioneight:
Ah, right, of course it's not that easy :)

What do you guys think? Should we parse the whole line or stick with the
simpler (with known limitations) version?

rebornix:
I have a dirty but quick way to handle this, while iterating the text, skip all escaped chars then check whether it's a `"`

```
let i = 0;
while (i < text.length) {
  if (text[i] === "\") {
    i = i+2;
  }

  if (text[i] == "\"") {
    return i;
  }
}
```

I didn't handle corner cases carefully but I think you can get the idea.

Besides, I don't have a strong opinion about postpoing the escaping handling or not, if we don't have a perfect solution right now, we can fix it later. It's up to you and @johnfn .

sectioneight:
OK, just pushed a new commit that pre-parses the line and handles your edge case properly. I'm not sure about performance implications, however -- is that something to be concerned about, or is that prematurely optimizing?

sectioneight:
This one is ready to go

src/matching/quoteMatcher.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  private char: string;
+  private start: number;
+  private corpus: string;
+  private quoteMap: boolean[] = [];
+
+  constructor(char: string, corpus: string, start: number) {
+    this.char = char;
+    this.start = start;
+    this.corpus = corpus;
```
rebornix: seems `this.corpus` is never used?

src/matching/quoteMatcher.ts
```diff
@@ -0,0 +1,44 @@
...
+      if (corpus[i] === QuoteMatcher.escapeChar) {
+        i += 1;
+        continue;
+      }
+      this.quoteMap[i] = corpus[i] === this.char;
+    }
+  }
+
+  findOpening(): number {
+    // First, search backwards to see if we could be inside a quote
```
rebornix: If we pass `start` to this method like `findClosing`, we don't need to store `start` as well, make sense?

rebornix:
only have some tiny suggestions. The logic and code look good to me so I'm willing to see it merged.

sectioneight:
@rebornix great feedback, updated.

src/actions/actions.ts
```diff
@@ -2399,6 +2400,81 @@ class MoveAClosingSquareBracket extends MoveInsideCharacter {
...
+
+    if (start === -1 || end === -1 || end === start) {
+      return {
+        start: position,
+        stop: position,
+        failed: true
+      };
+    }
+    return {
+      start: new Position(position.line, start + (this.includeSurrounding ? 0 : 1)),
```
johnfn: Can you use position objects and getRight()/getLeft() here?

sectioneight: For sure.

johnfn:
The one big thing that we're missing here is `i"` in a situation like this:

`"one " two |"three"` (where | is the start of the block cursor)

You currently select the inner " two " rather than the "three" to the right. I'm pretty sure that Vim actually determines which of the two potential quote sections is _inside_ the quotes rather than outside and selects that one. Could you add in that case?

This should be good to merge with that.

sectioneight:
Sounds good. Master tests are currently broken, I'm going to wait until that's resolved to push anything else here.

sectioneight:
@johnfn updated. Note that there are test cases that currently fail, as master is currently red.

johnfn:
Perfect! Thanks for the awesome work, @sectioneight!

PR #478, opened 2016-07-19
Set correct cursor and selection after code format.

rebornix:
Yay! We love PRs! 🎊

Please include a description of your change & check your PR against this list, thanks:

- [x] Commit message has a short title & issue references
- [x] Commits are squashed 
- [x] It builds and tests pass (e.g `gulp tslint`)

After code format, the cursor should be set to the first character of the first line of original selection.

PR #486, opened 2016-07-20
Screen lines and characters.

rebornix:
Yay! We love PRs! 🎊

- [x] Commits are squashed 

We are adding APIs for screen lines/characters in Insider and this PR is integrating those APIs for following commands

- [x] `g0`
- [x] `g^`
- [x] :1234: `g$`
- [x] `gm`
- [x] :1234: `gk`
- [x] :1234: `gj`
- [x] :1234: `H`
- [x] `M`
- [x] :1234: `L`

As @johnfn mentioned while I did #478, the way we implment Movement in VSCodeVim is calculating the target Position or Selection and update the View in a central place (modelHandler). We are likely doing this from the very beginning.

While we are exposing commands in VS Code, we try to avoid exposing View Model information as they are way too complex and always require additional work. Take `scroll to next screen line` as example, we want VSCodeVim just provides the Position information and handle the movment ourselves, but for other extension authors or just users who want to customize key bindings, they have to leverage other APIs to move the cursor to target position.

So in this PR, I'm calling VS Code API to do the real movment or selection and return the correct Position/Movement information after the command is executed.

### DIFF
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -1715,6 +1715,12 @@ class MoveScreenLineEnd extends MoveByScreenLine {
 }
 
 @RegisterAction
+class MoveScreenLienEndNonBlank extends MoveByScreenLine {
+  keys = ["g", "_"];
+  movementType = "wrappedLineLastNonWhitespaceCharacter";
+}
+
+@RegisterAction
 class MoveScreenLineCenter extends MoveByScreenLine {
   keys = ["g", "m"];
   movementType = "wrappedLineColumnCenter";
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -1664,7 +1664,7 @@ abstract class MoveByScreenLine extends BaseMovement {
   public async execAction(position: Position, vimState: VimState): Promise<Position | IMovement> {
     await vscode.commands.executeCommand("cursorMove", {
       to: this.movementType,
-      inSelectionMode: vimState.currentMode !== ModeName.Normal,
+      select: vimState.currentMode !== ModeName.Normal,
       noOfLines: this.noOfLines
     });
 
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -1652,6 +1652,114 @@ class MoveLineBegin extends BaseMovement {
 }
 
 @RegisterAction
+class MoveScreenLineBegin extends BaseMovement {
+  modes = [ModeName.Normal, ModeName.Visual, ModeName.VisualLine];
+  keys = ["g", "0"];
+  viewPosition = "wrappedLineStart";
+  /**
+   * This paramter will be consumed only when `to` is `lineUp` or `lineDown`.
+   * For other screen line movements, we are always operating on the same screen line.
+   * So we make its default value as 0.
+   */
+  noOfLines = 0;
+
+  public async execAction(position: Position, vimState: VimState): Promise<Position | IMovement> {
+    await vscode.commands.executeCommand("cursorMove", {
+      to: this.viewPosition,
+      inSelectionMode: vimState.currentMode !== ModeName.Normal,
+      noOfLines: this.noOfLines
+    });
+
+    if (vimState.currentMode === ModeName.Normal) {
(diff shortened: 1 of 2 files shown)


rebornix:
As we already hold on new features on Code for July release, I postpone `zh`, `zH`, etc commands to later weeks and it's tracked by #511. So I removed `WIP` prefix and it's ready for review @johnfn @jpoon .

src/actions/actions.ts
```diff
@@ -1652,6 +1652,114 @@ class MoveLineBegin extends BaseMovement {
...
+    await vscode.commands.executeCommand("cursorMove", {
+      to: this.viewPosition,
+      inSelectionMode: vimState.currentMode !== ModeName.Normal,
+      noOfLines: this.noOfLines
+    });
+
+    if (vimState.currentMode === ModeName.Normal) {
+      return Position.FromVSCodePosition(vscode.window.activeTextEditor.selection.active);
+    } else {
+      /**
```
johnfn: This is a big bummer; it will slow down macros and . considerably. 

I know that it's not your fault at all, but any additional griping you can do in the direction of VSCode would be helpful 😄

rebornix: I agree it's a bummer when users use this command quite often. Do we have any benchmark (or you have ideas about how) about commands to help measure how bad it is?

If it turns out to be unacceptablely slow, I'd like to push it hard from Code side.

johnfn: I'm sure we can make one 😉 

Upon reflection, a mild slowdown here is acceptable, because not too many people use these commands (they are a bit esoteric). At this point, I think that a history API is priority number 1, and that we can safely bump the priority of getting actual positions down.

Right now the thing I would kill for is the history API. We are really suffering a lot because of that.

src/actions/actions.ts
```diff
@@ -1652,6 +1652,114 @@ class MoveLineBegin extends BaseMovement {
 }
 
 @RegisterAction
+class MoveScreenLineBegin extends BaseMovement {
```
johnfn: In accordance with #519 (sorry for referring to an issue that was clearly made after your PR!), could you make this into an abstract base class?

src/actions/actions.ts
```diff
@@ -1652,6 +1652,114 @@ class MoveLineBegin extends BaseMovement {
...
+  modes = [ModeName.Normal, ModeName.Visual, ModeName.VisualLine];
+  keys = ["g", "0"];
+  viewPosition = "wrappedLineStart";
```
johnfn: Can we rename this to `movementType`?

src/actions/actions.ts
```diff
@@ -1652,6 +1652,114 @@ class MoveLineBegin extends BaseMovement {
...
+  /**
+   * This paramter will be consumed only when `to` is `lineUp` or `lineDown`.
```
johnfn: Typo: "parameter"

Also I think "This parameter is used only when `to` is `lineUp` or `lineDown`." is a sufficient comment here.

src/actions/actions.ts
rebornix: Let me make History the 1st feature request for next sprint.

johnfn:
When you fix up my minor comments this LGTM.

I would like to hold off until vscode 1.4 ships with the new features before merging this however, as I suspect a lot of our users are not using the insiders version, and we want to support them as well.

src/actions/actions.ts
rebornix: Will do. I'm naming it as `viewPosition` as what we do in Code, still remember the `viewColumn` naming I used before :)

rebornix:
@johnfn fixed per your comments.

Actually you don't need to wait for 1.4 as these commands have no side effect in any version below, that's saying even if you are using an old Code, when you run these new Code commands, Code will just eat them and emit nothing.

This is also a good way to test our new features unlocked by Code. People who use Insider can try these new features first and may report issues if there is any bug. The key point is, making sure newly added features have no side effect and are backwards compatible.

johnfn:
Thanks @rebornix!

PR #490, opened 2016-07-21
Fix % movement when not on opening character

### LINKED ISSUE #480: % should move to end of matching set, not beginning
Typing `%` moves the cursor to the opening matching set, not the close.

Where `_x_yz` represents the cursor on `x` in `xyz`, and → means pressing `%`.

Vim:
* `_t_his is (a {inner} test)` → `this is (a {inner} test_)_`
* `this is (_a_ {inner} test)` → `this is (a {inner_}_ test)`

VSCodeVim:
* `_t_his is (a {inner} test)` → `this is _(_a {inner} test)`
* `this is (_a_ {inner} test)` → `this is (a _{_inner} test)`

### Technical details:

* VSCode Version: 1.3.1
* VsCodeVim Version: 0.1.3
* OS: Windows 10 64 bit

sectioneight:
Vim's behavior is to move to the closing match, if available. Prior to
this commit, we were moving to the opening char.

Fixes #480

### DIFF
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -2224,7 +2224,9 @@ class MoveToMatchingBracket extends BaseMovement {
 
       for (let i = position.character; i < text.length; i++) {
         if (PairMatcher.pairings[text[i]]) {
-          return new Position(position.line, i);
+          // We found an opening char, now move to the matching closing char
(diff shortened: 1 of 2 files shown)


jpoon:
Awesome possum @sectioneight 👯

PR #493, opened 2016-07-21
CommandFold should be available in Normal mode

aminroosta:
This PR changes CommandFold to be used in Normal mode, just like all other fold commands.

Note that i am explicitly setting `vimState.currentMode = ModeName.Normal`, the reason is that *vscode* sometimes changes to Visual mode when folding the block:

Fore example this works fine.
screen shot 2016-07-21 at 4 15 59 pm
screen shot 2016-07-21 at 4 16 13 pm

But when the cursor is on the beginning line it goes to Visual mode!

screen shot 2016-07-21 at 4 16 30 pm
screen shot 2016-07-21 at 4 16 42 pm

### DIFF
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -1094,12 +1094,12 @@ class CommandDot extends BaseCommand {
 
 @RegisterAction
 class CommandFold extends BaseCommand {
-  modes = [ModeName.Visual, ModeName.VisualLine];
+  modes = [ModeName.Normal];
   keys = ["z", "c"];
 
   public async exec(position: Position, vimState: VimState): Promise<VimState> {
     await vscode.commands.executeCommand("editor.fold");
-
+    vimState.currentMode = ModeName.Normal;
     return vimState;
   }
 }
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -1094,7 +1094,7 @@ class CommandDot extends BaseCommand {
 
 @RegisterAction
 class CommandFold extends BaseCommand {
-  modes = [ModeName.Normal];
+  modes = [ModeName.Normal, ModeName.Visual, ModeName.VisualLine];
   keys = ["z", "c"];
 
   public async exec(position: Position, vimState: VimState): Promise<VimState> {
@@ -1121,12 +1121,12 @@ class CommandCenterScroll extends BaseCommand {
 
 @RegisterAction
 class CommandUnfold extends BaseCommand {
-  modes = [ModeName.Normal];
+  modes = [ModeName.Normal, ModeName.Visual, ModeName.VisualLine];
   keys = ["z", "o"];
 
   public async exec(position: Position, vimState: VimState): Promise<VimState> {
     await vscode.commands.executeCommand("editor.unfold");
-
+    vimState.currentMode = ModeName.Normal;
     return vimState;
   }
 }


src/actions/actions.ts
```diff
@@ -1094,12 +1094,12 @@ class CommandDot extends BaseCommand {
 
 @RegisterAction
 class CommandFold extends BaseCommand {
-  modes = [ModeName.Visual, ModeName.VisualLine];
+  modes = [ModeName.Normal];
```
rebornix: I suppose `zc` also works in Visual/Visual Line mode, so we just need to add Normal mode to the list.

rebornix:
BTW, per Vim's document, `zc` is used for fold close while `zf` is used for folding. So maybe `CommandFold` is not correct?

johnfn:
But why did you remove Visual and VisualLine from the list?

aminroosta:
@johnfn You are right it seems vim also supports `zc` and `zo` in visual mode.
Also i see that vim switches to Normal mode after these commands (`zc` & `zo`).
I will update the PR accordingly.

@rebornix I think the current behavior of `zc` is correct, as i understand `zf` is used to create manual folds.

I'm also interested in implementing `za` to toggle folds.
But it seems vscode has no such API to access fold regions.
Do you have any suggestions on if or how `za` could possibly be implemented?

@rebornix i think this is another use case for having Apis for folded regions. [related vscode discussion](https://github.com/Microsoft/vscode/pull/2971#issuecomment-234172524).

johnfn:
Awesome, LGTM.

re: `za`, it's probable that it's simply not possible. Check with @rebornix, and if you can't find it, file an issue on the vscode Github. We have an issue open already for missing Vim features: https://github.com/Microsoft/vscode/issues/8997

rebornix:
@aminroosta I'll take a deep look into Vim folding doc and see what kind of API it requires then I'd like to create a separate issue for it.

aminroosta:
@rebornix Do you think if it's a bad idea to figure out folded origins by using #486 `MoveDownByScreenLine` and then recording cursor position and then going back?!

@sandy081
This is one example why i think https://github.com/Microsoft/vscode/issues/2771 Does not really address what vim emulators need. If we had information about folded regions we could easily implement all `cursorMove` commands.

@rebornix 
I want to implement `za` and the only workaround seems using `cursorMove: up` api and then recording cursor position and going back. do you think it worth it?

rebornix:
@aminroosta Sandeep's solution is reasonable as we are always trying to avoid exposing View Model to users. The reason that we want to know if it's a folded area in the case of `za`, we need to determine whehter we should call `editor.fold` or `editor.unfold`, right? It looks like a shortcut somehow. Do you think it would make sense that Code support a new command shortcut, maybe called `toggleFold`?

It's an open question. If there are more commands relying Fold Area information, I think providing this info directly will make everything simpler.

rebornix:
@aminroosta lastly, I like the idea of moving up then down to detect whether it's inside a folded area. As this operation is real fast and it follows a heavy Fold/UnFold command, I think it's a good workaround.

aminroosta:
@rebornix I think having information on folding areas gives vim extension writers more power and control on how to move the cursor. I can assure you that there will be other cases where this will bee needed 😉 

> I like the idea of moving up then down to detect whether it's inside a folded area.  

When do you think #486  will be ready, i will add support for `za` after it gets merged.

aminroosta:
> we are always trying to avoid exposing View Model to users.

I think in this case vscode should expose this information, to me it's one of the basic informations you should be able to ask the TextEditor, just like cursor postion or selection stop and end.

aminroosta:
screen shot 2016-07-25 at 1 44 14 pm
screen shot 2016-07-25 at 1 44 06 pm

@rebornix For example this workaround will not work for the above case. But it's better than nothing :)

aminroosta:
@rebornix @sandy081 Here is another example where folding area information is needed:

`zj`:
Move downwards to the start of the next fold.  A closed fold
is counted as one fold. When a count is used, repeats the command [count] times.
This command can be used after an operator.

                                                        
`zk`:
Move upwards to the end of the previous fold.  A closed fold
                is counted as one fold. When a count is used, repeats the command [count] times.
                This command can be used after an operator.

rebornix:
I see, folding info is a must in this case. Since we already held on feature implementation for July Iteration, I'll create a separate issue to track this and see if we can make it in August. Thanks again for your input, they help a lot for both Vim and Code!

PR #494, opened 2016-07-21
Fix #438 - Limit the number of matches, and try to only recalculate when the searchString changes, or the document changes

### LINKED ISSUE #438: Incremental search is laggy in large files
If you try to search for something in our massive actions.ts file, or another large file, VSCodeVim lags. This is unsurprising because the current search approach is to parse the entire file on every keystroke.

"So don't write 2000 line files, johnfn"

Shush, that's not the point. The point is there are lots of easy win optimizations here. I did the dumbest thing that worked for the MVP. 😉

roblourens:
`document.version` is incremented each time the document is changed. Using a regex might be faster too, and since `/` is supposed to be a regex search, that's another thing that could be done. But you'll need to handle invalid regexes and stuff and I didn't want to do it in this change.

### DIFF
--- a/src/mode/modeHandler.ts
+++ b/src/mode/modeHandler.ts
@@ -128,50 +128,63 @@ export class VimSettings {
   useCtrlKeys     = false;
 }
 
+export enum SearchDirection {
+  Forward = 1,
+  Backward = -1
+}
+
 export class SearchState {
+  private static readonly MAX_SEARCH_RANGES = 1000;
+
   /**
    * Every range in the document that matches the search string.
    */
-  public matchRanges: vscode.Range[] = [];
+  private _matchRanges: vscode.Range[] = [];
+  public get matchRanges(): vscode.Range[] {
+    return this._matchRanges;
+  }
+
+  private _searchCursorStartPosition: Position;
+  public get searchCursorStartPosition(): Position {
+    return this._searchCursorStartPosition;
+  }
+
+  private _matchesDocVersion: number;
+  private _searchDirection: SearchDirection = SearchDirection.Forward;
 
   private _searchString = "";
   public get searchString(): string {
     return this._searchString;
   }
 
-  public searchCursorStartPosition: Position;
-
-  /**
-   * 1  === forward
-   * -1 === backward
-   */
-  public searchDirection = 1;
-
-
   public set searchString(search: string){
-    this._searchString = search;
-
-    this._recalculateSearchRanges();
+    if (this._searchString !== search) {
+      this._searchString = search;
+      this._recalculateSearchRanges(/*forceRecalc=*/true);
+    }
   }
 
-  private _recalculateSearchRanges(): void {
+  private _recalculateSearchRanges(forceRecalc?: boolean): void {
     const search = this.searchString;
 
     if (search === "") { return; }
 
-    // Calculate and store all matching ranges
-    this.matchRanges = [];
+    if (this._matchesDocVersion !== TextEditor.getDocumentVersion() || forceRecalc) {
+      // Calculate and store all matching ranges
+      this._matchesDocVersion = TextEditor.getDocumentVersion();
+      this._matchRanges = [];
 
-    for (let lineIdx = 0; lineIdx < TextEditor.getLineCount(); lineIdx++) {
-      const line = TextEditor.getLineAt(new Position(lineIdx, 0)).text;
+      for (let lineIdx = 0; lineIdx < TextEditor.getLineCount() && this._matchRanges.length < SearchState.MAX_SEARCH_RANGES; lineIdx++) {
+        const line = TextEditor.getLineAt(new Position(lineIdx, 0)).text;
 
-      let i = line.indexOf(search);
+        let i = line.indexOf(search);
 
-      for (; i !== -1; i = line.indexOf(search, i + search.length)) {
-        this.matchRanges.push(new vscode.Range(
-          new Position(lineIdx, i),
-          new Position(lineIdx, i + search.length)
-        ));
+        for (; i !== -1 && this._matchRanges.length < SearchState.MAX_SEARCH_RANGES; i = line.indexOf(search, i + search.length)) {
+          this._matchRanges.push(new vscode.Range(
+            new Position(lineIdx, i),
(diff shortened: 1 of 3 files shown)


src/mode/modeHandler.ts
```diff
@@ -129,50 +129,63 @@ export class VimSettings {
   useCtrlKeys     = false;
 }
 
+export enum SearchDirection {
```
johnfn: Big thumbs up on making this an enum. There's no need to retain my old 1 and -1. The default values are fine.

src/mode/modeHandler.ts
```diff
@@ -129,50 +129,63 @@ export class VimSettings {
...
+
 export class SearchState {
+  private static readonly MAX_SEARCH_RANGES = 1000;
+
   /**
    * Every range in the document that matches the search string.
    */
-  public matchRanges: vscode.Range[] = [];
+  private _matchRanges: vscode.Range[] = [];
+  public get matchRanges(): vscode.Range[] {
```
johnfn: You refactored this into a getter but I don't think you ever used the getter?

johnfn:
This is a really awesome change! I'll get to looking over it more seriously and merging it when I get back from vacation early next week.

rebornix:
It looks perfect!

src/mode/modeHandler.ts
roblourens: The public prop was used elsewhere, but only to read, so I changed it to a public getter. This and some of the other changes are just opinionated cleanup and totally not necessary, so I can revert them if you folks prefer the original way! I just like to restrict public access as much as possible.

src/mode/modeHandler.ts
roblourens: Thanks! The -1 and +1 are nice because elsewhere you're multiplying two directions to compute the 'effective direction', and that makes it easy. I like that pattern personally.

src/mode/modeHandler.ts
```diff
@@ -185,41 +198,41 @@ export class SearchState {
...
+      return { pos: Position.FromVSCodePosition(this._matchRanges[0].start), match: true };
     } else {
-      for (let matchRange of this.matchRanges.slice(0).reverse()) {
+      for (let matchRange of this._matchRanges.slice(0).reverse()) {
         if (matchRange.start.compareTo(startPosition) < 0) {
           return { pos: Position.FromVSCodePosition(matchRange.start), match: true };
         }
       }
 
       // TODO(bell)
```
roblourens: I'm not sure what any of these TODO's mean :)

src/mode/modeHandler.ts
rebornix: I do like the idea of accessibilty restriction and per https://github.com/Microsoft/TypeScript/issues/2845, accessors for the same member name must specify the same accessibility, we can't have a public `getter` and a private `setter` at the same time, so accessing the private member to do modification is the best way.

rebornix: Above comment only works for https://github.com/VSCodeVim/Vim/pull/494/files#diff-a7723f32fe6b05e135838f84e5c23c0fR176 . Even though `matchRanges` is readonly, you can't re-initialize it but you can always change its content, like pushing more elements to it.

johnfn: @roblourens I would suggest to use the `readonly` modifier in that case - it's a new modifier introduced in Typescript 2.0 😄

src/mode/modeHandler.ts
johnfn: Ah, I see. Hmm, if we're using the values of an enum directly, I'm not sure if this change actually improves readability, since that is not really standard. What do you think?

src/mode/modeHandler.ts
johnfn: This means that Vim rings a bell here (or flashes the screen or otherwise indicates an error has occurred).

src/mode/modeHandler.ts
roblourens: I think it improves readability from the outside, more than having to determine the magic numbers 1 and -1. The particular values are only used internally.

src/mode/modeHandler.ts
roblourens: Well, we still reinitialize it internally

johnfn: Ah, my bad, I assumed readonly was getter without setter.

src/mode/modeHandler.ts
```diff
@@ -129,50 +129,63 @@ export class VimSettings {
...
-  public searchDirection = 1;
-
-
   public set searchString(search: string){
-    this._searchString = search;
-
-    this._recalculateSearchRanges();
+    if (this._searchString !== search) {
+      this._searchString = search;
+      this._recalculateSearchRanges(/*forceRecalc=*/true);
```
johnfn: I like the idea here, but I'd suggest to take it one step further and pass in `{ forceRecalc: true }`.

src/mode/modeHandler.ts
```diff
@@ -129,50 +129,63 @@ export class VimSettings {
...
 
-      let i = line.indexOf(search);
+        let i = line.indexOf(search);
 
-      for (; i !== -1; i = line.indexOf(search, i + search.length)) {
-        this.matchRanges.push(new vscode.Range(
-          new Position(lineIdx, i),
-          new Position(lineIdx, i + search.length)
-        ));
+        for (; i !== -1 && this._matchRanges.length < SearchState.MAX_SEARCH_RANGES; i = line.indexOf(search, i + search.length)) {
```
johnfn: Instead of duplicating the MAX_SEARCH_RANGES check I would suggest to label the outer for loop with `outer:` and then do something like:

`if (this._matchRanges.length > SearchState.MAX_SEARCH_RANGES) { break outer; }`

within the inner loop.

johnfn:
Sorry for the slower responses, I was on vacation.

This change is awesome. Thanks a bunch for taking it on. I have a few minor suggestions, and after that we should be good to go here.

johnfn:
@roblourens Again, thanks for the awesome work on this PR. This is definitely one of the things that has been nagging at me - this is a huge improvement!

PR #496, opened 2016-07-21
Merge history changes into a single operation. Fixes #427

### LINKED ISSUE #427: Constant inserting should be merged into one single Insert Operation in history
### Issue Summary

When we enter Insert Mode and start typing, each character will be marked a single insert operation and saved in history tracker. Then if we do `undo`, these inserting operations will be reverted one by one. This is not how Vim does

### Steps to Reproduce

1. Open a file
2. Press `a` to enter Insert Mode
3. Type `abcdefhjkkl` then press `esc` and `u`

undo

These characters should be removed immediately together instead of one by one.

### Technical details:

* VSCode Version: Insider 1.4.0
* VsCodeVim Version: Latest
* OS:  OSX

infogulch:
Adds a method `HistoryStep.merge` that merges all of `this.changes` into as few `DocumentChange`s as possible, and call it in `finishCurrentStep()`. This makes undoing much faster.

I did some manual testing and I'm pretty sure the logic is sound, but it should probably be reviewed carefully.

See #427.

### DIFF
--- a/src/history/historyTracker.ts
+++ b/src/history/historyTracker.ts
@@ -33,17 +33,10 @@ export class DocumentChange {
    * Run this change.
    */
   public async do(undo = false): Promise<void> {
-    const rangeStart = this.start;
-
     if ((this.isAdd && !undo) || (!this.isAdd && undo)) {
-      await TextEditor.insert(this.text, rangeStart, false);
+      await TextEditor.insert(this.text, this.start, false);
     } else {
-      const rangeStop = rangeStart.advancePositionByText(this.text);
-
-      await TextEditor.delete(new vscode.Range(
-        rangeStart,
-        rangeStop
-      ));
+      await TextEditor.delete(new vscode.Range(this.start, this.end()));
     }
   }
 
@@ -52,6 +45,13 @@ export class DocumentChange {
    */
   public async undo(): Promise<void> {
     return this.do(true);
+  }
+
+  /**
+   * the position after advancing start by text
+   */
+  public end(): Position {
+    return this.start.advancePositionByText(this.text);
   }
 }
 
@@ -92,6 +92,51 @@ class HistoryStep {
     this.isFinished  = init.isFinished || false;
     this.cursorStart = init.cursorStart || undefined;
     this.marks     = init.marks || [];
+  }
+
+  /**
+   * merge collapses individual character changes into larger blocks of changes
+   */
+  public merge(): void {
+    if (this.changes.length < 2) {
+      return;
+    }
+    // merged will replace this.changes
+    var merged: DocumentChange[] = [];
+    // manually reduce() this.changes with variables `current` and `next`
+    // we can't use reduce() directly because the loop can emit multiple elements
+    var current = this.changes[0];
+    for (const next of this.changes.slice(1)) {
+      if (current.text.length === 0) {
+        // current is eliminated, replace it with top of merged, or adopt next as current
+        // see also add+del case
+        if (merged.length > 0) {
+          current = merged.pop();
+        } else {
+          current = next;
+          continue;
+        }
+      }
+      // merge logic. also compares start & end() Positions to ensure this is the same location
+      if (current.isAdd && next.isAdd && current.end().isEqual(next.start)) {
+        // merge add+add together
+        current.text += next.text;
+      } else if (!current.isAdd && !next.isAdd && next.end().isEqual(current.start)) {
+        // merge del+del together, but in reverse so it still reads forward
+        next.text += current.text;
+        current = next;
+      } else if (current.isAdd && !next.isAdd && current.end().isEqual(next.end())) {
+        // collapse add+del into add. this might make current.text.length === 0, see beginning of loop
+        current.text = current.text.slice(0, -next.text.length);
+      } else {
+        // del+add must be two separate DocumentChanges. e.g. start with "a|b", do `i<backspace>x<esc>` you end up with "|xb"
+        // also handles multiple changes in distant locations in the document
+        merged.push(current);
+        current = next;
+      }
+    }
+    merged.push(current);
+    this.changes = merged;
   }
 }
 
@@ -356,6 +401,8 @@ export class HistoryTracker {
     }
 
     this.currentHistoryStep.isFinished = true;
+
+    this.currentHistoryStep.merge();
 
     this.currentHistoryStep.marks = this.updateAndReturnMarks();
   }


infogulch:
I merged, would you rather I rebase?

johnfn:
The way I would test this would be to mash keys while also mashing backspace for about 30 seconds, then hit esc and then i and then do it again a few times. Then use u and c-r to see if all my gibberish was properly maintained. Would you mind doing a test like that? 

As you pointed out, I am very cautious about making changes to history tracking, since that's one thing I definitely don't want to break.

In any case, this is a really awesome and big change in terms of VSCodeVim usability! Thanks for the contribution!

infogulch:
I just pushed again, fixing a bug (more missed optimization) and adding comments to make the merge code a bit more understandable. Also, since I added method `DocumentChange.end()`, I refactored `DocumentChange.do()` to use it.

It seems very stable so far. One way to test it that gives a bit of visibility into what it's doing is to add `console.log(this.changes);` at the beginning and end of the `HistoryStep.merge()` method. I'll see if I can add some tests like you described.

One part that I haven't tested is where there are multiple `DocumentChange`s in different parts of the same document in a single `HistoryStep`. I'm not 100% sure this is possible, and this commit *should* support it, but I don't know how to trigger this condition.

johnfn:
> One part that I haven't tested is where there are multiple DocumentChanges in different parts of the same document in a single HistoryStep. I'm not 100% sure this is possible, and this commit should support it, but I don't know how to trigger this condition.

Yeah, this is tough. Possibly using search and replace with ctrl-f?

infogulch:
Hmm, the VSCode replace function doesn't appear to call `finishCurrentStep()`, so merge is never called in this case.

src/history/historyTracker.ts
```diff
@@ -93,6 +93,50 @@ class HistoryStep {
...
+    if (this.changes.length < 2) {
+      return;
+    }
+    // create new changes list
+    var merged: DocumentChange[] = [];
+    // iterate through the changes, saving the first outside the loop
+    var iter = (this.changes[Symbol.iterator]());
+    var prev = iter.next().value;
+    for (const change of iter) {
+      if (prev.text.length === 0) {
```
rebornix: this `prev` represents changes that we are going to merge or merging, maybe we can give it a more intuitive name?

src/history/historyTracker.ts
```diff
@@ -93,6 +93,50 @@ class HistoryStep {
...
+        prev.text += change.text;
+      } else if (!prev.isAdd && !change.isAdd && change.end().isEqual(prev.start)) {
+        // merge deletions together, but in reverse so it still reads forward
+        change.text += prev.text;
+        prev = change;
+      } else if (prev.isAdd && !change.isAdd && prev.end().isEqual(change.end())) {
+        // collapse addition+deletion into addition
+        prev.text = prev.text.slice(0, -change.text.length);
+      } else {
+        // deletion+addition must be two separate DocumentChanges
```
rebornix: Maybe I miss something, can you help me understand why `del+add` is different from `add+del`?

src/history/historyTracker.ts
infogulch: This loop is basically an `Array.reduce()`, but we can't use that directly because it can output more than one DocumentChange. e.g. changes in different parts of the document or a del+add (see below). reduce passes two arguments to the function: `previous` and `current`.

Here we have `prev` and `change` instead.

src/history/historyTracker.ts
infogulch: Say we start with `a|b`. Type `ix<backspace><esc>`, and it goes through `ax|b` then to `a|b`. Transpose the x and backspace and it's very different: `|b` then `x|b`.

src/history/historyTracker.ts
```diff
@@ -93,6 +93,50 @@ class HistoryStep {
...
+   * merge collapses individual character changes into larger blocks of changes
+   */
+  public merge(): void {
+    if (this.changes.length < 2) {
+      return;
+    }
+    // create new changes list
+    var merged: DocumentChange[] = [];
+    // iterate through the changes, saving the first outside the loop
+    var iter = (this.changes[Symbol.iterator]());
```
johnfn: Did you consider something like this?

```
prev = this.changes[0]
for (const change of this.changes) {```

File: src/history/historyTracker.ts (hunk 1 above)
Comment (johnfn) [inline src/history/historyTracker.ts]: Ah, I see. Since we can't use reduce, I'd suggest names that more closely map to what's going on here. Maybe `currentChange` for `prev` and `nextChange` or perhaps `toMerge` for `change`.

**Comment by johnfn** [comment]
This is completely awesome. It will definitely improve everyone's Vim experience. If you can fix up the few minor things I mentioned, we should be good to go here.

File: src/history/historyTracker.ts (hunk 3 above)
Comment (infogulch) [inline src/history/historyTracker.ts]: Yes I considered that at first, but then on the first iteration `prev` and `change` would be the same object. `change` needs to skip the first element, which is why I make an iterator, get the first element, then iterate through the rest.

Other options include using `var prev = this.changes.splice(0, 1)[0];` or maybe `var prev = this.changes[0]; for (const next of this.changes.slice(1))`. I'd prefer the slice method just because it doesn't alter the original array.

I went with the manual iterator because it doesn't change the original array, it only loops through the array once, and it doesn't create another object. I fully admit these are quite arbitrary reasons, two of which pretty clearly premature optimization. :smile: 

What would you prefer?

Comment (johnfn) [inline src/history/historyTracker.ts]: Oh, right right.

Of all the ideas listed, I prefer

```
prev = this.changes[0];
for (const next of this.changes.slice(1)) {
```

I think it's the most readable. 😄

src/history/historyTracker.ts
infogulch: I think I'll go for `previous` and `current` to match most of the examples on [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce), after making it a bit clearer that this loop is a manual reduce().

src/history/historyTracker.ts
infogulch: I want to note that there is a `TextEditor.replace` method that would be able to do both actions in one call, but the current DocumentChange does not support this (edge?)case right now and I didn't want to muck around too much.

infogulch:
Updated to fix the stuff you mentioned. Tests pass fine locally.

johnfn:
I just realized the most beautiful possible way to handle getting the first element of a list and the rest would have been:

`let [current, ...rest] = changes`

rather than that splicing thing we did.

I'm gonna just merge this because I'm impatient. I'll fix it on master sometime.

johnfn:
Anyways, again, thanks for the awesome pull request. This is really great work.

PR #499, opened 2016-07-23
Fix instructions for setting key bindings

positron:


johnfn:
Oops, thanks @positron!

PR #505, opened 2016-07-25
Turn around for cursor problem

rebornix:
Yay! We love PRs! 🎊

- [x] Commits are squashed 

Cursor update was wrong if we don't pass correct filename into ModeHanlder initialization.


johnfn:
Thanks for the quick fix @rebornix!

Issue #463, opened 2016-07-16
.vimrc support


johnfn:
Support an actual .vimrc file alongside our current settings.json. 

Allow users to specify a location for the .vimrc insite the settings.json file.

Support basic .vimrc commands.

rebornix:
Yup, assign this to me as I'm working on a prototype.

johnfn:
sick!

jpoon:
Re-opening this -- we don't have .vimrc support.

johnfn:
I'm gonna close this as although it's true that we don't support .vimrc, I believe our settings.json support is a solid equivalent, except better because it has autocomplete and documentation popups. 

If anyone disagrees, feel free to open a conversation on this issue.

kristiandupont:
Well, I have a shared vimrc file for vim and vsvim. I would have loved to include that from here as well.

wfxr:
I also expect vimrc support feature. Almost everyone using vscodevim knows how to config vimrc. So documentation popups feature is not that important as far as I see. 

I agree setting.json is good, but I think reuse and DRY means much.  I can use vimrc in vsvim, ideavim, neovim, etc., but I cant't use vscode setting.json anywhere else. That's the problem.

Maybe support both setting.json and vimrc is a better way?

johnfn:
I see your argument. Reopening this issue.

pietvanzoen:
+1 for this feature. Is there any update?

johnfn:
It's certainly on my radar, however it is currently deprioritized because:

* There are a lot of other issues with more thumbs up that we need to get to first.
* No functionality is actually lacking. It's just annoying not to have. 

If a bunch of people thumbs up the original issue, I'll certainly reconsider :)

polunzh:
+1 for this feature. Is there any update?

johnfn:
@polunzh The update is the post I made 9 days ago right above you. :P 

If you're curious, we pretty much tackle issues in this order: https://github.com/VSCodeVim/Vim/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc

atedja:
Just discovered Visual Studio Code, and immediately installed this plugin only to discover my .vimrc isn't being recognized :cry: I can't use vim mode without my vimrc, too many keybindings that I have developed over the years and too lazy to port them over to this plugin.

Chillee:
[...]

PR #508, opened 2016-07-25
Vim Settings

### LINKED ISSUE #481: :set hlsearch not working
Do you need help or have a question? Ask us on [Slack](https://vscodevim-slackin.azurewebsites.net/) 👫.
Found a bug? Please fill out the sections below 👍.

### What did you do?
I was trying to ":set hlsearch"
I would my expect my search would be highlighted. 
### What happened instead?
Not highlighted.
### Technical details:

* VSCode Version:1.3.1
* VsCodeVim Version: 0.1.3
* OS: Windows 10

### LINKED ISSUE #487: `w` should/could respect `wordSeparators` settings
My `wordSeparators` settings doesn't have `$`.

```js
  // Characters that will be used as word separators when doing word related navigations or operations
  "editor.wordSeparators": "`~!@#%^&*()-=+[{]}\\|;:'\",.<>/?",
```

### What did you do?

I edited my `settings.json` so that `wordSeparators` does *not* have `$`.
Then
The cursor in a file was here
```
foo$ + bar$
^
|
```
I used `w` in normal mode to skip to the next word.

Cursor to be here
```
foo$ + bar$
     ^
     |
```

### What happened instead?

```
foo$ + bar$
   ^
   |
```

### Technical details:

* VSCode Version: 1.3.1
* VsCodeVim Version: 0.1.3
* OS: Mac OS X El Capitan

rebornix:
- [x] Commits are squashed 

This feature is related to #433, #463, #487 and #481. The basic idea is maintaining a singleton of Vim Settings. This Vim Seting object reads configuration from User Settings, `.vimrc`, Code configuration then our own (VSCodeVim) default value.

- [x] Singleton Vim Settings
- [x] Load `vim.{xxx}` section
- [x] Map Vim configure to Code if there is any overlap
- [x] Fix #487
- [x] Fix #481
- [ ] `.vimrc` loading
- [ ] Auto update (monitoring both Code's configuration and `.vimrc`)

### DIFF
--- a/package.json
+++ b/package.json
@@ -166,6 +166,30 @@
                 "vim.scroll": {
                     "type": "number",
                     "description": "Number of lines to scroll with CTRL-U and CTRL-D commands."
+                },
+                "vim.tabstop": {
+                    "type": "number",
+                    "description": "Number of spaces that a <Tab> in the file counts for."
+                },
+                "vim.expandtab": {
+                    "type": "boolean",
+                    "description": "In Insert mode: Use the appropriate number of spaces to insert a <Tab>."
+                },
+                "vim.iskeyword": {
+                    "type": "string",
+                    "description": "keywords contain alphanumeric characters and '_'"
+                },
+                "vim.ignorecase": {
+                    "type": "boolean",
+                    "description": "Ignore case in search patterns."
+                },
+                "vim.smartcase": {
+                    "type": "boolean",
+                    "description": "Override the 'ignorecase' option if the search pattern contains upper case characters."
+                },
+                "vim.hlsearch": {
+                    "type": "boolean",
+                    "description": "When there is a previous search pattern, highlight all its matches."
                 }
             }
         }
--- a/package.json
+++ b/package.json
@@ -161,35 +161,33 @@
                 },
                 "vim.useSolidBlockCursor": {
                     "type": "boolean",
-                    "description": "Use a non blinking block cursor."
+                    "description": "Use a non blinking block cursor.",
+                    "default": false
                 },
                 "vim.scroll": {
                     "type": "number",
-                    "description": "Number of lines to scroll with CTRL-U and CTRL-D commands."
-                },
-                "vim.tabstop": {
-                    "type": "number",
-                    "description": "Number of spaces that a <Tab> in the file counts for."
-                },
-                "vim.expandtab": {
-                    "type": "boolean",
-                    "description": "In Insert mode: Use the appropriate number of spaces to insert a <Tab>."
+                    "description": "Number of lines to scroll with CTRL-U and CTRL-D commands.",
+                    "default": 20
                 },
                 "vim.iskeyword": {
                     "type": "string",
-                    "description": "keywords contain alphanumeric characters and '_'"
+                    "description": "keywords contain alphanumeric characters and '_'",
+                    "default": "/\\()\"':,.;<>~!@#$%^&*|+=[]{}`?-"
                 },
                 "vim.ignorecase": {
                     "type": "boolean",
-                    "description": "Ignore case in search patterns."
+                    "description": "Ignore case in search patterns.",
+                    "default": true
                 },
                 "vim.smartcase": {
                     "type": "boolean",
-                    "description": "Override the 'ignorecase' option if the search pattern contains upper case characters."
+                    "description": "Override the 'ignorecase' option if the search pattern contains upper case characters.",
+                    "default": true
                 },
                 "vim.hlsearch": {
                     "type": "boolean",
-                    "description": "When there is a previous search pattern, highlight all its matches."
+                    "description": "When there is a previous search pattern, highlight all its matches.",
+                    "default": true
                 }
             }
         }
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -8,6 +8,7 @@ import { Position } from './../motion/position';
 import { PairMatcher } from './../matching/matcher';
 import { QuoteMatcher } from './../matching/quoteMatcher';
 import { Tab, TabCommand } from './../cmd_line/commands/tab';
+import { Configuration } from './../configuration/configuration';
 import * as vscode from 'vscode';
 
 const controlKeys: string[] = [
@@ -1487,7 +1488,10 @@ class CommandMoveHalfPageDown extends BaseMovement {
   keys = ["ctrl+d"];
 
   public async execAction(position: Position, vimState: VimState): Promise<Position> {
-    return new Position(Math.min(TextEditor.getLineCount() - 1, position.line + vimState.settings.scroll), position.character);
+    return new Position(
+      Math.min(TextEditor.getLineCount() - 1, position.line + <number> Configuration.getInstance().get("scroll")),
+      position.character
+    );
   }
 }
 
@@ -1496,7 +1500,7 @@ class CommandMoveHalfPageUp extends BaseMovement {
   keys = ["ctrl+u"];
 
   public async execAction(position: Position, vimState: VimState): Promise<Position> {
-    return new Position(Math.max(0, position.line - vimState.settings.scroll), position.character);
+    return new Position(Math.max(0, position.line - <number> Configuration.getInstance().get("scroll")), position.character);
   }
 }
 
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -1489,7 +1489,7 @@ class CommandMoveHalfPageDown extends BaseMovement {
 
   public async execAction(position: Position, vimState: VimState): Promise<Position> {
     return new Position(
-      Math.min(TextEditor.getLineCount() - 1, position.line + <number> Configuration.getInstance().get("scroll")),
+      Math.min(TextEditor.getLineCount() - 1, position.line + Configuration.getInstance().scroll),
       position.character
     );
   }
@@ -1500,7 +1500,7 @@ class CommandMoveHalfPageUp extends BaseMovement {
   keys = ["ctrl+u"];
 
   public async execAction(position: Position, vimState: VimState): Promise<Position> {
-    return new Position(Math.max(0, position.line - <number> Configuration.getInstance().get("scroll")), position.character);
+    return new Position(Math.max(0, position.line - Configuration.getInstance().scroll), position.character);
   }
 }
 
--- a/src/cmd_line/commands/setoptions.ts
+++ b/src/cmd_line/commands/setoptions.ts
@@ -1,0 +1,68 @@
+"use strict";
+
+import * as node from "../node";
+import { Configuration } from '../../configuration/configuration';
+
+export enum SetOptionOperator {
+  /*
+   * Set string or number option to {value}.
+   * White space between {option} and '=' is allowed and will be ignored.  White space between '=' and {value} is not allowed.
+   */
+  Equal,
+  /*
+   * Toggle option: set, switch it on.
+   * Number option: show value.
+   * String option: show value.
+   */
+  Set,
+  /*
+   * Toggle option: Reset, switch it off.
+   */
+  Reset,
+  /*
+   * Add the {value} to a number option, or append the {value} to a string option.
+   * When the option is a comma separated list, a comma is added, unless the value was empty.
+   */
+  Append,
+  /*
+   * Subtract the {value} from a number option, or remove the {value} from a string option, if it is there.
+   */
+  Subtract
+}
+
+export interface IOptionArgs extends node.ICommandArgs {
+  name?: string;
+  operator?: SetOptionOperator;
+  value?: string | number | boolean;
+}
+
+export class SetOptionsCommand extends node.CommandBase {
+  protected _arguments: IOptionArgs;
+
+  constructor(args : IOptionArgs) {
+    super();
+    this._name = 'setoptions';
+    this._shortName = 'option';
+    this._arguments = args;
+  }
+
+  get arguments() : IOptionArgs {
+    return this._arguments;
+  }
+
+  async execute(): Promise<void> {
+    switch (this._arguments.operator) {
+      case SetOptionOperator.Set:
+        Configuration.getInstance().set(this._arguments.name!, true);
+        break;
+      case SetOptionOperator.Reset:
+        Configuration.getInstance().set(this._arguments.name!, false);
+        break;
+      case SetOptionOperator.Equal:
+        Configuration.getInstance().set(this._arguments.name!, this._arguments.value);
+        break;
+      default:
+        break;
+    }
+  }
+}
--- a/src/cmd_line/commands/setoptions.ts
+++ b/src/cmd_line/commands/setoptions.ts
@@ -65,13 +65,13 @@ export class SetOptionsCommand extends node.CommandBase {
   async execute(): Promise<void> {
     switch (this._arguments.operator) {
       case SetOptionOperator.Set:
-        Configuration.getInstance().set(this._arguments.name!, true);
+        Configuration.getInstance()[this._arguments.name!] = true;
         break;
       case SetOptionOperator.Reset:
-        Configuration.getInstance().set(this._arguments.name!, false);
+        Configuration.getInstance()[this._arguments.name!] = false;
         break;
       case SetOptionOperator.Equal:
-        Configuration.getInstance().set(this._arguments.name!, this._arguments.value!);
+        Configuration.getInstance()[this._arguments.name!] = this._arguments.value!;
         break;
       default:
         break;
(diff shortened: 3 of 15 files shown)


src/configuration/configuration.ts
```diff
@@ -0,0 +1,277 @@
...
+
+/**
+ * Vim option override sequence.
+ * 1. `:set {option}` on the fly
+ * 2. VS Code Vim configuration section && overlap configurations between Code and Vim
+ * 3. TODO .vimrc.
+ * 4. VSCodeVim flavored Vim option default values
+ */
+export class Configuration {
+  private static _instance: Configuration;
```
johnfn: Use optional types!

private static _instance: Configuration | null;

rebornix:
Currently you will see some problems as we have bugs in Code https://github.com/Microsoft/vscode/issues/9731.

johnfn:
> .vimrc loading

I have no qualms at all about you leaving this to another PR.

> Auto update (monitoring both Code's configuration and .vimrc)

This would be very awesome, but low priority. Most people expect to have to restart their editor anyways.

package.json
```diff
@@ -145,6 +145,30 @@
                 "vim.scroll": {
                     "type": "number",
                     "description": "Number of lines to scroll with CTRL-U and CTRL-D commands."
+                },
+                "vim.tabstop": {
```
johnfn: You don't know how happy it makes me to see we support all this stuff. My god, it's like we're actually making a real Vim plugin!!!

src/actions/actions.ts
```diff
@@ -1247,7 +1248,10 @@ class MCommandMoveHalfPageDown extends BaseMovement {
   keys = ["ctrl+d"];
 
   public async execAction(position: Position, vimState: VimState): Promise<Position> {
-    return new Position(Math.min(TextEditor.getLineCount() - 1, position.line + vimState.settings.scroll), position.character);
+    return new Position(
+      Math.min(TextEditor.getLineCount() - 1, position.line + <number> Configuration.getInstance().get("scroll")),
```
johnfn: I think the number cast can go inline like so

`.get<number>("scroll")`

Or something like that???

src/configuration/configuration.ts
```diff
@@ -0,0 +1,280 @@
...
+  public static allOptions: { [key: string]: BaseOptionDetails } = {};
+}
+
+export function RegisterOption(optionDetails: typeof BaseOptionDetails): void {
+  let newOption = new optionDetails();
+  newOption.initialize();
+  OptionMap.allOptions[newOption.id] = newOption;
+}
+
+@RegisterOption
```
johnfn: After having to deal with this pattern for 3000 lines of action.ts, I am beginning to regret it. Fixing that file will be hard, but fixing this one will be easy!!! 😀 

My suggestion would instead to be to make an interface called BaseOptionDetails that looks like this

```
interface BaseOption { //yep, use templating plz
    vimName: string;
    vscodeName: string?;
    defaultValue: OptionValue;
}
```

Then do something like

```
var options = [
    { vimName: "useSolidBlockCursor", defaultValue: false },
    { vimName: "useCtrlKeys", defaultValue: false },
    // ETC.
]
```

Then restructure the rest of the code around that.

I can explain exactly why I want to do this, but, long story short, classes are very prohibitive/clunky for our purposes, and we don't really use them for all that much that basic objects can't do.

src/configuration/configuration.ts
```diff
@@ -0,0 +1,280 @@
...
+    }
+  }
+}
+
+@RegisterOption
+class OptionInsertSpaces extends BaseOptionDetails {
+  id = "expandtab"; // aka insertSpaces
+  defaultValue = false;
+
+  fetchCodeConfiguration() {
```
johnfn: There is a bunch of duplication between this and OptionTabStop that you're going to need to remove as well.

src/actions/actions.ts
```diff
@@ -1247,7 +1248,10 @@ class MCommandMoveHalfPageDown extends BaseMovement {
   keys = ["ctrl+d"];
 
   public async execAction(position: Position, vimState: VimState): Promise<Position> {
-    return new Position(Math.min(TextEditor.getLineCount() - 1, position.line + vimState.settings.scroll), position.character);
+    return new Position(
```
johnfn: :+1:

src/cmd_line/subparsers/setoptions.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  if (option.startsWith("no")) {
+    // :se[t] no{option}  Toggle option: Reset, switch it off.
+    return {
+      name: option.substring(2, option.length),
+      operator: node.SetOptionOperator.Reset
+    };
+  }
+
+  if (option.includes('=')) {
```
johnfn: "foo=bar" should not parse as a single word, but rather 3 separate words: ["foo", "=", "bar"]. Same with ":".

I realize Vim has some slightly bizarre rules about whitespace, but I think it's okay to be more permissive than they are if it is also easier to code it that way.

src/configuration/configuration.ts
```diff
@@ -0,0 +1,280 @@
...
+
+    targetOption.value = value;
+  }
+
+  /**
+   * Return a value of option from vim settings.
+   * @param vim option.
+   * @param defaultValue A value should be returned when no value could be found, is `undefined`.
+   */
+  get(option: string, defaultValue?: OptionValue): OptionValue | undefined {
```
johnfn: I don't actually like this API. I would prefer to have every setting listed off sort of like so

```
class Configuration {
    tabstop: number;
    smartcase: boolean;
    ignorecase: boolean;
}
```

etc.

The advantages here are 

1. All settings accesses will be strongly typed
2. Autocomplete will work for our settings

src/configuration/configuration.ts
rebornix: The one you suggest is what exactly I did at first but later on I found it became difficult to add handlers then I switch to our Action inheritance magic, lol ...

But now it seems I'm over designing.

johnfn: Haha, yeah. The logic will have to go somewhere. I just want to separate the logic from the data.

rebornix:
Did a huge refactor to VimSettings. I can't imagine before that we can make this real simple with a little bit hack of Accessor Decorator.

With the new design, we only have one single Class/Object maintaining all Vim options' definition, intialization and updating. It's called `Configuration`.

## Access `Configuration`
Use `Configuration.getInstance()`, it's the only way of getting an object of Configuration and yeah we make it a Singleton. 

`getInstance()` is the only public method in `Configuration`.

## Access a single Vim option
1. Get: `Configuration.getInstance().{OptionName}`, eg, `Configuration.getInstance().tabstop`.
2. Set: `Configuration.getInstance().{OptionName} = {NewValue}`, eg, `Configuration.getInstance().tabstop = 4`.
3. Bracket notation: eg, `Configuration.getInstance()["tabstop"]`, `Configuration.getInstance()["tabstop"] = 4`.

Lastly, we will support `.vimrc` later on and the update of `.vimrc` should trigger `Configuration` to update automatically.

rebornix:
Fix #279 as it's real easy to implement based on current Setting design, just 20 lines of code change.

johnfn:
Hey @rebornix is this good to go? Or are you still working on it?

rebornix:
I updated the code after our last sync up on Slack. Feel free to review and let me know what else you think can be improved.

src/configuration/configuration.ts
```diff
@@ -0,0 +1,84 @@
...
+export class Configuration {
+  private static _instance: Configuration | null;
+
+  constructor() {
+    /**
+     * Load Vim options from User Settings.
+     */
+    let vimOptions = vscode.workspace.getConfiguration("vim");
+    /* tslint:disable:forin */
+    // Disable forin rule here as we make accessors enumerable.`
```
johnfn: Can we use for of here as well?

src/configuration/configuration.ts
```diff
@@ -0,0 +1,84 @@
...
+
+  public static getInstance(): Configuration {
+    if (Configuration._instance == null) {
+      Configuration._instance = new Configuration();
+    }
+
+    return Configuration._instance;
+  }
+
+  useSolidBlockCursor: boolean = false;
```
johnfn: Yess, this is great.

src/configuration/configuration.ts
```diff
@@ -0,0 +1,84 @@
...
+ * 4. If user doesn't set the option explicitly
+ *    a. we don't have a similar setting in Code, initialize the option as default value.
+ *    b. we have a similar setting in Code, use Code's setting.
+ *
+ * Vim option override sequence.
+ * 1. `:set {option}` on the fly
+ * 2. TODO .vimrc.
+ * 2. `vim.{option}`
+ * 3. VS Code configuration
+ * 4. VSCodeVim flavored Vim option default values
```
jpoon: Can you add this to the `readme` explaining which settings take priority?

src/configuration/configuration.ts
```diff
@@ -0,0 +1,84 @@
...
+  ignorecase: boolean = true;
+  smartcase: boolean = true;
+
+  @overlapSetting({ codeName: "tabSize", default: 8})
+  tabstop: number;
+
+  @overlapSetting({ codeName: "insertSpaces", default: false})
+  expandtab: boolean;
+
+  @overlapSetting({ codeName: "wordSepatators", default: "/\\()\"':,.;<>~!@#$%^&*|+=[]{}`?"})
```
jpoon: Spelling. `wordSeparators`

src/cmd_line/commands/setoptions.ts
```diff
@@ -0,0 +1,83 @@
...
+    this._name = 'setoptions';
+    this._shortName = 'option';
+    this._arguments = args;
+  }
+
+  get arguments() : IOptionArgs {
+    return this._arguments;
+  }
+
+  async execute(): Promise<void> {
```
johnfn: Do something like if (!this._arguments.name) { return; } (and probably throw an error as well) so you dont have to use ! below

src/mode/modeHandler.ts
```diff
@@ -174,21 +167,39 @@ export class SearchState {
       this._matchesDocVersion = TextEditor.getDocumentVersion();
       this._matchRanges = [];
 
+      /* Decide whether the search is case sensitive.
+       * If ignorecase is false, the search is case sensitive.
+       * If ignorecase is true, the search should be case insenstive.
```
jpoon: Spelling: case `insensitive`

src/mode/modeHandler.ts
```diff
@@ -174,21 +167,39 @@ export class SearchState {
...
+       * If both ignorecase and smartcase are true, the search is case sensitive only when the search string contains UpperCase character.
+       */
+      let ignorecase = Configuration.getInstance().ignorecase;
+
+      if (ignorecase && Configuration.getInstance().smartcase && /[A-Z]/.test(search)) {
```
jpoon: Is this logic necessary?

johnfn:
Instead of the object being a singleton, is it possible to export an instance of the class (rather than the class itself) and then use that directly?

I'm not actually sure if this is possible, but if it is it would improve the code a little. :-)

src/cmd_line/commands/setoptions.ts
```diff
@@ -0,0 +1,83 @@
...
+    switch (this._arguments.operator) {
+      case SetOptionOperator.Set:
+        Configuration.getInstance()[this._arguments.name!] = true;
```
jpoon: What does the `!` at the end mean?

src/mode/modeHandler.ts
```diff
@@ -174,21 +167,39 @@ export class SearchState {
...
+        ignorecase = false;
+      }
+
+      const regex = new RegExp(search.replace(/[\-\[\]{}()*+?.,\\\^$|#\s]/g, "\\$&"), ignorecase ? 'gi' : 'g');
```
johnfn: What is this regex? If we'll ever use it again, could you pull it into a constant somewhere? In either case, could you give it a better name?

jpoon:
@rebornix Sweeeet. I added a couple of minor comments but LGTM!

src/mode/modeHandler.ts
```diff
@@ -174,21 +167,39 @@ export class SearchState {
...
-        for (; i !== -1; i = line.indexOf(search, i + search.length)) {
+        while (result) {
           if (this._matchRanges.length >= SearchState.MAX_SEARCH_RANGES) {
             break outer;
           }
 
-          this._matchRanges.push(new vscode.Range(
-            new Position(lineIdx, i),
-            new Position(lineIdx, i + search.length)
+          this.matchRanges.push(new vscode.Range(
```
johnfn: Holy smokes, this PR includes smartcase as well? Heck yeah!

src/cmd_line/commands/setoptions.ts
johnfn: It's a new feature related to nullable types in Typescript 2.0. 

this._arguments.name is an optional type - it's either a string or undefined. The ! converts the type to a string.

johnfn:
This is great! I think we together (ok, mostly you!) have arrived at a really good design for settings that should be really pleasant to add to in the future. Fix my few minor complaints and we should be good to go here.

src/configuration/configuration.ts
```diff
@@ -0,0 +1,84 @@
...
+    for (const option in this) {
+      const vimOptionValue = vimOptions[option];
+      if (vimOptionValue !== null && vimOptionValue !== undefined) {
+        this[option] = vimOptionValue;
+      }
+    }
+  }
+
+  public static getInstance(): Configuration {
+    if (Configuration._instance == null) {
```
jpoon: If somebody updates their workspace config, being a singleton, will this class retain the old value?

johnfn: I'm ok with a restart for settings change if you are.

johnfn: (We could do hot reloading as a feature later on...)

johnfn:
And please ping me if I don't respond to a finished PR for a day or two. It's a little hard for me to sort through the PRs to know which ones are done, and I definitely don't want to leave anything that is done hanging around. :)

src/configuration/configuration.ts
jpoon: I'm ok with it. Just need to document it.

johnfn: Hm, what's the alternative?

jpoon: Nevermind, we read use this A LOT in our code. The alternative I had in mind initially was to read the configs from the file every time.

src/cmd_line/commands/setoptions.ts
jpoon: thanks

src/configuration/configuration.ts
rebornix: I'd love to `for of` if possible but `for of` is specific to collections, which requires the object has a [Symbol.iterator] property. So I just kept using `for in` to avoid making it complex.

src/mode/modeHandler.ts
rebornix: It's how smart case works. Whether we should use `i` is decided by user's `ignorecase`, `smartcase` configuration and the search keyword.

rebornix:
Update per comments and resolve conflicts.

jpoon:
ship it.

src/mode/modeHandler.ts
```diff
@@ -145,6 +146,7 @@ export class VimState {
    * by us or by a mouse action.
    */
   public whatILastSetTheSelectionTo: vscode.Selection;
+<<<<<<< b9a0b15742532ecc54e66ee04c417d82e70d29dc
```
johnfn: Wait, there are merge conflicts here.

src/mode/modeHandler.ts
```diff
@@ -556,6 +577,7 @@ export class ModeHandler implements vscode.Disposable {
     });
   }
 
+<<<<<<< b9a0b15742532ecc54e66ee04c417d82e70d29dc
```
johnfn: More conflicts

src/mode/modeHandler.ts
rebornix: I start to doubt myself as it happens several times. I remember correctly yesterday that I did remove these lines by `dd` and save them then use `git add . & git rebase --continue`. After that, I even search `====` globally as I made this kind of mistakes before, but damn, it still happens.

I'll try to track this down next time I do rebasing. The only potential catch came to my mind now is I always reload the window as jsdiff is hanging, I don't if this leads to the unsuccessful saving.

johnfn: Be careful, you want to use `git add . && git rebase --continue`. Using a single & will cause a process to be backgrounded. 

But yeah, it's annoying and it eventually happens to everyone. No worries :)

moopie:
Any update on the .vimrc loading feature?

PR #514, opened 2016-07-25
Make <delete> repeatable in Normal Mode. Fix #394

### LINKED ISSUE #394: delete N characters under and after the cursor with Del key
### Issue Summary

Now we only support `x` to delete characters while we can use `Del` on Windows/Linux platform to archive the target.

We only need to register this key in contribution and map it to `x` operator. I'd like to mark it as `easy` and `help_wanted`!

octref:
Yay! We love PRs! 🎊

- [x] Commits are squashed 

Previously `<delete>` key works in Normal Mode but is not repeatable.
This PR makes it repeatable, so commands like `5<delete>` would work like `5x`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,7 +55,7 @@
             {
                 "key": "Delete",
                 "command": "extension.vim_delete",
-                "when": "editorTextFocus"
+                "when": "editorTextFocus && vim.mode == 'Normal Mode'"
             },
             {
                 "key": "ctrl+r",
--- a/package.json
+++ b/package.json
@@ -53,6 +53,11 @@
                 "when": "editorTextFocus"
             },
             {
+                "key": "Delete",
+                "command": "extension.vim_delete",
+                "when": "editorTextFocus"
+            },
+            {
                 "key": "ctrl+r",
                 "command": "extension.vim_ctrl+r",
                 "when": "editorTextFocus"
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -1960,6 +1960,11 @@ class ActionDeleteChar extends BaseCommand {
 }
 
 @RegisterAction
+class ActionDeleteCharWithDeleteKey extends ActionDeleteChar {
+  keys = ["<delete>"];
+}
+
+@RegisterAction
 class ActionDeleteLastChar extends BaseCommand {
   modes = [ModeName.Normal];
   keys = ["X"];
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -1960,8 +1960,24 @@ class ActionDeleteChar extends BaseCommand {
 }
 
 @RegisterAction
-class ActionDeleteCharWithDeleteKey extends ActionDeleteChar {
+class ActionDeleteCharWithDeleteKey extends BaseCommand {
+  modes = [ModeName.Normal];
   keys = ["<delete>"];
+  canBePrefixedWithCount = true;
+  canBeRepeatedWithDot = true;
+
+  public async execCount(position: Position, vimState: VimState): Promise<VimState> {
+    // N<del> is a no-op in Vim
+    if (vimState.recordedState.count !== 0) {
+      return vimState;
+    }
+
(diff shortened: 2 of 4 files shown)


octref:
BTW, there are a few unrelated tests failing, but they also fail on the master branch without my changes. So I ignored them.

src/actions/actions.ts
```diff
@@ -1960,6 +1960,11 @@ class ActionDeleteChar extends BaseCommand {
 }
 
 @RegisterAction
+class ActionDeleteCharWithDeleteKey extends ActionDeleteChar {
+  keys = ["<delete>"];
+}
```
rebornix: There difference between delete and x is the way they handle numeric prefix. `Nx` will perform `x N times but `N<Del>` does nothing.

package.json
```diff
@@ -53,6 +53,11 @@
                 "when": "editorTextFocus"
             },
             {
+                "key": "Delete",
+                "command": "extension.vim_delete",
+                "when": "editorTextFocus"
```
rebornix: We are reimplementing delete on in Normal mode, so you may want to check Vim state here like https://github.com/VSCodeVim/Vim/blob/master/package.json#L68

src/actions/actions.ts
octref: I thought `N<Del>` to delete the next 5 chars should be the expected behavior.

So, in Normal Mode, is this

`<Del>` -> removes next char
`N<Del>` -> do nothing

what I should do instead?

package.json
octref: Thanks for the hint!

src/actions/actions.ts
rebornix: Yup, that's Vim's behavior and it's awkward, but let's do this.

test/mode/normalModeTests/commands.test.ts
```diff
@@ -61,13 +68,20 @@ suite("Mode Normal", () => {
...
 
     newTest({
-      title: "Can handle x at end of line",
+      title: "Can handle 'x' at end of line",
       start: ['one tw|o'],
       keysPressed: '^llxxxxxxxxx',
       end: ['|'],
     });
 
     newTest({
```
rebornix: Also add a case for `N<delete>` as the behavior is strange in Vim.

src/actions/actions.ts
```diff
@@ -1960,6 +1960,19 @@ class ActionDeleteChar extends BaseCommand {
 }
 
 @RegisterAction
+class ActionDeleteCharWithDeleteKey extends ActionDeleteChar {
+  keys = ["<delete>"];
+
+  public async execCount(position: Position, vimState: VimState): Promise<VimState> {
+    if (vimState.recordedState.count !== 0) {
```
johnfn: This is actually impossible, as 0 by itself is a motion, so it can't prefix any motion.

johnfn:
Fix up my one suggestion and we should be good to go here! Thanks for the PR, @octref!

src/actions/actions.ts
octref: It seems to me when I press `<delete>`, it goes to `execCount` with a 0 count (default value for an int).
In your BaseCommand, you have this line to set it to 1 if it's 0:
`let timesToRepeat = this.canBePrefixedWithCount ? vimState.recordedState.count || 1 : 1;`

johnfn: The fact that Ndelete does nothing is so bizarre that I totally misread this. This is fine.

johnfn:
Ah, I'm sorry for doing this retroactively, but you caught us in the middle of a refactor. 

Would you mind fixing this up by making both delete classes extend from an abstract base class rather than using inheritance? I'm trying to move away from big inheritance trees; see #519.

octref:
No worries. Refactoring is good.
In work now. Will do that tonight.

octref:
@johnfn 
Done. Sorry for the delay. Had some company events for the last few days.

johnfn:
LGTM. Thanks, @octref!

PR #515, opened 2016-07-25
Implement increment and decrement operators

### LINKED ISSUE #513: Support ctrl+a and ctrl+x to increment/decrement numbers
* VSCode Version: 1.3.1
* VsCodeVim Version: 0.1.3
* OS: Ubuntu 16.04

sectioneight:
This adds support for single or numeric-prefix control+x and control+a.

Numeric modes supported:
- [x] Decimal
- [x] Hex
- [x] Octal

Fixes #513

### DIFF
--- a/extension.ts
+++ b/extension.ts
@@ -179,7 +179,7 @@ export async function activate(context: vscode.ExtensionContext) {
     showCmdLine("", modeHandlerToEditorIdentity[new EditorIdentity(vscode.window.activeTextEditor).toString()]);
   });
 
-  'rfbducw['.split('').forEach(key => {
+  'rfbducw[ax'.split('').forEach(key => {
     registerCommand(context, `extension.vim_ctrl+${key}`, () => handleKeyEvent(`ctrl+${key}`));
   });
 
--- a/package.json
+++ b/package.json
@@ -93,6 +93,16 @@
                 "when": "editorTextFocus && vim.useCtrlKeys"
             },
             {
+                "key": "ctrl+a",
+                "command": "extension.vim_ctrl+a",
+                "when": "editorTextFocus"
+            },
+            {
+                "key": "ctrl+x",
+                "command": "extension.vim_ctrl+x",
+                "when": "editorTextFocus"
+            },
+            {
                 "key": "left",
                 "command": "extension.vim_left",
                 "when": "editorTextFocus && vim.mode != 'Insert Mode'"
--- a/package.json
+++ b/package.json
@@ -95,7 +95,7 @@
             {
                 "key": "ctrl+a",
                 "command": "extension.vim_ctrl+a",
-                "when": "editorTextFocus"
+                "when": "editorTextFocus && vim.useCtrlKeys"
             },
             {
                 "key": "ctrl+x",
--- a/package.json
+++ b/package.json
@@ -100,7 +100,7 @@
             {
                 "key": "ctrl+x",
                 "command": "extension.vim_ctrl+x",
-                "when": "editorTextFocus"
+                "when": "editorTextFocus && vim.useCtrlKeys"
             },
             {
                 "key": "left",
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -2712,7 +2712,7 @@ abstract class IncrementDecrementNumberAction extends BaseMovement {
       // '-' doesn't count as a word, but is important to include in parsing the number
       if (text[start.character - 1] === '-') {
         start = start.getLeft();
-        word = text[start.character - 1] + word;
+        word = text[start.character] + word;
       }
       // Strict number parsing so "1a" doesn't silently get converted to "1"
       const num = NumericString.parse(word);
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -2694,3 +2694,76 @@ class ToggleCaseAndMoveForward extends BaseMovement {
     return position.getRight();
   }
 }
+
+abstract class IncrementDecrementNumberAction extends BaseMovement {
+  modes = [ModeName.Normal];
+  canBePrefixedWithCount = true;
+
+  offset: number;
+
+  public async execActionWithCount(position: Position, vimState: VimState, count: number): Promise<Position> {
+    count = count || 1;
+    const text = TextEditor.getLineAt(position).text;
+
+    // They may start on a number, so check the current word boundaries first.
+    let wordEnd = position.getCurrentWordEnd(true);
+    let wordStart = position.getWordLeft(true);
+    do {
+      // '-' doesn't count as a word, but is important to include in parsing the number
+      if (text[wordStart.character - 1] === '-') {
+        wordStart = wordStart.getLeft();
+      }
+      const word = text.substring(wordStart.character, wordEnd.character + 1);
+      // Strict number parsing so "1a" doesn't silently get converted to "1"
+      const num = Number(word);
+
+      if (!isNaN(num)) {
+        return this.replaceNum(num, this.offset * count, wordStart, wordEnd);
+      }
+
+      if (wordEnd.isLineEnd()) {
+        // We got to the end of the line and didn't find anything that looks like a number
+        break;
+      }
+
+      // Move forward to the next word
+      wordStart = wordStart.getWordRight();
+      wordEnd = wordStart.getCurrentWordEnd();
+    } while (true);
+
+    // No usable numbers, return the original position
+    return position;
+  }
+
+  public async replaceNum(start: number, offset: number, startPos: Position, endPos: Position): Promise<Position> {
+    const oldWidth = start.toString().length;
+    start += offset;
+    const newNum = start.toString();
+
+    const range = new vscode.Range(startPos, endPos.getRight());
+
+    if (oldWidth === newNum.length) {
+      await TextEditor.replace(range, newNum);
+    } else {
+      // Can't use replace, since new number is a different width than old
+      await TextEditor.delete(range);
+      await TextEditor.insertAt(newNum, startPos);
+      // Adjust end position according to difference in width of number-string
+      endPos = new Position(endPos.line, endPos.character + (newNum.length - oldWidth));
+    }
+
+    return endPos;
+  }
+}
+
+@RegisterAction
+class IncrementNumberAction extends IncrementDecrementNumberAction {
+  keys = ["ctrl+a"];
+  offset = +1;
+}
+
+@RegisterAction
+class DecrementNumberAction extends IncrementDecrementNumberAction {
+  keys = ["ctrl+x"];
+  offset = -1;
+}
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -2707,31 +2707,20 @@ abstract class IncrementDecrementNumberAction extends BaseMovement {
     const text = TextEditor.getLineAt(position).text;
 
     // They may start on a number, so check the current word boundaries first.
-    let wordEnd = position.getCurrentWordEnd(true);
-    let wordStart = position.getWordLeft(true);
-    do {
+
+    for (let { start, end, word } of Position.IterateWords(position.getWordLeft(true))) {
       // '-' doesn't count as a word, but is important to include in parsing the number
-      if (text[wordStart.character - 1] === '-') {
-        wordStart = wordStart.getLeft();
+      if (text[start.character - 1] === '-') {
+        start = start.getLeft();
+        word = text[start.character - 1] + word;
       }
-      const word = text.substring(wordStart.character, wordEnd.character + 1);
       // Strict number parsing so "1a" doesn't silently get converted to "1"
       const num = NumericString.parse(word);
 
       if (num !== null) {
-        return this.replaceNum(num, this.offset * count, wordStart, wordEnd);
+        return this.replaceNum(num, this.offset * count, start, end);
       }
-
-      if (wordEnd.isLineEnd()) {
-        // We got to the end of the line and didn't find anything that looks like a number
-        break;
-      }
-
-      // Move forward to the next word
-      wordStart = wordStart.getWordRight();
-      wordEnd = wordStart.getCurrentWordEnd();
-    } while (true);
-
+    }
     // No usable numbers, return the original position
     return position;
   }
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -2,6 +2,7 @@ import { VimSpecialCommands, VimState, SearchState } from './../mode/modeHandler
 import { ModeName } from './../mode/mode';
 import { TextEditor } from './../textEditor';
 import { Register, RegisterMode } from './../register/register';
+import { NumericString } from './../number/numericString';
 import { Position } from './../motion/position';
 import { PairMatcher } from './../matching/matcher';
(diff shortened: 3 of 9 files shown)


src/actions/actions.ts
```diff
@@ -2694,3 +2694,76 @@ class ToggleCaseAndMoveForward extends BaseMovement {
...
+    if (oldWidth !== newNum.length) {
+      // Can't use replace, since new number may be different width than old
+      await TextEditor.delete(range);
+      await TextEditor.insertAt(newNum, startPos);
+    } else {
+      await TextEditor.replace(range, newNum);
+    }
+
+    // Advance to the end of the word in case it's longer
+    return Position.FromVSCodePosition(endPos.translate(0, newNum.length - oldWidth));
```
sectioneight: Not sure about this, feels weird to create an intermediate object just to move a character. Thoughts?

sectioneight:
"." isn't working quite right, need to track that down

src/actions/actions.ts
```diff
@@ -2694,3 +2694,76 @@ class ToggleCaseAndMoveForward extends BaseMovement {
...
+  canBePrefixedWithCount = true;
+
+  offset: number;
+
+  public async execActionWithCount(position: Position, vimState: VimState, count: number): Promise<Position> {
+    count = count || 1;
+    const text = TextEditor.getLineAt(position).text;
+
+    // They may start on a number, so check the current word boundaries first.
+    let wordEnd = position.getCurrentWordEnd(true);
```
rebornix: I'm worried about relying on word boundaries as users are allowed to customize word separators, see https://github.com/VSCodeVim/Vim/issues/487 . Will this be a problem?

sectioneight: Interesting, yes that could be a problem. Is there going to be a different API to get "builtin" word boundary matching? Should I just use a custom regex for this movement?

src/actions/actions.ts
```diff
@@ -2694,3 +2694,76 @@ class ToggleCaseAndMoveForward extends BaseMovement {
...
+      wordStart = wordStart.getWordRight();
+      wordEnd = wordStart.getCurrentWordEnd();
+    } while (true);
+
+    // No usable numbers, return the original position
+    return position;
+  }
+
+  public async replaceNum(start: number, offset: number, startPos: Position, endPos: Position): Promise<Position> {
+    const oldWidth = start.toString().length;
```
rebornix: this won't work for `0x9` as well, maybe we should pass the length into this method as well.

sectioneight: Woah, I had no idea Vim support hex increments.

I think to handle this we can check if the string starts with "0x" and then add a helper to convert it to base16 and prefix with "0x".

Are there other numeric formats that Vim supports?

src/actions/actions.ts
rebornix: We can use `getWord` as it always reads from user settings or default, I suppose want to want to get is something starts with numbers or `+` and `-`, a custom regex may do the trick.

src/actions/actions.ts
rebornix: I mean when we handle hex value, `lenght` of its deximal value is not a right one.

src/actions/actions.ts
sectioneight: Looks like all of the current Position methods that take regexp's are private -- do you think I should change the visibility, or subclass `Position` to something like `NonCustomizablePosition` that does **not** respect user settings for word boundaries?

johnfn: If a user is insane enough to define numbers as word separators, these motions make no semantic sense (is 1234 4 separate words or one? If we select the whole thing with visual mode and do ctrl a, what should happen?) 

I'm completely fine with this being undefined behavior.

sectioneight: Sounds good to me.

src/actions/actions.ts
sectioneight: Looks like JS doesn't parse octals like Vim does. I think for simplicity's sake I'll start with just handling hex.

Are there other cases where parsing numbers respecting bases would be useful? I think the sanest way is to create a class that handles preserving the string representation, but I'm not sure where to put it.

sectioneight:
Updated with a `NumericString` class to handle hex -- I have no idea if that's the right place to put it or if I should just make some private inner class (do those even exist in TS?)

sectioneight:
The only thing left that this PR does not implement (trying to understand how useful it is):

given
```
0x0
```

hitting `ctrl+x`

gives you

```
0xffffffffffffffff
```

johnfn:
@sectioneight that seems of very little use. It doesn't make a lot of sense to guess how many bits are in the user's variable. 😛

extension.ts
```diff
@@ -179,7 +179,7 @@ export async function activate(context: vscode.ExtensionContext) {
     showCmdLine("", modeHandlerToEditorIdentity[new EditorIdentity(vscode.window.activeTextEditor).toString()]);
   });
 
-  'rfbducw['.split('').forEach(key => {
+  'rfbducw[ax'.split('').forEach(key => {
```
sectioneight: is there any specific ordering to the strings here? should we alphabetize them? reverse-alphabetize? put our favorite ones first?

package.json
```diff
@@ -93,6 +93,16 @@
                 "when": "editorTextFocus && vim.useCtrlKeys"
             },
             {
+                "key": "ctrl+a",
```
sectioneight: Note that there's already a `ctrl+a` action in vscode, which seems to hop between the beginning of the line and the first non-whitespace char on the line.

I don't really want to disable that (least surprising), but people may be confused if they ctrl+a on a line and some text doesn't match so their cursor is moved to the beginning.

test/testSimplifier.ts
```diff
@@ -162,6 +162,9 @@ function tokenizeKeySequence(sequence: string): string[] {
 
     if (char === '>') {
       isBracketedKey = false;
+      if (key.startsWith("<c-")) {
+        key = "ctrl+" + key.substring(3, key.length - 1);
```
sectioneight: so, this is pretty gross, and i'm not sure if we want people to be able to turn `<c-abc>` into `ctrl+abc`. but at least it removes the `TODO hack` from the other place.

extension.ts
johnfn: hah! there's no order, but yeah, this thing is pretty dumb. honestly i think we should parse them directly out of package.json and DRY this up, but that's a job for another day.

package.json
johnfn: Do the thing like we do elsewhere in package.json where we only enable these 2 things if ctrlKeys is turned on.

sectioneight: I wasn't quite sure what that was for, since a lot (some?) of the other ones don't have that guard.

sectioneight: Ok, I added it here but not for `ctrl+x`... where did we land? Both? Neither?

src/number/numericString.ts
```diff
@@ -0,0 +1,36 @@
+export class NumericString {
```
johnfn: I think you could actually write this whole thing very concisely (less than 10 lines) with 3 separate regexes. Just do grouped matches to grab the values and then send them through parseInt. Something like this:

```{ 
    8: /^0([0-7]+),
    10: /^(\d+)/,
    16: /0x([\da-fA-F]+)/
}```

File: src/actions/actions.ts [hunk 8]
```diff
@@ -2694,3 +2695,76 @@ class ToggleCaseAndMoveForward extends BaseMovement {
...
+  offset: number;
+
+  public async execActionWithCount(position: Position, vimState: VimState, count: number): Promise<Position> {
+    count = count || 1;
+    const text = TextEditor.getLineAt(position).text;
+
+    // They may start on a number, so check the current word boundaries first.
+    let wordEnd = position.getCurrentWordEnd(true);
+    let wordStart = position.getWordLeft(true);
+    do {
```
johnfn: Would you mind abstracting out a function in the Position class called something like IterateWords in the same style of IterateDocument and IterateLineStart?

src/actions/actions.ts
```diff
@@ -2694,3 +2695,76 @@ class ToggleCaseAndMoveForward extends BaseMovement {
...
+  }
+
+  public async replaceNum(start: NumericString, offset: number, startPos: Position, endPos: Position): Promise<Position> {
+    const oldWidth = start.toString().length;
+    start.value += offset;
+    const newNum = start.toString();
+
+    const range = new vscode.Range(startPos, endPos.getRight());
+
+    if (oldWidth === newNum.length) {
```
johnfn: I think it's better to only have the else case here.

src/actions/actions.ts
```diff
@@ -2694,3 +2695,76 @@ class ToggleCaseAndMoveForward extends BaseMovement {
...
+    let wordEnd = position.getCurrentWordEnd(true);
+    let wordStart = position.getWordLeft(true);
+    do {
+      // '-' doesn't count as a word, but is important to include in parsing the number
+      if (text[wordStart.character - 1] === '-') {
+        wordStart = wordStart.getLeft();
+      }
+      const word = text.substring(wordStart.character, wordEnd.character + 1);
+      // Strict number parsing so "1a" doesn't silently get converted to "1"
+      const num = NumericString.parse(word);
```
johnfn: Can you have parse return string | undefined rather than NumericString | undefined?

src/number/numericString.ts
sectioneight: Good call, updated with what I think you meant

src/actions/actions.ts
sectioneight: We need the functionality below, so that when we re-hydrate it it respects the base and the prefix (otherwise the numbers all come out decimal)

src/actions/actions.ts
sectioneight: I'm cool with that, although visually it does look worse.

src/actions/actions.ts
johnfn: Oh, duh.

src/actions/actions.ts
johnfn: Ugh, that's annoying. Alright, we can keep it this way for now.

src/mode/modeHandler.ts
```diff
@@ -503,8 +503,6 @@ export class ModeHandler implements vscode.Disposable {
   }
 
   async handleKeyEvent(key: string): Promise<Boolean> {
-    if (key === "<c-r>") { key = "ctrl+r"; } // TODO - temporary hack for tests only!
```
johnfn: Did someone fix this?

src/number/numericString.ts
```diff
@@ -0,0 +1,32 @@
...
+  prefix: string;
+
+  private static matchings: { regex: RegExp, base: number, prefix: string }[] = [
+    { regex: /^0([0-7]+)$/, base: 8, prefix: "0"},
+    { regex: /^(\d+)$/, base: 10, prefix: ""},
+    { regex: /^0x([\da-fA-F]+)$/, base: 16, prefix: "0x"},
+  ];
+
+  static parse(input: string): NumericString | null {
+    for (let matcher of NumericString.matchings) {
```
johnfn: You can use destructuring here

`for (let { regex, base, prefix } of NumericString.matchings)`

src/number/numericString.ts
johnfn: Yeah, LGTM, though maybe I slightly exaggerated how short it was going to be.

test/testSimplifier.ts
```diff
@@ -162,6 +162,9 @@ function tokenizeKeySequence(sequence: string): string[] {
 
     if (char === '>') {
       isBracketedKey = false;
+      if (key.startsWith("<c-")) {
```
sectioneight: @johnfn this is where I "fixed" it -- lmk if this looks sane

src/mode/modeHandler.ts
sectioneight: See https://github.com/VSCodeVim/Vim/pull/515/files#diff-bf5fd3f226adf3a0b4a4f0224184aa7fR165

src/number/numericString.ts
sectioneight: Ah, neat. Updating.

package.json
johnfn: I think both is right. It would be weird for one to work by default but not the other.

test/testSimplifier.ts
johnfn: Ahh, I understand now.

Here's the thing. We have to have this hack because currently we're inconsistently handling ctrl keys between our tests and our infrastructure. We should be using &lt;c-*&gt; everywhere. 

As insane as it sounds, I'd actually prefer you to NOT fix it here and instead make the hack bigger, so that it becomes an increasingly large eye-sore which will cause people (mostly me) to want to fix it the correct way.

src/actions/actions.ts
```diff
@@ -2694,3 +2695,65 @@ class ToggleCaseAndMoveForward extends BaseMovement {
...
+
+  public async execActionWithCount(position: Position, vimState: VimState, count: number): Promise<Position> {
+    count = count || 1;
+    const text = TextEditor.getLineAt(position).text;
+
+    // They may start on a number, so check the current word boundaries first.
+
+    for (let { start, end, word } of Position.IterateWords(position.getWordLeft(true))) {
+      // '-' doesn't count as a word, but is important to include in parsing the number
+      if (text[start.character - 1] === '-') {
```
johnfn: Use `word` here rather than text, start and end.

test/testSimplifier.ts
sectioneight: Gotcha, updated.

src/mode/modeHandler.ts
```diff
@@ -504,6 +504,8 @@ export class ModeHandler implements vscode.Disposable {
 
   async handleKeyEvent(key: string): Promise<Boolean> {
     if (key === "<c-r>") { key = "ctrl+r"; } // TODO - temporary hack for tests only!
+    if (key === "<c-a>") { key = "ctrl+a"; } // TODO - temporary hack for tests only!
+    if (key === "<c-x>") { key = "ctrl+x"; } // TODO - temporary hack for tests only!
```
johnfn: Beautiful. In the worst way possible.

src/actions/actions.ts
sectioneight: I can't, because I'm testing **outside** of the returned word (since `-` isn't included in the word boundaries)

package.json
sectioneight: Done.

src/actions/actions.ts
johnfn: Oh geez, that actually seems like it could have some bad interactions with people who DONT use - as a word delimiter. It's also really annoying. Oh well, it's fine for now.

johnfn:
Ok, sweet. Just remove the hack fix and we should be good to go.

sectioneight:
Remove which hack fix? I think I already undid the changes to tokenization.

johnfn:
Yup, you're right, I must have been looking at an old version of the code. 

Sweet. LGTM!

johnfn:
(Thanks for putting up with all my crazy requests! 😁 )

sectioneight:
haha all good, I'm the same way on code reviews at work

PR #525, opened 2016-07-27
Substitute with no range or marks

### LINKED ISSUE #512: Support :s/foo/bar/g style substitutions when text is selected
* VSCode Version: 1.3.1
* VsCodeVim Version: 0.1.3
* OS: Ubuntu 16.04

rebornix:

Fix #512. 

We support followings substitute commands

- [x] `:s/old/new/flags`
- [x] `'<,'>s/old/new/flags`

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -332,7 +332,7 @@ Status | Command | Description
 :white_check_mark:   | :1234:  ={motion}           | filter the lines that are moved over through 'equalprg'
    | :1234:  ==	                 |    filter N lines through 'equalprg'
 :white_check_mark:  |    {visual}=                | filter the highlighted lines through 'equalprg'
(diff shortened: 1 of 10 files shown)


johnfn:
The way that it works in vim is if you have a selection, Vim automatically inserts `'<,'>`, which means "run this ex command on a selection." However, you can backspace that and run an ex command not on a selection, if you made a mistake.

Would it be possible to insert those characters into the ex command line in vscode only if there's a selection, then check for them (rather than the existence of a selection) to see if the ex command operates on a selection or not?

rebornix:
@johnfn yeah, I didn't implement `<>` in the first beginning. Let me bring it in.

rebornix:
Support subsitution with no range or marks (now we only support `'<` and `'>`).

johnfn:
@rebornix is this done? sorry, i think i missed your previous message!

rebornix:
yup, it contains everything in description.

PR #545, opened 2016-08-02
update roadmap

rebornix:
Update roadmap to align with implemented commands.

@sectioneight I saw you marked those commands you added previously as `:warning:`, but I don't know whether it's due to Code's API limit or not. Do you mind marking them as `:warning: :x:` if it's blocked by Code? Or you can just point it out here and I'll update my PR.


johnfn:
I believe @sectioneight was just cautious about marking things as complete without fully testing them and they are in fact done. :) 

Thanks for updating the grid, @rebornix!

sectioneight:
Yea, just being cautious. Things should be all working except for some
visual selection issues that I have a WIP PR open to fix

PR #546, opened 2016-08-02
Add support for S

### LINKED ISSUE #536: Shift+S doesn't work
In normal mode in VIM, i can press SHIFT+S on any line, and it'll clear the line, put the cursor at the correct indentation and switch to INSERT mode. This doesn't happen here.

It'd be fantastic if this could be added.

I'm currently using version 0.1.4.

Thanks

glibsm:
Fixes #536


src/actions/actions.ts
```diff
@@ -1440,6 +1440,24 @@ class CommandInsertNewLineBefore extends BaseCommand {
 }
 
 @RegisterAction
+class CommandClearLine extends BaseCommand {
+  modes = [ModeName.Normal];
+  keys = ["S"];
+
+  public async exec(position: Position, vimState: VimState): Promise<VimState> {
+    const lineStart = position.getFirstLineNonBlankChar();
```
johnfn: Why firstLineNonBlankChar? S eats up the whole line, start to end.

glibsm: @johnfn That wasn't my understanding from #536, or from my usage of vim.

For example, consider a sample function:

```
 def func(one, two):
     print o|ne
     print two
```

By pressing `S` I would expect to replace the second line with the correct indent, which this accomplishes.

Just nuking the whole line seems less useful, and not what my terminal vim does.

I'm happy to change to first char, but is that really desired?

glibsm: @johnfn did some more testing and digging through manuals. `c-c` aka `S` is defined to "replace the whole line".

However, my 7.3 VIM and Atom vim-plus replace to the first non-blank character as I have described. 

VSCode `c-c` does the whole line replace. In order to be consistent within VSCode, we should either make `S` do whole line, or update `c-c` to point to first non-white.

rebornix:
My two cents here:

1. About the implementation. Since `S` is just a Synonym for `cc` and we do support `cc` in Code (a `c` operator then a `c` movement), you may want to just leverage these two existing actions to implement `S`. This way we can avoid difference between synonyms.
2. About the behavior. Both you and @johnfn are correct, it depends on `autoindent` option. If `autoindent` it turned on, it will always keep the indentation of first changed lines. If it's off, it will delete the whole line as @johnfn described. 

I think right now you just keep `S` the same as current Code's `cc` (by leverage existing actions) and once #508 is merged, you can send another PR to get the indentation problem fixed? I'll say that's another feature :)

rebornix:
Lastly, please update the roadmap about `S`.

src/actions/actions.ts
johnfn: Heh, for a second there I was confused because when I see c-c I normally thing ctrl-c. Took me a sec to realize you meant `cc`!

I'm running Vim 7.4 on OSX and I see the "full line including whitespace" behavior. It's interesting that your Vim has different behavior. In cases like this I think it's best to defer to the manual and remove the whole line.

If you want "change starting at whitespace", there's always `^C`, which is exactly the same number of characters as `cc` :-)

johnfn: Ahh, @rebornix cleared this up beautifully. 

Thanks, @rebornix!

johnfn:
👍  on what @rebornix said.

Also, if you want you could do numeric prefixes on S as well. Not a requirement for getting this pulled in - we can always do it later, of course.

glibsm:
@rebornix Thank you for the explanation, that makes total sense. Completely forgot about the `autoindent` .

I'll change this PR for `S` to be an alias of `cc` and update the roadmap.

johnfn:
Ready to merge?

glibsm:
@johnfn ready if you are.

johnfn:
Looks great to me. Props on doing the numeric prefix as well! :-)

johnfn:
... Oh wait, there are merge conflicts. Could you fix those? Then we are really good to go. 😛

glibsm:
Whoops. Yeah, let me fix those

glibsm:
@johnfn should be good to go now without conflicts.

johnfn:
Awesome! Thanks, @glibsm!

mnsr:
Not sure if i should put this here, but the implementation of shift+S doesn't match standard VIM.

Shift+S adds indentation, in vscodevim it does not.

So if if i'm inside a function like this:

```
function Test() {
// << Shift+S should put the correct indentation here and switch to insert mode
}

```

johnfn:
@mnsr could you please open up an issue for that bug? Thanks. :)

mnsr:
@johnfn yeah np

glibsm:
As discussed earlier, it's a bug with auto indent. Has it been fully implemented already?

rebornix:
@glibsm sorry for the late response and yes, it's already covered by #629. Kudos to @sectioneight .

sectioneight:
Ha, small world. @glibsm and I sit right next to each other at work :)

PR #547, opened 2016-08-02
Support gd

johnfn:
Finish off #530

### DIFF
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -1337,6 +1337,20 @@ class CommandExitVisualLineMode extends BaseCommand {
   }
 }
 
+@RegisterAction
+class CommandGoToDefinition extends BaseCommand {
+  modes = [ModeName.Normal];
+  keys = ["g", "d"];
+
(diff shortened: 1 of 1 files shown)


src/actions/actions.ts
```diff
@@ -1343,10 +1343,28 @@ class CommandGoToDefinition extends BaseCommand {
...
 
+    // Unfortuantely, the above does not necessarily have to have finished executing
+    // (even though we do await!). THe only way to ensure it's done is to poll, which is
+    // a major bummer.
+
+    await new Promise(resolve => {
+      let interval = setInterval(() => {
+        const positionNow = Position.FromVSCodePosition(vscode.window.activeTextEditor.selection.start);
+
+        if (!startPosition.isEqual(positionNow)) {
```
jordanlewis: Will it be a problem if the `gd` happens to be a no-op movement?

johnfn: Ah, that was rather mindless of me. Opened #553 as a self-reminder.

PR #551, opened 2016-08-02
Fix incorrect cursor location after deleting linebreak (fixes #550)

### LINKED ISSUE #550: In insert mode, deleting at the start of a line should not move the cursor to the end of the new line
Given the following text and cursor in insert mode:

```
abcd
|efgh
```

Pressing backspace should result in the following text and cursor placement:

```
abcd|efgh
```

Instead, the cursor is placed at the *end* of the new line:

```
abcdefgh|
```

* VSCode Version: 1.3.1
* VsCodeVim Version: 0.1.4
* OSX 10.11

This behavior appears to be defined [here](https://github.com/VSCodeVim/Vim/blob/master/src/actions/actions.ts#L598). This seemed like an obvious bug report, but I couldn't find an existing one - sorry if this is a duplicate!

thomasboyt:
This fixes the issue I reported in #550, and adds a test to prevent regressions.

### DIFF
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -590,13 +590,15 @@ class CommandInsertInInsertMode extends BaseCommand {
     if (char === "<backspace>") {
       if (position.character === 0) {
         if (position.line > 0) {
+          const prevEndOfLine = position.getPreviousLineBegin().getLineEnd();
+
           await TextEditor.delete(new vscode.Range(
             position.getPreviousLineBegin().getLineEnd(),
             position.getLineBegin()
           ));
 
-          vimState.cursorPosition      = position.getPreviousLineBegin().getLineEnd();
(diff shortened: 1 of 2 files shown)


thomasboyt:
Looks like the 8 failing tests here are also failing in master.

src/actions/actions.ts
```diff
@@ -590,13 +590,15 @@ class CommandInsertInInsertMode extends BaseCommand {
     if (char === "<backspace>") {
       if (position.character === 0) {
         if (position.line > 0) {
+          const prevEndOfLine = position.getPreviousLineBegin().getLineEnd();
```
rebornix: :100: This fix is truly representative. We should be carefull with `TextEditor.delete`, etc as they have kind of `side effect`.

johnfn:
Thanks for the PR, @thomasboyt!

PR #560, opened 2016-08-03
Fix test regression

rebornix:
Yay! We love PRs! 🎊

As title. Since we still have test failure in local (those `god name` ones), sometimes we fail to realize our code change brings more failure.


rebornix:
The first build job passed while the second said `1) register Can use two registers together: ...` failed.

jpoon:
I restarted the failed build job. We'll see if that helps.

johnfn:
Hmm, looks like it may be a real be unrelated bug on OSX. In any case, it's not related to this PR, which is fine. Thanks @rebornix!

PR #575, opened 2016-08-05
Fix binding of control-keys

sectioneight:
Seems to be a regression from recent refactor. Tested manually in debug
extension. Prior to this commit, ctrl+a/ctrl+x didn't work. Now they do.

### DIFF
--- a/extension.ts
+++ b/extension.ts
@@ -189,7 +189,7 @@ export async function activate(context: vscode.ExtensionContext) {
 
   for (let { key } of packagejson.contributes.keybindings) {
     if (key.startsWith("ctrl+")) {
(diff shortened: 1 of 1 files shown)


sectioneight:
cc @johnfn I think this was caused by e5499335c3b74f4b8c28961090b88f3e3768c95f, since the `key` variable now already has the `ctrl+` prefix in it.

jpoon:
@sectioneight beat me to it. thanks! :+1:

johnfn:
Oh yikes! Thanks. This is a testing blindspot unfortuantely.

johnfn:
Good thing you caught this before our next release...

PR #577, opened 2016-08-06
OS X non-global key repeat fix

jimray:
The current README has the user disable OS X’s “press and hold” feature at a global level. It’s possible to disable this on a per-app basis, though, provided you know the name of an app’s .plist file.

This change will disable “press and hold” only for VSCode, not globally.


johnfn:
Thanks, @jimray!

Issue #479, opened 2016-07-19
Insert-mode selections do not work


tobia:
### What did you do?

Select some text while in insert mode, for example with the mouse or Shift+arrow.

Then type something that operates on the text (that works in vanilla VS Code) for example `(` to surround with parentheses.

### What did you expect to happen?

I expected the editor to go into some sort of Insert-mode selection, more similar to regular VS Code selection that Vim modes, where I could type into the selection to replace it, or type `(`, `[`, `{`, `'`, `"`, backtick, etc. to surround the selection with pairs of characters.

### What happened instead?

The editor went straight into Vim visual mode, where those keys don't do anything.

### Technical details:

* VSCode Version: 1.3.1
* VsCodeVim Version: 0.1.3
* OS: Mac OS X

rebornix:
@tobia just one stupid question, this is not related to Vim, right? It's all about how we reuse VS Code functionality in Insert mode?

tobia:
@rebornix It's a condition where this extension removes functionality that was present in vanilla VS Code (modeless selection + keys to wrap in parentheses) without providing a replacement.

tobia:
It also breaks snippets that have variable parts, unless I'm missing something.

killercup:
I've also had problems with that. There are some vim plugins to surround a selection with braces/quotes/etc, and it would be great if there was a nice way to do this in vscode as well.

That being that, I'd also be happy with an option to _not_ enter visual mode when selecting text with the cursor in insert mode. That may sound crazy, as there will be two different "select text" modes, but it would be a way to preserve more built-in vscode behaviour.

tobia:
Vim itself has this mode, at least in some of its GUI flavors. It's entered by selecting with the mouse or with shift+arrows while in insert mode, is labeled `-- (insert) VISUAL --` and it goes back to insert mode when exited or acted upon, not to normal mode.

See for example http://stackoverflow.com/questions/16203524/vims-insert-visual-mode

everpointer:
I kind of have the same problem, when using snippets with variable label. When snippet tab stops at label variable, it turned into visual mode which broke the snippets operation.
I keep uninstalling vim extension just because I can't use snippet well.

johnfn:
@everpointer ahh, that is definitely an issue. will definitely see if we can get snippets working for you!

rebornix:
@johnfn I suppose https://github.com/Microsoft/vscode/issues/8093 covers this issue, right?

[...]

PR #587, opened 2016-08-10
[WIP] Multi-Cursor Mode

johnfn:
The most requested feature in the history of VSCodeVim... Coming to a Visual Studio Code near you! 💥 🎆 💣 💀 

### Todo:

- [x] all motions
- [x] d
- [x] i
- [x] c
- [x] y
- [x] I broke insert in insert mode.. :( 
- [x] Only the current cursor is kept within bounds.
- [x] d$ with two cursors on same line breaks
- [x] Insert with 2 cursors on same line has trouble
- [x] I think that count actions are broken
- [x] o and O don't work and probably other text-inserting actions don't work either.
- [x] see if i can rewrite the text edit collapsing to take advantage of builtin vscode functions
- [x] backspace doesn't work over newlines
- [x] when 2 block cursors combine, they turn into visual 
- [x] esc can't be repeated for every cursor, because it's not idempotent. But it can't be run only once because every cursor needs to shift position. Interesting.
- [x] When you do 0 on a line with 2 cursors, it doesn't go into visual multi-cursor mode. 
- [x] Paste is broken
- [x] u should set ALL cursors back.
- [x] go through commands and figure out which ones plausibly could work with multi-cursor mode
- [x] Visual Block Mode has issues now - I should add tests
- [x] 50 failing tests

### Stuff which should be fixed before release
- [x] n/N doesn't work (some cursors are lost)
- [x] jj doesn't work
- [x] r doesn't work (this is super weird!)
- [x] Cursors can stack and then multiple operations are applied at the same location (this is because allCursors is an array, not a set. Should it be a set? I think it should be a set.)
- [x] Clicking in multi-cursor mode causes wonky behavior
- [x] selection breaks after visual block mode delete
- [x] y doesn't work
- [ ] . doesn't work

### Stuff which is not crucial for release but should be fixed at some point:
- [ ] double actions like dd yy cc don't work. I knew this design decision would come back to bite me eventually. 
- [ ] vaw doesn't work, and probably other word motions don't work either. (except not all, because apaprently `va(` works.)
- [ ] visual line multi cursor doesn't even exist
- [ ] J doesn't work because it doesn't use the same text insertion mechanisms 
- [ ] Y doesn't work either
- [ ] C doesn't work
- [ ] S doesn't work
- [ ] s doesn't work
- [ ] marks work in an insane way that doesn't really work at all.

This is extra tricky. If you leave a mark `a` in multi cursor mode, we should probably store an array of all the positions that you left. Then if you are in normal mode, but then do `ma`, you would go into multi cursor mode at where all the a marks were left.

This also makes me wonder if I could do something like `gma` (syntax pending) to add an additional location to the marks list. That seems really confusing for people other than me, however.

- [ ] substitution commands s/blah/foo/g don't really work across multiple selections.
- [ ] Replace mode is broken


johnfn:
Update: This is now more-or-less feature complete and I think it's pretty cool. (What other multi cursor mode allows you to do a / search independently on each cursor??)

Now there are just some bugs I need to hunt down and a bunch of test cases that are failing for some reason...

ZombineDev:
Hi and thanks for the good work! Can you tell me if this PR will provide support for [this](https://github.com/terryma/vim-multiple-cursors) style multiple cursors?

johnfn:
Hi @ZombineDev! I haven't directly ported that plugin or anything, so don't expect those shortcuts to work necessarily, but the general functionality appears to be the same. Is there anything in particular that you're hoping we add?

ZombineDev:
TBH, I'm not sure what featureset this PR is meant provide, so I don't know how it compares with that vim plugin. What I find most useful is  using `Ctrl+n`, `Ctrl+p` and `Ctrl+x` for selecting, deselecting and skipping the next word occurrence (though IIRC VS Code 1.4 added something similar). Another quite powerful thing is `MultipleCursorsFind` which allows you to place multiple cursors at all matches of a regular expression. You can see from the gifs at the plugin's readme page how it is meant to be used.

johnfn:
Alright, this is done!

I mean, there's still a bunch of work left to do, but it works reasonably well, and preserves all existing functionality, somehow, which is somewhat of a miracle.

Anyways, @rebornix @jpoon @xconverge @sectioneight or anyone else who has time, feel free to give me feedback, though I realize that this PR is probably rather intimidating to review. 

The overview of what I did is, essentially:

* Wrap all commands in a giant for loop that loops over all cursors and runs the command on each one individually.
* Rewrite all insert/delete/replace actions such that they return transformation objects rather than being applied immediately. We then apply them all in parallel in an `edit()` block. (This is necessary because otherwise previous changes could affect the positions of later ones.)

rebornix:
This pr fixes #790 as well.

rebornix:
We need to update `engine` section to `^1.5.0` to have `setLastHistoryEndPosition` and get Travis passed.

rebornix:
Verified: This PR fixes code snippet issue mentioned in https://github.com/VSCodeVim/Vim/issues/479#issuecomment-233681126

johnfn:
Ah yeah I need to do updates for Travis. Also, I didn't even intend to fix #479. That's awesome.

sectioneight:
This looks great @johnfn. I can't say I grok every line of it, though. If there's a particular section you'd like some extra eyes on I'm happy to help.

johnfn:
Wrapped up in #811.

oliverjhn:
Is this ever going to be revisited? It seems like it kind of got left half-finished....

GRURoot:
curl https://api.cloudflare.com/client/v4/accounts/$ACCOUNT_ID/tokens/permission_groups \
    -H "Authorization: Bearer $CLOUDFLARE_API_TOKEN"

PR #591, opened 2016-08-11
fix C or cc kill the empty line

### LINKED ISSUE #556: C or cc will kill the empty line
### What did you do?

I have these three line of codes:

```
export const options = {

};
```

The second line is empty, without any tab or spaces, my cursor on line 2 and then I'm hitting `C` or `cc`.

Just simply turn into `INSERT` mode like vim do

### What happened instead?

VsCodeVim kill the second line, and the cursor goes before `}`;

```
export const options = {
};
```

### Technical details:

* VSCode Version: 1.3.1
* VsCodeVim Version: *0.1.4*
* OS: OS X 10.11.6

shotaAkasaka:
fixes #556

### DIFF
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -1030,14 +1030,17 @@ export class ChangeOperator extends BaseOperator {
 
     public async run(vimState: VimState, start: Position, end: Position): Promise<VimState> {
         const isEndOfLine = end.character === TextEditor.getLineAt(end).text.length - 1;
-        const state = await new DeleteOperator().run(vimState, start, end);
-        state.currentMode = ModeName.Insert;
+        let state = vimState;
 
         // If we delete to EOL, the block cursor would end on the final character,
         // which means the insert cursor would be one to the left of the end of
         // the line.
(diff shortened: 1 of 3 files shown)


src/actions/actions.ts
```diff
@@ -1030,14 +1030,17 @@ export class ChangeOperator extends BaseOperator {
...
 
         // If we delete to EOL, the block cursor would end on the final character,
         // which means the insert cursor would be one to the left of the end of
         // the line.
         if (isEndOfLine) {
+          state = await new DeleteOperator().run(vimState, start, end);
+          state.currentMode = ModeName.Insert;
           state.cursorPosition = state.cursorPosition.getRight();
+        } else {
+          state.currentMode = ModeName.Insert;
```
rebornix: I suppose you want to move the delete operation here instead of `isEndOfLine` branch?

shotaAkasaka: Thanks for your comment. it fixed bea29a2.

johnfn:
Great change! Thanks, @shotaAkasaka.

PR #592, opened 2016-08-11
Sentence selection

rebornix:
Yay! We love PRs! 🎊

The code becomes complex when dealing with Vim' special logic of leading/trailing spaces handling.

### DIFF
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -2308,7 +2308,7 @@ class MovePreviousSentenceBegin extends BaseMovement {
   keys = ["("];
 
   public async execAction(position: Position, vimState: VimState): Promise<Position> {
-    return position.getPreviousSentenceBegin();
+    return position.getSentenceBegin({forward: false});
   }
 }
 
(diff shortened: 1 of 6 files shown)


src/actions/actions.ts
```diff
@@ -2971,6 +2974,93 @@ class MovementIBigWordTextObject extends MovementAWordTextObject {
...
+
+  public async execAction(position: Position, vimState: VimState): Promise<IMovement> {
+    let start: Position;
+    let stop: Position;
+
+    const currentSentenceBegin = position.getSentenceBegin({forward: false});
+    const currentSentenceNonWhitespaceEnd = currentSentenceBegin.getCurrentSentenceEnd();
+
+    if (currentSentenceNonWhitespaceEnd.isBefore(position)) {
+      // The cursor is on a trailing white space.
```
rebornix: The thing is Vim has a weird mechanism of handling leading or trailing spaces. The logic should be shared between words, sentences or paragraphs but since Vim is handling word, sentence and paragraph 's boundary differently, I still don't find a good way to reduce these redundant `if-else` statements. Will try later.

PR #597, opened 2016-08-12
Add support Y in visual mode

### LINKED ISSUE #573: shift-Y support in visual mode
Currently Y does nothing. From the vim docs:

    {Visual}["x]y		Yank the highlighted text [into register x] (for
    			{Visual} see |Visual-mode|).  {not in Vi}

    {Visual}["x]Y		Yank the highlighted lines [into register x] (for
    			{Visual} see |Visual-mode|).  {not in Vi}

So Y yanks lines, while y yanks arbitrary text.

In visual line mode y and Y should be equivalent, so this is probably quite easy there, but I don't know about non-line mode.

shotaAkasaka:
Fixes #573

### DIFF
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -929,6 +929,16 @@ export class YankOperator extends BaseOperator {
 }
 
 @RegisterAction
+export class ShiftYankOperatorVisual extends BaseOperator {
+    public keys = ["Y"];
+    public modes = [ModeName.Visual];
+
+    public async run(vimState: VimState, start: Position, end: Position): Promise<VimState> {
+      return await new YankOperator().run(vimState, start, end);
+    }
+}
+
+@RegisterAction
 export class DeleteOperatorXVisual extends BaseOperator {
     public keys = ["x"];
     public modes = [ModeName.Visual, ModeName.VisualLine];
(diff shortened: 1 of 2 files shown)


jpoon:
Noice! 🎉 .

johnfn:
If you directly extend the YankOperator, then this will act as an operator, meaning it will be able to be combined with motions to do actions (d is an operator because you can do dw, and c is an operator because you can do c$, etc.) 

That's not what you want here. You want to create and execute a new YankOperator, running it across the entire current line.

Make sense? :)

shotaAkasaka:
@johnfn Thanks for yout comment.
I have to create anad execute new YankOperator,like that?

```
@RegisterAction
export class ShiftYankVisualModeOperator extends BaseOperator {
    public keys = ["Y"];
    public modes = [ModeName.Visual];

    public async run(vimState: VimState, start: Position, end: Position): Promise<VimState> {
      // execute 
    }
}
```

johnfn:
I mean something more along the lines of

`new YankOperator().run(vimState, position.getLineStart(), position.getLineEnd())`

(That is untested code...)

johnfn:
There we go. :) 

Thanks for the PR, @shotaAkasaka!

PR #608, opened 2016-08-15
change  "_" underscore motions direction, fix #602

### LINKED ISSUE #602: Fix for "_" underscore motions with counts
EDIT: Whoops, apparently this feature is implemented, but instead of down, if you put a count in front of `_` it goes up instead (so `2_` goes two lines up instead of down). If that could be fixed that would be appreciated!

--

I just found out about this vim feature! `_` and `^` aren't the same thing, as `_` supports the use of a count.

From the vim docs:

`  _  <underscore>         [count] - 1 lines downward, on the first non-blank character |linewise|.`

So `2_` would bring you to the first non-blank character 2 lines down.

If that could be supported that would be great!

* VSCode Version: 1.4.0
* VsCodeVim Version: 0.1.6
* OS: Windows 10 Pro

htwenning:
Yay! We love PRs! 🎊

- [ ] Commit message has a short title & issue references
- [ ] Each commit does a logical chunk of work.  
- [ ] It builds and tests pass (e.g `gulp tslint`)

### DIFF
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -1897,7 +1897,7 @@ class MoveUpUnderscore extends BaseMovement {
   keys = ["_"];
 
   public async execActionWithCount(position: Position, vimState: VimState, count: number): Promise<Position | IMovement> {
-    return position.getUpByCount(Math.max(count - 1, 0))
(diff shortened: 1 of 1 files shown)


johnfn:
Wonderful! Thanks, @htwenning!

PR #627, opened 2016-08-19
Allow regex in / search

### LINKED ISSUE #572: Find functionality does not support regex
### What did you do?

In a JS file, from normal mode:
`/\bfor\b.*\bin\b`[Enter]

Incremental search should match segments of lines with a for...in loop. e.g. `for (var a in stuff) {`

### What happened instead?

Search opened, all of my keystrokes appeared on the line, but matching did not appear to use regex at all.

Did not match anything at any point after typing `\b`.

Also tried using just `.*` patterns and that didn't work either.

### Technical details:

* VsCodeVim Version: 0.1.4
- VSCode Version: Code - Insiders 1.5.0-insider (1a7ca5a77bf6b63951e499ea93b7316d36118ade, 2016-08-05T07:19:53.669Z)
- OS Version: Windows_NT ia32 10.0.14393

sectioneight:
Fixes #572

### DIFF
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -796,7 +796,7 @@ export class CommandSearchBackwards extends BaseCommand {
   isMotion = true;
 
   public async exec(position: Position, vimState: VimState): Promise<VimState> {
-    vimState.searchState = new SearchState(SearchDirection.Backward, vimState.cursorPosition);
+    vimState.searchState = new SearchState(SearchDirection.Backward, vimState.cursorPosition, "", { isRegex: true });
     vimState.currentMode = ModeName.SearchInProgressMode;
 
     return vimState;
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -782,7 +782,7 @@ export class CommandSearchForwards extends BaseCommand {
   isMotion = true;
 
   public async exec(position: Position, vimState: VimState): Promise<VimState> {
-    vimState.searchState = new SearchState(SearchDirection.Forward, vimState.cursorPosition, "", true);
+    vimState.searchState = new SearchState(SearchDirection.Forward, vimState.cursorPosition, "", { isRegex: true });
     vimState.currentMode = ModeName.SearchInProgressMode;
 
     return vimState;
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -782,7 +782,7 @@ export class CommandSearchForwards extends BaseCommand {
   isMotion = true;
 
   public async exec(position: Position, vimState: VimState): Promise<VimState> {
-    vimState.searchState = new SearchState(SearchDirection.Forward, vimState.cursorPosition);
+    vimState.searchState = new SearchState(SearchDirection.Forward, vimState.cursorPosition, "", true);
     vimState.currentMode = ModeName.SearchInProgressMode;
 
     return vimState;
--- a/src/mode/modeHandler.ts
+++ b/src/mode/modeHandler.ts
@@ -179,6 +179,7 @@ export class SearchState {
 
   private _matchesDocVersion: number;
   private _searchDirection: SearchDirection = SearchDirection.Forward;
+  private isRegex: boolean;
 
   private _searchString = "";
   public get searchString(): string {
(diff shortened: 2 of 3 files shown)


johnfn:
My big concern is I'm not quite sure how well Vim regexp maps to JS regexp. Would you mind doing a little research to see if it is a fairly close mapping or not?

sectioneight:
Sounds good. Hopefully they're close...

johnfn:
I hope so too. Or close enough that you can use a meta regex to translate the normal regex to a js re... oh god, what am I saying?

rebornix:
@sectioneight the regex we use here is the same as Vim Substitution https://github.com/VSCodeVim/Vim/blob/7ac106d8708a1ad9918102abf0e59dbb0609cb4b/src/cmd_line/commands/substitute.ts#L64 . The catch about Regex is no one is following the standard (if there is a standard).

For now I'm using JS's regex in substitution, I'm not quite sure if ppl are satisfied with it but I don't see any complains yet (maybe they don't notice its existence yet).

You may want to write some JS regex to translate Vim's regex to JS's regex ...

sectioneight:
Yes, I've found:

https://github.com/othree/eregex.vim/blob/master/plugin/eregex.vim

To translate Vim regex to PCRE, but of course JS is not PCRE.

There's also https://en.wikipedia.org/wiki/Comparison_of_regular_expression_engines but it's pretty high level.

Sounds like a good first pass would be to use JS regex for search. But I'm open to suggestions.

johnfn:
Yiiiikes. If it's too complicated, let's just do JS regex for now and see if anyone complains. 😉

rebornix:
This PR is good to go IMO.

johnfn:
It still says WIP on it - @sectioneight is this good to go?

sectioneight:
Still need to add a test or two, but it sounds like you guys approve of the approach.

johnfn:
I do; I'm very much a fan of 80% solutions, especially when people very rarely use the other 20%.

sectioneight:
As am I. Will get some tests in tomorrow and ping you for review.

sectioneight:
Ok, ready for review @johnfn

src/actions/actions.ts
```diff
@@ -782,7 +782,7 @@ export class CommandSearchForwards extends BaseCommand {
   isMotion = true;
 
   public async exec(position: Position, vimState: VimState): Promise<VimState> {
-    vimState.searchState = new SearchState(SearchDirection.Forward, vimState.cursorPosition);
+    vimState.searchState = new SearchState(SearchDirection.Forward, vimState.cursorPosition, "", true);
```
johnfn: Minor, but I'm trying to move towards a style where we explicitly name all our boolean arguments. Something like `{ isRegex: true }` is what I'm thinking here.

johnfn:
Only one small comment.

I think this PR is going to have a big impact for such a small LOC difference! :)

src/actions/actions.ts
sectioneight: Sounds good.

sectioneight:
I'm going to update one more thing on this PR. Right now if you have a partial regex (e.g. `/a(`) it throws an error in the console. I need to try/catch the `SyntaxError` and just do a plain match if it doesn't compile.

src/mode/modeHandler.ts
```diff
@@ -280,10 +295,11 @@ export class SearchState {
     }
   }
 
-  constructor(direction: SearchDirection, startPosition: Position, searchString = "") {
+  constructor(direction: SearchDirection, startPosition: Position, searchString = "", { isRegex = false } = {}) {
```
sectioneight: I'm not really sure about this syntax (still a TS noob). Is this what you were thinking?

johnfn: This is actually pure ES6 JS syntax - it's parameter destructuring. I would say to do it without the default object assignment, so just `{ isRegex = false }`

johnfn: (Which will destructure the object immediately and assign isRegex = false if it can't find a value)

johnfn: Yeah, that looks fine to me!

sectioneight:
OK, I think we're finally ready to go here

sectioneight:
Ping for review

src/mode/modeHandler.ts
```diff
@@ -214,7 +215,21 @@ export class SearchState {
         ignorecase = false;
       }
 
-      const regex = new RegExp(search.replace(SearchState.specialCharactersRegex, "\\$&"), ignorecase ? 'gi' : 'g');
+      let searchRE = search;
+      if (!this.isRegex) {
+        searchRE = search.replace(SearchState.specialCharactersRegex, "\\$&");
```
johnfn: This is very cool. I never knew about \$&.

johnfn:
Thanks for the ping. Feel free to ping me on here/slack if I don't review within ~24h. That is almost certainly an indicator that it just slipped through the cracks.

johnfn:
And thanks for the PR, of course :)

shaunluttin:
Does this PR support searching for new lines? If so, how?

simahao:
@shaunluttin ,i have same problem. :(

johnfn:
No, it doesn't support newline search yet. Feel free to open an issue.

PR #629, opened 2016-08-19
Respect indentation on cc and S

### LINKED ISSUE #626: Shift+S doesn't add indentation
* VSCode Version: 1.4.0
* VsCodeVim Version: latest
* OS: Win10

The implementation of Shift+S doesn't match standard VIM.

Shift+S is supposed to:
1. Clear entire line under cursor
2. Add indentation
3. Switch to insert mode.

Currently, vscodevim doesn't do #2.

So if if i'm inside a function like this:

```
function Test() {
  var greeting = "hello";
// << Shift+S should put the correct indentation here (2 spaces) and switch to insert mode
}

```

sectioneight:
This uses a new `autoindent` setting (default true) to determine whether
a line replacement should cut to the beginning of the line, or the first
non-whitespace char.

Fixes #626

Two things to note:

- We're not actually doing any smart indentation, we're just assuming they want to keep the original indentation that was on the line. This appears to be how Vim behaves as well.
- In Vim, `autoindent` is not enabled by default, but I think most people enable it and VSCode obviously has it on by default as well.

### DIFF
--- a/test/mode/modeNormal.test.ts
+++ b/test/mode/modeNormal.test.ts
@@ -1153,7 +1153,7 @@ suite("Mode Normal", () => {
       title: "Can do S",
       start: ["    one", "    tw|o", "    three"],
       keysPressed: "2S",
-      end: ["    one", "|"]
+      end: ["    one", "        |"]
     });
 
     newTest({
@@ -1208,6 +1208,14 @@ suite("Mode Normal", () => {
       start: ["one <blink>he|llo</blink> two"],
       keysPressed: "cat",
       end: ["one | two"],
+      endMode: ModeName.Insert
+    });
+
+    newTest({
+      title: "Respects indentation with cc",
+      start: ["{", "  int| a;"],
+      keysPressed: "cc",
+      end: ["{", "  |"],
       endMode: ModeName.Insert
     });
 });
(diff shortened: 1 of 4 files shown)


test/mode/modeNormal.test.ts
```diff
@@ -1153,7 +1153,7 @@ suite("Mode Normal", () => {
       title: "Can do S",
       start: ["    one", "    tw|o", "    three"],
       keysPressed: "2S",
-      end: ["    one", "|"]
+      end: ["    one", "        |"]
```
sectioneight: This is the weirdness mentioned in the PR. I don't know enough about `testSimplifier` to understand why the behavior is different there.

sectioneight:
Even weirder, it works correctly on travis.

sectioneight:
Confirmed that Vim's behavior is not to try to find the "correct" level of indentation, but just to respect what's currently on the line. Which is great, because it makes this change very simple.

johnfn:
Thanks, @sectioneight!

glibsm:
> Confirmed that Vim's behavior is not to try to find the "correct" level of indentation, but just to respect what's currently on the line.

😮 

image

@johnfn I will be waiting for your written apology ;)

Just kidding. Glad we got this cleared this up.

johnfn:
@glibsm hahaha, yeah, now that we have the ability to flip it on/off via settings, I'm not as concerned. :)

Issue #613, opened 2016-08-15
:wq doesn't work when there are changes (does nothing)


Mehuge:
------

Please *thumbs-up* 👍 this issue if it personally affects you! You can do this by clicking on the emoji-face on the top right of this post. Issues with more thumbs-up will be prioritized.

-----

### What did you do?

`:wq<CR>`

### What did you expect to happen?

`:wq<CR>` should perform a `:w` then a `:q`

If I `:w` first (to save changes) or if no changes have been made, then `:wq` does quit

### What happened instead?

nothing at all, changes not saved, editor window not closed

### Technical details:

* VSCode Version: 1.4.0
* VsCodeVim Version: 0.1.7
* OS: Windows

johnfn:
It works now! :)

PR #630, opened 2016-08-19
Fix bug #613 ":wq command dows not work"

Platzer:
Bug fix is currently not working with Visual Studio Code - Insiders
>Version 1.5.0-insider
>Commit 3d9622208b2167921b6664e04fd1fbd51507a4a0  

it seems that the `isDirty` property of the `vscode.TextDocument` is set false for newly created document. Going to check it tomorrow with normal vscode.

### DIFF
--- a/src/cmd_line/commands/write.ts
+++ b/src/cmd_line/commands/write.ts
@@ -37,7 +37,7 @@ export class WriteCommand extends node.CommandBase {
     return this._arguments;
   }
 
-  async execute(modeHandler : ModeHandler) {
+  async execute(modeHandler : ModeHandler) : Promise<void> {
     if (this.arguments.opt) {
       util.showError("Not implemented.");
       return;
@@ -74,7 +74,7 @@ export class WriteCommand extends node.CommandBase {
     }
   }
 
-  private async save(modeHandler : ModeHandler) {
+  private async save(modeHandler : ModeHandler) : Promise<void> {
     await this.activeTextEditor.document.save().then(
       (ok) => {
         modeHandler.setupStatusBarItem('"' + path.basename(this.activeTextEditor.document.fileName) +
--- a/src/cmd_line/commands/write.ts
+++ b/src/cmd_line/commands/write.ts
@@ -37,7 +37,7 @@ export class WriteCommand extends node.CommandBase {
     return this._arguments;
   }
 
-  execute(modeHandler : ModeHandler) : void {
+  async execute(modeHandler : ModeHandler) {
     if (this.arguments.opt) {
       util.showError("Not implemented.");
       return;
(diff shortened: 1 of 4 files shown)


Platzer:
It was too late yesterday didn't get it how `isDirty` and `isUntitled` are ment to be used.
The changes don't include the save as dialog, should we consider a implementation in the `WriteCommand`?

johnfn:
No save as dialog is required because Vim doesn't use one. :) 

Do you think this is good to go, or do you need any help?

Platzer:
I would like to implement the tests for `w`, `q` and `wq` but i need some time to figure out how the test framework works ;-) and how to implement it best.

Platzer:
hey guys (@johnfn, @rebornix) i'm not sure if i get how async work (i'm realy new to ts / js programming), could you please have a look if i'm doing something stupid here?

johnfn:
Nothing looks horrifically stupid to me. :) Anything in particular you want us to look at?

Platzer:
No nothing special, just a bit insecure about the async/await promise stuff ;). I just cleaned up the test code a bit (cut out the ugly polling). I think it's good to go.

johnfn:
:) you aren't the only one, we were having a discussion in slack about exactly that earlier today!

src/cmd_line/commands/write.ts
```diff
@@ -37,7 +37,7 @@ export class WriteCommand extends node.CommandBase {
     return this._arguments;
   }
 
-  execute(modeHandler : ModeHandler) : void {
+  async execute(modeHandler : ModeHandler) : Promise<void> {
```
johnfn: Actually, this one doesn't need to be async because you don't do any awaiting inside it.

Platzer: i need it to return the encapsulating promise from the `activeTextEditor.document.save()` call in the save method. otherwise i wasn't able to call `write` cmd and `quit` cmd synchronously.

src/cmd_line/commands/write.ts
```diff
@@ -56,27 +56,26 @@ export class WriteCommand extends node.CommandBase {
...
-              modeHandler.setupStatusBarItem(e.message);
-            } else {
-              this.save(modeHandler);
-            }
-          });
-        } else {
-          modeHandler.setupStatusBarItem(accessErr.message);
-        }
+    try {
+      fs.accessSync(this.activeTextEditor.document.fileName, fs.W_OK);
```
rebornix: Hey can you help me understand why we use `sync` instead of `async` methods here?

johnfn:
Let me know if this one is good to go, @Platzer!

src/cmd_line/commands/write.ts
Platzer: Hi @rebornix, with my restricted type/javascript skills i couldn't return the `Save` method Promise, and if i'm not able to wait until the editor has saved the file i can not close it with the close-cmd.

Platzer:
@johnfn I would say so, but i've got the developer sickness > never believe my code is done (especially on new terrain)!

src/cmd_line/commands/write.ts
johnfn: I'm not super concerned about blocking during a save... should I be?

johnfn:
Looks fine to me - thanks, @Platzer!

PR #634, opened 2016-08-21
updating the undo tree when using bracket operators slightly

xconverge:
Yay! We love PRs! 🎊

Please include a description of your change & check your PR against this list, thanks:

- [X] Commit message has a short title & issue references
- [X] Each commit does a logical chunk of work.  
- [x] It builds and tests pass (e.g `gulp tslint`)

More info can be found by clicking the "guidelines for contributing" link above.

PR #636, opened 2016-08-22
Ctrl+a and Ctrl+x now create undo points correctly and can be repeate…

### LINKED ISSUE #633: Undo behavior with {}[]()<> does not act exactly as it should
@johnfn took a quick stab at this issue in 7ac106d but the behavior does not fully match the way vim works yet.

When writing something where you open and close the brackets then fill in the interior, a new undo point should be created.

ex:
_if(true){
}_

then filling in 

if(true){
    _int i = 0;_
}

an undo point is created in the first step during insert mode, so pressing u once will get back to the first step, and pressing u again will clear it all.

------

-----

### Technical details:

* VSCode Version: 1.4.0
* VsCodeVim Version: master as of 7ac106d8708a1ad9918102abf0e59dbb0609cb4b
* OS: OSX

xconverge:
Will probably close #633 after this and PR #634 get merged

### DIFF
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -3543,6 +3543,7 @@ class ToggleCaseAndMoveForward extends BaseMovement {
 abstract class IncrementDecrementNumberAction extends BaseMovement {
   modes = [ModeName.Normal];
   canBePrefixedWithCount = true;
+  canBeRepeatedWithDot = true;
 
   offset: number;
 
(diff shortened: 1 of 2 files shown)


johnfn:
The proper way to do this would just be to make them into Commands rather than Movements. Movements do not create undo points, but Commands do.

xconverge:
Hmm ok, I tried changing it to inherit from BaseCommand since that was my first thought, but it was doing something funky somewhere later down the line...I will check again

johnfn:
Ready to go?

xconverge:
good to go

johnfn:
Great! Thanks again, @xconverge!

PR #645, opened 2016-08-24
fix for extra character when double click mouse selection

### LINKED ISSUE #641: Mouse double click selection selects an extra character
For example:

test[]

double click test

"test[" is selected. Only "test" should be selected.

Should be able to double click test, then "y", and "p" should only paste "test" not "test["

xconverge:
fix #641 forgot to put that in the commit

ps: someday I will stop using the mouse...today is not the day


johnfn:
How does this work with click dragging?

xconverge:
seems to work fine/similar to vim

johnfn:
Awesome, thanks for the great work as always, @xconverge!

Issue #644, opened 2016-08-23
Undo history not affected by movement


spicydonuts:
### What did you do?

Example 1: Comment out a line, move to another line, comment it out, repeat repeat, hit `u`
Example 2: type an if-block with a body, press escape, hit `u`

### What did you expect to happen?

Changes on separate lines should be treated as separate events in the undo history.

### What happened instead?

Large, multi-line blocks of changes seem to be treated as one undo event, inconsistent with both vim and vscode's normal undos.

### Technical details:

* VSCode Version: 1.4.0
* VsCodeVim Version: 0.1.7
* OS: macOS 10.11.6

xconverge:
For example 1, how are you moving to another line, are you using arrow keys?

spicydonuts:
Doesn't matter which.  I've tried both, plus clicking with the mouse.

johnfn:
It's interesting to me how we've gotten a large blast of fixes for our undo history over the last week or so, when it's been out for many months without any substantial changes. Thanks @xconverge and @spicydonuts for being fastidious! :)

PR #646, opened 2016-08-24
Undo behavior when position changes using arrows or mouse

xconverge:
This is to address issue #644 

I am not sure I like this approach and want to see if I can do something with cursor ending position vs start position so that it applies to mouse AND arrows without needing to register the keys the way I am.

### DIFF
--- a/package.json
+++ b/package.json
@@ -130,22 +130,22 @@
             {
                 "key": "left",
                 "command": "extension.vim_left",
-                "when": "editorTextFocus && vim.mode != 'Insert Mode'"
+                "when": "editorTextFocus"
             },
             {
                 "key": "right",
                 "command": "extension.vim_right",
-                "when": "editorTextFocus && vim.mode != 'Insert Mode'"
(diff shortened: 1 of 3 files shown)


johnfn:
This is cool, but be very careful that the arrow keys continue to work for e.g. navigating autocomplete or the quick fix menu.

xconverge:
Yep you are right, I am going to come up with a way to use the cursor position from the last history step instead so that it applies to mouse and arrow keys without changing arrow key behavior, it will be much cleaner

xconverge:
So this is weird but was the best way I could think of solving it easily

this handles mouse and arrow movements without impacting any of the key handlers or anything

it mostly looks for an unexplained/handled jump in position via arrow/mouse/bicycle and says ok that means its time to create a new undo point.

johnfn:
It is a bit weird but not much weirder than a lot of the other fixes we've had to do. So I'm okay with it. :) 

Thanks for the PR @xconverge!

PR #647, opened 2016-08-24
Fix missleading readme instruction

AntonAderum:
Tiny fix to README that had missleading instructions.


johnfn:
Thanks @AntonAderum!

PR #649, opened 2016-08-25
improves bracket undo behavior when vscode autocloses brackets

xconverge:
Yay! We love PRs! 🎊

### DIFF
--- a/src/mode/modeHandler.ts
+++ b/src/mode/modeHandler.ts
@@ -24,6 +24,7 @@ import { Position } from './../motion/position';
 import { RegisterMode } from './../register/register';
 import { showCmdLine } from '../../src/cmd_line/main';
 import { Configuration } from '../../src/configuration/configuration';
+import { PairMatcher } from './../matching/matcher';
 
 export enum VimSpecialCommands {
   Nothing,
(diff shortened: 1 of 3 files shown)


src/mode/modeHandler.ts
```diff
@@ -1095,8 +1096,25 @@ export class ModeHandler implements vscode.Disposable {
...
           break;
-        default:
-          return false;
+      }
+
+      // This section is mostly for vscode auto closing brackets without a user inserting them
+      const letterToTheRight = TextEditor.getLineAt(vimState.cursorPosition).text[vimState.cursorPosition.character];
+      if (letterToTheRight !== undefined) {
+        switch (key) {
+          case "{":
```
johnfn: Generally when there are giant switch statements that's a sign that you could better represent your stuff as a data structure. In this case, maybe something like this:

```
const pairings = { 
    "{" : "}",
    "[" : "]",
    "<" : ">",

    /* etc... */
}
```

johnfn: I think this idea would also let you DRY up this code a little bit. It's a little repetitive right now.

xconverge:
Ok definitely a _little_ cleaner now

johnfn:
Do you think you can leverage this?

https://github.com/VSCodeVim/Vim/blob/master/src/matching/matcher.ts

xconverge:
I should be able to... this looks perfect...

johnfn:
Wonderful, thank you!

PR #655, opened 2016-08-26
fix #652

### LINKED ISSUE #652: Repeated search of the same text doesn't work
### What did you do?
Want to search for words in my file.

Put the exact keys you pressed, if relevant.
Pressed "/" to filled in text to search for.
It searched for first matching word.  
But when I press "/" again and hit <return>, it should automatically continue searching the file for next match.

When pressing "/" and hitting <return> it should automatically search for the last text that was searched for.

### What happened instead?
Pressing "/" and hitting <return> the second or more time does nothing.  You have to manually fill in the same text each time.

### Technical details:

* VSCode Version: 1.4
* VsCodeVim Version: 0.17
* OS: Windows 10 64-bit

xconverge:
Not exactly how you guys were discussing it, but I wanted it to start saying searching at the bottom when you first press /. I couldn't figure out a cleaner way with just moving initialization of the search state around...

### DIFF
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -614,6 +614,16 @@ class CommandInsertInSearchMode extends BaseCommand {
       vimState.currentMode = ModeName.Normal;
       vimState.cursorPosition = searchState.getNextSearchMatchPosition(searchState.searchCursorStartPosition).pos;
 
+      // Repeat the previous search if no new string is entered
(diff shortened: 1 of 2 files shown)


johnfn:
Eh, I hate being pedantic, but I feel like the other way (where we move the initialization to when you hit enter) would be a little cleaner. What's the problem that you ran into?

PR #659, opened 2016-08-27
fix #510

### LINKED ISSUE #510: Unexpected cursor shape with Go To Definition
### What did you do?

Open two source files where one file references a type in the other file. I used .fs files; one with a record type that includes a record type from the other file.

Leave both files in Insert mode. From one file use F12 Go To Definition to navigate to the other file. Then press a vim mode command such as insert or append.

I expected to be able to Go To Definition and see the cursor indicating the current vim mode.

### What happened instead?
The cursor indicates normal mode but typing a mode command inserts the letter.

### Technical details:

* VSCode Version: 1.3.1 e6b4afa53e9c0f54edef1673de9001e9f0f547ae
* VsCodeVim Version: 0.1.3
* OS: macOS El Capitan 10.11.5

### Extensions installed
EditorConfig 0.3.1
Ionide-FAKE 1.1.0
Ionide-Paket 1.2.1
Ionide-fsharp 2.2.7
Path intellisense 1.0.2
PowerShell 0.6.1
C# 1.3.0
Mono Debug 0.10.20
yo 0.8.5
vscodevim 0.1.3

xconverge:
We update everything better now when text editors are changed

### DIFF
--- a/extension.ts
+++ b/extension.ts
@@ -12,6 +12,7 @@ import { showCmdLine } from './src/cmd_line/main';
 import { ModeHandler } from './src/mode/modeHandler';
 import { TaskQueue } from './src/taskQueue';
 import { Position } from './src/motion/position';
+import './src/globals';
 
 interface VSCodeKeybinding {
   key: string;
@@ -71,7 +72,7 @@ export async function getAndUpdateModeHandler(): Promise<ModeHandler> {
   const activeEditorId = new EditorIdentity(vscode.window.activeTextEditor);
 
   if (!modeHandlerToEditorIdentity[activeEditorId.toString()]) {
-    const newModeHandler = new ModeHandler(false, activeEditorId.fileName);
+    const newModeHandler = new ModeHandler(activeEditorId.fileName);
 
     modeHandlerToEditorIdentity[activeEditorId.toString()] = newModeHandler;
     extensionContext.subscriptions.push(newModeHandler);
@@ -223,6 +224,12 @@ async function handleKeyEvent(key: string): Promise<void> {
 }
 
 async function handleActiveEditorChange(): Promise<void> {
+
+  // Don't run this event handler during testing
+  if(isTesting){
+    return;
+  }
+
   if (vscode.window.activeTextEditor !== undefined) {
     const mh = await getAndUpdateModeHandler();
     mh.updateView(mh.vimState, false);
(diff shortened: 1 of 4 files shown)


xconverge:
So this works pretty well...however....it bombs on the tests.

I tracked it to the fact that now that I am calling getAndUpdateModeHandler(), it will create modehandlers that have withTesting set to FALSE for the tests.

Is there a way to disable this event handler for just the tests? :/ 

Any advice would be great...

johnfn:
Can you move our isTesting flag to a global somewhere? That's something I should have done back when I made modehandler per-editor rather than global itself.

xconverge:
where would be the best place to put a global like that in extension.ts? And then where is the top most level of testing where I could set it, testSimplifier.ts?

johnfn:
You may actually want to start a new file called globals.ts and follow the same pattern i did in taskqueue.ts to export it. 

testSimplifier is OK, but not all of our tests use that infrastructure. Maybe there's a better place somewhere? I'm not sure!

xconverge:
got it, let me know how it looks

extension.ts
```diff
@@ -12,6 +12,7 @@ import { showCmdLine } from './src/cmd_line/main';
 import { ModeHandler } from './src/mode/modeHandler';
 import { TaskQueue } from './src/taskQueue';
 import { Position } from './src/motion/position';
+import './src/globals';
```
johnfn: Instead of importing everything from globals into the namespace, can you do something like

`import { globals } from './src/globals'` 

and then check `globals.isTesting`?

src/globals.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * globals.ts hold some globals used throughout the extension
+ */
+
+// true for running tests, false during regular runtime
+declare var isTesting: boolean;
```
johnfn: Can you use let here? 

Also the best way to do this is:

`export let isTesting = false;`

(though actually export the object, not just the boolean!

johnfn:
More small comments.

src/globals.ts
xconverge: which object, are you saying I should make a globals class that is exported?

export class globals {
  public isTesting: boolean = false;
}

xconverge:
Let me know if that hits any of the marks you were referring to...

johnfn:
This looks great. Can you just fix up those merge conflicts and we should be good to go here.

xconverge:
woops bad merge..

johnfn:
Another great change. 0.1.8 is shaping up to be the most bug free release in VSCodeVim history! (sort of.)

PR #661, opened 2016-08-29
[WIP] fixes #638 and fixes #563

### LINKED ISSUE #563: Ctrl-A Handled Incorrect
### What did you do?
With texts:

<kbd>a</kbd> = 1.9

<kbd>a</kbd> means cursor is on letter "a"

Press Ctrl-A

Got:

a = <kbd>2</kbd>.9

### What happened instead?

Got:

a = 1.1<kbd>0</kbd>

### Technical details:

* VSCode Version: 1.3.1
* VsCodeVim Version: 0.1.4
* OS: Windows 10.0.10586 (x64)

### LINKED ISSUE #638: Ctrl-A/X not working correctly if the numbers aren't surrounded by separators
### What did you do?
I used Ctrl-A on something like `width: 2em;` or `abc1`.

They should have changed to `width: 3em;` and `abc2`.

### What happened instead?
Nothing

### Technical details:
none

* VSCode Version: 1.4.0
* VsCodeVim Version: 0.1.7
* OS: Mac OSX 10.11.6 (15G31)

xconverge:
So this is not very elegant... but it works...Any feedback would be great

I am almost positive that what you guys would LIKE is to fix a bunch of the searchers and getword start and end to be better for numbers...

My main motivation for doing it this way vs improving the searchers is for "in word" numbers. ie. abc1

@sectioneight Any thoughts here...

### DIFF
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -3571,20 +3571,90 @@ abstract class IncrementDecrementNumberAction extends BaseCommand {
   public async exec(position: Position, vimState: VimState): Promise<VimState> {
     const text = TextEditor.getLineAt(position).text;
 
-    for (let { start, end, word } of Position.IterateWords(position.getWordLeft(true))) {
-      // '-' doesn't count as a word, but is important to include in parsing the number
-      if (text[start.character - 1] === '-') {
-        start = start.getLeft();
-        word = text[start.character] + word;
-      }
-      // Strict number parsing so "1a" doesn't silently get converted to "1"
-      const num = NumericString.parse(word);
+    if (text.length === 0) {
+      return vimState;
+    }
 
-      if (num !== null) {
-        vimState.cursorPosition = await this.replaceNum(num, this.offset * (vimState.recordedState.count || 1), start, end);
-        return vimState;
+    let numFound = false;
+    let numStart = 0;
+    let numEnd = 0;
+    let numString = "";
+    let negative = false;
+    let isHex = false;
+
+    // Character to start searching at (beginning of this word)
+    let startSearchPos = vimState.cursorPosition.getWordLeft(true).character;
+
+    // Was the previous char a negative sign?
+    // Found a potential negative number, record the start position just incase
+    if (text[startSearchPos - 1] === '-') {
+      numStart = startSearchPos - 1;
+      negative = true;
+    }
+
+    // Handle the case where this is the first character on the line, check for negative
+    if (text[startSearchPos] === '-') {
+      negative = true;
+      startSearchPos += 1;
+    }
+
+    // Iterate over characters and compute the number
+    // Needs to handle hex, octal, negative, ints
+    for (let i = startSearchPos; i < text.length; i++) {
+      const char = text[i];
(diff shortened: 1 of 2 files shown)


sectioneight:
So I haven't worked through all the edge cases, but is there a reason we couldn't express this as a regular expression? We're basically looking for the position of numeric/hex/octal digits within a word:

```js
let numerals = new RegExp(/(-)?(0x)?([0-9a-f]+)/);
```

And just look for that in the position starting after the cursor?

xconverge:
Hmm I thought 1ab|c would increment 1 as well but it looks like that's not true...

Another edge case: if the number happens to be 0x, you can't increment/decrement it

I will take another look, this PR is super ugly so I will take a 2nd stab at it to see what the **minimally invasive** way might look like and if there are any unsolvable edgecases with searching

xconverge:
I am not very good with regex, thus the brute force... I am hoping someone can suggest something a bit cleaner with the implementation before.

I am not sure what you guys want to do, I wouldn't merge this but it works pretty well... I think I will move on to something else and leave this behind :/

sectioneight:
@xconverge I'm actually OK with this approach. I'm just wondering if it's something that could be solved with less code.

If you want to abandon, I'm happy to take this on when I have free time (although full disclosure: September is a very busy month for me, and I don't know exactly when that will be).

johnfn:
I would really prefer to continue to use the regex capturing groups idea that @sectioneight and I had if at all possible - I think it's probably the most elegant way to solve this problem.

Could you just add two new regexes to the NumericString matchings dictionary - one for words like 5xxx and one for words like 5.5?

xconverge:
Yea I will try to give it a 2nd try with the original method but additional regexes

xconverge:
Is this on better track? The regexes are still wrong, but I detect anything that has letters and numbers, gets the letters as a prefix and as a postfix if they are there


```
 private static matchings: { regex: RegExp, base: number, prefix: string }[] = [
    { regex: /^([-+])?0([0-7]+)$/, base: 8, prefix: "0"},
    { regex: /^([-+])?(\d+)$/, base: 10, prefix: ""},
    { regex: /^([-+])?0x([\da-fA-F]+)$/, base: 16, prefix: "0x"},
    { regex: /^[a-z0-9]+$/i, base: 10, prefix: ""}
  ];

  static parse(input: string): NumericString | null {
    for (const { regex, base, prefix } of NumericString.matchings) {
      const match = regex.exec(input);
      if (match == null) {
        continue;
      }

      let findLetters = /[a-z]+/i;
      let findPrefix = /^[^\d]+(?=[0-9]+)/i;
      let findPostfix =/^[0-9]+(?=[^\d]+)/i;

      let newPrefix = prefix;
      let newPostfix = "";

      if(findLetters.exec(match[0]) !== null){
        newPrefix = findPrefix.exec(match[0]).toString();
        newPostfix = findPostfix.exec(match[0]).toString();
        match[0] = match[0].replace(/^[a-z]+/i, '');
      }

      return new NumericString(parseInt(match[0], base), base, newPrefix, newPostfix);
    }
    return null;
  }
```

PR #664, opened 2016-08-29
Document autoindent option

sectioneight:
So people know they can turn it on/off.


johnfn:
Thanks @sectioneight!

PR #671, opened 2016-08-30
Implement :wa[ll] command (write all)

mleech:
Implements a basic `:wa[ll]` command (http://vimdoc.sourceforge.net/htmldoc/editing.html#:wall)
It currently just triggers vscode's `workbench.action.files.saveAll` command, so the bang version (`wa!`) doesn't overwrite readonly files.

Yay! We love PRs! 🎊

### DIFF
--- a/src/cmd_line/commands/wall.ts
+++ b/src/cmd_line/commands/wall.ts
@@ -1,0 +1,34 @@
+"use strict";
+
+import * as vscode from "vscode";
+import * as node from "../node";
+
+export interface IWallCommandArguments extends node.ICommandArgs {
+  bang?: boolean;
+  range?: node.LineRange;
+}
+
+//
+//  Implements :wall (write all)
+//  http://vimdoc.sourceforge.net/htmldoc/editing.html#:wall
+//
+export class WallCommand extends node.CommandBase {
+  protected _arguments : IWallCommandArguments;
+
+  constructor(args : IWallCommandArguments) {
+    super();
+
+    this._name = 'wall';
+    this._shortName = 'wa';
+    this._arguments = args;
+  }
+
+  get arguments() : IWallCommandArguments {
+    return this._arguments;
+  }
+
+  async execute() : Promise<void> {
+    // TODO : overwrite readonly files when bang? == true
+    await vscode.workspace.saveAll(false);
+  }
+}
--- a/src/cmd_line/subparsers/wall.ts
+++ b/src/cmd_line/subparsers/wall.ts
@@ -1,0 +1,25 @@
+"use strict";
+
+import * as node from "../commands/wall";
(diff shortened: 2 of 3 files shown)


johnfn:
Is this a WIP? I still see some TODO comments.

mleech:
@johnfn I think it is useful in it's current form as partial implementation of `:wa` (at least it'd be useful for me 😃 `:wa` is ingrained in my muscle memory). The TODOs are more to call out what's missing re mimicking Vim fully. I'm a bit of a vscode noob so I'm not sure if they're possible with the current API.

src/cmd_line/commands/wall.ts
```diff
@@ -0,0 +1,35 @@
...
+    this._shortName = 'wa';
+    this._arguments = args;
+  }
+
+  get arguments() : IWallCommandArguments {
+    return this._arguments;
+  }
+
+  async execute() : Promise<void> {
+    // TODO : ignore untitled documents
```
rebornix: You can access `textEditor.document.isUntitled` to see if it's untitled.

src/cmd_line/subparsers/wall.ts
```diff
@@ -0,0 +1,25 @@
...
+import {Scanner} from '../scanner';
+import {VimError, ErrorCode} from '../../error';
+
+export function parseWallCommandArgs(args : string) : node.WallCommand {
+  if (!args) {
+    return new node.WallCommand({});
+  }
+  var scannedArgs : node.IWallCommandArguments = {};
+  var scanner = new Scanner(args);
+  const c = scanner.next();
```
rebornix: `c` can be `EOF` right?

rebornix:
Two tiny comments and all others are good to me.

johnfn:
LGTM; however ping @markrendle about @rebornix's comment about EOF. And can you bring this up to date with master?

mleech:
Thanks @johnfn, it should be now up to date with master. 
Also, I found the `vscode.workspace.saveAll(includeUntitled: boolean)` function in the vscode namespace, so I'm using that instead as it's cleaner and handles ignoring the untitled files.
Re @rebornix's EOF concern in the parsing code - that code is copied from the quit command and seems to be working. The call to `scanner.next()` doesn't get hit unless there are characters after the main command (`:wa` or `:wall`) so the call should always return something other than EOF.

rebornix:
@mleech I see, thanks for helping understand the code better :) LGTM as well.

jpoon:
Thanks @mleech

mleech:
@jpoon yeah you're right, it would be a better name. I mistakenly thought all the other classes/files were named after the long version of the vim command - hence `wall`. I'll try and fix it up when I get some spare time.

Issue #410, opened 2016-07-08
Cannot use ; and , to search forward and backward in f{char} F{char} t{char} and T{char}


jingkangy91:
Do you need help or have a question? Ask us on Slack (https://vscodevim-slackin.azurewebsites.net/) 👫. 
Found a bug? Please fill out the sections below 👍.

### Issue Summary

A summary of the issue -- what blew up? what major feature are we missing? 
Cannot use ; and , to search forward and backward in f{char} F{char} t{char} and T{char}

### Steps to Reproduce

1. First I use "f{char}" to find a character,
2. then I cannot use ";" or "," to search forward or backward which is very convenient in Vim.

Any other info e.g. Why do you consider this to be a bug? What did you expect to happen instead?

### Technical details:

* VSCode Version: 1.3.0
* OS: Windows 10

johnfn:
This is done! Thanks @aminroosta!

PR #674, opened 2016-08-30
Implement ; and ,

aminroosta:
Related Issue #410 

@rebornix @johnfn please let me know if you want me to change something in this PR.

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -54,8 +54,8 @@ Status | Command | Description
 :white_check_mark:   |:1234:  F{char}	| to the Nth occurrence of {char} to the left
 :white_check_mark:   |:1234:  t{char}	| till before the Nth occurrence of {char} to the right
 :white_check_mark:   |:1234:  T{char}	| till before the Nth occurrence of {char} to the left
-   |:1234:  ;	| repeat the last "f", "F", "t", or "T" N times
-   |:1234:  ,	| repeat the last "f", "F", "t", or "T" N times in opposite direction
+:white_check_mark:   |:1234:  ;	| repeat the last "f", "F", "t", or "T" N times
+:white_check_mark:   |:1234:  ,	| repeat the last "f", "F", "t", or "T" N times in opposite direction
 
 ## Up-down motions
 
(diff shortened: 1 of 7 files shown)


johnfn:
First of all, thank youuu for implementing this! This is a big one that a lot of people want, and I'm super excited to get it in VSCodeVim!

I have some suggestions for improvement. Let me know what you think.

1. Do not have a BaseRepeatableMovement subclass. 
2. In order to determine the reverse, simply make a mapping from motions to reverse motions. It might look something like this:

```
const reverseMotionMapping = {
    MoveFindForward: MoveFindBackward,
    MoveFindBackward: MoveFindForward,

    /* etc... */ 
}
```

Now you don't have to duplicate the keypress assignment and all the information about reversing is clustered together.

aminroosta:
@johnfn As always great feedback :+1:
PR is now updated accordingly.

johnfn:
Last suggestion: instead of `VimState.lastRepeatableMovement = this` inside the body of the exec statement, my suggestion would be to have a property on BaseMotion called something like `repeatableWithSemicolonOrComma` (feel free to come up with a better name if you wish) and set that to true on the related motions (tTfF).

Then after we execute movements, check if that flag is set on the movement we just ran and if so *then* run `VimState.lastRepeatableMovement = this`.

Kind of hard to explain exactly why I think this is better but I like a more declarative approach of specifying important characteristics of actions as properties on the class rather than in the body of the function (where it's harder to find). I think it improves readability.

aminroosta:
@johnfn No wonder that this project has a readable and clean codebase ;)
PR updated accordingly.

johnfn:
:) 

Last thing - can you actually make it a property on the class directly, sort of like right under

```
class MyActionThing extends BaseMovement {
   keys = ["f", "<character>"];
```

you have

`    repeatableWithSemicolonOrComma = true`

?

aminroosta:
@johnfn repeatableWIthSemicolonOrComma should only be set to true if the action is successfully performed and there is no operator like "dfa"

aminroosta:
@johnfn Based on our discussion in slack channel PR updated.

johnfn:
Last thing, I promise! (Sorry for the delay!)

Can you condense it down to one line in the 4 cases where it appears?

`return !vimState.recordedState.operator || !(isIMovement(result) && result.failed);`

Then we should legitimately be good to go here.

aminroosta:
@johnfn Done 😄

johnfn:
I'm ready to merge this, I just need it to be up to date with master.

aminroosta:
@johnfn Done

Issue #677, opened 2016-08-30
z+t and z+b not adopted


alexandrudima:
Testing Microsoft/vscode#11147

VSCode Version 1.5.0-insider
Commit: 18f7fe90aa470ddc632c80388f84bf511afb7078
VSCodeVim Commit: 1185bceee27e09562d63a040808c5dac5227af9b
OS: Windows

Config: `"vim.useCtrlKeys": true`

`z`+`t`, `z`+`b` don't appear to do anything

johnfn:
Whoa, didn't realize that this was possible now. Thanks for the heads up, @alexandrudima! (Also, thanks for all the issues recently!)

jpoon:
Related Issue: https://github.com/VSCodeVim/Vim/issues/210

Issue #210, opened 2016-04-30
scroll-cursor commands (zt, zb) don't work


stapelberg:
Vim has handy commands for scrolling without moving the cursor, e.g. `zt`, which scrolls the screen so that the line with the current cursor is placed as the first line. E.g., I often do actions like `/main(<CR>zt`, so that I can see as much of the main function as possible.

Would it be possible to implement `zt`, `zb` and `zz`?

Thanks!

johnfn:
zz has been done. zt and zb are impossible until https://github.com/Microsoft/vscode/issues/2771 😞

jpoon:
On 0.2.0:

1) `zz` works
2) `zt` and `zb` don't work. We do have `CommandTopScroll` and `CommandBottomScroll` which are called but the window does not update.

rebornix:
@jpoon I moved the master bits to extensions folder and tried, both `zt` and `zb` work perfectly. I don't know whether we did anything wrong with compiling, this is really weird.

jpoon:
@rebornix Hmm..yeah, I see the same thing. Works in master, doesn't work in released version. Super strange. As there's been a fair amount of changes since 0.2.0, shall we release a new version and see if that magically resolves this issue?

rebornix:
@jpoon If you generate a visx and install it manually, it works as well. Let's wait and see next version!

johnfn:
This appears to be done for me. :)

Issue #676, opened 2016-08-30
Ctrl+F moves cursor down but does not scroll editor


alexandrudima:
Testing Microsoft/vscode#11147

VSCode Version 1.5.0-insider
Commit: 18f7fe90aa470ddc632c80388f84bf511afb7078
VSCodeVim Commit: 1185bceee27e09562d63a040808c5dac5227af9b
OS: Windows

Config: `"vim.useCtrlKeys": true`

## Expected
![ctrl-f-expected](https://cloud.githubusercontent.com/assets/5047891/18089950/a47c145c-6ec2-11e6-98ce-c5312d8ae98e.gif)

## Actual
 * cursor moves but the editor is not scrolled

![ctrl-f-actual](https://cloud.githubusercontent.com/assets/5047891/18089928/8c229b06-6ec2-11e6-9876-cff4c5918753.gif)

alexandrudima:
The exact same problem with Ctrl-B

Issue #675, opened 2016-08-30
Ctrl+D appears to scroll more than half window lines downwards (same as Ctrl+F?)


alexandrudima:
Testing Microsoft/vscode#11147

VSCode Version 1.5.0-insider
Commit: 18f7fe90aa470ddc632c80388f84bf511afb7078
VSCodeVim Commit: 1185bceee27e09562d63a040808c5dac5227af9b
OS: Windows

Config: `"vim.useCtrlKeys": true`

## Expected
![ctrl-d-expected](https://cloud.githubusercontent.com/assets/5047891/18089789/ea9c43cc-6ec1-11e6-9903-5119e45a3427.gif)

## Actual
 * cursor moves more than 1/2 a screen & the editor is not scrolled

![ctrl-d-actual](https://cloud.githubusercontent.com/assets/5047891/18089796/eec399a0-6ec1-11e6-86ed-add559bea2a6.gif)

alexandrudima:
The exact same problem with Ctrl-U

rebornix:
@alexandrudima it's more or less our workaround before we have `editorScroll`, using 20 lines (Vim's default config) instead of dynamic half page. Since we already have `editorScroll`, this issue should be gone once we fix the side effect issue there.

PR #681, opened 2016-08-30
Editor Scroll

### LINKED ISSUE #672: Ctrl-E and Ctrl-I stop working as soon as the cursor should move
Testing Microsoft/vscode#11147

VSCode Version 1.5.0-insider
Commit: 18f7fe90aa470ddc632c80388f84bf511afb7078
VSCodeVim Commit: 1185bceee27e09562d63a040808c5dac5227af9b
OS: Windows

Config: `"vim.useCtrlKeys": true`

In these gifs I'm doing Ctrl-E but the exact problem exists for Ctrl-Y.

## Expected
ctrl-e-expected

## Actual
ctrl-e-actual

rebornix:
Add editor scroll commands. The trick in this PR is disabling view updating for editor scrolling as Code already does it for us, re-updating the view will lead to unexpected result of layout redraw.

This PR should fix #672, #673, #675, #676 .

### DIFF
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -1724,6 +1724,28 @@ class CommandCenterScroll extends BaseCommand {
 }
 
 @RegisterAction
+class COmmandTopScroll extends BaseCommand {
+  modes = [ModeName.Normal];
+  keys = ["z", "t"];
+
+  public async exec(position: Position, vimState: VimState): Promise<VimState> {
+    await vscode.commands.executeCommand('revealLine', {lineNumber: 1, at: 'top'});
+    return vimState;
+  }
+}
+
+@RegisterAction
+class CommandBottomScroll extends BaseCommand {
+  modes = [ModeName.Normal];
+  keys = ["z", "b"];
+
+  public async exec(position: Position, vimState: VimState): Promise<VimState> {
+    await vscode.commands.executeCommand('revealLine', {lineNumber: 10, at: 'bottom'});
+    return vimState;
+  }
+}
+
+@RegisterAction
 class CommandGoToOtherEndOfHighlightedText extends BaseCommand {
   modes = [ModeName.Visual, ModeName.VisualLine];
   keys = ["o"];
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -1724,7 +1724,7 @@ class CommandCenterScroll extends BaseCommand {
 }
 
 @RegisterAction
-class COmmandTopScroll extends BaseCommand {
+class CommandTopScroll extends BaseCommand {
   modes = [ModeName.Normal];
   keys = ["z", "t"];
 
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -1,7 +1,7 @@
 import { VimSpecialCommands, VimState, SearchState, SearchDirection, ReplaceState } from './../mode/modeHandler';
 import { ModeName } from './../mode/mode';
 import { VisualBlockInsertionType } from './../mode/modeVisualBlock';
-import { TextEditor } from './../textEditor';
+import { TextEditor, EditorScrollByUnit, EditorScrollDirection, CursorMovePosition, CursorMoveByUnit } from './../textEditor';
 import { Register, RegisterMode } from './../register/register';
 import { NumericString } from './../number/numericString';
 import { Position } from './../motion/position';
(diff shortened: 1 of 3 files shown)


rebornix:
Fix #677 as well. After reading @sandeep 's source code now I understand `lineNumber` parameter correctly.

src/actions/actions.ts
```diff
@@ -1466,6 +1510,28 @@ class CommandCenterScroll extends BaseCommand {
 }
 
 @RegisterAction
+class COmmandTopScroll extends BaseCommand {
```
mleech: Minor casing issue - there's an uppercase `O` in `CommandTopScroll`

rebornix: Ooops, I've put too much time on `shift` ...

johnfn:
I've mentioned before that I dislike updating the view outside of our updateView function, though in the past I never really drew a hard line. However, I've actually started to take steps to minimize this behavior within my multi-cursor-mode branch, and I think it would be prudent to begin to draw the line and say we want to do that everywhere.

The way I'd suggest to do something like this would be that for actions that can adjust the view, make simple interfaces like `{ type: 'scrollViewToTop' }` `{ type: 'scrollViewToBottom' }` etc. Then make a property on VimState which is something like viewChanges. Then read off this array inside of the updateView function.

rebornix:
@johnfn ha I was expecting this feedback on this :) 

I do agree that any side effect should happen in a centralized place like updateView but sometimes it's difficult. The reason that Code won't provide a pure command for us to call (I mentioned quite a bit times) is a pure command without any side effect may be not useful to other callers and way too Vim specific.

The top 1st command/API that conflicts with your principle (making commands pure/ putting view updating all in viewUpdate) is `vscode.window.activeTextEditor.edit()`. We leveraged this API to archive `delete` and `replace` but they will update the view immediately once they are executed. This makes it difficult to draw the hard line.

I'm not against your proposal and actually I do want to make it this way in VSCodeVim but we do have limitations. I'll read your multi-cursor PR to grab ideas from you and update this PR accordingly.

johnfn:
For the pure, just scroll and do nothing else commands, I think it's possible! But the other ones will be more difficult, you're right. Kind of a bummer.

jpoon:
This would also fix https://github.com/VSCodeVim/Vim/issues/210

rebornix:
@johnfn @jpoon just followed Johnfn's suggestions, postpone Code commands which have side effect to view layout and only run them when we trigger our inhouse `updateView`, making it a little bit similar to https://github.com/VSCodeVim/Vim/pull/587/files#diff-a7723f32fe6b05e135838f84e5c23c0fR371 .

Look forward to your comments on this.

src/actions/actions.ts
```diff
@@ -609,16 +635,49 @@ class CommandDeleteIndentInCurrentLine extends BaseCommand {
   }
 }
 
+class CommandCtrlE extends CommandEditorScroll {
+  keys = ["<C-e>"];
+  to = "down";
```
jpoon: Can we create enums for `to` and `by` values?

src/mode/modeHandler.ts
```diff
@@ -1088,6 +1100,15 @@ export class ModeHandler implements vscode.Disposable {
 
     vscode.window.activeTextEditor.setDecorations(this._caretDecoration, rangesToDraw);
 
+    if (this.vimState.postponedCodeViewChanges.length > 0) {
```
jpoon: is this outside `if` necessary?

src/mode/modeHandler.ts
```diff
@@ -33,6 +33,11 @@ export enum VimSpecialCommands {
   Dot
 }
 
+export class ViewChange {
+  public name: string;
```
jpoon: nit: maybe rename this to `command`? As that'll be inline with how it's used in `executeCommand`.

src/actions/actions.ts
rebornix: Sure, that's definitely necessary.

rebornix:
@jpoon comments addressed!

src/textEditor.ts
```diff
@@ -224,3 +224,61 @@ export class TextEditor {
...
+export const RevealLineAtArgument = {
+  Top: 'top',
+  Center: 'center',
+  Bottom: 'bottom'
+};
+
+/**
+ * Positions in the view for cursor move command.
+ */
+export const CursorMovePosition = {
```
johnfn: You could actually just use string union types here, which I like better (stronger typechecking).

`type CursorMovePosition = 'left' | 'right' | 'up' | 'down' | ...`

rebornix: yes, string union type can easily validate string value assignment but if we use that, then we still need to write plain text string while initialization in `action.ts`, which is what we want to avoid, right?

PR #682, opened 2016-08-31
[WIP] fixes issue needing to press i or a multiple times to switch modes

xconverge:
Yay! We love PRs! 🎊

### DIFF
--- a/src/mode/modeHandler.ts
+++ b/src/mode/modeHandler.ts
@@ -597,9 +597,6 @@ export class ModeHandler implements vscode.Disposable {
           }
         } else {
           if (this._vimState.currentMode !== ModeName.Insert) {
-            // Clear out the actions that were run that caused this mode to transition.
-            // This is mostly for commands that need to be the first key (i, a, etc.) will get triggered properly after this
-            this._vimState.recordedState.actionsRun = [];
             this._vimState.currentMode = ModeName.Normal;
             this.setCurrentModeByName(this._vimState);
           }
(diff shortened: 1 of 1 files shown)


jpoon:
@xconverge Can you share a repo of the bug you are trying to fix?

xconverge:
What do you mean? This was mostly addressing an issue I have been battling for a long time where sometimes after using cmd+c, cmd+v or the mouse in certain ways, when finishing, it takes 3 presses of i or a to enter insert mode.

This specific commit only fixes on case, so I was trying to move the clearing of actionKeys to this.setCurrentModeByName instead.

johnfn:
@jpoon easiest way is to visually select a word, do ctrl-v, and then try to get into insert mode - you have to press i a few times.

I bugged @xconverge about this already, but although this fix appears to work I'd like it to be a bit more general by running every time a mode change happens. An alternative solution would be to simply override ctrl-v with `*p`.

xconverge:
Yea so I moved it to running it every time a mode changes but the tests fail, so I am working on figuring out why they fail and how to address it

jpoon:
Ah, thanks; that's exactly the explanation (ie. repro steps) I was looking for

rebornix:
As the repro steps are intuitive and descriptive, I think we can have a test case to cover it? @xconverge

xconverge:
The problem is I can easily fix that one case (this PR as is for example), but still encounter the issue many times in other ways (that I do not have the exact repro steps for)

johnfn:
Making a test case for this wouldn't be straightforward. The reason it happens is because we don't handle `ctrl-v` in Vim. Once we intercept it (e.g. to make a test for it), the problem will go away.  :P

jpoon:

PR #683, opened 2016-08-31
Use Angle Bracket Notation (Fixes #64)

### LINKED ISSUE #64: renaming keys to vim nomenclature
Vim uses names like `<cr>` (`return`), `<C-m>` (`ctrl+m`). This will be required later to process keys, for example for the `:normal` command.

@jpoon Thoughts?

jpoon:
Easing myself back into the project after a short hiatus with this simple change. 

Update: Translating key presses to Vim's angle bracket notation:

`ctrl+*` becomes `<C-*>`
`escape` becomes `<Esc>`
`backspace` becomes `<BS>`
`delete` becomes `<Del>`

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ On OS X, open Terminal and run the following command:
 defaults write com.microsoft.VSCode ApplePressAndHoldEnabled -bool false
 ```
 
-#### How can I bind `jj` to `<escape>`?
+#### How can I bind `jj` to `<Esc>`?
 
 1. Add the following to `settings.json` (open the Command Pallete and search for "User Settings"):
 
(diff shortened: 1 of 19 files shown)


johnfn:
Could you actually fix this in the more general way by obliterating ctrl+* from the codebase entirely and replacing all such instances with <c-*>? 

It would fix our longest standing bug #64 :)

jpoon:
Argh, just realized this will break backwards compatibility for those that have used a `otherModesKeyBindings` settings.

johnfn:
True, but we don't actually document how to do that anywhere (intentionally, since I knew this was gonna happen eventually...)

Ideally we could just show an error to people who we detect have ctrl+* keybindings - say "you need to update your old keybindings to match the new format"

We could even handle both in the keybindings parser.

jpoon:
That should do it. Moved the angle bracket translation logic to util.ts where it is now called by:

1) `extension.ts` where the vscode key presses come in
2) `remapper.ts` to remap keys for insertModeKeyBindings and otherModesKeyBindings (ie. we have backwards compat)

README.md
```diff
@@ -42,26 +42,26 @@ On OS X, open Terminal and run the following command:
...
+#### How can I bind `jj` to `<Esc>`?
 
 1. Add the following to `settings.json` (open the Command Pallete and search for "User Settings"):
 
    ```
            {
                "before": ["j", "j"],
-               "after": ["<escape>"]
+               "after": ["<Esc>"]
```
johnfn: Can we make this `<esc>` to stay in line with Vim?

jpoon: Are you referring to the capitalization? [VimDocs](http://vimdoc.sourceforge.net/htmldoc/intro.html#<Esc>) have it as a capital `E` so following suit.

johnfn: I didn't know that! Okay, that's fine.

src/actions/actions.ts
```diff
@@ -413,7 +413,7 @@ class CommandEsc extends BaseCommand {
     ModeName.SearchInProgressMode,
     ModeName.Replace
   ];
-  keys = ["<escape>"];
+  keys = ["<Esc>"];
```
johnfn: This must have been annoying to do. Thanks.

I would love to extract these out into a string enum at some point.

PR #692, opened 2016-09-01
fix quote match issue in visual mode

zjhmale:

this PR should fix a trivial issue that when using `cmd+x` on OSX or `ctrl+x` on Windows after matching quotes like `"`, `'` and quasiquote.

before

before

after

after

### DIFF
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -3412,6 +3412,13 @@ abstract class MoveQuoteMatch extends BaseMovement {
 
     let startPos = new Position(position.line, start);
     let endPos = new Position(position.line, end);
+    let fullAction = vimState.currentFullAction;
+    let fullActionLength = fullAction.length;
+    if (fullAction[fullActionLength - 3] === "v"
+        && fullAction[fullActionLength - 2] === "i"
+        && ["\"", "'", "`"].indexOf(fullAction[fullActionLength - 1]) > -1) {
+      endPos = new Position(position.line, end + 1);
+    }
     if (!this.includeSurrounding) {
       startPos = startPos.getRight();
       endPos = endPos.getLeft();


johnfn:
Hmm, does this preserve behaviors such as `vi"d`?

zjhmale:
@johnfn i think `vi"d` is also cool.

screenshot

johnfn:
I'm confused - in the after picture on the original PR, the block cursor on the right is clearly on the " after vi". (This is why I am concerned.) However, in the gif you just sent, the block cursor is on the "g". How is that possible? You did vi" for both, right?

zjhmale:
@johnfn oh sorry, i did not compile it after i rebase the upstream, yeah there is issues with `vi"d` now 😅 , it will also delete the right quote.
screenshot

maybe i should to predicate wether we are using shortcuts `cmd+x` or `ctrl+x` after `vi"` ?

rebornix:
I suppose the issue exists because Editor (whether it's Code or Atom) has no idea bout the cursor management in Vim. But @zjhmale is on the track, we'd better override the behavior `cmd/ctrl+x` if it's in selection mode? That's possible as we are accessing system clipboard already.

zjhmale:
@rebornix cool, overriding default behavior is quit reasonable, and any examples or start point about how to override default behavior? thx in advance!

zjhmale:
and which is really interesting that it is totally fine to use `cmd/ctrl+x` when matching delimiters such as parenthesis, brackets, braces and angle brackets.

screenshot

johnfn:
@zjhmale all you need to do is add a new command for the ctrl-x key (though I don't think we've ever bound a cmd key for cmd-x before - you may need to investigate that one) and make it run a delete operation.

zjhmale:
@johnfn cool, i will figure it out, thx!

jpoon:

PR #702, opened 2016-09-02
Fix `#` and `*` Behaviour

jpoon:
Properly fixes https://github.com/VSCodeVim/Vim/issues/569. 

Behaviour of `#` and `*` is expected to be:
1) If cursor selects whitespace, the word closest to right is chosen
2) If there is no word to the right of the cursor, a VIM error is thrown
3) Otherwise, searches for currently selected word.

### DIFF
--- a/src/textEditor.ts
+++ b/src/textEditor.ts
@@ -4,6 +4,7 @@ import * as vscode from "vscode";
 import { ModeHandler } from './mode/modeHandler';
 import { Position } from './motion/position';
 import { Configuration } from './configuration/configuration';
+import { Globals } from './globals';
 
 export class TextEditor {
   // TODO: Refactor args
@@ -138,6 +139,31 @@ export class TextEditor {
 
   static getText(selection: vscode.Range): string {
     return vscode.window.activeTextEditor.document.getText(selection);
+  }
+
+  /**
+   *  Retrieves the current word at position.
+   *  If current position is whitespace, selects the right-closest word
+   */
+  static getWord(position: Position) : string | undefined {
+    let start = position;
+    let end = position.getRight();
+
+    const char = TextEditor.getText(new vscode.Range(start, end));
+    if (Globals.WhitespaceRegExp.test(char)) {
+      start = position.getWordRight();
+    } else {
+      start = position.getWordLeft(true);
+    }
+    end = start.getCurrentWordEnd(true).getRight();
+
+    const word = TextEditor.getText(new vscode.Range(start, end));
+
+    if (Globals.WhitespaceRegExp.test(word)) {
+      return undefined;
+    }
+
+    return word;
   }
 
   static isFirstLine(position : vscode.Position): boolean {
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -670,21 +670,17 @@ class CommandStar extends BaseCommand {
   isMotion = true;
   canBePrefixedWithCount = true;
 
-  public static GetWordAtPosition(position: Position): string {
-    const start = position.getWordLeft(true);
-    const end   = position.getCurrentWordEnd(true).getRight();
-
-    return TextEditor.getText(new vscode.Range(start, end));
-  }
-
   public async exec(position: Position, vimState: VimState): Promise<VimState> {
-    const currentWord = CommandStar.GetWordAtPosition(position);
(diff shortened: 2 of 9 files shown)


src/textEditor.ts
```diff
@@ -140,6 +141,32 @@ export class TextEditor {
...
+      start = position.getWordRight();
+    } else {
+      start = position.getWordLeft(true);
+    }
+    end = start.getCurrentWordEnd(true).getRight();
+
+    const word = TextEditor.getText(new vscode.Range(start, end));
+
+    if (whitespaceRegex.test(word)) {
+      throw VimError.fromCode(ErrorCode.E348);
```
johnfn: I'm beginning to think we should move away from `throw` and `try/catch`. The more I use them the more I feel like they're just a goto statement to some arbitrary point in the codebase. 

I'd much rather prefer making the return type something like `string | undefined` to indicate failure and then handling that at the call site.

src/textEditor.ts
```diff
@@ -140,6 +141,32 @@ export class TextEditor {
     return vscode.window.activeTextEditor.document.getText(selection);
   }
 
+  /**
+   *  Retrieves the current word at position.
+   *  If current position is whitespace, selects the right-closest word
+   */
+  static getWord(position: Position) : string {
+    const whitespaceRegex = new RegExp("^ *$");
```
johnfn: Now that we have a settings object, could we put this in there? I don't think anyone sane will ever want to change it, but it'll be nice to have in one location. (I have a bunch of TODO(whitespace) in anticipation of this moment, if you wanna clean those up as well.)

src/textEditor.ts
```diff
@@ -140,6 +141,32 @@ export class TextEditor {
...
+    let start = position;
+    let end = position.getRight();
+
+    const char = TextEditor.getText(new vscode.Range(start, end));
+    if (whitespaceRegex.test(char)) {
+      start = position.getWordRight();
+    } else {
+      start = position.getWordLeft(true);
```
johnfn: Not necessary for this PR, but I really grow to dislike this bare booleans being passed around. They are very hard to read. Much easier to read would be `position.getWordLeft({ inclusive: true })`

src/textEditor.ts
jpoon: I prefer `throw` as it's a bit more declarative of what's going on. It doesn't matter in this case as the caller ignores the error, but eventually what would be nice is to show this at the status bar like how vim does.

src/textEditor.ts
jpoon: Sorry, what `settings` object? Are you referring to moving the regex or the entire function?

src/textEditor.ts
jpoon: Agreed :)

src/textEditor.ts
johnfn: I mean the Configuration object: https://github.com/VSCodeVim/Vim/blob/master/src/configuration/configuration.ts

And just this regex.

jpoon: I don't feel like it belongs there. configuration is for Vim settings right? I don't want people fiddling with this regex.

src/textEditor.ts
johnfn: I see what you're saying, but hailing from other languages with optionals, IMO returning an empty optional is as declarative as throwing an error. I like it because now the return value indicates "this function can fail" and the caller is forced to deal with it, or else get type errors.

As it is, anyone else who calls this function will have to remember that it throws or else they'll get into trouble. 

I could see putting the status message on VimState actually. This seems nice to me because then we'd retain the behavior of updateView taking a VimState and rendering it to screen.

What do you think?

jpoon: > As it is, anyone else who calls this function will have to remember that it throws or else they'll get into trouble.

Fair point. Coming from c#, the compiler would complain if you didn't properly catch an exception.

src/textEditor.ts
johnfn: I was kinda feeling like maybe you would want to, but I can't think of any concrete cases, so yeah, let's not worry about that for now. 

Can you just dump it on here then? https://github.com/VSCodeVim/Vim/blob/master/src/globals.ts

jpoon: Moved it to `globals`. Ended up leaving the name (ie. did not change the class to `constants`)

PR #705, opened 2016-09-04
Fix race condition with switching active text editor.

johnfn:
When you `ctrl-w h` into another editor, the undo history states of the two editors would become intertwined. It could also cause VSCodeVim to crash if they were big enough, because it'd attempt to diff them against each other and take a really long time.

### DIFF
--- a/extension.ts
+++ b/extension.ts
@@ -247,9 +247,11 @@ async function handleActiveEditorChange(): Promise<void> {
 
   taskQueue.enqueueTask({
     promise: async () => {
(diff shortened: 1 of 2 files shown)


johnfn:
(Ping @xconverge just in case you're curious. It's important that we use the TaskQueue as a global locking mechanism to ensure that two updates don't run simultaneously and cause weird issues. I should have noticed this when you submitted your PR, though; it's a bit of a subtle issue!)

Issue #716, opened 2016-09-06
j+j (j + k) works not as expected in 0.1.9


ScenK:
"vim.insertModeKeyBindings": [ { "before": ["j", "k"], "after": ["\<esc\>"] } ]

j+j (j + k) works not as expected in 0.1.9, j + k could not exit insert mode actually -- unless i press \<esc\>. See screen recode.

![untitled](https://cloud.githubusercontent.com/assets/2987542/18260617/f12062c2-7421-11e6-8fa6-855fbabadde9.gif)

What did you do?

j + k

What did you expect to happen?

Change modes

What happened instead?

jk was inserted

Technical details:

VSCode Version: 1.5.0-insider
VsCodeVim Version: 0.1.9
OS: OSX

deshiknaves:
Yep, same issue on Windows 10. I'm sure my mac machine will be the same.

jpoon:
Definitely a bug. Case insensitivity got all screwed up with the `remapper`, fix on its way.

deshiknaves:
Sweet. I just realized that I can work around this using:

`"vim.insertModeKeyBindings": [
        {
            "before": ["j", "j"],
            "after": ["escape"]
        }
   ],`

jpoon:
Releasing another version of the extension shortly.

PR #721, opened 2016-09-06
Remapper fixes

jpoon:
* Fix for #716. In the config, you can now set `<Esc>` or `<esc>`
* Moved angle notation stuff to it's own class

### DIFF
--- a/extension.ts
+++ b/extension.ts
@@ -7,14 +7,13 @@
  */
 
 import * as vscode from 'vscode';
-import * as util from './src/util';
 import * as _ from "lodash";
 import { showCmdLine } from './src/cmd_line/main';
 import { ModeHandler } from './src/mode/modeHandler';
 import { TaskQueue } from './src/taskQueue';
 import { Position } from './src/motion/position';
 import { Globals } from './src/globals';
-
+import { AngleBracketNotation } from './src/notation';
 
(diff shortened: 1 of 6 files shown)


src/actions/notation.ts
```diff
@@ -0,0 +1,27 @@
+
+export class AngleBracketNotation {
+  private static _angleBracketNotationMap = {
+    'ctrl+' : 'C-',
+    'escape': 'Esc',
+    'backspace': 'BS',
+    'delete': 'Del',
+  };
+
+  public static Translate(key: string): string {
```
johnfn: Can you call this something like normalize?

johnfn: Also does this normalize eg esc to Esc and ESC to Esc?

johnfn:
Check my 2 comments but LGTM otherwise if it fixes the jj problem.

johnfn:
Why is xconverge's commit showing up in the list of commits?

src/actions/notation.ts
jpoon: Done and added tests to verify.

jpoon:
@johnfn hmm, GitHub doesn't seem to display the merges very well. In any case, I burned the old branch. All is well.

Issue #666, opened 2016-08-29
Give us feedback: Vim keybindings in debugging console


johnfn:
It looks like the latest version of VSCode will give us the option to use Vim keybindings in the debugging console as well as the main text editor.

Is this something you think should be on by default or off by default? Please use the 😆  reaction for 'on' and the 😕  reaction for 'off'. (You can also leave a comment.)

rebornix:
I see it as a newly introduced or bug as I remember previously `type` command doesn't cover debugging console area, it's for `editor` :(

johnfn:
@rebornix feel free to join in on the conversation at https://github.com/Microsoft/vscode/issues/10451 as I feel the same way.

PR #723, opened 2016-09-06
Disable Vim Mode in Debug Repl

rebornix:
Yay! We love PRs! 🎊

This PR demonstrates the idea of disabling Vim keybindings in Debug Repl as Debug Repl is now inside the Editor Repl. It doesn't mean we have to merge this in, just want to show team how it works and what's the catch.


jpoon:
Damn...that is a lot of if/else statements we need for this :(

rebornix:
@jpoon yup, that's for other Code team members to see the dirty workaround with current Code behavior. It's real dirty.

johnfn:
Can you do all of these just once inside keyPressed in modeHandler? Or is there something I'm missing?

jpoon:
> Can you do all of these just once inside keyPressed in modeHandler? 

That's what I was thinking too. But, with the way the code is now, it shows how *bad* of a hack it is and VSCode should really fix it on their end :)

rebornix:
@johnfn As we catch Code's keystrokes like `Backspace`, I don't know how to delegate back this keystroke to Code. If that's possible, then the solution is quite clean somehow. I may need to do some searching or maybe you guys already have ideas.

jpoon:
I believe all keystrokes go through

```
```

johnfn:
> it shows how bad of a hack it is and VSCode should really fix it on their end

If @rebornix's solution is really the only way to go about this, IMO it shows more that we architected our own code poorly. :/

jpoon:
Related: https://github.com/VSCodeVim/Vim/issues/666

rebornix:
@jpoon @johnfn now the code looks more reasonable :) We are going to mitigate this issue (like disabling debug panel from editor pane in extension host) this sprint but now this fix can get rid of the issue real quick, even though the additional `inDebugRepl ` check is still dirty.

package.json
```diff
@@ -45,127 +45,127 @@
...
         "key": "right",
         "command": "extension.vim_right",
-        "when": "editorTextFocus && vim.mode != 'Insert Mode'"
+        "when": "editorTextFocus && vim.mode != 'Insert Mode' && !vim.debugInput"
       },
       {
         "key": "up",
         "command": "extension.vim_up",
-        "when": "editorTextFocus && vim.mode != 'Insert Mode'"
+        "when": "editorTextFocus && vim.mode != 'Insert Mode' && !vim.debugInput"
```
jpoon: Is it possible to move this check to a more central location? Maybe inside 'handlekey'?

rebornix: The catch is we can't delegate back modifiers like `ctrl` to Code, that's also why we need to put `vim.mode != 'Insert Mode'` here :(

jpoon: Ah right. Brain fart. 

Change is ugly :( but LGTM.

jpoon: Oh wait. Cause these ctrl modifiers are configured they *should* already go through handlekeys right?

PR #731, opened 2016-09-07
Add support for appending to named registers

dominic31:
Named registers should be case insensitive, with lower case denoting replace register contents, and upper case denoting append to register contents.

[Vim documentation, Named Registers](http://vimdoc.sourceforge.net/htmldoc/change.html#quote_alpha)

### DIFF
--- a/src/register/register.ts
+++ b/src/register/register.ts
@@ -61,7 +61,7 @@ export class Register {
         if (typeof existingText === 'string') {
           content = existingText + (existingText.endsWith('\n') ? '' : '\n') + content;
         } else {
-          content = existingText.concat(content.split('\n'))
+          content = existingText.concat(content.split('\n'));
         }
       } else {
         if (typeof existingText === 'string') {
--- a/src/register/register.ts
+++ b/src/register/register.ts
@@ -90,7 +90,7 @@ export class Register {
       throw new Error(`Invalid register ${register}`);
     }
 
-    // Register names are case-insensitive.
+    // Retrieving register content is case-insensitive
     if (/^[A-Z]$/.test(register)) {
       register = register.toLowerCase();
     }
--- a/src/register/register.ts
+++ b/src/register/register.ts
@@ -61,7 +61,7 @@ export class Register {
         if (typeof existingText === 'string') {
           content = existingText + (existingText.endsWith('\n') ? '' : '\n') + content;
         } else {
-          content = existingText.join('\n') + '\n' + content;
+          content = existingText.concat(content.split('\n'))
         }
       } else {
         if (typeof existingText === 'string') {
--- a/src/register/register.ts
+++ b/src/register/register.ts
@@ -42,7 +42,7 @@ export class Register {
    * register ".
    */
   public static put(content: string | string[], vimState: VimState): void {
-    const register = vimState.recordedState.registerName;
+    let register = vimState.recordedState.registerName;
 
     if (!Register.isValidRegister(register)) {
       throw new Error(`Invalid register ${register}`);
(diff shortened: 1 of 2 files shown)


src/register/register.ts
```diff
@@ -52,6 +52,27 @@ export class Register {
...
+    // Upper-case register name denotes append.
+    if (/^[A-Z]$/.test(register)) {
+     register = register.toLowerCase();
+     let existingText = Register.registers[register] && Register.registers[register].text;
+     if (existingText) {
+      if (typeof content === 'string') {
+        if (typeof existingText === 'string') {
+          content = existingText + (existingText.endsWith('\n') ? '' : '\n') + content;
+        } else {
+          content = existingText.join('\n') + '\n' + content;
```
johnfn: Careful. The reason that content can be a string[] is because of yanking in visual block mode - you copy an entire block, and then paste the whole block later. I'm don't think that flattening down that array is the right thing to do here, because then you lose that block.

src/register/register.ts
```diff
@@ -63,12 +84,17 @@ export class Register {
...
    */
   public static async get(vimState: VimState): Promise<IRegisterContent> {
-    const register = vimState.recordedState.registerName;
+    let register = vimState.recordedState.registerName;
 
     if (!Register.isValidRegister(register)) {
       throw new Error(`Invalid register ${register}`);
     }
 
+    // Register names are case-insensitive.
```
johnfn: This comment is a little bit misleading. Probably you want to say something like `retrieving register content is case-insensitive.`

johnfn:
Thanks for the PR, @dominic31! Looks good to me, just a few concerns about how we handle visual block mode copy/paste.

src/register/register.ts
dominic31: Yes I was unsure about this. The approach I took preserves the type of the new content, so string + string[] -> string[], string[] + string -> string.

Are you saying that if either existingText or content is string[] then the result should be string[]?

dominic31:
@johnfn Thanks for the comments! I've pushed some changes, hopefully it looks better now.

johnfn:
Ergh.

After investigating registers a little more, I believe that I have actually implemented registers incorrectly in the case of visual block mode! It seems like when you copy in visual block mode, you simply copy each line separated by a newline. This is different than what we are doing, where if you copy in visual block mode, you actually copy an array of each line.

I would encourage you to go ahead and fix this, but actually I am anticipating some annoying merge conflicts with a branch I have been working on for a while #587 which adds yet another register case type. 

If you really want to, you could first copy my register changes into here from my PR, and then work on top of them to replace the `string[]` case with `string`. If that sounds too complicated, I don't blame you; I'd just wait until I merge my PR in (optimistically) a few days.

dominic31:
> It seems like when you copy in visual block mode, you simply copy each line separated by a newline.

Where does this happen? When appending to registers in visual block mode, I concatenate the arrays,

jpoon:

dominic31:
@jpoon I was waiting for a reply to my last comment, else it is as far as I can see ready.

jpoon:
@dominic31 My bad. Can you fix the conflicts? Then we can give it a quick review and merge.

src/register/register.ts
```diff
@@ -52,6 +52,27 @@ export class Register {
       clipboard.copy(content);
     }
 
+    // Upper-case register name denotes append.
+    if (/^[A-Z]$/.test(register)) {
```
jpoon: Can you move this into it's own separate static function (`isNamedRegister`), similar to `isValidRegister`?

johnfn:
I'm really sorry I didn't respond to that comment! I'm getting spammed with notifications from Github constantly and it makes it really difficult to find the stuff that matters. :(

jpoon:
@Platzer submitted https://github.com/VSCodeVim/Vim/pull/974 which incompases these same changes. I'm closing this PR in favor of @Platzer mainly as there hasn't been any movement on this in the last week.

PR #733, opened 2016-09-08
Doc for enabling repeating j/k for Insider build

octref:


rebornix:
@octref you did what I'm thinking about, kudos!

PR #736, opened 2016-09-08
Read command

dominic31:
Initial support for read and read!.

[Vim documentation, read](http://vimdoc.sourceforge.net/htmldoc/insert.html#:read)
[Vim documentation, read!](http://vimdoc.sourceforge.net/htmldoc/insert.html#:read!)

Range and opt arguments are not yet implemented.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,259 +1,259 @@
 {
-    "name": "vim",
-    "displayName": "Vim",
-    "description": "Vim emulation for Visual Studio Code",
-    "icon": "images/icon.png",
-    "version": "0.1.10",
-    "publisher": "vscodevim",
-    "galleryBanner": {
-        "color": "#a5c9a2",
-        "theme": "light"
-    },
-    "license": "MIT",
-    "keywords": [
-        "vim",
-        "vi",
-        "vscodevim",
-        "vsc-vim"
+  "name": "vim",
+  "displayName": "Vim",
+  "description": "Vim emulation for Visual Studio Code",
+  "icon": "images/icon.png",
+  "version": "0.1.10",
+  "publisher": "vscodevim",
+  "galleryBanner": {
+    "color": "#a5c9a2",
+    "theme": "light"
+  },
+  "license": "MIT",
+  "keywords": [
+    "vim",
+    "vi",
+    "vscodevim",
+    "vsc-vim"
+  ],
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/VSCodeVim/Vim.git"
+  },
+  "homepage": "https://github.com/VSCodeVim/Vim",
+  "bugs": {
+    "url": "https://github.com/VSCodeVim/Vim/issues"
+  },
+  "engines": {
+    "vscode": "^1.0.0"
+  },
+  "categories": [
+    "Other"
+  ],
+  "activationEvents": [
+    "*"
+  ],
+  "main": "./out/extension",
+  "contributes": {
+    "commands": [
+      {
+        "command": "extension.showCmdLine",
+        "title": "Vim: Show Command Line"
+      }
     ],
-    "repository": {
-        "type": "git",
-        "url": "https://github.com/VSCodeVim/Vim.git"
-    },
-    "homepage": "https://github.com/VSCodeVim/Vim",
-    "bugs": {
-        "url": "https://github.com/VSCodeVim/Vim/issues"
-    },
-    "engines": {
-        "vscode": "^1.0.0"
-    },
-    "categories": [
-        "Other"
+    "keybindings": [
+      {
+        "key": "Escape",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -233,6 +233,7 @@
         "postinstall": "node ./node_modules/vscode/bin/install && gulp init"
     },
     "dependencies": {
+		"child-process": "^1.0.2",
```
rebornix: Indent :)

test/cmd_line/read.test.ts
```diff
@@ -0,0 +1,24 @@
...
+
+  setup(async () => {
+    await setupWorkspace();
+    modeHandler = new ModeHandler();
+  });
+
+  teardown(cleanUpWorkspace);
+
+  test("Can read shell command output", async () => {
+    await runCmdLine('r! echo "hey ho"', modeHandler);
```
rebornix: I suppose the result of this command differs on different platform, which currently breaks the test on Windows.

rebornix:
Thanks for taking things that are currently not on our list :) Just drop some comments and can you update the roadmap in your PR as well, especially add `Note` to it if `child-process` is not OS-compatible (I'm not sure).

test/cmd_line/read.test.ts
dominic31: I just didn't realise that Windows prints the "s in echo "..". I modified the test so that it passes on Windows too.

dominic31:
I've updated the roadmap and dropped one comment at least :)

src/cmd_line/commands/read.ts
```diff
@@ -0,0 +1,81 @@
...
+      return await this.getTextToInsertFromFile();
+    } else if (this.arguments.cmd && this.arguments.cmd.length > 0) {
+      return await this.getTextToInsertFromCmd();
+    } else {
+      throw Error('Invalid arguments');
+    }
+  }
+
+  async getTextToInsertFromFile() : Promise<string> {
+    // TODO: Substitute utf8 with current file encoding, couldn't find this anywhere in vscode's api.
```
jpoon: Is there an issue tracking this with VSCode? Some initial Google says there's some config at the workspace level where you can set the encoding. Maybe we can read that config?

test/cmd_line/read.test.ts
```diff
@@ -0,0 +1,24 @@
...
+
+  setup(async () => {
+    await setupWorkspace();
+    modeHandler = new ModeHandler();
+  });
+
+  teardown(cleanUpWorkspace);
+
+  test("Can read shell command output", async () => {
+    await runCmdLine('r! echo heyho', modeHandler);
```
jpoon: What happens with a more complex example? 

`:r! echo hey\\n\\tho`

jpoon:
@dominic31 awesome work! dropped a few comments.

test/cmd_line/read.test.ts
dominic31: 1. This works for me on Ubuntu, probably fails on Windows though :)
2. The error that is thrown is caught in `runCmdLine()` that attempts to display an error message the status bar, this is the same mechanism that would handle  `:thisexcommanddoesnotexist`. I don't see anything in the status bar though, I guess that's another issue.

dominic31:
@jpoon thanks for the comments! I've left some replies.

test/cmd_line/read.test.ts
jpoon: 1. Can you replace the existing test with `:r! echo hey\\n\\tho`? Helps test more cases :)
2. Yeah, that's a backlog work item to display it on the status bar

jpoon:
LGTM.

* If you choose not to include the detection of the current file encoding, can you file a separate issue so we can track it?
* Update the test

test/cmd_line/read.test.ts
dominic31: 1. Ah, that's what you meant. Done.

dominic31:
I've updated the unit test. 

About the encoding, there's an [open issue](https://github.com/Microsoft/vscode/issues/824) on the VS Code project for supporting encoding related functions through the api. It hasn't got past the backlog yet. Shall I open an issue here that refers to that issue?

jpoon:
Yes please!

One last thing, can you merge with master and resolve the conflicts?

dominic31:
Sorry, I've gotten things a little muddled up here, the encoding parameter in fs.readFile() is of course the encoding of the external file that gets read in, not the current file in VS Code. Looking at the vim docs for read (link in the first comment), the encoding can be specified in the opt arguments that I haven't yet implemented. I'll update the comment in getTextToInsertFromFile().

Should I open an issue for implementing opt args, or just mention it in the roadmap?

johnfn:
I'm okay with this not having ++opt, that can be a separate task. :)

PR #752, opened 2016-09-10
fixes #750

### LINKED ISSUE #750: A "broken link" in README.md

-----

In the "Supported Options" section in README.md, there is a link to the 155th line of `package.json`, but that line has changed:

* https://github.com/VSCodeVim/Vim/blob/f3127d9c09ab9156d97717b3fda8881e0b850e69/package.json#L155 (old)
* https://github.com/VSCodeVim/Vim/blob/master/package.json#L155 (now)

xconverge:
minor, might want to change the way we do this for the future, especially since there is the extension contribution section now


johnfn:
Actually we can just point them at the contributions section of the plugin page in VSCode right? Since that was added in 1.5.0.

PR #755, opened 2016-09-11
build on extension/test launch

jpoon:
Request for comment cc: @johnfn @rebornix 

Force a build via `gulp` on extension launch/test. I tend to forget to run `gulp watch` manually; often times I would change the source, run it, and wonder why the behaviour didn't change before the duh moment.

With this change,

* Pro: Guaranteed that the TS is compiled; no need to run `gulp watch` or `npm compile` manually and check console if things compiled correctly. auto-runs tslint.
* Con: Adds build time (5s right now) before you can launch the extension/run tests

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,30 +1,28 @@
 // A launch configuration that compiles the extension and then opens it inside a new window
 {
-	"version": "0.1.0",
-	"configurations": [
-		{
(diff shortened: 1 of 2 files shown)


rebornix:
LGTM. I put the check in pre-commit but definitely this change is necessary.

PR #767, opened 2016-09-14
fixes #739

### LINKED ISSUE #739: cw doesn't work on empty line
### What did you do?

```js
if (true)
|// Nothing other than newline here
}
```

Press `cw`

```js
if (true) {
|// Nothing other than newline here
}
```

in insert mode.

### What happened instead?

```js
if (true) {

}|
```

In insert mode.

### Technical details:

* VSCode Version: 1.5.0
* VsCodeVim Version: 0.1.10
* OS: OS X 10.11.6

xconverge:
Weird bug, weird fix, but I am ok with it if you are

Yay! We love PRs! 🎊

### DIFF
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -1281,19 +1281,13 @@ export class ChangeOperator extends BaseOperator {
         const isEndOfLine = end.character === TextEditor.getLineAt(end).text.length - 1;
         let state = vimState;
 
-        // If only EOL on line do nothing except change to insert
-        if (TextEditor.getLineAt(start).text.length < 1) {
-          vimState.cursorPosition = vimState.cursorStartPosition;
-          vimState.currentMode = ModeName.Insert;
-          return vimState;
-        }
-
         // If we delete to EOL, the block cursor would end on the final character,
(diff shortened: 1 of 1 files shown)


johnfn:
How does this handle, say, `c}` in the same situation?

I think this may be a special case on `w`, not `c`.

xconverge:
hah ok so you are right, I fixed this on w, however c} does not work right (and did not before so I will fix that as well since you brought it up)

rebornix:
Just one comment and you may also want to write a simple test to cover this scenario.

johnfn:
I don't understand why this behavior would happen for `cw` but not `dw`. I see two possibilities:

1. We have another special case handling thing in `d`. If so, they should both be moved into `w`. 
2. Something very strange and worrying is going on. If so, figure out the source of the strangeness and worrisomeness.

xconverge:
@johnfn what is the special case I have in "c" now?

johnfn:
My bad, @xconverge. I misread your PR. (And also I completely missed your reply. I have trouble with Github's notification system...) It looks good to me. If you bring it up to date, I'll merge it! :)

PR #768, opened 2016-09-14
Insert Previous text

rebornix:
Yay! We love PRs! 🎊

This PR implements `ctrl-A` and `ctrl-@` in Insert Mode, which inserts previously change. Instead of maintaining the change in history ourselves, I'm currently just leveraging `documentContentChangeEvent`. This way when we type `ab` and intellisense helps us autocomplete it to `abstract`, we can get the whole `abstract` through `contentChange`, not just `ab`.

@johnfn @xconverge this is totally a different way of handing last change which we currently maintained in HistorySteps, so for now I just put the plain array in HistoryTracker. Please take a look and see if it can inspire you on how to handle history.

The only catch here is there might be racing condition between `documentChange` and `type`. So basically if we write test cases for this feature and run case automatically, or a human types real quick, you might see bugs as `ctrl-A` may come before all `documentChange`.

### DIFF
--- a/extension.ts
+++ b/extension.ts
@@ -14,6 +14,7 @@ import { taskQueue } from './src/taskQueue';
 import { Position } from './src/motion/position';
 import { Globals } from './src/globals';
 import { AngleBracketNotation } from './src/notation';
+import { ModeName } from './src/mode/mode';
 
 interface VSCodeKeybinding {
   key: string;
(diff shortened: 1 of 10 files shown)


rebornix:
@johnfn Agree, personally I want to do the refactor as well since with content change, the tab complete issue we mentioned in https://github.com/Microsoft/vscode/issues/1431 and https://github.com/Microsoft/vscode/issues/10801 will be gone. Besides, we can directly operate on content changes instead of calculating all the key strokes. We don't need Code's customization on Undo Stop and history API as we can merge content change and pick what we need by will.

The only problem right now is the racing condition between handlers for `docuementTextContentChange` and `type`, like when we type `ab<Tab>` then `c`, we can't make sure when we handle `c`, we've already received all content change made by `ab<Tab>`. But this issue is already addressed and let's see what we can do these few days.

johnfn:
Okay! The history tracker refactor will be a very significant change; I definitely encourage you to start a new PR on that one. 

Honestly, I think it could actually be quite difficult for you to do since you didn't implement it in the first place, so no shame if you just want to wait for me to get to it!

rebornix:
I'd like to have this merged to avoid additional manual merging (like cursor related change) then see how to refactor History Tracker.

Right now I have 5 test cases to cover this feature but that's not enough as Code Snippet has some issues with cursor position yet, and auto-complete is not testable.

rebornix:
Have it merged first, there might be following code reviews or suggestions and let's see.

Issue #780, opened 2016-09-18
Add "+ register for accessing the clipboard


bdchauvette:
Please *thumbs-up* 👍 this issue if it personally affects you! You can do this by clicking on the emoji-face on the top right of this post. Issues with more thumbs-up will be prioritized.

-----

### What did you do?

1. Select text, e.g. using `V`
2. `"+y`

### What did you expect to happen?

The selected text to be copied to the system clipboard (see Vim documentaiton for [`quoteplus`](http://vimdoc.sourceforge.net/htmldoc/gui_x11.html#quoteplus))

### What happened instead?

Nothing 😞 

### Technical details:

* VSCode Version:  **1.5.2**
* VsCodeVim Version: **0.1.10**
* OS: **Windows 10**

-----

Vim has two clipboard registers, `"+` and `"*`. VsCodeVim currently only supports `"*`.

I think on most platforms, the registers point to the same clipboard text, but they behave differently on Linux: `"+` access the X selection `CLIPBOARD` (used for e.g. ctrl+c/x/v), while `"*` access the X selection `PRIMARY` (used by middle-clicking).

I'm used to using the `"+` register to access the clipboard, , so it would be great if VsCodeVim could add support for it 😅 

I don't think it's really worth it (or even possible?) to emulate the semantics of the different registers, but it would be nice to have basic support for accessing the clipboard from both `"+` and `"*`.

bdchauvette:
If you think it's worth implementing, I can take a stab at fixing this.

Seems like it should only require a few changes (lines [51][], [68][], and [97][] in `register/register.ts`) from

[51]: https://github.com/VSCodeVim/Vim/blob/99a7c434246047ac7f91aabe99804925b7af5a3c/src/register/register.ts#L51
[68]: https://github.com/VSCodeVim/Vim/blob/99a7c434246047ac7f91aabe99804925b7af5a3c/src/register/register.ts#L68
[97]: https://github.com/VSCodeVim/Vim/blob/99a7c434246047ac7f91aabe99804925b7af5a3c/src/register/register.ts#L97

```js
if (register === '*') {
```

to something like

```js
if (register === '*' || register === '+') {
```

Alternatively, it could be done with a regex like:

```js
const isClipboardRegister = /\+|\*/;


if (isClipboardRegister.test(register)) {
```

[...]

PR #782, opened 2016-09-19
Support "+ system clipboard register (#780)

bdchauvette:
This adds support for accessing the system clipboard using the `"+` register, addressing #780.

The `"+` register is basically an alias for the `"*` register, so I didn't make any changes to [how the `useClipboard` setting is handled](https://github.com/VSCodeVim/Vim/blob/master/src/mode/modeHandler.ts#L372).

### DIFF
--- a/src/register/register.ts
+++ b/src/register/register.ts
@@ -15,8 +15,9 @@ export enum RegisterMode {
 };
 
 export interface IRegisterContent {
-  text        : string | string[];
-  registerMode: RegisterMode;
+  text                : string | string[];
+  registerMode        : RegisterMode;
+  isClipboardRegister?: boolean;
 }
 
 export class Register {
@@ -31,14 +32,18 @@ export class Register {
    */
   private static registers: { [key: string]: IRegisterContent } = {
     '"': { text: "", registerMode: RegisterMode.CharacterWise },
-    '*': { text: "", registerMode: RegisterMode.CharacterWise },
-    '+': { text: "", registerMode: RegisterMode.CharacterWise }
+    '*': { text: "", registerMode: RegisterMode.CharacterWise, isClipboardRegister: true },
+    '+': { text: "", registerMode: RegisterMode.CharacterWise, isClipboardRegister: true }
   };
 
-  private static systemClipboardRegisters: string[] = ["*", "+"];
+  public static isClipboardRegister(registerName: string): boolean {
+    const register = Register.registers[registerName];
 
-  public static isSystemClipboardRegister(register: string): boolean {
-    return Register.systemClipboardRegisters.indexOf(register) !== -1;
+    if (register && register.isClipboardRegister) {
+      return true;
+    }
+
+    return false;
   }
 
   public static isValidRegister(register: string): boolean {
@@ -56,13 +61,14 @@ export class Register {
       throw new Error(`Invalid register ${register}`);
     }
 
-    if (Register.isSystemClipboardRegister(register)) {
+    if (Register.isClipboardRegister(register)) {
       clipboard.copy(content);
     }
 
     Register.registers[register] = {
-      text        : content,
-      registerMode: vimState.effectiveRegisterMode(),
+      text               : content,
+      registerMode       : vimState.effectiveRegisterMode(),
+      isClipboardRegister: Register.isClipboardRegister(register),
     };
   }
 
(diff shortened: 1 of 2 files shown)


bdchauvette:
@rebornix Awesome! Just let me know what part you'd like commented, and I'll add it when I get a chance :+1:

src/register/register.ts
```diff
@@ -30,9 +31,16 @@ export class Register {
...
   private static registers: { [key: string]: IRegisterContent } = {
     '"': { text: "", registerMode: RegisterMode.CharacterWise },
-    '*': { text: "", registerMode: RegisterMode.CharacterWise }
+    '*': { text: "", registerMode: RegisterMode.CharacterWise },
+    '+': { text: "", registerMode: RegisterMode.CharacterWise }
   };
 
+  private static systemClipboardRegisters: string[] = ["*", "+"];
+
+  public static isSystemClipboardRegister(register: string): boolean {
```
johnfn: Instead of this, make IRegisterContent have another optional field called `isClipboardRegister` and put the value on there.

bdchauvette: Good call! I've switched the check to look up the `isClipboardRegister` property on the register.

(For consistency's sake, I also renamed the check from `Register.isSystemClipboardRegister` to just `Register.isClipboardRegister`. Saves a few keystrokes too!)

src/register/register.ts
```diff
@@ -30,9 +32,20 @@ export class Register {
...
     '"': { text: "", registerMode: RegisterMode.CharacterWise },
-    '*': { text: "", registerMode: RegisterMode.CharacterWise }
+    '*': { text: "", registerMode: RegisterMode.CharacterWise, isClipboardRegister: true },
+    '+': { text: "", registerMode: RegisterMode.CharacterWise, isClipboardRegister: true }
   };
 
+  public static isClipboardRegister(registerName: string): boolean {
+    const register = Register.registers[registerName];
+
+    if (register && register.isClipboardRegister) {
```
johnfn: Can you just do `return register && register.isClipboardRegister`?

bdchauvette: I actually tried that initially, but tsc doesn't like it very much:

```
> vim@0.1.10 vscode:prepublish c:\Users\Ben\Code\vscode-vim
> node ./node_modules/vscode/bin/compile -p ./
src/register/register.ts(42,12): error TS2322: Type 'boolean | undefined' is not assignable to type 'boolean'.
  Type 'undefined' is not assignable to type 'boolean'.
```

It works fine if I change the signature to `isClipboardRegister(string): boolean | undefined`, but I wasn't sure if that was preferable to having a messier function body.

johnfn: Can you show me the exact line(s) of code that you're using to get that error?

bdchauvette:
```typescript
  public static isClipboardRegister(registerName: string): boolean {
    const register = Register.registers[registerName];
    return register && register.isClipboardRegister;
  }
```

capture

I think it's being caused by the `isClipboardRegister` property being optional: the property has a type signature of `boolean | undefined`, but the `Register.isClipboardRegister` method expects to only return booleans. 

If we try to return the property without first converting it to a `boolean`, it's possible for the value to `undefined` (b/c it's optional), which violates the signature for the checking method.

(I think) this would also explain why it works if I change the allowed return type to `boolean | undefined`.

I'm pretty new to typescript, so I'm not sure what the idiomatic way to handle this kind of thing is 😕

bdchauvette: Each of these allow the code to compile without a hitch:

```ts
// Cast to boolean at runtime
return register && Boolean(register.isClipboardRegister);

// or:
return register && !!register.isClipboardRegister;
```

```ts
// Type assertion
return register && (register.isClipboardRegister as boolean);
```

```ts
// No return type annotation
public static isClipboardRegister(registerName: string) {
  const register = Register.registers[registerName];
  return register && register.isClipboardRegister;
}
```

johnfn: Ohhhhh. It's because it's an optional property, which has type `boolean | undefined`, which we are returning directly. That makes sense... in a weird way. :) 

I guess TypeScript is pointing out that we did a dumb thing by saying the type was `isClipboardRegister?: boolean` since now we can't distinguish `undefined` from `false` in an easy way.

Best is probably to change the type to `isClipboardRegister: boolean`, and add it to all registers. Then this code will work and also our `undefined` vs `false` case will go away. 

TypeScript caught a (minor) bug! Yay, TypeScript!

bdchauvette: Yay, TypeScript! 🙌 

Looks like I just broke Travis, though. Any ideas what's going on there? 😕

jpoon: ¯\_(ツ)_/¯ restarted the Travis build.

johnfn:
Annnd it's good! 

Thanks @bdchauvette for the PR! :)

bdchauvette:
Awesome! Thanks for such a great project :D

PR #796, opened 2016-09-23
Add autoindent to README, fix hlsearch default

srenatus:
I believe [hlsearch is false by default](https://github.com/VSCodeVim/Vim/blob/214619d8d61524960c2c4d099a6a120e3a0e91a2/src/configuration/configuration.ts#L56).

Trivial change, haven't run tests 😃


rebornix:
@srenatus thanks for the help!

PR #811, opened 2016-09-28
Multi-Cursor Mode v 2.0

johnfn:
The other one got way out of control because I was trying to refactor the entire codebase. Undoing all that insanity and cutting it down to basics!


.vscode/settings.json
```diff
@@ -13,8 +13,12 @@
 			}
 	],
 
+	"vim.hlsearch": false,
```
jpoon: TBH, I feel like we shouldn't be checking in these settings as they are very much user-specific.

src/actions/actions.ts
```diff
@@ -254,6 +258,17 @@ export abstract class BaseCommand extends BaseAction {
    */
   isCompleteAction = true;
 
+  supportsMultiCursor = true;
+
+  multicursorIndex: number | undefined = undefined;
```
jpoon: Nitpicking: I feel like you should capitalize the 'c'

src/actions/actions.ts
```diff
@@ -254,6 +258,17 @@ export abstract class BaseCommand extends BaseAction {
...
+
+  /**
+   * In multi-cursor mode, do we run this command for every cursor, or just once?
+   */
+  public runsOnceForEveryCursor(): boolean {
```
jpoon: Can this be a property instead?

jpoon: Ah, I see why you did it this way.

src/actions/actions.ts
```diff
@@ -269,12 +284,42 @@ export abstract class BaseCommand extends BaseAction {
...
+      for (let i = 0; i < timesToRepeat; i++) {
+        vimState = await this.exec(position, vimState);
+      }
+
+      return vimState;
+    }
 
-    for (let i = 0; i < timesToRepeat; i++) {
-      vimState = await this.exec(position, vimState);
+    let timesToRepeat  = this.canBePrefixedWithCount ? vimState.recordedState.count || 1 : 1;
```
jpoon: This is the same as line 288? Can we execute this once outside the if block?

src/actions/actions.ts
```diff
@@ -269,12 +284,42 @@ export abstract class BaseCommand extends BaseAction {
...
+    let resultingCursors: Range[] = [];
+    let i              = 0;
+
+    const cursorsToIterateOver = vimState.allCursors
```
jpoon: Mind leaving a comment here about what's happening?

src/actions/actions.ts
johnfn: hmm, i am being inconsistent here...

src/actions/actions.ts
johnfn: I actually explained what was going on over here: 

https://github.com/VSCodeVim/Vim/pull/811/files#diff-a7723f32fe6b05e135838f84e5c23c0fR965

Unfortunately I had to have similar logic in 2 places but I think the alternative is even worse. :P

src/actions/actions.ts
johnfn: Yeah, good point.

.vscode/settings.json
johnfn: My settings.json is objectively the best though... 😉 

Okay, okay, we should probably add it to .gitignore. :P

PR #815, opened 2016-10-01
Fix issue where could not use I while in visual mode

xconverge:
Yay! We love PRs! 🎊


src/actions/actions.ts
```diff
@@ -2112,8 +2112,8 @@ class CommandGoToDefinition extends BaseCommand {
 
 @RegisterAction
 class CommandInsertAtFirstCharacter extends BaseCommand {
-  modes = [ModeName.Normal];
-  mustBeFirstKey = true;
+  modes = [ModeName.Normal, ModeName.Visual];
+  mustBeFirstKey = false;
```
johnfn: Are you sure that changing this to `false` doesn't cause any regressions?

xconverge: @johnfn I am definitely not sure...but I know it is harming this particular action

I am not sure what else I can do other than run the tests that pass and cross my fingers..

johnfn: I *think* the reason was (in the case of i) so that `ciw` would not cause an Insert to trigger and make things get really confused.

I'm not sure if that applies here, though! I'm not sure if `I` is inside of another Vim command somewhere... I don't think it is, but then again, it's a bit scary...

xconverge: @johnfn  I actually think it is wrong to be first key, in Vim if you have a line of text, and in the middle press 4At<Esc> it will append 4 t's to the end. With this set to true that definitely can't happen right?

xconverge: Same with I using 4It*

johnfn: Hmm, yeah, that's a good point. Plus, I can't think of any actions that have `A` in the middle of them.

johnfn:
Thanks again, @xconverge!

PR #816, opened 2016-10-01
Fix to allow A while in visual mode

xconverge:
Yay! We love PRs! 🎊


johnfn:
This is the Visual Block Mode A. You want the Insert Mode A, right?

xconverge:
@johnfn no idea what I was doing here, really don't

Same as the other one though, I can find any regressions but can't say I am fluent enough with the commands to know exactly how they could work on all of the edge cases

johnfn:
Been thinking about this more and I think it's okay. Could you just remove the line 

`mustBeFirstKey = false;`

entirely (as `mustBeFirstKey` defaults to `false` so there's no reason to set it). 

Same thing with your other PR.

johnfn:
Thanks, @xconverge!

PR #818, opened 2016-10-02
Clean up typings

jpoon:
* What started as a small itty bitty change in `extension.ts` ended up with a full cleaning of our typings
* Clean up typings. Namely, typescript definitions should not be included in the source. They are defined in `typings.json` and installed with `npm install`


johnfn:
LGTM. I actually checked in the typings a while back because we had some stuff that hadn't been added to the official typings, but that's no longer the case.

jpoon:
Thanks @johnfn!

PR #827, opened 2016-10-04
X eats eol

xconverge:
This one was actually super annoying, let me know what you think of the way I addressed it. The original report was from user iivvoo on slack

"since we're talking about obscure issues: In vscode you can click one character past the last character (sort of the newline character?)
[6:30]  
x-ing it will join two lines
[6:31]  
sometimes I click way past the characters to the right to put the cursor at the end, 'x'-ing to remove the last character doesn't have the expected behaviour"

Yay! We love PRs! 🎊


johnfn:
This looks good to me. Thanks for fixing this annoying issue, @xconverge! :)

PR #835, opened 2016-10-04
[WIP] Use new transformation style in delete and paste

johnfn:
Refactoring this massive codebase into this new style is... tough.

The more we do it, though, the more awesome multi-cursor becomes. And also, the more consistent our codebase becomes. Which is nice for anyone who wants to add stuff to it. 😉 

### Todo

- [x] 35 failing tests...
- [x] 32 failing tests
- [x] 30 failing tests
- [x] 26 failing tests
- [x] 22 failing tests
- [x] 21 failing tests
- [x] 20 failing tests
- [x] 19 failing tests
- [x] 18 failing tests
- [x] 16 failing tests
- [x] 13 failing tests
- [ ] 1 failing test!

### DIFF
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -3266,28 +3266,28 @@ class ActionReplaceCharacter extends BaseCommand {
   modes = [ModeName.Normal];
   keys = ["r", "<character>"];
   canBeRepeatedWithDot = true;
-  runsOnceForEachCountPrefix = true;
+  runsOnceForEachCountPrefix = false;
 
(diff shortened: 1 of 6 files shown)


xconverge:
I don't really see where diff is used in a lot of cases? is there an obvious base class implementation somewhere I am missing? Trying to understand why this is good :)

johnfn:
@xconverge the diffs are all used in `executeCommand`

Essentially now the way we do cursor position updates is by a diff instead of by an exact location. This is because if you're running a command on lots of cursors in multi-cursor mode, previous inserts could offset your current location, which would invalidate locations you calculated.

PR #852, opened 2016-10-06
Add new logo icon

kevincoleman:
New logo icon file
Updated banner color to match logo
logo source design files in `design/`


xconverge:
This is awesome, thanks @kevincoleman

jpoon:
Mind actually putting it all under /images?

jpoon:
@kevincoleman are the edges of the icon also #208ecd? As we're changing the banner color to #208ecd on https://marketplace.visualstudio.com/items?itemName=vscodevim.vim, want to make sure that they don't bleed into the icon.

kevincoleman:
@jpoon Ahhh. Good point. The edges are the also #208ecd, so that won’t work too well.

Options:
add a white border or background around the icon design to give it somewhat of a “sticker” feel.
pick another banner color so as to avoid the bleed issue.
any other ideas?

jpoon:
Let's go with the easy one -- different banner color.

jpoon:
Sweeet. Thanks @kevincoleman .

Issue #846, opened 2016-10-06
Closing Default Settings window with :q doesn't work


FSMaxB:
Please *thumbs-up* 👍 this issue if it personally affects you! You can do this by clicking on the emoji-face on the top right of this post. Issues with more thumbs-up will be prioritized.

-----

### What did you do?

* In the menu: File > Preferences > User Settings
* Selected the "Default Settings" tab.
* Pressed `:q`

### What did you expect to happen?

The tab should be closed.

### What happened instead?

Nothing.

### Technical details:

* VSCode Version: 1.5.3
* VsCodeVim Version: 0.3.0
* OS: Windows 8.1

### What works:

* Closing all tabs with `:qa`
* `<ctrl-p> >View: Close Editor`

dkalowsk:
👍

xconverge:
@johnfn sorry I forget to say fixes this in the commit message but #855 fixed this

johnfn:
Thanks for letting me know!

PR #855, opened 2016-10-07
Unnecessary quit check on untitled files

xconverge:
Is there a reason this was here? This PR is more of a question, but could be merged if it makes sense...

Shouldnt :q on a new untitled file work? On MacVim if I open it and immediately do :q it quits

This is also causing #846

### DIFF
--- a/src/cmd_line/commands/quit.ts
+++ b/src/cmd_line/commands/quit.ts
@@ -29,10 +29,6 @@ export class QuitCommand extends node.CommandBase {
   }
 
   async execute() : Promise<void> {
-    if (this.activeTextEditor.document.isUntitled && !this._arguments.bang) {
-      throw error.VimError.fromCode(error.ErrorCode.E32);
-    }
-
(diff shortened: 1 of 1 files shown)


xconverge:
duh... I need to check if it is dirty...

Platzer:
Thats there because VIM also do nothing when you try to `:q`uit a unnamed file.
grafik

But i think the `Default Settings` File is `isUntitled = true` because i don't think it is an actual file?

xconverge:
If I open vim, and have an empty file, and do :q, it will quit...

Platzer:
Yes you are right, it is a unnecessary check.

rebornix:
LGTM.

xconverge:
That being said, the isDirty check that comes after this seems to be funky..I will look into it

xconverge:
Yep!

johnfn:
Awesome, thanks @xconverge!

Issue #842, opened 2016-10-05
j/k moving diagonally


octref:
As you see it's moving diagonally.

![j](https://cloud.githubusercontent.com/assets/4033249/19123350/26489276-8ae3-11e6-8790-72321cffdb97.gif)
### Technical details:
- VSCode Version: 1.6 Insiders
- VsCodeVim Version: 0.3
- OS: OS X 10.11.6

xconverge:
I cant reproduce this, does it happen all of the time for you?

octref:
Seems to happen if I'm moving from outside of a `{}` to inside.

![jk](https://cloud.githubusercontent.com/assets/4033249/19128887/120a9220-8afa-11e6-859d-7998e2b24e10.gif)

johnfn:
I can't reproduce this either. Do you have any other extensions running? Also, what language is that? Is it TypeScript?

octref:
@johnfn @xconverge 

Figured out the problem. This happens when indentation is tab.

octref:
Now it makes sense. The fix would be make a tab `tabSize` long instead of 1 char long when calculating movement.

xconverge:
@octref right on! Reproduced the issue, thanks for finding it

rebornix:
@octref good catch.

We need take `tabStop` into consideration while calculating `desiredColumn`.

xconverge:
@johnfn is working on this in #859

octref:
@xconverge This seems to be fixed so closing it.

johnfn:
This was not fixed. It still appears to be a problem in master.

jpoon:
Given the old age of this bug, the lack of activity, and that I can't repro, closing as this bug has likely been fixed on the latest release.

PR #859, opened 2016-10-08
[WIP] Fix whitespace/tab j/k issues

johnfn:
This is an implementation that sort of works but is annoying. 

Does anyone know a better way to do this?
- @rebornix, others - `tabstop` does not always report the correct width of a tab - it seems like the setting can be overridden in-editor somehow. Do you know how to get the right width?
- anyone (but maybe @rebornix knows..?) - is there some provided way to move up/down through a document that has spaces and tabs while keeping the cursor on the right column?

### DIFF
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -2333,12 +2333,19 @@ class MoveDown extends BaseMovement {
   doesntChangeDesiredColumn = true;
 
   public async execAction(position: Position, vimState: VimState): Promise<Position> {
-    return position.getDown(vimState.desiredColumn);
+    const { newPosition, newDesiredColumn } = position.getDownWhileObservingTabstops(vimState.desiredColumn);
+
+    vimState.desiredColumn = newDesiredColumn;
+    return newPosition;
   }
 
   public async execActionForOperator(position: Position, vimState: VimState): Promise<Position> {
+    const { newPosition, newDesiredColumn } = position.getDownWhileObservingTabstops(vimState.desiredColumn);
+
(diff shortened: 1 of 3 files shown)


xconverge:
@johnfn I have some stuff stashed from yesterday that is done EXACTLY the same way as this but I didnt get to finish the implementation

What you are looking for is 
```
vscode.window.activeTextEditor.options.tabSize;
```

I added a getTabSize() to TextEditor that returned this

johnfn:
Aww haha man sorry @xconverge! Good find on the tabSize thing though. Perhaps we should use that in Configuration. 

You can always let me know on slack or whatever if you're working on a bigger feature. Also don't be afraid of opening WIP PRs earlier rather than later. :)

rebornix:
@johnfn I missed this PR for a long time. Yes @xconverge is right, that reading config from `activeTextEditor.options` is always correct. About moving cursor up and down without calculating desired column ourselves is command `cursorMove`, you have seen that quite a few times when I implemented those scrolling features.

xconverge:
@johnfn @rebornix Is this fixed somehow (assuming vscode reports different character values now?) assuming #842 is now closed?

johnfn:
How could this be fixed?? o_O

xconverge:
My thought is that vscode may report a tab char as more than 1?

johnfn:
That would be weird because then what would it mean to insert at character 1 on a line filled with tabs? 

meh, if it works it works :P

jpoon:
Spring cleaning. Closing this PR as there hasn't been movement in awhile. Feel free to re-open.

PR #863, opened 2016-10-10
Remove unused modehandlers

xconverge:
Ignore the extra commits, I will clean up my fork after this PR...

What do you think of something like this?


johnfn:
Looks good to me. Does it fix a bug? :P

xconverge:
It mitigates some other stuff that is really bad, this gives you a chance to close a tab and reopen it for a file instead of restarting vscode

PR #865, opened 2016-10-10
Remove unused modehandlers when tabs are closed

xconverge:
Done cleaning up my git...sorry for the noise in the closed PRs

This mitigates when bad things happen in a modehandler, you can close the tab and when you reopen the file you can start again...

### DIFF
--- a/extension.ts
+++ b/extension.ts
@@ -127,6 +127,11 @@ export async function activate(context: vscode.ExtensionContext) {
     }, 0);
   });
 
+  vscode.workspace.onDidCloseTextDocument((event) => {
+    // Remove modehandler for closed document
+    delete modeHandlerToEditorIdentity[event.fileName + vscode.window.activeTextEditor.viewColumn];
+  })
+
   registerCommand(context, 'type', async (args) => {
     taskQueue.enqueueTask({
       promise: async () => {


jpoon:
3rd times a charm ;)

extension.ts
```diff
@@ -127,6 +127,11 @@ export async function activate(context: vscode.ExtensionContext) {
     }, 0);
   });
 
+  vscode.workspace.onDidCloseTextDocument((event) => {
+    // Remove modehandler for closed document
+    delete modeHandlerToEditorIdentity[event.fileName + vscode.window.activeTextEditor.viewColumn];
```
jpoon: I'm not very familiar with what this does. When is an editorIdentity created with a key `event.fileName + vscode.window.activeTextEditor.viewColumn`?

xconverge: It is created in getAndUpdateModehandler based on ID.

If a more handler with that I'd already exists it will use it and if not it creates a new one. You can see this by creating a new file, entering insert, then closing the file and creating a new untitled file(must have same name ex. Untitled-1.)

This just removes a MH after a tab is closed.

This mitigates if bad bugs happen, you have a chance of closing and opening the tab instead of restarting vscode.

johnfn:
Thanks for the PR, @xconverge :)

PR #894, opened 2016-10-11
Macro

rebornix:
Yay! We love PRs! 🎊

Instead of tracking keystrokes, we track the translation of keystrokes, which are actions or document changes.

### DIFF
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -660,8 +660,13 @@ class CommandRecordMacro extends BaseCommand {
   public async exec(position: Position, vimState: VimState): Promise<VimState> {
     const register = this.keysPressed[1];
     vimState.recordedMacro = new RecordedState();
-    vimState.recordedMacro.registerName = register;
-    Register.putByKey(new RecordedState(), register);
+    vimState.recordedMacro.registerName = register.toLocaleLowerCase();
+
+    if (!/^[A-Z]+$/i.test(register) || !Register.has(register)) {
+      // If register name is upper case, it means we are appending commands to existing register instead of overriding.
+      Register.putByKey(new RecordedState(), register);
+    }
+
     vimState.isRecordingMacro = true;
     return vimState;
   }
@@ -685,7 +690,8 @@ export class CommandQuitRecordMacro extends BaseCommand {
   keys = ["q"];
 
   public async exec(position: Position, vimState: VimState): Promise<VimState> {
-    Register.putByKey(vimState.recordedMacro, vimState.recordedMacro.registerName);
+    let existingMacro = (await Register.getByKey(vimState.recordedMacro.registerName)).text as RecordedState;
+    existingMacro.actionsRun = existingMacro.actionsRun.concat(vimState.recordedMacro.actionsRun);
     vimState.isRecordingMacro = false;
     return vimState;
   }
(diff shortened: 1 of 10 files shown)


src/register/register.ts
```diff
@@ -15,7 +15,7 @@ export enum RegisterMode {
 };
 
 export interface IRegisterContent {
-  text               : string | string[];
+  text               : string | string[] | RecordedState;
```
johnfn: Shouldn't it be `RecordedState[]`?

src/register/register.ts
```diff
@@ -70,7 +77,8 @@ export class Register {
     };
   }
 
-  public static putByKey(content: string | string[], register = '"', registerMode = RegisterMode.FigureItOutFromCurrentMode): void {
+  public static putByKey(content: string | string[] | RecordedState, register = '"',
```
johnfn: We should probably do `type RegisterContent = string | string[] | RecordedState` at this point.

src/register/register.ts
rebornix: I directly append each action to `RecordedState.runActions` so we don't necessarily need an array.

src/register/register.ts
rebornix: Good point

src/actions/actions.ts
```diff
@@ -611,6 +611,14 @@ class CommandInsertRegisterContent extends BaseCommand {
 
     if (register.text instanceof Array) {
        text = (register.text as string []).join("\n");
+    } else if (register.text instanceof RecordedState) {
+      vimState.recordedState.transformations.push({
```
johnfn: Can we just immediately produce the stringified macro right here instead of shipping it off to some text transformation thing? The transformations are supposed to only be when you're actually changing the content in the text editor.

johnfn: If we're actually performing a macro at this point, then we absolutely should use a transformation.

But it looked to me like we were just storing the content of the macro in a register, which doesn't require any text transformations.

rebornix: hmn, I agree. In this particular case, we just need to insert all the keystrokes.

rebornix: @johnfn just played with Vim again and found that running `Ctrl-R {register}` is actually performing the macro stored in `{register}`.

For example, you recored `ab<BS>c<Esc>` in register `a`, which represents `Enter Insert Mode, insert b, backspace, insert c`. Repeating this macro in Normal mode will just insert `c` and return back to Normal Mode. If you are in Insert Mode and run `<C-r> a`, what gets inserted is `ac`. The first `a` is translated to `Insert a as it's Insert Mode` and then perform all following keystrokes.

rebornix:
Change the code a little bit and fix several corner cases. @johnfn @jpoon @xconverge I think it's good to review then have it merged as I'm not a frequent Macro user so I'd like to have it released first and test in prod :)

xconverge:
This is awesome!

test/testUtils.ts
```diff
@@ -84,4 +84,8 @@ export async function cleanUpWorkspace(): Promise<any> {
 export function setTextEditorOptions(tabSize: number, insertSpaces: boolean): void {
   Configuration.getInstance().tabstop = tabSize;
   Configuration.getInstance().expandtab = insertSpaces;
+  let options = vscode.window.activeTextEditor.options;
+  options.tabSize = tabSize;
```
rebornix: This should be moved once we can really update Code's configuration through API, which relies on #913. So for now we still have to update the cache option in active text editor.

PR #913, opened 2016-10-13
[Post 1.0] Two way syncing of Vim and Code's configuration

rebornix:
Yay! We love PRs! 🎊

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
         "url": "https://github.com/VSCodeVim/Vim/issues"
     },
     "engines": {
-        "vscode": "^1.5.0"
+        "vscode": "^1.6.0"
     },
     "categories": [
         "Other"
--- a/src/configuration/configuration.ts
+++ b/src/configuration/configuration.ts
@@ -110,7 +110,7 @@ function overlapSetting(args: {codeName: string, default: OptionValue, codeValue
               codeValue = args.codeValueMapping[value];
             }
 
-            await vscode.workspace.getConfiguration("editor").update(propertyKey, codeValue, true);
+            await vscode.workspace.getConfiguration("editor").update(args.codeName, codeValue, true);
           },
           isRunning: false
         });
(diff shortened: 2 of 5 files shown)


rebornix:
@johnfn @xconverge 

With this PR, you should be able to manage configs through EX command and if it has overlap with Code, Code's configuration will be override and the new config will be stored in user's `setting.json` file.

Besides, it requires Code 1.6. Please let me know when do you think is the best time to merge it.

johnfn:
So this will actually directly update the user's settings.json if they do something like `:set blah=true`? That's really cool! Maybe we should prompt the user first (they might not expect that). 

I'm otherwise fine to merge this! You guys are the ones with stats about 1.6 percentage usage after all. 😉

rebornix:
@johnfn @xconverge @jpoon 

We've released 1.6 for a week and in these 7 days, 75% of the extension activation are from Code 1.6+, 15% are from 1.5.3. More detailed info are like below

image

In latest 3 days, usage of 1.6+ increased to 80% but we still have 10% from 1.5.3.

image

I'll keep an eye on the data in the following week but since our 0.3.x breaks quite often and there are still part of users are using Code 1.5.3 or even below, my personal suggestion about our Code engine upgrade is like

* Have selection related bug fixed merged and have a new release. Target 1.5
* Have Macro merged and have a RC release. Still target 1.5
* Wait for issues and feedback about selection or macro or anything else. Fix them and once it's stable, have our RTM release: `1.0`
* Have this PR merged and target 1.6+ for all following releases.

jpoon:
It looks like we'll be leaving about 15% of users behind. Seems like a large percentage, but in the 4-day span between the 2 graphs, 1/4 of those people upgraded (it went from 20.2% to 14.8%)!

FWIW, I'm fine with leaving these people behind, but that release plan sounds like the "safer" thing to do :)

johnfn:
Easy suggestion: Show some sort of error dialog if the user is not on 1.6.0. Then we can merge this whenever.

Issue #901, opened 2016-10-12
Visual selection does not work with the "go to matching brace" % command.


warrenhodg:
### What did you do?

Shift-V, $, %

### What did you expect to happen?

Start selecting code, go to line end where the opening brace was, then select everything from where I started up until the closing brace.

### What happened instead?

Visual mode was cancelled, and the cursor navigated to the closing brace.

### Technical details:

* VSCode Version: 1.6.0
* VsCodeVim Version: 0.3.6
* OS: MacOS Sierra

johnfn:
This works for me on master. New version coming soon. :)

vedran:
I believe that I am able to reproduce this issue on 0.5.2. Is it expected to be fixed now?

![image](https://d17oy1vhnax1f7.cloudfront.net/items/3L0C0R2J2Q1d0a3Y2l1i/Screen%20Recording%202017-01-31%20at%2008.28%20PM.gif?v=0274b9b9)

Steps:

1) Start at the beginning of the line
2) Press v
3) Press $
4) Press %

**What is expected**

Select the entire function, to the matching curly brace, in visual select mode.

**What happens**

The visual mode goes into a bit of a weird state, as if it restarted from the end of the line. The cursor does not move. Notice on my second attempt in the gif when I move up and down, it's as if the visual block started from the end of the line.

**Workaround**

Press h to go left one position _after_ pressing $, but before pressing %. For example:

![image](https://d17oy1vhnax1f7.cloudfront.net/items/030f1N030t1n1w2l2Z17/Screen%20Recording%202017-01-31%20at%2008.39%20PM.gif?v=24bfbd17)

VSCode Version: 1.8.1
VsCodeVim Version: 0.5.2
OS: MacOS Sierra


Thanks!

Issue #896, opened 2016-10-12
When you click on a panel you're not focused on, the cursor position goes to the last cursor position rather than the location of the click.


johnfn:
Please *thumbs-up* 👍 this issue if it personally affects you! You can do this by clicking on the emoji-face on the top right of this post. Issues with more thumbs-up will be prioritized.

-----

xconverge:
This was closed because it is a dupe of #794 not because it was fixed, correct?

johnfn:
@xconverge It certainly appears to be fixed for me..?

xconverge:
Working for me now...I haven't changed anything...probably a race condition...

johnfn:
Amusingly, it just stopped working for me. 

Maybe I did some sort of condition in the code where it could only work for one contributor to VSCodeVim at a time?

steinso:
I still have the same issue. For me it works correctly if it is the same file open in both panes, but not if it is two different files.

Issue #895, opened 2016-10-12
Cursor behavior is flakey with same file open in first and second panel


johnfn:
Please *thumbs-up* 👍 this issue if it personally affects you! You can do this by clicking on the emoji-face on the top right of this post. Issues with more thumbs-up will be prioritized.

-----

tvhung83:
The issue remains on v0.3.7.

johnfn:
Ah, you're right... bummer

rebornix:
I'll say it's gone in 0.3.8.

Issue #892, opened 2016-10-11
Indenting in visual line mode then dot


mreishus:
------

Please *thumbs-up* 👍 this issue if it personally affects you! You can do this by clicking on the emoji-face on the top right of this post. Issues with more thumbs-up will be prioritized.

-----

### What did you do?

Scenario 1: `Vjjj>...`

### What did you expect to happen?

Select 4 lines.  `>` should indents those 4 line one extra tabstop and take me out of visual mode.
Each press of `.` should indent those lines one more tabstop.

### What happened instead?

Pressing `>` indented the lines, but left me in visual mode. (It moved me from visual line mode to visual mode).  My selection is graphically hidden, but it's 'still there'.   Pressing `.` did nothing instead of indenting again.

-----

### What did you do?

Scenario 2: `Vjjj>j`

### What did you expect to happen?

Select 4 lines.  `>` should indent those 4 lines one extra tabstop and take me out of visual mode.
`j` should move me down to the second of those 4 lines.

### What happened instead?

`>` indented, but it did not take me out of visual mode.  (It moved me from visual line mode to visual mode).   `j` expanded my selection to 5 lines. (With a graphical error in dark theme, one character has a light background)

### Technical details:

* VSCode Version: 1.6.0
* VsCodeVim Version: 0.3.5
* OS:  Windows 7

Both of these scenarios should work with `v` as well as `V` - just tested in vim.

johnfn:
This has been fixed!

PR #919, opened 2016-10-14
update internal cursor position per Code selection change

rebornix:
Yay! We love PRs! 🎊

### DIFF
--- a/src/mode/modeHandler.ts
+++ b/src/mode/modeHandler.ts
@@ -496,6 +496,18 @@ export class ModeHandler implements vscode.Disposable {
       return;
     }
 
+    if (!e.kind || e.kind === vscode.TextEditorSelectionChangeKind.Command) {
+      // Whether e.kind is undefined or Command, it means the selection change is from Code.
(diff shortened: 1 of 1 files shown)


xconverge:
Does this fix a problem and then create all of the original problems that we closed before we made this change? It makes sense to me, but I imagine some of the issues closed from the other day will reopen.

rebornix:
let me give them a quick check.

rebornix:
Verified  #872 #877 #882 #892 #895 #896, they are still good.

Verified #856, which we claimed it was fixed while actually it was still half broken without this PR.

#901 I'm not sure if I'm doing the right keystrokes, on both Master and my PR, it still doesn't work.

xconverge:
LGTM then, thanks for checking those

What portion of the code was causing #882 then...

johnfn:
@xconverge I _think_ the problem was that it would always go into that complex selection logic and eventually trigger `updateView`. Now that happens much more rarely. 

So I think we're good here.

PR #944, opened 2016-10-19
New increment without separators

xconverge:
Yay! We love PRs! 🎊

@sectioneight @johnfn took another stab at this one, last attempt was https://github.com/VSCodeVim/Vim/pull/661

### DIFF
--- a/src/number/numericString.ts
+++ b/src/number/numericString.ts
@@ -2,13 +2,13 @@ export class NumericString {
   radix: number;
   value: number;
   prefix: string;
-  postfix: string;
+  suffix: string;
 
   private static matchings: { regex: RegExp, base: number, prefix: string }[] = [
     { regex: /^([-+])?0([0-7]+)$/, base: 8, prefix: "0" },
     { regex: /^([-+])?(\d+)$/, base: 10, prefix: "" },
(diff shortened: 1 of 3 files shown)


src/number/numericString.ts
```diff
@@ -15,18 +17,53 @@ export class NumericString {
...
+
+      // Regex to find any trailing characters after the number
+      let findSuffix = /[^\d]*$/i;
+
+      let newPrefix = prefix;
+      let newSuffix = "";
+      let newNum = input;
+
+      // Only use this section if this is a number surrounded by letters
+      if (findNondigits.exec(input) !== null && NumericString.matchings[NumericString.matchings.length - 1].regex === regex) {
```
xconverge: Would like feedback on how I can get an ID to only run this for 1 of the 4 regex situations

johnfn: What do you mean? I don't understand.

xconverge: Right here I am only doing this logic if it is the last regex in the list (none of the others matched, so the last one is a catchall for anything that contains any number) It seems like a hack to depend on the last regex in the list to be this catchall...

xconverge:
This should be good to go unless someone has some more feedback, I tested it a bit more and am happy enough with it

rebornix:
LGTM.

PR #959, opened 2016-10-20
Delegate everything back to Code if it's Insert Mode and Multi Cursor.

### LINKED ISSUE #843: Characters inserted in wrong section when using Snippets

--------
### What did you do?

Set-up: empty file, insert mode, trigger with multiple sections - example below
1. Type a snippet trigger (in my case: `rnc`), and press tab.
2. Type some characters for the first section of the snippet.
3. Press `Enter` OR `Tab` to select the next snippet section.
4. Type anything - the first character of what you typed will be appended to all the instances of the previous section.

GIF of problem: 
oct-06-2016 09-59-17

The characters I typed the second section would only occur in the second section.
### What happened instead?

The first character I typed in the second section is added to the first section, and is not typed in the second section
### Technical details:
- VSCode Version: 1.5.3
- VsCodeVim Version: 0.3.0
- OS: OSX 10.11.5
### Other details

*_Snippet used: *_

``` json
    "reactNativeComponent": {
        "prefix": "rnc",
        "body": [
            "import * as React from 'react';",
            "import { StyleSheet, Text, View } from 'react-native';",
            "",
            "interface I${1:componentName}Props {}",
            "",
            "class ${1:componentName} extends React.Component<I${1:componentName}Props, void> {",
            "\tpublic render() {",
            "\t\treturn (",
            "\t\t\t<View style={styles.container}>",
            "\t\t\t\t<Text>$0</Text>",
            "\t\t\t</View>",
            "\t\t);",
            "\t}",
            "}",
            "",
            "const styles = StyleSheet.create({",
            "\tcontainer: {} as React.ViewStyle,",
            "});",
            "",
            "export default ${1:componentName};"
        ],
        "description": "Creates a React Native component with styles and props"
    },
```

rebornix:
Yay! We love PRs! 🎊

Fix #843.

### DIFF
--- a/extension.ts
+++ b/extension.ts
@@ -164,6 +164,11 @@ export async function activate(context: vscode.ExtensionContext) {
       promise: async () => {
         const mh = await getAndUpdateModeHandler();
 
+        if (mh.vimState.isMultiCursor && mh.vimState.currentMode === ModeName.Insert) {
+          await vscode.commands.executeCommand('default:type', args);
+          return;
+        }
+
         if (compositionState.isInComposition) {
           compositionState.composingText += args.text;
         } else {
@@ -178,6 +183,11 @@ export async function activate(context: vscode.ExtensionContext) {
     taskQueue.enqueueTask({
       promise: async () => {
         const mh = await getAndUpdateModeHandler();
+
+        if (mh.vimState.isMultiCursor && mh.vimState.currentMode === ModeName.Insert) {
+          await vscode.commands.executeCommand('default:replacePreviousChar', args);
+          return;
+        }
 
         if (compositionState.isInComposition) {
           compositionState.composingText = compositionState.composingText.substr(0, compositionState.composingText.length - args.replaceCharCnt) + args.text;
--- a/package.json
+++ b/package.json
@@ -73,7 +73,7 @@
             {
                 "key": "Backspace",
                 "command": "extension.vim_backspace",
-                "when": "editorTextFocus && !inDebugRepl"
+                "when": "editorTextFocus && !inDebugRepl && !vim.multiCursorSlient"
             },
             {
                 "key": "Delete",
--- a/src/mode/modeHandler.ts
+++ b/src/mode/modeHandler.ts
@@ -1506,6 +1506,8 @@ export class ModeHandler implements vscode.Disposable {
     vscode.commands.executeCommand('setContext', 'vim.mode', this.currentMode.text);
     vscode.commands.executeCommand('setContext', 'vim.useCtrlKeys', Configuration.getInstance().useCtrlKeys);
     vscode.commands.executeCommand('setContext', 'vim.platform', process.platform);
+    vscode.commands.executeCommand('setContext', 'vim.multiCursorSlient',
+      this.vimState.isMultiCursor && this.vimState.currentMode === ModeName.Insert);
   }
 
   async handleMultipleKeyEvents(keys: string[]): Promise<void> {


rebornix:
This PR changes the behavior of multi cursor so it's quite dangerous, look forward to @johnfn 's feedback.

johnfn:
What about insert mode actions like `<C-w>`?

johnfn:
Other than that though I think this is fine.

rebornix:
@johnfn good catch, special keys in Insert Mode break as they are still using the old position, sharing the same root cause I put here https://github.com/VSCodeVim/Vim/issues/843#issuecomment-255260044 .

This makes life difficult. If we sync cursor position, the behavior is wrong if you run `tab` multiple times. If we don't, special keys in Insert Mode in MultiCusor Mode is still broken.

I spent at least one hour with following double multi cursor check in selectionChangeHandler 

```
    if (this.vimState.isMultiCursor) {
      // AAAAAARGGHHHHH - johnfn

      return;
    }

...
...

    if (this.vimState.isMultiCursor) {
      return;
    }
```

but with no luck. This is the best solution in my mind with my limited knowledge of current multi cursor stuff.

johnfn:
Can we do the default:type stuff, but then also sync up the cursor positions after it finishes?

rebornix:
@johnfn , let say we have two sections to fill in as below

```
var ${0} = 1;
var ${1} = 2;
```

Once the code snippet is inserted into the document, the cursor is at `${0}`, after we type`for`, we press `tab` then the cursor is at `${1}` and you type `bar`. Then if we type `tab` again, the cursor is moved back to `foo` and `foo` itself is selected.

Our internal cursor position breaks the last behavior somehow.

johnfn:
Yeah, I am coming to understand this is a rather complicated problem.

maxcan:
Is there any chance of this getting merged? Right now vscodevim and snippets are basically unusable together

johnfn:
@maxcan I believe this PR as it stands will not be able to be merged. The idea is probably correct but the amount of revision needing to be done is rather large. :/

maxcan:
@johnfn sounds reasonable.  Is there an easy way to temporarily disable the plugin for a certain buffer or pain besides wholesale disabling the plugin using the extensions sidebar?  if there was a command like that, it would make my current work around: disable, reload window, use snippets, re-enable, reload window a lot more convenient.

vavans:
+1 we can't use snippets or multi cursor with vscodevim. I think it is one of the most important issue for now.

johnfn:
Yeah, I agree. When I find the time, snippets are first on my list to fix.

poiati:
Any progession here? We are unable to use any multicursor snippet. Thanks!

Chillee:
@poiati Is that a recent regression or has it always been like that? The recent VSCode update introduced some kind of change that's caused a bunch of issues. See here: 

https://github.com/Microsoft/vscode/issues/28853

poiati:
@Chillee I'm not sure, noticed it today. My VSCode version is 1.13.0. The issue you linked isn't associated with a version, is the 1.13 affected? Tks!

Chillee:
@poiati It's only present in version 1.13, I'm pretty sure.

Chillee:
@poiati Ok I'm sorry that's not completely true. It's not as broken as it can be now, but there's still some bugs related to backspace.

Chillee:
I'm going to close this PR as there's been no progress and we aren't sure this is the right way to fix the relevant issues.

PR #962, opened 2016-10-21
remove leading spaces when <Esc>... is pressed #685

### LINKED ISSUE #685: In insert mode, should remove leading spaces if <ESC> or <Ctrl-c> is pressed
### What did you do?

For this file, in insert mode:
```js
function foo() {
  var bar = 1;
  |
}
```

If I press `<ESC>` or `<Ctrl-c>`, it should be:
```js
function foo() {
  var bar = 1;
|
}
```

### What happened instead?

```js
function foo() {
  var bar = 1;
 |
}
```

It's in normal mode. The third line has two spaces. Cursor is on second space.

### Technical details:

* VSCode Version: 1.4.0 
* VsCodeVim Version: 0.1.7
* OS: OS X 10.11.6

Zzzen:
fix #685 

I found it annoying that <Esc> doesn't remove leading spaces in insert mode. Today, I'm going to fix it myself, contributing to this aaaaaawesome project. You guys have done a great job. 👍

### DIFF
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -874,6 +874,21 @@ class CommandEscInsertMode extends BaseCommand {
 
   public async exec(position: Position, vimState: VimState): Promise<VimState> {
     vimState.cursorPosition = position.getLeft();
+
+    // only remove leading spaces inserted by vscode.
+    // vscode only inserts them when user enter a new line,
+    // ie, o/O in Normal mode or \n in Insert mode.
+    const lastActionBeforeEsc = vimState.currentFullAction[vimState.currentFullAction.length - 2];
+    if (['o', 'O', '\n'].indexOf(lastActionBeforeEsc) > -1 &&
+        vscode.window.activeTextEditor.document.languageId !== 'plaintext' &&
+        /^\s+$/.test(TextEditor.getLineAt(position).text)) {
+      vimState.recordedState.transformations.push({
+        type: "deleteRange",
+        range: new Range(position.getLineBegin(), position.getLineEnd())
+      });
+      vimState.cursorPosition = position.getLineBegin();
+    }
+
     vimState.currentMode = ModeName.Normal;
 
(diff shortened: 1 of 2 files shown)


xconverge:
This is not the case for all filetypes, if a file is not named, it does not do this. Not a big deal though...

rebornix:
From my observation, the leading white spaces are only deleted if they are added by Vim's autoindent feature. It means if the indentation is added by users and then you press `<ESC>`, they won't be deleted. That explains why you can't experience this feature in an unnamed file as Vim has no idea about indentation.

So right now, the question is how we detect whether the white spaces are added by users or not? And how we avoid deleting users' input.

rebornix:
I think that's maybe why I say there is no easy issue any more ;(

Zzzen:
Yeah, this is a hard one. It should be easy to get current file type from VSCode. But how to avoid deleting spaces input by users is difficult. 😥 where do we store user keyboard input? And I noticed that if you create newlines by pressing Enter and then press Esc, vim will clear current line and lines above (Though VSCode has remove the trailing spaces, nice)

Zzzen:
It seems that removing leading spaces is triggered only if the last action is inserting new line, ie, pressing o in normal mode, press Enter in insert mode.

Zzzen:
How about this? Only when the last action before `<Esc>` is `"o" || "O" || "\n"` and language identifier is not `"plaintext"` will we clear current line. Your guidance will be helpful. @rebornix

rebornix:
@Zzzen I love your trick here, even though it might not be the same implementation as Vim. To avoid confusing other maintainers, do you mind adding comments before the change then we should be good to go.

Zzzen:
That should look better now. One problem remains that I cannot create a test case it takes effect since filename is created randomly and vscode regards them as plaintext. It seems that currently we are not able to switch language mode programmatically.

rebornix:
@Zzzen LGTM thanks!

PR #918, opened 2016-10-13
Numbered register reloaded

Platzer:
Implements basic functionality for numbered registers (#500).
- `0` for last yanked text
- `1` to `9` delete history

Issue #500, opened 2016-07-24
Support numbered registers


cpetrov:
One common use case is to yank a text [1], delete something and paste the yanked text. Currently the yanked text (1) will be overwritten by the deleted text in the unnamed register. Last yank should be stored in the numbered register `0`.

Another useful feature is accessing and cycling through last deleted/changed text by using registers >=1. Using e.g. `u.` works for cycling through these registers because of a special behavior of the repeat command, increasing the used register on each repeat. 

[Vim documentation: Numbered registers "0 to "9](http://vimdoc.sourceforge.net/htmldoc/change.html#quote1)
[Vim documentation: redo register](http://vimdoc.sourceforge.net/htmldoc/undo.html#redo-register)

rebornix:
It's a quite valid feature request, thanks for reporting it @cpetrov

johnfn:
Absolutely a valid request. Putting it on the roadmap for 0.3 because I feel like it's one of the lesser-used features of Vim. If more people thumbs it up then I'll bump it in priority (and you can always feel free to take a whack at it in the mean time!)

philipmat:
Registers are actually one of the *most* used features in vim, both in user operation (yanking to named register is a way of handling multiple clipboards, macro recordings save content into registers, etc) and in vim programming, though you're probably not going to implement VimL :)

johnfn:
@philipmat I don't doubt that they get a good amount of use - the problem is that *so many* features get a good amount of use. 

But hey, this is a democracy - if this gets more thumbs up, we'll definitely get to it sooner. :)

PR #974, opened 2016-10-24
Numbered, upper case and multicursor register

Platzer:
Implements basic functionality for numbered registers (#500)
- `0` for last yanked text
- `1` to `9` delete history

Implements appending to register with uppercase letter
- for `single` to `single` cursor register
- for `single` to `multi` cursor register
- for `multi` to `multi` cursor register (if line count match)

gif

Closes PR #918 because i was to stupid to rebase that branch...

### DIFF
--- a/src/register/register.ts
+++ b/src/register/register.ts
@@ -35,6 +35,7 @@ export class Register {
    */
   private static registers: { [key: string]: IRegisterContent } = {
     '"': { text: "", registerMode: RegisterMode.CharacterWise, isClipboardRegister: false },
+    '.': { text: "", registerMode: RegisterMode.CharacterWise, isClipboardRegister: false },
     '*': { text: "", registerMode: RegisterMode.CharacterWise, isClipboardRegister: true },
     '+': { text: "", registerMode: RegisterMode.CharacterWise, isClipboardRegister: true },
     '0': { text: "", registerMode: RegisterMode.CharacterWise, isClipboardRegister: false },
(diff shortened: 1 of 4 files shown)


Platzer:
@jpoon as you reviewed #918, looking forward the get your remarks on that one.

src/register/register.ts
```diff
@@ -50,33 +62,172 @@ export class Register {
...
+      } else {
+        Register.putNormalRegister(content, register, vimState);
+      }
+    }
+  }
+
+  /**
+   * Puts the content at the specified index of the multicursor Register.
+   *
+   * `REMARKS:` This Procedure asume that you pass an valid register.
```
jpoon: > asume

spelling

jpoon: Same spelling mistake in the other comments as well.

src/register/register.ts
```diff
@@ -50,33 +62,172 @@ export class Register {
...
+   * `REMARKS:` This Procedure asume that you pass an valid uppercase register.
+   */
+  private static appendNormalRegister(content: RegisterContent, register: string, vimState: VimState): void {
+    let appendToRegister = Register.registers[register.toLowerCase()];
+    let currentRegisterMode = vimState.effectiveRegisterMode();
+
+    // Check if appending to a multicursor register or normal
+    if (typeof appendToRegister.text === "object") {
+      if (appendToRegister.registerMode === RegisterMode.CharacterWise && currentRegisterMode === RegisterMode.CharacterWise) {
+        (appendToRegister.text as string[]).forEach(element => {
```
jpoon: Why not move this outside the `if` block as it's the same as line 218.

src/register/register.ts
```diff
@@ -146,10 +316,49 @@ export class Register {
         })
       );
 
-      Register.registers[register].text = text;
-    }
+      // Harmonize newline character
+      text = text.replace(/\r\n/g, '\n');
```
jpoon: Can we read `files.eol` to use the proper line ending of the file? https://code.visualstudio.com/Docs/customization/userandworkspace

Platzer: These lines only apply when we read from the system clipboard. The problem is vscode automaticaly transforms `\n` to `\r\n` when your EOL is set to `CRLF`, so when we're on a windows machine the EOL is `\r\n` if i would paste that vscode transforms it to `\r\r\n` and you've got two linebreaks in the editor. 
My first idea was to get the current EOL setting from the active editor but i didn't found the vscode API, you can set it but i don't found where i can read the setting??? 
As i can't go that way, my next idea was check the OS if we're on windows replace the newline char with the vscode-vim internal newline ('\n' we use it everywhere), but then i thought safe me the check and replace it everytime because on windows it fixes the double newline and on unix systems it causes no harm because there is no `\r\n` in the clipboard.

What do you think, is it better to add the OS check so it is more obvious that it should only apply to windows machines?

jpoon: The following should work:

`let eol = vscode.workspace.getConfiguration('files').get("eol", "\n");`

and then you can add this as a new member to `configuration.ts` to expose it to your current functionality.

Platzer: So when i think about it the "The default end of line character" (comment for the setting) is not the problem, i would need the "current system clipboard end of line character" because i had to romve the `\r` only on windows because vscode adds it when `\n` is inserted. And in case of unix system replacing `\n` with `\n` is unnecessary.

jpoon: Ah, very interesting behaviour. I repro that on my mac and I see what you mean, in that case using `\n` is the way to go.

jpoon:
Thanks @Platzer!

PR #988, opened 2016-10-25
Implemented initial EasyMotion motion

Metamist:
Yay! We love PRs! 🎊

### DIFF
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -5198,81 +5198,116 @@ if (Configuration.getInstance().easymotion) {
 
     public async exec(position: Position, vimState: VimState): Promise<VimState> {
       const searchChar = this.keysPressed[3];
-      var searchState: SearchState;
-      if (searchChar === " ") {
-        searchState = new SearchState(SearchDirection.Forward, vimState.cursorPosition, " {1,}", { isRegex: true });
-      }
-      else {
-        searchState = new SearchState(SearchDirection.Forward, vimState.cursorPosition, searchChar);
-      }
-
-      var ranges = searchState.matchRanges;
-      if (ranges.length === 0) {
-        return vimState;
-      }
-
-      var filtered: EasyMotion.Match[] = [];
-
-      var totalLineLength = 0;
-      var prevLine = -1;
-      var cursorIndex = 0;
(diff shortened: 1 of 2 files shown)


Metamist:
SearchState has a limit at 1000 total matches.
Should implement my own, range based SearchState, that only searches near the cursor instead of the whole file.

johnfn:
In fact, if you want to improve SearchState to do the 1000 matches nearest the cursor as well, that would be really useful.

johnfn:
This is really awesome work, though! Is it done and ready for feedback?

src/actions/actions.ts
```diff
@@ -5187,3 +5188,145 @@ class ActionOverrideCmdAltUp extends BaseCommand {
...
+        if (Math.abs(pos.line - position.line) > 100) {
+          continue;
+        }
+
+        if (prevLine === -1) {
+          prevLine = pos.line;
+        }
+        if (pos.line !== prevLine) {
+          const line = TextEditor.getLineAt(new Position(prevLine, 0)).text;
+          totalLineLength += line.length;
```
xconverge: No need to declare a var here if it is just to append length

src/actions/actions.ts
```diff
@@ -5187,3 +5188,145 @@ class ActionOverrideCmdAltUp extends BaseCommand {
...
+      }
+
+      filtered.sort((a: EasyMotion.Match, b: EasyMotion.Match): number => {
+        var diffA = cursorIndex - a.index;
+        var diffB = cursorIndex - b.index;
+
+        var absDiffA = Math.abs(diffA);
+        var absDiffB = Math.abs(diffB);
+
+        if (a.index > cursorIndex) {
```
xconverge: Can you comment all of your edge and corner cases? I don't know WHY a lot of these checks are here

Metamist: I’m going to improve and simplify the sorting since it’s just unnecessarily complex right now, but I’ll make sure I comment it.

Metamist:
SearchState runs its search function before isRegex is able to be defined. My last commit fixes this, but could this affect other actions that depend on the current isRegex: true, which is broken? Namely `CommandSearchBackwards` and `CommandSearchForwards`?

PR #989, opened 2016-10-26
Fix for visual line behaving funky when going from bottom up

xconverge:
Here you go @rebornix 

Sorry about that!


rebornix:
neat!

PR #993, opened 2016-10-26
Implemented EasyMotion plugin functionality

Metamist:
Yay! We love PRs! 🎊

This implements the common motions found in the [EasyMotion vim plugin](https://github.com/easymotion/vim-easymotion), which is s, w, e, b. More to come later.
Now also supports the motions t, T, f, F.


src/actions/actions.ts
```diff
@@ -5287,3 +5288,287 @@ class ActionOverrideCmdAltUp extends BaseCommand {
     return vimState;
   }
 }
+
+
+// Only register the EasyMotion actions if the configuration is set
+if (Configuration.getInstance().easymotion) {
+  class BaseEasyMotionCommand extends BaseCommand {
```
johnfn: Could this class be made abstract?

src/easymotion/easymotion.ts
```diff
@@ -0,0 +1,408 @@
...
+      }
+    }
+
+    return markers;
+  }
+
+  /**
+   * Search and sort using the index of a match compared to the index of position (usually the cursor)
+   */
+  public sortedSearch(position: Position, searchString = "", options: EasyMotion.SearchOptions = {}): EasyMotion.Match[] {
```
johnfn: Instead of passing in `isRegex: true`, simply make the type of searchString `string | RegExp`. Then you can do `if (typeof searchString === "string")` and TypeScript is intelligent enough to infer that within the curly braces of that if, `searchString` is now a string (similarly with the RegExp).

src/actions/actions.ts
```diff
@@ -5287,3 +5288,287 @@ class ActionOverrideCmdAltUp extends BaseCommand {
...
+      // Search for the beginning of all words before the cursor
+      return vimState.easyMotion.sortedSearch(position, "\\w{1,}", {
+        isRegex: true,
+        max: position,
+      });
+    }
+  }
+
+  @RegisterAction
+  class CommandEscEasyMotionMode extends BaseCommand {
```
johnfn: Can you consolidate this with our normal CommandEsc stuff? Just be like `if (vimState.currentMode === ModeName.EasyMotionMode)` and then do the custom handling you have here.

src/actions/actions.ts
```diff
@@ -5287,3 +5288,287 @@ class ActionOverrideCmdAltUp extends BaseCommand {
...
+    modes = [ModeName.EasyMotionMode];
+    keys = ["<character>"];
+
+    public async execAction(position: Position, vimState: VimState): Promise<Position> {
+      var key = this.keysPressed[0];
+      if (!key) {
+        return position;
+      }
+
+      // "nail" refers to the accumulated depth keys
```
johnfn: Why is it called `nail`? :P

src/easymotion/easymotion.ts
```diff
@@ -0,0 +1,408 @@
...
+    this._vimState.currentMode = ModeName.EasyMotionMode;
+  }
+
+  /**
+   * Exit EasyMotion mode and clean up
+   */
+  public exitMode() {
+    this._vimState.currentMode = ModeName.Normal;
+
+    this.accumulation = "";
```
johnfn: Instead of doing this (which requires you to remember to add new properties here as well), you can just do something like `vimState.easyMotionObject = new EasyMotion()`

johnfn:
Pretty amazing stuff. It's awesome that you did this! 

BEST VIM PLUGIN EVER!!!

src/actions/actions.ts
Metamist: I honestly have no idea, I couldn't find a decent word to describe what it does, which is why I am referring to it as "accumulated depth keys". Got a better name? :)

src/actions/actions.ts
xconverge: I actually disagree

Originally I was going to make the same comment, however I then thought this was more modular if/when he moves it to its own file

Why do we have a Replace vs VisualReplace sort of thing, behavior is different so its a different action

src/easymotion/easymotion.ts
Metamist: Vimstate already has `vimstate.easymotion = new EasyMotion(vimstate)`

src/easymotion/easymotion.ts
Metamist: Good idea, I was following the StateSearch method. This suggestion applies to StateSearch as well!

src/actions/actions.ts
```diff
@@ -5287,3 +5288,287 @@ class ActionOverrideCmdAltUp extends BaseCommand {
...
+  class BaseEasyMotionCommand extends BaseCommand {
+    public getMatches(position: Position, vimState: VimState): EasyMotion.Match[] {
+      throw new Error("Not implemented!");
+    }
+
+    public getMatchPosition(match: EasyMotion.Match, position: Position, vimState: VimState): Position {
+      return match.position;
+    }
+
+    public processMarkers(matches: EasyMotion.Match[], position: Position, vimState: VimState) {
```
Metamist: You're right about it belonging in `EasyMotion` object, but then I would have to pass in the `getMatchPosition` as a callback in order to map the matches differently on each action (e, T, t have different match mappings)

kidproquo:
Awesome! This is a really cool plugin. Thanks for doing this!

src/easymotion/easymotion.ts
Metamist: Recreating the whole object instead of individual variables just puts unnecessary load on the garbage collector. Any particular reason you want it this way instead?

johnfn: Yes. Until we have actual profiling data that show that memory consumption or gc cycles are issues, we shouldn't worry at all about performance. Code quality is much more important.

My concern with the way we're doing here is that every time you add another property to the EasyMotion object, you also have to remember to clear it out in exitMode(). This is an easy way to end up adding bugs when someone comes back to modify this class months down the line.

I follow this pattern in ModeHandler with RecordedState - look for the multiple instances of `new RecordedState()` - because I got bitten by similar bugs. :)

johnfn:
Also, I don't have any problem with defaulting easymotion to on. :)

johnfn:
Thanks for the amazing work, @Metamist!

wangsongiam:
Do we have a shortcut for this motion? I've tried 'ctrl+alt+\', but It doesn't work.

hastebrot:
@wangsongiam `backslash`, `backslash`, `s` (\\s) worked for me.

Metamist:
@wangsongiam, you can use most basic motions with EasyMotion, such as `s, e, w, b, f, F, t, F`, with `s` being the most useful one IMO. All these motions are prefixed with `\\`

wangsongiam:
@Metamist worked, thanks! but can I change the prefix `\\` to something else? also; It would be helpful if we can make other link hints style(black and red are not very recognizable for some light themes), or we can configure the style like [vimium](http://vimium.github.io/)

src/easymotion/easymotion.ts
```diff
@@ -0,0 +1,376 @@
...
+      var cache = this.svgCache[code];
+      if (cache) {
+        return cache;
+      }
+
+      const width = code.length * 8 + 1;
+      var uri = vscode.Uri.parse(
+        `data:image/svg+xml;utf8,<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 ${width} ` +
+        `13" height="14" width="${width}"><rect width="${width}" height="14" rx="2" ry="2" ` +
+        `style="fill: ${backgroundColor};"></rect><text font-family="Consolas" font-size="14px" ` +
```
dsifford: Consolas font isn't installed on my macbook either. Why not just lookup the font-family being used in the user-settings?

Metamist: @dsifford, Good point, this was the first time I had ever touched anything regarding extensions for VSCode, so I didn't know I could look it up. If someone who has more experience regarding this can fix it that would be great!

dsifford: @Metamist Not even totally sure my suggestion is possible. I'm a "conveted-atom-user-in-progress" so I have very little experience on the ins and outs of VSCode for now.

That said, an even easier solution for now would be to just give a bunch of fallback font options ending with `monospace`. 

So, for example, `font-family: Consolas, Monaco, 'Courier New', monospace;` for example might work in a wider range of machines.

PR #994, opened 2016-10-26
SearchState - Fixed isRegex not set on search

Metamist:
Yay! We love PRs! 🎊

Currently, when a SearchState object is defined, the searching begins before this.isRegex is set. This is because the setter, `searchString` runs the search function, and it is set in the constructor before isRegex is set, which means it will be undefined during the search.
This PR makes sure isRegex is set and available during the search by defining it before searchString.


rebornix:
thanks @Metamist

PR #997, opened 2016-10-27
Improved performance of visual block replace by a lot

xconverge:
This just removes iterating over every character and instead iterates over lines. Still works with blank lines and partial lines as well.


johnfn:
This looks much better to me. Thanks!

PR #1006, opened 2016-10-28
fix mouse clicking past EOL

xconverge:
I think that this is pretty good...

@rebornix @johnfn

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,6 +48,16 @@
                 "when": "editorTextFocus && !inDebugRepl"
             },
             {
+                "key": "End",
+                "command": "extension.vim_end",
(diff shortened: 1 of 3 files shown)


johnfn:
I'm concerned mostly just because this has been a persistently difficult bug to wring out. Have you tested your solution to make sure it didn't introduce any weird side effects?

xconverge:
I am concerned as well, but its driving me nuts

I know that its another patch

rebornix:
eol

Basically I like the idea of handling the redraw of selection when ppl clicking EOL, but I often had above experience with selection with this kind of change, it seems like some racing condition.

PR #1007, opened 2016-10-28
Make create-multicursor commands repeatable

Platzer:
Makes `Cmd-D`, `Cmd-Alt-Up` and `Cmd-Alt-Down` repeatable by setting the `runsOnceForEachCountPrefix` property to `true`.
So you can type `3<Cmd-Alt-Down>` and end up with 4 multi-cursor because 3 where created beneath your main cursor.


johnfn:
Very cool! Thanks, @Platzer!

PR #1008, opened 2016-10-28
fixes #983

### LINKED ISSUE #983: Cursor does not preserve width when moving to shorter lines
### What did you do?
move with j and k   to lines of different length
cursor stays at the same column position, or at max length if the line is shorter, but goes back to the previous column position when the line is long enough
### What happened instead?
cursor stays at max position
### Technical details:
this explains it better than words

* VSCode Version: 1.6.1
* VsCodeVim Version: 0.4.0
* OS: macOS

xconverge:
@rebornix @johnfn 

I git blamed this change and couldn't find a good reason for it being there


rebornix:
Does this mitigate https://github.com/VSCodeVim/Vim/issues/842 as well?

xconverge:
That's from tabs though so I don't think so

johnfn:
Looks simple enough...

Issue #727, opened 2016-09-06
cc on end of line eats currentline


octref:
### What did you do?

For this js file:

``` js
if (true) {|
}
```

Pressed:
`o<C-c>cc`
### What did you expect to happen?

``` js
if (true) {
  |
}
```
### What happened instead?

``` js
if (true) {
  |}
```

Also related to #685 , as when I press `o<C-c>`, it ends up in

``` js
if (true) {
 | // Two whitespaces here, cursor before second one
}
```

Then if I do `cc`, the newline is eaten.

`cc` for the following file works fine:

``` js
if (true) {
| // No trailing space
}
```
### Technical details:
- VSCode Version: 1.4.0
- VsCodeVim Version: master branch
- OS: OS X 10.11.6

I set `autoindent` to true.

octref:
Happend to see #696 and can confirm cw also eats EOL in latest master.

SamOfFinland:
I can take a look at this issue.

johnfn:
By all means, go ahead @SamOfFinland :)

Chillee:
This issue seems to be fixed @jpoon .

xconverge:
I can't repro this anymore, I think it may be fixed as well. Feel free to reopen!

PR #1015, opened 2016-10-31
Fix for cc to prevent deletion of empty line

SamOfFinland:
Fix for issue:
"cc on end of line eats currentline" #727

One character deletes with some actions ("cc" or "dl")  on an empty line can delete current line. Added small check into delete operator so that line won't be deleted.

### DIFF
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -1621,7 +1621,18 @@ export class DeleteOperator extends BaseOperator {
           end   = end.getLineEnd();
         }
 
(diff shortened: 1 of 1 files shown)


johnfn:
This logic should actually be inside of the `l` and `cc` actions rather than inside the DeleteOperator.

jpoon:
Spring cleaning. Closing this PR as there hasn't been movement in awhile. Feel free to re-open addressing John's comments.

PR #1029, opened 2016-11-03
Fixed "d" and "D" in multicursor mode

Platzer:
In Multi-Cursor mode "d" in Visual Mode and "D" in Normal-Mode is not working.
Added the current multicursorindex to the `Register.Put` method.

### DIFF
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -3700,7 +3700,7 @@ class ActionJoin extends BaseCommand {
 
     let resultLine = lineOne + (addSpace ? " " : "") + lineTwoTrimmedStart;
 
-    let newState = await new DeleteOperator(this.multicursorIndex).run(
+    let newState = await new DeleteOperator().run(
       vimState,
       position.getLineBegin(),
       lineTwo.length > 0 ?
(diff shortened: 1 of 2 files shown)


Platzer:
I don't get it why that one is failing with my commit:
grafik
On my local machine it is working, i don't think is related to my changes but i cannot figure out why it is not working.

xconverge:
@Platzer same problem on my PR as well

jpoon:
Can you add tests around this? Thanks.

johnfn:
Yeah, my bad for not getting to this earlier. The failing test concerns me. If you can figure that one out (and yeah add another test maybe?) this should be good to go.

johnfn:
I'm actually not sure if this was necessary; I think the bug ended up being somewhere else. Can you check what's currently on master and see if this change is required?

Platzer:
@johnfn, the deleting part is fixed in current release but putting into multi cursor register is still broken! would be fixed with this PR. I will try to have a look this week!

Platzer:
@johnfn, travis is failing cause some npm stuff, but tests are okay.

johnfn:
Sorry for taking so long on this, @Platzer. I've merged it in now.

PR #1037, opened 2016-11-04
Hide the info line in issue template

octref:
Learned this trick from TypeScript's issue template: https://github.com/Microsoft/TypeScript/issues/new

The top line won't be rendered, but will be visible when issue gets reported.


johnfn:
Oh man is this useful.

jpoon:
THanks @octref. We can probably do the same for the PR template -- a lot of the PRs just have that default message and nothing else :(

PR #1071, opened 2016-11-19
fixes #1065

### LINKED ISSUE #1065: Please support CTRL-O {command}
### What did you do?
image
I enabled `useCtrlKeys` ,Because I want to to use the `Ctrl-O` (escapes user to do one normal-mode command, and then return to the insert mode.)

Execute {command} and return to Insert mode
### What happened instead?
Just like pressed the `Enter` Key instead of what I expected.

* VSCode Version: 1.7.1
* VsCodeVim Version: 0.4.2
* OS: 10.12.1 (16B2555)

xconverge:
@johnfn Do you thank that the getter way is cleaner than this? This doesn't seem too bad to me but it is relatively specific for this command.

### DIFF
--- a/package.json
+++ b/package.json
@@ -182,6 +182,11 @@
                 "when": "editorTextFocus && vim.useCtrlKeys && !inDebugRepl"
             },
             {
+                "key": "ctrl+o",
+                "command": "extension.vim_ctrl+o",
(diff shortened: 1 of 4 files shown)


johnfn:
Yeah, this is fine. Can you add a test and then we should be good to go here.

xconverge:
will do, thanks

xconverge:
@johnfn good to go

PR #1085, opened 2016-11-23
fixes #1084

### LINKED ISSUE #1084: Bracket pairs don't get deleted when you delete the first bracket
### What did you do?
https://gyazo.com/f6e305ea0d7afce408329e6ebf72402d
This happens for character pairs like [], {}, '', etc.

Both characters in the pair to deleted.

### What happened instead?
Only one character was deleted.

### Technical details:

* VSCode Version: 1.7.2
* VsCodeVim Version: 0.4.3
* OS: Windows

xconverge:
Any thoughts on this? Seems like a positive effect?

When you enter a quote or parenthesis, vscode has an option to auto add the closing portion.

With this, when you remove the opening char, and the closing char is the immediate next character, it will remove the closing char as well.

### DIFF
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -527,6 +527,20 @@ export class CommandInsertInInsertMode extends BaseCommand {
           range: new Range(selection.start as Position, selection.end as Position),
         });
       } else {
+        // Check for matching character that was auto inserted ("" or [] and delete the closing char if the opening char is deleted)
+        const text = TextEditor.getLineAt(position).text;
+        const charToMatch = text[position.character - 1];
+        const toFind = PairMatcher.pairings[charToMatch];
+
+        if (toFind !== undefined) {
+          if (text[position.character] === toFind.match) {
+            vimState.recordedState.transformations.push({
+              type: "deleteRange",
+              range: new Range(position, new Position(position.line, position.character + 1)),
+            });
+          }
+        }
+
         vimState.recordedState.transformations.push({
           type: "deleteText",
           position: position,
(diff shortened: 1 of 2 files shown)


xconverge:
I am also considering putting an if statement around this to only do this if editor.autoClosingBrackets is set to true (which is default currently)

johnfn:
Ideally we could delegate the work of deleting back to VSCode who would do the correct thing. Who knows what other auto-pairing characters they will add? 

If there's no easy way to do that, then this solution could work.

xconverge:
Was my first thought as well, however since we had some matchers already I figured this was a possible solution. I am not sure what that call to delegate back to vscode would look like

johnfn:
I think the matcher stuff is for the da" actions and stuff like that? I actually have no idea what it would look like either lol.

xconverge:
Yea but the matchers have useful maps already of these paired characters, so as long as those stay updated this should work no problem

xconverge:
Interesting, here is a call in vscode, so you are right, really this should be delegated. There should be a way in the api to call delete 1 char left and have this invoked

public static deleteLeft(config: CursorConfiguration, model: ICursorSimpleModel, cursor: SingleCursorState): EditOperationResult {
		let r = this.autoClosingPairDelete(config, model, cursor);

rebornix:
@xconverge we have deleteLeft, deleteRight, deleteAllLeft and deleteAllRight in Code. The only catch with the first two is that they might not work perfectly in multi cursor mode (technical debt, I'm working on improving all commands that were implemented before we had multi cursor mode).

xconverge:
Yea and then there is no way for me to use them right?

johnfn:
Yah, the problem is that we need to know the position of where the cursor is going to be for multi cursor mode, and that's not provided by the moveRight/Left functions afaik.

Issue #1108, opened 2016-12-04
[cd]it inside of tag which includes another empty tag


DanEEStar:
Please *thumbs-up* 👍 this issue if it personally affects you! You can do this by clicking on the emoji-face on the top right of this post. Issues with more thumbs-up will be prioritized.

-----

### What did you do?

See this test case which works in Vim:

```
    newTest({
      title: "Can do cit inside of a tag with another non closing inside tags",
      start: ["<blink>hello<br>wo|rld</blink>"],
      keysPressed: "cit",
      end: ["<blink>|</blink>"],
      endMode: ModeName.Insert
    })
```

### Technical details:

* VSCode Version: 1.7.2
* VsCodeVim Version: current master in time of writing
* OS: Mac

I am working on a solution by the way...

johnfn:
@DanEEStar Could you open up a PR just so everyone knows what you're working on? Feel free to mark it [WIP].

johnfn:
(This one actually seems fairly complex. It looks like it involves parsing HTML. So if you want to work on something easier, feel free...)

PR #1117, opened 2016-12-05
new algorithm for [cd]it, fixes #971 and #1108

### LINKED ISSUE #971: Deleting or changing multi-line tag blocks does not work

--------
### What did you do?

Open an HTML file and move the cursor to a multi-line tag block, e.g. the `d` in this `<div>` tag:

``` html
<div>
    <p>foo bar</p>
</div>
```

Type `dit` or `cit` to delete or edit the multi-line content of the `div` tag.

The entire contents of the `div` tag should be deleted, so we're only left with a single line. The cursor moves to the opening `<` of the closing `</div>`, and for `cit`, it also enters insert mode.
### What happened instead?

Nothing
### Technical details:
- VSCode Version: 1.6.1
- VsCodeVim Version: 0.3.8
- OS: Arch Linux

DanEEStar:
The algorithm is based on the same algorithm implemented
in the [IntelliJ IdeaVim plugin](https://github.com/JetBrains/ideavim)

I had to get access to the `document` so I wrote a getter in textEditor.ts. I am not so sure about that one.

There are still some minor things, where [cd]it does not work like the original Vim.
I am planning to get to them later.

Already in this form this feature should be usable.

### DIFF
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -5673,12 +5673,11 @@ abstract class MoveTagMatch extends BaseMovement {
   protected includeTag = false;
 
   public async execAction(position: Position, vimState: VimState): Promise<IMovement> {
-    const text = TextEditor.getLineAt(position).text;
-    const tagMatcher = new TagMatcher(text, position.character);
-    const start = tagMatcher.findOpening(this.includeTag);
-    const end = tagMatcher.findClosing(this.includeTag);
+    const tagMatcher = new TagMatcher(position);
+    const startPos = tagMatcher.findOpening(this.includeTag);
+    const endPos = tagMatcher.findClosing(this.includeTag);
 
-    if (start === undefined || end === undefined) {
+    if (startPos === undefined || endPos === undefined) {
       return {
         start: position,
         stop: position,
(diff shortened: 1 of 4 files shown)


johnfn:
Thanks for the pull request, @DanEEStar!

I would like to merge this, but I have to admit that I find it very difficult to understand what the code is doing. Keeping track of all of the indexOfs, lastIndexOfs and character positions is very hard. 

I think that a conceptually easier way to attack this problem would be to tokenize the string into either text or tags, so that something like "<p><div>blah</div>blah blah<p>" would turn into `["<p>", "<div>", "blah blah", "<p>"]. Then it's easy to traverse the tokens and understand where we need to run our operators.

I would also suggest to use (more) regexes to match the tags rather than using indexOf and lastIndexOf. Regexes are nice because they give you a contextual clue to the thing that you're trying to match.

DanEEStar:
Thanks for the input, @johnfn 

I will look into that and try to refactor some things. It will try to do it over the weekend.

DanEEStar:
@johnf
The suggestion with tokenized string sounds good, but wouldn't we loose the context of the cursor position once we have the token array? I think that could become very messy.

So I tried to simplify the existing code, but honestly after thinking about some edge cases, I don't think I can simplify it easily. What happens if the cursor starts inside of the end tag, directly on the ending slash tag, etc...

Are you willing to try to understand the code as it is once again?
After debugging it some more I don't actually think it is that hard to understand.
But I leave it to your judgement.  :relaxed:

johnfn:
I understand what you're saying but I think you have a slight bias because you just wrote it. 😉  I'm trying to keep in mind people who will be trying to understand and modify our code a year down the line. :)

Yes, I didn't fully spec out the code that I had in mind. It would probably look a little more like this:

`[ { text: "<div>", start: 0 }, { text: "blah", start: 4 } ]`

etc.

Then you could use the start properties to figure out where to put the cursor.

theamazingfedex:
Any update on this?

johnfn:
The one to keep an eye on is this one: https://github.com/VSCodeVim/Vim/pull/1454

Someone (prob me) just needs to fix that final issue

PR #1126, opened 2016-12-06
Updated useCtrlKeys default value

Mxbonn:
The README stated that the useCtrlKeys default is `false` while the default in package.json defines it as `true`. This PR makes the readme consistent with the settings in package.json.


johnfn:
That's weird, I would swear I set it to false. Hm. Well, no one has complained, so let's keep it on 😉

Issue #1130, opened 2016-12-06
highlighting with * asterisk


chulander:
Please *thumbs-up* 👍 this issue if it personally affects you! You can do this by clicking on the emoji-face on the top right of this post. Issues with more thumbs-up will be prioritized.

-----

### What did you do?
pressed * on the word
Put the exact keys you pressed.
*
### What did you expect to happen?
I expected highlighting to occur on all words that match similar to use :/mysearchword
### What happened instead?
the behavior works as I can press 'n' to move to next match, however highlighting doesn't work. I tried on regular vim and it works there.

### Technical details:

* VSCode Version: Version 1.8.0-insider (1.8.0-insider)
* VsCodeVim Version: *[please ensure you are on the latest]*0.4.9
* OS: Mac Sierra

xconverge:
Did you do :nohl or something at one point by any chance? (trying to pin it down...)

xconverge:
And you don't have "vim.hlsearch": false, in your settings?

Do you EVER see any highlighting when searching? Using /shajdasjkdha and n or N?

chulander:
apologies, there is no issue and it's was my own confusion for not restarting vsCode after enabling **hlsearch** and setting the **searchHighlightColor **. I realized after my personal macbook did not have the same problem as my company-issued one that it must be me.

thanks for your help xconverge

PR #1134, opened 2016-12-07
minor fixes for # and * after using :nohl

xconverge:
Could be related to #1130, but I am unable to reproduce that specific issue yet so will not close it.

Highlighting works fine for me using * and #, this fixes highlighting ONLY in the case that :nohl had been used previously at some point.


johnfn:
Ah yeah, I was vaguely considering this a possibility.

PR #1141, opened 2016-12-09
Reselect visual implemented (gv)

xconverge:
Let me know what you think, works on visual, visual line, and visual block selections


johnfn:
Wonderful! You've single handedly knocked out two of our biggest request items in the last few days!

PR #1151, opened 2016-12-11
control key bindings respect the useCtrlKey setting

xwvvvvwx:
These keybindings are all used in the default bindings (on linux) and so should respect the `vim.useCtrlKeys` setting.

### DIFF
--- a/package.json
+++ b/package.json
@@ -119,7 +119,7 @@
       {
         "key": "ctrl+r",
         "command": "extension.vim_ctrl+r",
-        "when": "editorTextFocus && !inDebugRepl"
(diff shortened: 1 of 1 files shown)


xconverge:
Makes sense to me!

xconverge:
@johnfn a handful of users have been talking about this in slack

johnfn:
It makes me sad because of the number of people who get confused by useCtrlKeys defaulting to off. But okay. I guess there's no other option.

xwvvvvwx:
@johnfn Thanks! 🙏 

And just as a note I think it would maybe make sense to have useCtrlKeys defaulting to on

PR #1175, opened 2016-12-18
Select inner vi( fix

xconverge:
Definitely helps that I have been adding tests for each one of these that gets closed...


johnfn:
Yes, tests are good. 😁

Issue #1156, opened 2016-12-14
Repeat command (.) is scoped to a file


vikramthyagarajan:
Please *thumbs-up* 👍 this issue if it personally affects you! You can do this by clicking on the emoji-face on the top right of this post. Issues with more thumbs-up will be prioritized.

-----

### What did you do?

Put the exact keys you pressed.
Open 2 files in the same workbench.
1. Delete a line using dd
2. Run the repeat command (.) in the same file. It works
3. Go to the next file and run the repeat command (.)
4. Nothing happens

### What did you expect to happen?

The last command I performed (dd) is repeated

### What happened instead?

Nothing. The repeat command only works within the scope of the file the action was performed in.

### Technical details:

* VSCode Version: 1.7.2
* VsCodeVim Version: 0.4.9
* OS: MacOS Sierra 10.12.1

xconverge:
@johnfn can close now!

johnfn:
Thanks @xconverge 😀

PR #1179, opened 2016-12-20
Implements Global state

### LINKED ISSUE #1081: Search is unique to file

-----

### What did you do?
Have a search query persist across different files
`/myFunction`
switch files
press `n`
Search would go to next occurrence of `myFunction` in new file 
### What happened instead?
Two possibilities:
1. If no previous search in that file -> nothing
2. If previous search in that file -> Searches for that
### Technical details:

* VSCode Version: 1.7.2
* VsCodeVim Version: 0.4.3
* OS: Windows 10

vikramthyagarajan:
Holds state of repeat "." actions and search globally (across editors), and added tests for each in their respective test files.

Should fix #1081 and #1156 issues

### DIFF
--- a/src/mode/modeHandler.ts
+++ b/src/mode/modeHandler.ts
@@ -139,6 +139,22 @@ export class VimState {
     GlobalState.previousFullAction = state;
   }
 
+  public get searchState(): SearchState | undefined {
+    return GlobalState.searchState;
+  }
+
+  public set searchState(state : SearchState | undefined) {
+    GlobalState.searchState = state;
+  }
+
+  public get searchStateIndex(): number {
+    return GlobalState.searchStateIndex;
+  }
+
+  public set searchStateIndex(state : number) {
+    GlobalState.searchStateIndex = state;
+  }
+
   /**
    * The position the cursor will be when this action finishes.
    */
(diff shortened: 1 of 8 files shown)


johnfn:
Wow, very impressive work!

src/mode/modeHandler.ts
```diff
@@ -124,6 +121,41 @@ export class VimState {
   public currentFullAction: string[] = [];
 
   /**
+   * Getters and setters for changing global state
+   */
+  public get searchStatePrevious(): SearchState[]{
```
johnfn: Instead of adding a bunch of shims for global state, can we just move these directly onto the globals object and allow consumers to use these getters/setters directly?

johnfn: So instead of saying `vimState.searchState` they would say `vimState.globalState.searchState`?

johnfn:
Minor request and we should be good to go here!

src/util.ts
```diff
@@ -30,6 +30,23 @@ export async function waitForCursorUpdatesToHappen(): Promise<void> {
...
 }
 
+/**
+ * FOR TESTING PURPOSES ONLY
+ * Waits for the tabs to change after a command like 'gt' or 'gT' is run.
+ * Sometimes it is not immediate, so we must busy wait
+ * Caveat: Use only it tests. This promise does not timeout. Awaiting this promise
+ * in extension code is dangerous
+ */
+export async function waitForTabChange(): Promise<void> {
```
vikramthyagarajan: Thank you :)

src/util.ts
```diff
@@ -30,6 +30,23 @@ export async function waitForCursorUpdatesToHappen(): Promise<void> {
   });
 }
 
+/**
+ * FOR TESTING PURPOSES ONLY
```
xconverge: if you want you can add a guard to use if Globals.isTesting is set to true, not a big deal

johnfn:
Thanks again for the awesome work.

xconverge:
And tests!

Issue #1181, opened 2016-12-20
Add documentation on how to map leader bindings


eastwood:
There has been some questions recently regarding mapping the `leader` bindings. This wasn't immediately obvious to me either and required a little code inspection. Hopefully, this PR addresses this issue and gives a quick insight for future users.

xconverge:
@johnfn can close now!

PR #1182, opened 2016-12-20
Adds documentation for adding leader bindings

eastwood:
#1181 
Quick documentation addition for insight on mapping leader bindings.


johnfn:
Thanks! You're right, we could have documented this better.

PR #1190, opened 2016-12-23
Index fixes

### LINKED ISSUE #1188: Search history and change list navigation require an extra press at the first and last index
If you search something, and search something else, using ```<up>``` and ```<down>``` to navigate search history requires an extra press at each limit of the history list.

This is also true for ```g ;``` and ```g ,``` to move through changelist

This has to do with when to increment/decrement searchStateIndex and changelistIndex near the limits.

https://github.com/VSCodeVim/Vim/commit/c8ee059f5d21c280175a21383314b5b977ecf647#diff-7addcd1a8a241060a73b1fe4abb39960R1465

https://github.com/VSCodeVim/Vim/commit/dd7aa1974a48c568156707d38e6594f6c21dba29#diff-7addcd1a8a241060a73b1fe4abb39960R2815

xconverge:
A few minor fixes, fixes #1188


src/actions/actions.ts
```diff
@@ -1465,12 +1465,25 @@ class CommandInsertInSearchMode extends BaseCommand {
...
+
       if (prevSearchList[vimState.globalState.searchStateIndex] !== undefined) {
         searchState.searchString = prevSearchList[vimState.globalState.searchStateIndex].searchString;
         vimState.globalState.searchStateIndex -= 1;
       }
     } else if (key === "<down>") {
       const prevSearchList = vimState.globalState.searchStatePrevious!;
+
+      // Preincrement if on boundary to prevent seeing the same search index twice
+      if (vimState.globalState.searchStateIndex === 0) {
```
johnfn: How come we increment twice when we start at 0? I don't quite follow.

xconverge: Otherwise index is 0 and you can press up and down alternating and never see a different result

xconverge:
@johnfn this one is actually pretty solid even if it looks weird

What it fixes:
- do a few searches, then use <up> and <down> to navigate them, when you get to the last search result or the first search result, you need to press up or down TWICE to get to the next one

- <BS> on first char of first word of first line, would move it to the last char

xconverge:
@johnfn this is weird, but I would rather it be merged than closed...

johnfn:
Sorry for taking so long on this one. I had it incorrectly categorized as a WIP in my mind, so I totally forgot to come back to it. 

The reason I hesitated here is this fix doesn't really make sense. The index value should be just an index into the array that you can walk up or down by pressing up or down respectively. I don't see anything about the structure of that task that requires a double increment at the borders, which makes me cautious about this PR. I would like the code to reflect the structure of the problem we try to solve. 

That being said, I don't really understand the search history code all too well, because I didn't implement it. So if you tell me this is the best way, I'll trust your judgement.

xconverge:
There 100% has to be a better way with a handful of lines to do the same thing, this is just covering that up

let me take a peek, I will @johnfn you when its good to go

xconverge:
@johnfn I think that this reads better. Works even better since if you make it to the end of history you get "" so you can enter your own search

The key was when entering "/" or "?" or reaching the end, allowing the index to go past the end of the limit.

johnfn:
Alright, looks good to me now!

PR #1191, opened 2016-12-24
[WIP] Surround extension

xconverge:
Just opening this so people know it is being worked on...

### DIFF
--- a/package.json
+++ b/package.json
@@ -337,6 +337,11 @@
           "description": "Enable the EasyMotion plugin for Vim.",
           "default": false
         },
+        "vim.surround": {
+          "type": "boolean",
+          "description": "Enable the Surround plugin for Vim.",
+          "default": false
+        },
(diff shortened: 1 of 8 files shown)


xconverge:
I need to figure out how to nicely use our actions, the way the plugin does it is pretty clever

It does a di [char] to copy the inner value to a register, then da [char] to remove the surrounds, paste the new ones, then paste the inner value back in, before restoring the original copy register to what it was before the command was initiated

I think that this is cool and will allow me to not need to reimplement a bunch of searching etc. I will spend some time on being able to execute [a, w] etc instead of doing it all in surround.ts so I can stick with DRY

bookercodes:
I'm really excited for this 👌. Any idea when it might land?

Great work!!

xconverge:
I will reopen if I get time to continue, but atleast it is a start with some ideas for someone else if not

PR #1227, opened 2017-01-13
Support "gf" in es6 import statements by adding the file extension

aminroosta:
@johnfn In the es6 import statements, the file extension is omitted, this PR tries to add the file extension automatically.

For example when your cursor is at `"../node"` the `gf` command will open `../node.ts`

`---------current file:  src/cmd_line/commands/file.ts--------- `
`import * as node from "../node";`

### DIFF
--- a/src/cmd_line/commands/file.ts
+++ b/src/cmd_line/commands/file.ts
@@ -2,6 +2,7 @@
 
 import * as vscode from "vscode";
 import * as path from "path";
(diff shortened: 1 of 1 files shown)


aminroosta:
@johnfn One test is failing, but it seems irrelevant to changes in this PR.

johnfn:
Ah, wonderful! Thanks for the PR! :)

Issue #862, opened 2016-10-09
o and O on indented line provide wrong indent / strip character


iivvoo:
### What did you do?

Given a piece of (python) code with indenting

```
    # comment _
```

where _ marks the cursor, pressing O should add a line above and 'o' should add an indented line below
### What did you expect to happen?

A line to be added with the proper indent
### What happened instead?

Inserting the line above (O) seems to remove a character. The cursor seems to start at column two and then backspace to cursor one. However, the first space I hit seems to be ignored. I have to add 5 spaces in stead of 4 to match my indent below

Inserting below (o) does something similar: The cursor starts initially at the right indent but then jumps back one column
### Technical details:
- VSCode Version: 1.5.2
- VsCodeVim Version: 0.3.3
- OS: Ubuntu 16.04 64bit

iivvoo:
Perhaps introduced by the fix for issue #574

iivvoo:
Additionally, when I create a new empty file and hit 'i' to insert, the first two letters I type get swapped.

xconverge:
@iivvoo I am unable to reproduce this...can you provide me a bigger snippet for a complete file that I can use for testing? o and O are behaving as they are supposed to in this file for me

```
#test
if(True):
  print("test")
  #test
  #test
```

Regarding your second issue for the new file, and swapping the first two letters...I can't reproduce this either, can you open its own issue? How are you opening a new file

xconverge:
ah I think I can see your 2nd issue with files acting funny, I was able to reproduce it kind of

johnfn:
Yeah, you're totally right. Gonna look into this...

johnfn:
Ugh, this is really tricky to reproduce when you want to reproduce it.

xconverge:
Is this fixed now? I can't reproduce it anymore

iivvoo:
I can't reproduce this anymore. Close?

inejge:
(I found this issue while searching for the closest match for the bug I'm seeing.)

![ovsshifto](https://cloud.githubusercontent.com/assets/1049870/22241237/d0332e80-e21e-11e6-9cf2-bcd4fba4807b.gif)

VSC 1.8.1, Linux 32-bit (Debian), Vim 0.5.2.

xconverge:
@johnfn this looks like it can be closed

Issue #1172, opened 2016-12-16
o and O behave erratically causing failing test


xconverge:
o and O do not always work correctly, I think that there is an issue with allowVSCodeToPropagateCursorUpdatesAndReturnThem

Does not work:

newTestOnly({
     title: "Can handle 'yy'",
     start: ['|one'],
     keysPressed: 'yyO<Esc>p',
     end: ["", "|one", "one"],
   });

Does work:

newTestOnly({
     title: "Can handle 'yy'",
     start: ['|one'],
     keysPressed: 'yyI\n<Esc>kp',
     end: ["", "|one", "one"],
   });

I think it is caused by https://github.com/Microsoft/vscode/issues/17429

PR #1249, opened 2017-01-26
Fix autoindent when opening a line above

inejge:
Re [#862](https://github.com/VSCodeVim/Vim/issues/862#issuecomment-274751266). I rewrote `CommandInsertNewLineAbove` to mirror `CommandInsertNewLineBefore`, just using `editor.action.insertLineBefore` as the editor command to execute.

### DIFF
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -3004,13 +3004,14 @@ class CommandInsertNewLineAbove extends BaseCommand {
   runsOnceForEveryCursor() { return false; }
 
   public async exec(position: Position, vimState: VimState): Promise<VimState> {
+    let indentationWidth = TextEditor.getIndentationLevel(TextEditor.getLineAt(position).text);
     vimState.currentMode = ModeName.Insert;
 
     vimState.recordedState.transformations.push({
       type: "insertText",
-      text: "\n",
+      text: TextEditor.setIndentationLevel("V", indentationWidth).replace("V", "\n"),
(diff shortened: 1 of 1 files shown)


inejge:
I don't really understand the test failure on first glance, will investigate later.

xconverge:
Well if you do a git blame you will get to the reasoning here

https://github.com/VSCodeVim/Vim/issues/1172

Can you just change the diff to do the indenting correctly? 

instead of 

```diff: new PositionDiff(-1, 0),```

do 

```diff: new PositionDiff(-1, charPositionItShouldBeDueToIndentation),```

johnfn:
I think we have a method on TextEditor to determine correct indentation.

inejge:
> Can you just change the diff to do the indenting correctly?

Just repositioning is not enough, a properly indented empty line must be inserted for this to work. I've changed the patch accordingly (the gymnastics with `replace` after `setIndentationLevel` are necessary because the latter doesn't want to indent a string consisting of a bare `"\n"` and keep the newline).

When invoking O for the first time, i get "invalid value for set cursor position..." in the console.

The tests are still failing for reasons I don't understand -- they're failing for me even without the patch.

xconverge:
The failing test is unrelated, not sure what is going on there, this LGTM now

xconverge:
An o and O test that check indentation would be great

johnfn:
This LGTM to me as well. Thank you for the PR, @inejge (and apologies for the slow responses this month!)

Issue #1236, opened 2017-01-20
Remapped otherModes keys occasionally still do old functions instead


rufusroflpunch:
Please *thumbs-up* 👍 this issue if it personally affects you! You can do this by clicking on the emoji-face on the top right of this post. Issues with more thumbs-up will be prioritized.

-----

### What did you do?

I entered the following keymappings:

```javascript
    "vim.otherModesKeyBindingsNonRecursive": [
        {
            "before": ["-"],
            "after": [],
            "commands": [
                {
                    "command": "workbench.action.previousEditor",
                    "args": []
                }
            ]
        },
        {
            "before": ["="],
            "after": [],
            "commands": [
                {
                    "command": "workbench.action.nextEditor",
                    "args": []
                }
            ]
        }
    ]
```

Then I opened multiple tabs and pressed "-" and "=" keys in normal mode.

### What did you expect to happen?

I would expect it to switch backward and forward between tabs every keypress.

### What happened instead?

I noticed that it would occasionally (maybe randomly) go up a line when pressing "-" instead of going to previous tab. Occasionally when pressing "=" it would do nothing. My theory is that it is triggering old functionality instead of using the remapped command (hence why "-" will occasionally go up a line).

### Technical details:

* VSCode Version: 1.8.1
* VsCodeVim Version: 0.4.10
* OS: macOS Sierra 10.12.1

PaulMilla:
I'm experiencing the same issue on Windows 7 Enterprise SP1 at a more consistent rate (exactly every other time) using the following keymappings:

    "vim.otherModesKeyBindingsNonRecursive": [
        {
            "before": ["J"],
            "after": [],
            "commands": [
                {
                    "command":"scrollLineDown",
                    "args":[]
                }
            ]
        },
        {
            "before": ["K"],
            "after": [],
            "commands": [
                {
                    "command":"scrollLineUp",
                    "args":[]
                }
            ]
        }
    ]

'J' will do the default 'join' action on one key press, followed by the correct 'scrollLineDown', followed again by the 'join' action and so on. 'K' similarly will scroll up on one keypress, followed by doing nothing on the next, followed by scrolling up again and so forth.

xconverge:
I think you fixed this in those PRs

PR #1259, opened 2017-02-01
Clear commandList after executing remapped commands

rufusroflpunch:
#1236 

Previously executed remapped commands were leaving the keys inside commandList, resulting in an unmatched command the next time, i.e. `["-","-"]` instead of `["-"]`. Clearing the commandList after executing the commands seems to resolve the issue.


rufusroflpunch:
This seems to cause a different bug all on it's own where commands >1 key are getting erased. Will come back to this later.

Issue #1171, opened 2016-12-16
Go to definition (gd) does not update vimstate cursor correctly if you jump to another file 


xconverge:
Since each file has it's own modeHandler, if you do gd and it goes to a different file that already has a mode handler, then press j or k, it will jump away to the files previous modeHandler position.

If the file you are jumping to does NOT already have a modehandler then it will work fine, but this can only happen once for that  file.

Relates to #1162, #1081, etc

johnfn:
Yeah, however I don't think this is a consequence of separate mode handlers. After all, mode handlers should handle a different position per file, so that when you c-w h c-w l it remembers the old one.

xconverge:
ah yea it is a selection changed type thing then?

johnfn:
Precisely. It's a particularly annoying thing because when you do gd, you need to somehow let VSCodeVim know not to update the cursor on the current mode handler but instead to wait for the next mode handler and then update the cursor on that one. That's why it's so annoying.

johnfn:
Woot.

PR #1262, opened 2017-02-01
Keep cursor position when `gd` opens a new file

grncdr:
This fixes the issue #1171, but is neither elegant nor robust. The tests are spitting out a _lot_ of noise like `invalid value for set cursor position. This is probably bad?` and `Error: Illegal value for line`which makes me quite nervous, but they seem to do this on the master branch as well.

There are also 2 failing tests (edit: these don't seem to fail on Travis 😕 ) , which seem like they _could_ be legitimate failures, but I really can't tell:

```
  2 failing
  1) Motions in Normal Mode Remembers a reverse search from another editor:
      AssertionError: Cursor CHARACTER position is wrong.
      + expected - actual
      -18
      +11

  2) Motions in Normal Mode Shares reverse search history from another editor:
      AssertionError: Cursor CHARACTER position is wrong.
      + expected - actual
      -18
      +11
```

As I understand it, this is meant to test that all editors share the the current search term when using `?`. E.g. `?blah<Enter><C-w><C-w>n` will jump backwards from the current cursor position to the first preceding instance of `blah`.
 
I did some manual testing of this feature locally (doing reverse searches in split editors working on the same file), and couldn't replicate the behaviour of the tests (that is, reverse search seems to work perfectly across editors).

### Checklist

- [x] Also went through https://gist.github.com/rebornix/d21d1cc060c009d4430d3904030bd4c1 for good measure.
- [ ] It builds and tests pass (e.g `gulp`)

### DIFF
--- a/extension.ts
+++ b/extension.ts
@@ -69,8 +69,7 @@ let extensionContext: vscode.ExtensionContext;
 let modeHandlerToEditorIdentity: { [key: string]: ModeHandler } = {};
 let previousActiveEditorId: EditorIdentity = new EditorIdentity();
 
-export async function getAndUpdateModeHandler(): Promise<ModeHandler> {
-  const oldHandler = modeHandlerToEditorIdentity[previousActiveEditorId.toString()];
+export function getModeHandlerForActiveEditor(): ModeHandler {
   const activeEditorId = new EditorIdentity(vscode.window.activeTextEditor);
 
   if (!modeHandlerToEditorIdentity[activeEditorId.toString()]) {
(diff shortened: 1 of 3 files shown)


johnfn:
Thanks for the PR. This one has been a thorn in our side for quite some time. I am currently traveling, but I will try to review it when I find some time!

grncdr:
Hm, after merging master the travis build fails, but with only _one_ of the errors I observed locally.

johnfn:
The two reverse search testcase failures were happening on my machine as well. I am not sure why, but those tests do particularly weird things with opening up new IDEs - maybe the API changed or something...

grncdr:
I've been using this day-to-day, and it's not quite right yet. It seems like the vim state cursor is not always accurately updated when switching windows. Will dig into it later.

xconverge:
Great job test driving it for a bit, that's appreciated :)

johnfn:
I finally managed to fix this. I don't blame you for not trying to track down the root cause; it was pretty difficult to find!

PR #1263, opened 2017-02-02
Clearing commandList for remapped commands

rufusroflpunch:
#1236

Clear commandList for any remapped commands that handled successfully. This resolves the issue of remapped command occasionally doing the original command instead of the remapped command.


johnfn:
I think this fixes a number of issues. Thanks for the PR, @rufusroflpunch!

Issue #1241, opened 2017-01-23
Quotes - doesn't delete both empty quotes in Insert mode.


bargeruns:
Please *thumbs-up* 👍 this issue if it personally affects you! You can do this by clicking on the emoji-face on the top right of this post. Issues with more thumbs-up will be prioritized.

-----

### What did you do?

In insert mode, type `'` or `"` to insert a pair of quotes. Then, press backspace to delete the quotes.

### What did you expect to happen?

_Both_ quotes should be deleted - this is the editor-default behavior, as well as the behavior when using the amVim extension.
### What happened instead?

Only one quote is deleted, so you have to either switch to Normal mode or use forward-delete to delete the trailing quote. This can make a simple typing mistake into an annoyance and interrupt flow.
### Technical details:

* VSCode Version: 1.9.0 (Insider)
* VsCodeVim Version: 0.5.1
* OS: Mac OS 10.12.3

xconverge:
@johnfn this can be closed due to #1267

Issue #1228, opened 2017-01-14
matching brace not deleted on backspace or ex:`x`


Walt-D-Cat:
Please *thumbs-up* 👍 this issue if it personally affects you! You can do this by clicking on the emoji-face on the top right of this post. Issues with more thumbs-up will be prioritized.

-----

### What did you do? 
Type any of the following keys <, (, {,  [ and the backspace or escape then `x`

Put the exact keys you pressed.

### What did you expect to happen?
The first brace character places a matching closing brace, without the extension, removing the left brace removes the closing brace if there are no intervening characters (including space). 
### What happened instead?
The closing brace remains.
### Technical details:

* VSCode Version: 0.4.10
* VsCodeVim Version: *[please ensure you are on the latest]* 1.7.1
* OS: windows 10 & ubuntu 16.10

johnfn:
I don't think that `x` should remove the other matching key. It certainly doesn't on my copy of Vim.

Backspace should though!

Walt-D-Cat:
Fair enough ... the `x` is behaviour handled by an extension in my vim, and it's mostly backspace (especially when miss-typing between `[` and `{` ).

xconverge:
@johnfn this can be closed due to #1267

xconverge:
@johnfn @jasonpoon

PR #1267, opened 2017-02-03
Delete matching bracket upon backspace

rufusroflpunch:
#1228 

When in insert mode, if a user is deleting an opening bracket, and a matching closing bracket is immediately after it, also delete the closing bracket. Corresponding test included.

### DIFF
--- a/src/matching/matcher.ts
+++ b/src/matching/matcher.ts
@@ -1,4 +1,5 @@
-import { Position } from './../motion/position';
+import { Position, PositionDiff } from './../motion/position';
+import * as vscode from 'vscode';
 
 /**
  * PairMatcher finds the position matching the given character, respecting nested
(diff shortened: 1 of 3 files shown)


xconverge:
related #1085 #1084 

At the time it made sense to put this off in hopes that it was part of the vscode api...

xconverge:
and here is the vscode issue

https://github.com/Microsoft/vscode/issues/17490

rufusroflpunch:
I didn't know that. If you still prefer to wait, I won't hem and haw about it.

xconverge:
I will let @johnfn decide, I think it would resolve a few open issues

The problem is without the proper vscode fix, multicursor stuff is tougher I think, do you know if your fix works with multi cursors?

Also, your way is nice!

rufusroflpunch:
Certainly seems to:

multicursor

xconverge:
Is there a way that you can wrap the functionality using the editor option of whether a user has the "editor.autoClosingBrackets": true?

I think that this is great, might be worth adding a comment to the vscode issue to see what the status is or if they have any thoughts

rufusroflpunch:
It now checks for `editor.autoClosingBrackets` setting before deleting closing bracket.

xconverge:
I would also add single and double quotes as well for this behavior. And < is not included in the default vscode behavior

rufusroflpunch:
I added double, single and weird-thing (`) quotes to the behavior and removed angle brackets. I added the quotes to the `PairMatcher.pairings`  object. I was worried about what downstream affects it might have on the text objects associated with quotes, but from my testing I can't see any problems.

xconverge:
regarding weird-thing (back-tick) and <, what I meant was if you disable the vim plugin, and try various ones, only some of them actually auto delete  ```( { [ ' "``` I think are all of them not ``` ` < ```

no big deal, would just leave it as is for now until @johnfn can give input

johnfn:
First of all, thanks for the PR. This is definitely an issue that has been annoying a lot of people for a long time.

Unfortunately, this problem is actually an annoying one. How does it work with stuff like you type in "(blah)" but then backspace blah and the (? We will somehow need to distinguish a ( that was just entered from a ( which was entered a few keystrokes ago. 

The reason that I have held off solving this is that the correct solution to the problem is to perform all text edits and deletes via vscode api, rather than in the manual way that we are currently doing. Sadly, I believe this would require a large refactor to do correctly, so it's probably better to just manually cover these cases for now.

rufusroflpunch:
You're welcome!

In this case @johnfn, it behaves the same if you enter "()" and backspace the "(", as well as "(blah)" and backspace the "(blah". In both cases, it removes the matching bracket. I believed this to be the desired behavior.

johnfn:
Whoops, I never realized that. Alright, this seems good to me then. :)

johnfn:
Oops, I always forget to release my comments. I have some small suggestions for improvement, but we should be fine here for the most part.

rufusroflpunch:
Thanks for suggestions @johnfn. Done.

xconverge:
this also fixes #1241

rufusroflpunch:
@johnfn Is there something I am missing with this?

johnfn:
Nope, it's fine :)

Dunno if I mentioned this, but apologies for any late responses - I am currently on vacation, hopping all around and not really using my laptop all too much.

Issue #1222, opened 2017-01-11
I can't get the ZZ example on the main page to work with current VS Code.


cmaughan:
Please *thumbs-up* 👍 this issue if it personally affects you! You can do this by clicking on the emoji-face on the top right of this post. Issues with more thumbs-up will be prioritized.

-----

### What did you do?

I added the example from the readme to handle ZZ

### What did you expect to happen?

ZZ would save and close the file

### What happened instead?

Nothing.  Other commands, such as replacing jk for <ESC> work fine.

### Technical details:

* VSCode Version: 1.7.2
* VsCodeVim Version: Latest
* OS: Win 10

johnfn:
You're right, it's not working currently. :/

Nogbit:
Missing buffers so much and all the commands to go with them....would love this to work to close files from the keyboard.

alexriedl:
A possible alternative until this gets fixed is:
`    { "key": "shift+z shift+z", "command": "workbench.action.closeActiveEditor",
                                  "when": "editorTextFocus && vim.mode == 'Normal'" },`

Only difference is it does not save the file before trying to close it.

xconverge:
@johnfn  @jpoon  this can be closed now due to #1145

jpoon:
Mmmk. Thanks!

PR #1272, opened 2017-02-05
fixes #1145

### LINKED ISSUE #1145: Three-letter otherModesKeyBindingsNonRecursive remappings fail to match

-----

### What did you do?

Added remapping with three letters, like:
```javascript
    {
      "before": [ "<leader>", "f", "s" ],
      "after": [],
      "commands": [{
        "command": "workbench.action.files.save",
        "args": []
      }]
    },
    {
      // Just some random letters to make sure it wasn't a problem with the leader key
      "before": [ "g", "y", "z" ],
      "after": [],
      "commands": [{
        "command": "workbench.action.quickOpen",
        "args": []
      }]
    }
  ]
```
Then pressed the key combination

The command to run.

### What happened instead?
The third letter's command (`s` in the case of `<leader>fs`) runs.

Also, when I press the first letter, it appears in the status line. But when I press the second letter, the status line goes back to blank.

If I change the remapping to two characters, it works as expected.

### Technical details:

* VSCode Version: 1.7.2
* VsCodeVim Version: 0.4.9, also tried master
* OS: Windows 10

I tried tracing the problem with the debugger, but other than seeing that `Remapper.sendKey` doesn't find a remapping in the three-character case, I couldn't figure out what was up. If someone has a better idea of where to look, I'd be happy to dig into it further. I think this is related to why `"<space>", "f", "s"` doesn't work in https://github.com/VSCodeVim/Vim/issues/1132.

xconverge:
should work...tested the few cases that were in the issue reported

### DIFF
--- a/src/mode/remapper.ts
+++ b/src/mode/remapper.ts
@@ -19,6 +19,11 @@ class Remapper {
   private _remappedModes: ModeName[];
   private _recursive: boolean;
 
+  /**
+   * Have the keys pressed so far potentially be a remap
+   */
+  public couldRemappingApply = false;
+
   constructor(configKey: string, remappedModes: ModeName[], recursive: boolean) {
     this._recursive = recursive;
     this._remappedModes = remappedModes;
@@ -120,8 +125,17 @@ class Remapper {
       }
 
       vimState.isCurrentlyPerformingRemapping = false;
-
       return true;
+    } else {
+      // Check to see if a remapping could potentially be applied when more keys are received
+      for (let remap of this._remappings) {
+        if (keys.join("") === remap.before.slice(0, keys.length).join("")) {
+          this.couldRemappingApply = true;
+          break;
+        } else {
+          this.couldRemappingApply = false;
+        }
+      }
     }
 
     return false;
--- a/src/mode/remapper.ts
+++ b/src/mode/remapper.ts
@@ -22,7 +22,10 @@ class Remapper {
   /**
    * Have the keys pressed so far potentially be a remap
    */
-  public couldRemappingApply = false;
+  private _couldRemappingApply = false;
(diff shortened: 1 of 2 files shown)


src/mode/remapper.ts
```diff
@@ -120,8 +125,17 @@ class Remapper {
       }
 
       vimState.isCurrentlyPerformingRemapping = false;
-
       return true;
+    } else {
+      // Check to see if a remapping could potentially be applied when more keys are received
+      for (let remap of this._remappings) {
```
johnfn: Is it possible to make this a return value rather than returning it implicitly as a class property?

xconverge: Want me to just make a getter for it? It is already used on lines 77 and 83

xconverge: or do you mean break off the function into a isRemappingPotential() function that takes the current keystrokes as an input?

johnfn: I mean could we return this value from `sendKey` somehow?

xconverge:
this also fixes #1222

src/mode/remapper.ts
xconverge: Unless I am missing something obvious that you can see pretty clearly...

It is definitely possible, the problem I see is that we call handleKeyEvent, which calls sendKey, then it calls handleKeyEventHelper which uses the property

I could make a data structure that returned a few properties but I don't think it will end up being cleaner...

sendkey only returning a boolean seems to make sense to me, I would really only change this to return a true/false/maybe with the data structure anyways

johnfn: I see what you mean, it would be annoying to thread through. OK, a getter is fine.

xconverge:
@johnfn should be good to go now, let me know if there is anything else

PR #1277, opened 2017-02-06
Support exact and inexact current word search

### LINKED ISSUE #1266: Search for current word is inexact

-----

### What did you do?

The following four commands are incorrectly implemented:
- <kbd>*</kbd> Forwards exact word search
- <kbd>#</kbd> Backwards exact word search
- <kbd>g</kbd><kbd>*</kbd> Forwards partial word search
- <kbd>g</kbd><kbd>#</kbd> Backwards partial word search

Exact word searches should search for exact words. eg. Pressing <kbd>*</kbd> on the string "he" should not highlight "hello".

Partial word searches should search within words for the search string.

[Vim wiki reference](http://vim.wikia.com/wiki/Searching#Searching_for_the_current_word)

### What happened instead?

Exact searches do partial search.
Partial search commands are not implemented.

### Technical details:

* VSCode Version: 1.9.0
* VsCodeVim Version: 0.5.2
* OS: OSX

rhys-vdw:
Rename `CommandStar` and `CommandHash` to `CommandSearchCurrentWordExactForward` and `CommandSearchCurrentWorldExactBackward`, which describe their effect. Also include regex word bounds in the search so that they act the same as in Vim.

Add `CommandSearchCurrentWordForward` and `…Backward` (commands `g*` and `g#` respectively), that match without word bounds.

Closes #1266

---

So this is mainly a copy+paste, I wasn't sure if there was an idiomatic approach to sharing behaviour between these two commands. Also I renamed the classes because it felt more sensible to name them after their effect than their keybindings (also it seemed strange to call the new one something like `CommandGStar`.

Let me know if there's anything stylistically or functionally wrong here, I'm happy to iterate on it.

### DIFF
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -1570,6 +1570,37 @@ class CommandCmdA extends BaseCommand {
   }
 }
 
+function searchCurrentWord(position: Position, vimState: VimState, direction: SearchDirection, isExact: boolean) {
+    const currentWord = TextEditor.getWord(position);
+    if (currentWord === undefined) {
+      return vimState;
+    }
+
+    // For an exact search we need to use a regex with word bounds.
+    const searchString = isExact
+      ? `\\b${currentWord}\\b`
+      : currentWord;
+
+    // Start a search for the given term.
+    vimState.globalState.searchState = new SearchState(
+      direction, vimState.cursorPosition, searchString, { isRegex: isExact }
+    );
+
+    // If the search is going left then use `getWordLeft()` on position to start
+    // at the beginning of the word. This ensures that any matches happen
+    // outside of the currently selected word.
+    const searchStartCursorPosition = direction === SearchDirection.Backward
+      ? vimState.cursorPosition.getWordLeft(true)
+      : vimState.cursorPosition;
+
+    vimState.cursorPosition = vimState.globalState.searchState.getNextSearchMatchPosition(searchStartCursorPosition).pos;
+
+    // Turn one of the highlighting flags back on (turned off with :nohl)
+    Configuration.hl = true;
+
+    return vimState;
+}
+
 @RegisterAction
 class CommandSearchCurrentWordExactForward extends BaseCommand {
   modes = [ModeName.Normal, ModeName.Visual, ModeName.VisualLine];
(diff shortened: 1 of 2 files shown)


src/actions/actions.ts
```diff
@@ -1583,6 +1583,34 @@ class CommandStar extends BaseCommand {
       return vimState;
     }
 
+    vimState.globalState.searchState = new SearchState(
```
johnfn: Normally we share methods with static methods on one of the relevant classes.

rhys-vdw:
Seems I need to update some tests... I started looking into this and found [this test](https://github.com/VSCodeVim/Vim/blob/ac871462d13ec36429753c33653fe3a844e5c2fd/test/mode/normalModeTests/motions.test.ts#L472-L477):

```ts
  newTest({
    title: "Can handle tricky *",
    start: ['|blah blahblah duh blah'],
    keysPressed: '*',
    end: ['blah blahblah duh |blah']
  });
```

Which superficially implies that the change I've made was already implemented? However, on further investigation this is actually behaving incorrectly. Either `*` or `#` will immediately jump to the final `blah`, but then pressing `n` will cycle through all `blah`s (including the two in `blahblah`). So the initial jump is incorrect.


**Vim**
vim

**VSCodeVim**
Note how it jumps to the end of line on either `*` or `#`. This is master (not my branch). I haven't fixed this specific problem.
vcodevim


I think I should review these few test cases and make sure they can't trigger false positives (and add more coverage for these commands). I need to fix other broken tests as well so

johnfn:
Actually, the thing is, like I mentioned, this behavior actually works correctly on my machine. So I feel like we're dealing with some config oddities rather than actual bugs. (I do remember covering this case many many months ago...)

rhys-vdw:
Hm. Yeah, that is weird. You did mention that and I had assumed at the time that you had misunderstood my report (since the code itself didn't seem to handle the correct behaviour). Even if I completely clear my config this still behaves as reported in #1266, so perhaps you have some configuration that is causing it to do the right thing?

Could you confirm whether you get the incorrect initial move with `*` and `#` that I show in the second gif above?

johnfn:
Oh yeah good point. My Vim related settings are:

```
    "vim.useCtrlKeys": true,
    "vim.useSolidBlockCursor": false,
    "vim.timeout": 1000,
    "vim.easymotion": true,
    "vim.showcmd": true,
    "vim.otherModesKeyBindings": [ ],
```

rhys-vdw:
Using your configuration made no difference.

johnfn:
Great! I have been noticing weirdness with the tests as well - sorry if that was confusing.

Thanks again for the PR.

PR #1278, opened 2017-02-07
avoid using user remapping in test mode

rufusroflpunch:
User remapping affect tests, resulting in unpredictable test breakage on local. If the remapping functionality starts to receive testing later, a new solution will probably need to be found.


jpoon:
Restarted Travis.

xconverge:
Lgtm

johnfn:
Oh, this is a good one. Thanks @rufusroflpunch!

PR #1279, opened 2017-02-07
fixes #1247

### LINKED ISSUE #1247: Highlighted character jumps around when highlighting text.
In the output window or editor, highlight some text (in normal mode)
gif
The cursor/selected char lags behind and sometimes jumps to the right when highlighting left.

* VSCode Version: 1.8.1
* VsCodeVim Version: 0.5.2
* OS: Windows 10

xconverge:
@johnfn any ideas on how this can be done better?

1. I assume there is a more elegant way to if else on the await? I was thinking some sort of if/else resolve thing?

2. Any ideas for a better solution to this issue? Basically if you have useSolidBlockCursor and then do a mouse selection, it does not always match the selection you selected if you press d for example


johnfn:
You could do `const update = vscode.workspace.blah` and then `if (blah) { await update; }`.

I don't understand why the solid cursor is messing stuff up though.

xconverge:
It is that await, when you drag out a mouse selection, it calls updateView over and over, getting the config option, and setting it that fast is not working in this case

xconverge:
yes it is, I am not sure a task queue will completely remedy it, it is still slow. It will fix the main bug, but it will still appear very laggy

johnfn:
So the reason I didn't pull this was because you said it seemed very laggy, which I didn't like - it sounded like a game killer. But I just realized you might not agree. Do you think this is good to be pulled?

xconverge:
No I don't think it's good to merge as is, I can close it too, but it is on the right track

xconverge:
@johnfn it looks like maybe your taskqueue stuff made this a bit better, but now my point of it still appearing laggy holds true

Say you do a big 10 line mouse selection, it is all queued up and executed in the right order, but it takes about 2x-4x as long to process through the whole queue as it does to do the mouse movement, so you see it do weird things in that time.

I don't think that this PR is the right way to solve it but it isn't too bad I don't think. I will think about it some more later

Issue #1255, opened 2017-01-29
Surround does not behave as it should in a few cases


xconverge:
Great work, your way of implementing the surround plugin is pretty nice! Here are some findings I had that I wanted to at least document. It is kind of nitpicking but also takes away quite a bit from the usage of the plugin when things are off by one etc, I anticipate some people might mention things on slack related to these.

Here are a few cases that I found surround does weird things:

1.
`te|st`
`ysaw"`
expected:` "test"`
actual: `t"est"`
(this only happens on the last line of a file I think...)

2.
`te|st`
`ysas"`
expected: `"test"`
actual: `test""`

3.
visual line select, [ ] is selection
```
[test
test]
```

`S"`
expected:
```
"
test
test
"
```
actual:
```
"test
test"
```

4.
Backspace when entering a tag should remove the previous character, not add <BS> (similar to searchstate)

5.
surround t should trigger the same thing as <, not surround with t

6.
Should not be able to surround using any character, and b and B should be brackets
check out this data structure for the only valid pairs besides tags, I think it is correct:

https://github.com/xconverge/Vim/blob/8c24283d2801596378f9d34dee4b27b592f8968f/src/surround/surround.ts#L29

kitsu:
I ran into the `t` for tag one already. Also some kind of feedback durring tag input would be nice.
Over all great feature though, the one thing I *really* missed most from Vim.

xconverge:
feedback during tag input was added to master, will show in the status bar atleast for now

xconverge:
fixed most in #1289

PR #1289, opened 2017-02-10
More surround fixes

xconverge:
Fixes the important ones from #1255, the only ones left have nothing to do with surround but more so how sentence objects and word objects work

- backspace when entering tags
- visual line then surround creates newlines before/after and puts cursor in right place
- surround t triggers tag input correctly
- added some shortcuts from the original surround plugin b = (, B = {, and r = [


johnfn:
Awesome! Thanks!

johnfn:
I'm so confused. I was 99% certain I merged this.

PR #1297, opened 2017-02-14
D in visual mode behaves like d

xlaech:
The delete function of the DeleteOperation Class already account for this case, if the Register Mode is set to LineWise.

### DIFF
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -1869,6 +1869,10 @@ export class DeleteOperatorVisual extends BaseOperator {
     public modes = [ModeName.Visual, ModeName.VisualLine];
 
     public async run(vimState: VimState, start: Position, end: Position): Promise<VimState> {
+      // ensures linewise deletion when in visual mode
+      // see special case in DeleteOperator.delete()
+      vimState.currentRegisterMode = RegisterMode.LineWise;
+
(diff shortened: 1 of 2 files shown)


xlaech:
Sorry for the earlier Code. Should have created a new branch for the first one as well :)

xconverge:
Nice, mind adding a test to the bottom of modeVisual.test.ts? you can use newTestOnly() to run just the test that you are making, then you can switch it back to newTest() before you commit so that all tests still run

xconverge:
Awesome!

src/actions/actions.ts
```diff
@@ -696,7 +696,9 @@ export class CommandOneNormalCommandInInsertMode extends BaseCommand {
 
   public async exec(position: Position, vimState: VimState): Promise<VimState> {
     vimState.returnToInsertAfterCommand = true;
-    return await new CommandEscInsertMode().exec(position, vimState);
+    return await new CommandEscInsertMode().exec(
```
johnfn: I don't think that this bit is relevant to the D fix, right? Could you remove it and include it in a separate PR?

johnfn: Wait nevermind, it doesn't really matter heh.

johnfn:
Excellent, and tests too. Awesome job. Thanks for the PR. :)

PR #1301, opened 2017-02-15
fixes travis with minor hack used in tests

xconverge:
wooo


xconverge:
@jpoon would be nice to get this one in to fix the build

it is an ugly hack but it is only used in tests and the functionality in those tests all works

johnfn:
Yeah fair enough. Nice job finding this.

jpoon:
Lol. Nice fix :)

PR #1303, opened 2017-02-15
Fix c on line beginning#1302

xlaech:
Changed the c Action to encounter for cases where the selection ends at the beginning of a line.
The tests revealed, that s is implemented using c(0,0).

Therefore i check, if this is a real selection before calling the getLeftThroughLineBreaks() function

### DIFF
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -2082,7 +2082,7 @@ export class ChangeOperator extends BaseOperator {
         const isEndOfLine = end.character === end.getLineEnd().character;
         const isBeginning =
                 end.isLineBeginning() &&
-                start.isLineBeginning() // to ensure this is a selection and not e.g. and s command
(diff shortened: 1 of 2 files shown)


src/actions/actions.ts
```diff
@@ -2080,13 +2080,16 @@ export class ChangeOperator extends BaseOperator {
 
     public async run(vimState: VimState, start: Position, end: Position): Promise<VimState> {
         const isEndOfLine = end.character === end.getLineEnd().character;
+        const isBeginning =
```
johnfn: Position has a function called `isLineBeginning()` - could you use that one?

xlaech:
Sorry i got excited :)

xlaech:
just wait... i need to make sure the tests pass again

johnfn:
Cool, thanks for the PR!

PR #1331, opened 2017-02-25
fixes #1327

### LINKED ISSUE #1327: Text input between quotes is slow
### What did you do?

1. Install `vim` extension.

2. open a large `js` file.  e.g : https://code.jquery.com/jquery-3.1.1.js

3. Typing `I` switch to insert mode.

4. Typing `"` and other letters (e.g: "Hello") at **first line**. The other letters **takes a long time** to appear.

gif

### Technical details:

* VSCode Version: 1.9.0
* VsCodeVim Version: 0.5.3
* OS: windows7 and mac

xconverge:
@rufusroflpunch this isn't the cleanest way to fix the problem but I think it is fine, what do you think?

The problem is we use nextPairedCharacter in the matcher in a few other places, and if it is not in the pairings, then it returns instantly. However, when we added quotes, this meant nextPairedCharacter would now iterate the document looking for a forward or backward matching char, this could probably be made to work correctly since in these cases quotes are typically pairs, but this solution seemed easier to me for now...

### DIFF
--- a/src/matching/matcher.ts
+++ b/src/matching/matcher.ts
@@ -17,11 +17,6 @@ export class PairMatcher {
     // useful for text objects.
     "<" : { match: ">",  nextMatchIsForward: true  },
     ">" : { match: "<",  nextMatchIsForward: false },
-    // These are useful for deleting closing and opening quotes, but don't seem to negatively
-    // affect how text objects such as `ci"` work, which was my worry.
-    '"' : { match: '"',  nextMatchIsForward: false  },
-    "'" : { match: "'",  nextMatchIsForward: false  },
-    "`" : { match: "`",  nextMatchIsForward: false  },
   };
 
   static nextPairedChar(position: Position, charToMatch: string, closed: boolean = true): Position | undefined {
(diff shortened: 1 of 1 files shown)


src/matching/matcher.ts
```diff
@@ -17,11 +17,6 @@ export class PairMatcher {
     // useful for text objects.
     "<" : { match: ">",  nextMatchIsForward: true  },
     ">" : { match: "<",  nextMatchIsForward: false },
-    // These are useful for deleting closing and opening quotes, but don't seem to negatively
-    // affect how text objects such as `ci"` work, which was my worry.
-    '"' : { match: '"',  nextMatchIsForward: false  },
```
johnfn: Can you leave these inside PairMatcher? Because we use PairMatcher's pairings in other places, like surround.

Can you just give them an attribute like `directionless: true`? which you check for down below instead?

xconverge: sure, but you shouldn't be using quotes in the pairmatcher for surround...there is no open or close, so how is doing a surround with " different than a surround with any other normal character?

johnfn: Actually you're right, I just duplicated that code in surround.vim. Yeahhhhhh. 

I still would like it to be here for completeness though. :)

xconverge: agreed :)

johnfn:
Perfect, thanks!

PR #1338, opened 2017-02-27
Fix #1318

### LINKED ISSUE #1318: r<tab> inserts "<tab>" instead of "\t"

-----

### What did you do?

`r<tab>`

Character under the cursor replaced with tab character (`"\t"`).

### What happened instead?

Character under the cursor replaced with the string `"<tab>"`.

Note that `s<tab><esc>` and `cl<tab><esc>` work as expected.

### Technical details:

* VSCode Version: 1.9.0
* VsCodeVim Version: 0.5.3
* OS: Windows 7

rebornix:
It has been a while since my last commit, so I'd like to start with PR again this time even though my changes don't touch any piece that I don't understand.

### DIFF
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -4456,6 +4456,15 @@ class ActionReplaceCharacter extends BaseCommand {
     let timesToRepeat = vimState.recordedState.count || 1;
     const toReplace = this.keysPressed[1];
 
+    /**
+     * <character> includes <BS>, <SHIFT+BS> and <TAB> but not any control keys,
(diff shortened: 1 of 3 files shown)


rebornix:
It's about delegating back commands that we don't want to handle ourselves, eg, Tab which depends on quite a few Code settings.

The only catch is probably that we are delegating things back Code instead of asking Code to find another proper registered Tab handler but since Code doesn't support that right now, we are in good shape.

johnfn:
LGTM. :)

Issue #1341, opened 2017-02-28
:sort not supported


jordan-heemskerk:
Please *thumbs-up* 👍 this issue if it personally affects you! You can do this by clicking on the emoji-face on the top right of this post. Issues with more thumbs-up will be prioritized.

-----

### What did you do?

Run the command `:sort`

### What did you expect to happen?

The entire file is sorted alphabetically. See [Sort lines](http://vim.wikia.com/wiki/Sort_lines).

### What happened instead?

Nothing.

### Technical details:

* VSCode Version: 1.9.1
* VsCodeVim Version: 0.5.3
* OS: Windows 10

I use this feature a lot and it would be super helpful to have in VSCode. I'll try to make a Pull Request, but I've never worked in TypeScript before.

johnfn:
Have you considered using something like https://marketplace.visualstudio.com/items?itemName=Tyriar.sort-lines?

If that doesn't work for you, then by all means make a PR. :)

jordan-heemskerk:
Gave it a shot, let me know what you think.

PR #1342, opened 2017-02-28
Add limited support for :sort

jordan-heemskerk:
- Works on a range
 - ! option is supported to reverse the sort
 - Other options not supported at this time

Fixes: #1341

### DIFF
--- a/src/cmd_line/subparsers/sort.ts
+++ b/src/cmd_line/subparsers/sort.ts
@@ -6,8 +6,6 @@ export function parseSortCommandArgs(args : string) : node.SortCommand {
 
   const reverse = args !== null && args.indexOf("!") >= 0;
 
-  return new node.SortCommand({
-    reverse: reverse,
-  });
+  return new node.SortCommand({ reverse });
 
 }
--- a/src/cmd_line/subparsers/sort.ts
+++ b/src/cmd_line/subparsers/sort.ts
@@ -1,0 +1,13 @@
+"use strict";
+
+import * as node from "../commands/sort";
+
+export function parseSortCommandArgs(args : string) : node.SortCommand {
+
+  const reverse = args !== null && args.indexOf("!") >= 0;
+
+  return new node.SortCommand({
+    reverse: reverse,
+  });
+
+}
--- a/src/cmd_line/commands/sort.ts
+++ b/src/cmd_line/commands/sort.ts
@@ -1,0 +1,64 @@
+"use strict";
+
+import * as vscode from "vscode";
+import * as node from "../node";
+import * as token from "../token";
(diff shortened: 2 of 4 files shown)


xconverge:
Hmmm, you could have used editor.action.sortLinesAscending and editor.action.sortLinesDescending I think?

src/cmd_line/subparsers/sort.ts
```diff
@@ -0,0 +1,16 @@
+"use strict";
+
+import * as node from "../commands/sort";
+
+export function parseSortCommandArgs(args : string) : node.SortCommand {
+
+  let reverse_ = false;
```
johnfn: Why the trailing underscore? Also you could write `const reverse = args !== null && args.indexOf("!") >= 0;`

jordan-heemskerk:
@xconverge That'd be super nice not to have to roll my own sorting routine. Is there a way we can support a line range whilst doing that?

For example: `1,3sort`, see [Line 41](https://github.com/VSCodeVim/Vim/pull/1342/files#diff-6adb5f89b8602fce4ac0927a942a9df8R41) in the new test file.

src/cmd_line/subparsers/sort.ts
jordan-heemskerk: No specific reason, my bad. Will update the conditional as well.

xconverge:
@jordan-heemskerk I think you would have to set the selection to contain those lines and then run the command

Might just be easier to keep the homegrown one though since a sort isn't too complicated

jordan-heemskerk:
@xconverge yeah, my thinking as well if that is the case. Especially since that sorting routine may grow more complex to support the [full option list](http://vimdoc.sourceforge.net/htmldoc/change.html#:sort) of `:sort`.

xconverge:
the failing test isn't you

jordan-heemskerk:
Yeah looks like `master` is in a bad state. I'll let you guys merge this once its happy again.

src/cmd_line/subparsers/sort.ts
```diff
@@ -0,0 +1,13 @@
+"use strict";
+
+import * as node from "../commands/sort";
+
+export function parseSortCommandArgs(args : string) : node.SortCommand {
+
+  const reverse = args !== null && args.indexOf("!") >= 0;
+
+  return new node.SortCommand({
+    reverse: reverse,
```
johnfn: Fun fact: `{ reverse: reverse }` can be better written in ES6 as `{ reverse }`. :)

johnfn:
Heh I forgot I can make little tweaks directly from Github. That's a new feature :) 

Alright, this LGTM! Thanks for the PR, @jordan-heemskerk 😄

PR #1346, opened 2017-03-01
regex match

rebornix:
Use regex to match brackets instead of iterating all characters one by one. Fix performance issues of inserting brackets in a large file.

cc @xconverge @johnfn

### DIFF
--- a/src/matching/matcher.ts
+++ b/src/matching/matcher.ts
@@ -1,5 +1,31 @@
 import { Position, PositionDiff } from './../motion/position';
+import { TextEditor } from "./../textEditor";
 import * as vscode from 'vscode';
+
+function escapeRegExpCharacters(value: string): string {
+  return value.replace(/[\-\\\{\}\*\+\?\|\^\$\.\,\[\]\(\)\#\s]/g, '\\$&');
+}
+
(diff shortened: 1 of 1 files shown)


johnfn:
Haha okay this is really interesting. I don't doubt that it's faster, but I have no idea why. A regex would _certainly_ be slower than a handrolled checking function... 

My best guess is that `Position.IterateDocument` was SUPER slow? That's kind of weird though.

rebornix:
@johnfn it's very likely Generators, which might not be optimized enough by V8.

rebornix:
```
		function norm() { 
			for (var i = 0; i < 10000000; i++);
			return 42;
		} 
		function *g() {
			for (var i = 0; i < 10000000; i++) {
				yield i;
			}
		}
		function gen() {
			for(const i of g());
			return 42;
		}
		console.time('normal');
		console.log(norm()); 
		console.timeEnd('normal');
		console.time('generator');
		console.log(gen());
		console.timeEnd('generator');
```

Result

```
generators.html:21 normal: 6.246ms
generators.html:24 generator: 578.637ms
```

johnfn:
Holy smokes. Yeah, that will do it!

Issue #1233, opened 2017-01-19
Relative line move with remapped hjkl keys not supported


bdauria:
### What did you do?
I remapped hjkl keys to ctsr, since I'm using a [french dvorak](https://bepo.fr/wiki/Accueil) keyboard. Relative moves work only with the default hjkl mapping 

### What did you expect to happen?
Relative line moves to follow the ctsr mapping. 

### What happened instead?
The only way to move from lines relatively is to use the default mapping: for instance 5j instead of 5t

### Technical details:

* VSCode Version: 1.8.1
* VsCodeVim Version: 0.4.10
* OS: MacOS

johnfn:
Are you saying that if you remap h to say c then 5c does not do 5h?

bdauria:
correct.

PR #1359, opened 2017-03-04
Perform remapped commands when prefix by a number

bdauria:
This PR allows remapped commands to perform correctly when prefixed by a count. 

The modified code makes sure that only the command uses its remapped version, and not the following text objects. 

For instance, if the command l is remapped to c, and w to é performing: 2liw will still be translated to 2ciw and not 2cié (this is the current behaviour for normal, non-count commands). 

The code uses keymappings defined in otherModesNonRecursive. Should recursive mappings be supported in this case as well ? 

This should close the following issue: https://github.com/VSCodeVim/Vim/issues/1233

### DIFF
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -21,6 +21,7 @@ import { FileCommand } from './../cmd_line/commands/file';
 import { QuitCommand } from './../cmd_line/commands/quit';
 import * as vscode from 'vscode';
 import * as clipboard from 'copy-paste';
+import { getOtherModesNonRecursiveMapping } from '../mode/remapper';
 
 
 const is2DArray = function<T>(x: any): x is T[][] {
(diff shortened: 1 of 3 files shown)


johnfn:
Not ignoring this, just waiting for some time to properly review it...

bdauria:
Sure, no worries take the time you need for this.

bdauria:
@johnfn The thing that stopped me from implementing the feature in modeHandler.ts, is the limitation that the command has to be 1 char long. That basically means that it's already too late for remapping '2s' for instance. 

We could handle this special case by adding a condition, and then remapping everything that is not a count prefix. 

What do you think ?

johnfn:
Thanks, @bdauria! If you have any trouble, feel free to ping me on Slack.

bdauria:
@johnfn I've pushed some changes. I basically reverted all the changes from the original PR and made the fix in ModeHandler instead. 
The command passed to the different remappers will now be without the detected count prefix (see: https://github.com/VSCodeVim/Vim/pull/1359/commits/b009d410ba5874ad587645a8d66d8540afae3419). 

as we discussed, moving the remapping part to getRelevantActions() would to big of a refactoring so not really in the scope of this PR.

johnfn:
Awesome. Thanks for the great work, @bdauria!

PR #1361, opened 2017-03-05
Update config options without restarting

xconverge:
This is really nice and was easy!

It exposes some weirdness with how we do remapping but o well.. let me know what you think

### DIFF
--- a/src/configuration/configuration.ts
+++ b/src/configuration/configuration.ts
@@ -33,6 +33,18 @@ class ConfigurationClass {
   private static _instance: ConfigurationClass | null;
 
   constructor() {
+    this.updateConfiguration();
+  }
+
+  public static getInstance(): ConfigurationClass {
+    if (ConfigurationClass._instance == null) {
+      ConfigurationClass._instance = new ConfigurationClass();
+    }
+
+    return ConfigurationClass._instance;
+  }
+
+  updateConfiguration() {
     /**
      * Load Vim options from User Settings.
      */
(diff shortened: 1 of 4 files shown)


src/configuration/configuration.ts
```diff
@@ -149,28 +153,28 @@ class ConfigurationClass {
   /**
    * Size of a tab character.
    */
-  @overlapSetting({ codeName: "tabSize", default: 8})
+  @overlapSetting({ codeName: "tabSize", default: 8 })
```
johnfn: Haha I see my weird spacing habits are starting to rub off on you.

johnfn:
Overall feedback: Instead of updating them, could you just recreate them entirely - construct new ones? Ideally we only have one code path: addNewRemappers(), which gets called immediately and whenever a config update happens, rather than the 2 code paths we have right now.

src/configuration/configuration.ts
xconverge: cmd+a then = :)

xconverge:
yep, makes sense

src/configuration/configuration.ts
johnfn: Whoa, I didn't know that it was the standard style... I thought I was doing something weird.

rebornix:
LGTM, the only thing that I think we can change is guarding Vim from flooded configuration changes.

johnfn:
Yeah this is awesome. It's super simple and an incredibly useful change. Imagine how many fewer questions we'll have to answer on Slack now 😉

Issue #1055, opened 2016-11-12
How to close vim mode ?


leecz:
i want to enable or disable vim mode easily.

johnfn:
This can be done from within VSCode. Just go to the extensions menu and click "Disable".

FSMaxB:
Although that requires a restart of VSCode.

johnfn:
Take that up with the VSCode team 😉

leecz:
谢谢你们！
thanks! 
这个插件非常棒，它让我从 vim 转到 VScode了！
this extension is so nice！I started to use VScode now.

artusvranken:
Any chance thi will get supported? I'm looking for a quick shortcut to enable/disable the extension. Manually opening the extension page and toggling the state makes for an impacting context switch while working.

jpoon:
https://github.com/VSCodeVim/Vim#vimdisableextension

artusvranken:
Ok, thank you.

So for people that don't know how to do this: You can [open the Command Palette by pressing F1](https://code.visualstudio.com/shortcuts/keyboard-shortcuts-windows.pdf). A text field will open where you can enter commands. Enter `toggleVim` to disable/enable Vim mode.

jpoon:
You can bind that command to whatever key you'd like:

```
{
  "command": "toggleVim",
  "key": "ctrl+k ctrl+v"
}
```

lqueryvg:
@jpoon  thanks, that's helpful.
But the big problem with that (and this is on any application and any platform...) is choosing a key mapping to use. It's too easy to choose a mapping which conflicts with something else, either in the application or the platform itself (layer vim on top of that and..., oh...).
I couldn't see an easy way to list all in-use key mappings in VS code.
Now that would be useful feature.

federicopessina:
In VSCode you can just go to where NORMAL, INSERT, VISUAL mode is displayed, at the bottom of VSCode window, and left-click on it. It will display -- VIM:DISABLED --

PR #1364, opened 2017-03-06
Toggle Vim

rebornix:
#1055 Then you can disable Vim temporarily

cc @jpoon @xconverge @johnfn . I'm happy that we don't have too many dirty checks in code but our package.json is laughing at us.

### DIFF
--- a/extension.ts
+++ b/extension.ts
@@ -252,8 +252,8 @@ export async function activate(context: vscode.ExtensionContext) {
       mh.updateView(mh.vimState, { drawSelection: false, revealRange: false });
     } else {
       let cursorStyle = await vscode.workspace.getConfiguration('editor').get('cursorStyle', 'line');
-      if (vscode.window.activeTextEditor) {
-        let options = vscode.window.activeTextEditor.options;
+      vscode.window.visibleTextEditors.forEach(editor => {
+        let options = editor.options;
         switch (cursorStyle) {
           case 'line':
             options.cursorStyle = vscode.TextEditorCursorStyle.Line;
@@ -268,8 +268,8 @@ export async function activate(context: vscode.ExtensionContext) {
           default:
             break;
         }
-        vscode.window.activeTextEditor.options = options;
-      }
+        editor.options = options;
+      });
       await vscode.commands.executeCommand('setContext', 'vim.active', Globals.active);
     }
   });
--- a/extension.ts
+++ b/extension.ts
@@ -125,6 +125,10 @@ export async function activate(context: vscode.ExtensionContext) {
   vscode.window.onDidChangeActiveTextEditor(handleActiveEditorChange, this);
 
   vscode.workspace.onDidChangeTextDocument((event) => {
+    if (!Globals.active) {
+      return;
+    }
+
     /**
      * Change from vscode editor should set document.isDirty to true but they initially don't!
      * There is a timing issue in vscode codebase between when the isDirty flag is set and
@@ -158,7 +162,7 @@ export async function activate(context: vscode.ExtensionContext) {
     }, 0);
   });
 
-  registerCommand(context, 'type', async (args) => {
+  overrideCommand(context, 'type', async (args) => {
     taskQueue.enqueueTask({
       promise: async () => {
         const mh = await getAndUpdateModeHandler();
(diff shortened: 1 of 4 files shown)


src/globals.ts
```diff
@@ -9,4 +9,6 @@ export class Globals {
   public static modeHandlerForTesting: any = undefined;
 
   public static WhitespaceRegExp = new RegExp("^ *$");
+
+  public static active = true;
```
xconverge: can you add a comment so people know what this means

extension.ts
```diff
@@ -238,15 +242,54 @@ export async function activate(context: vscode.ExtensionContext) {
     registerCommand(context, keybinding.command, () => handleKeyEvent(`${ bracketedKey }`));
   }
 
+  registerCommand(context, 'toggleVim', async () => {
+    Globals.active = !Globals.active;
+    if (Globals.active) {
```
xconverge: I imagine somewhere in here we want to do this for each mh we currently have?

xconverge: Or does having it just here impact all mh correctly already?

xconverge:
maybe want to clear out the status bar too? or put [vim-disabled] or something

jpoon:
this feels so dirty. so we really need/want this change? the fact that we *can* do it doesn't necessarily mean we *should*. 

i personally would never use this and it seems sufficient to disable this through the extensions UI...

johnfn:
The one use case I know of is snippets, which we don't currently support. I've heard of people restarting their editor and disabling vim just to use a few snippets, which is kind of insane. With this and xconverges work making restarts unnecessary for config changes, I can see this being used by some people. 

Agreed that it's not the idea solution for any individual problem we have though.

xconverge:
It is in VsVim, I have used the vsvim version on numerous occasions.

My use case is a coworker comes over and wants to explain something etc, as soon as they touch the keyboard, crazy stuff happens if they don't use vim. I personally will never use this for myself, it is only for that use case.

johnfn:
@xconverge I think what @jpoon is saying is that this doesn't offer much advantage over just disabling the extension in the sidebar.

xconverge:
Ok so what if I want to interrupt my coworker and type out a function?

I see what you guys are saying, but I don't think it is that dirty, and vscode is slow to disable/reload

johnfn:
He _could_ just press `i` for you, hehe.

I'm ambivalent, but I'm never _really_ against adding new features. 😉  (Oh, except useful ones, of course. _ducks_ )

xconverge:
I have tried pressing i for coworkers, but then inevitably they go to copy paste or something and it just becomes silly for both of us

johnfn:
Haha and then it cuts off the last character and you get in a fighting match that ends in attempting to install spacemacs for 20 minutes?

extension.ts
rebornix: Nice catch!

src/globals.ts
rebornix: Addressed.

jpoon:
My question is do people do this so often that they need a keyboard shortcut to save them the 2 seconds it would take to do through the UI?

rebornix:
My use case is I'm trying to avoid reload VS Code as possible

* VSCode source code is large and once Code is reloaded, it takes quite a while for TypeScript Language Service to warm up. 
* Tasks will be terminated when reloading Code
* You can't reload as you are in a debugging session. etc, etc.

If I disable Vim for VS Code workspace, sometimes I forgot to bring it back as it's costly.

@jpoon It would be a good feature request to have a new handler called `disableExtension` in `extension.ts` but that requires thorough thinking. I'm not happy with the code either, it's nasty and I can wait for more ppl's feedback or request for this.

xconverge:
I think we should leverage #1425 here and add the vim.use< tag to every binding. Then we can use the Global.isActive flag to set context

Chillee:
I think this would be useful. There are several cases where the plugin as it currently stands has very painful behavior to deal with (namely snippets and folds), and it would be very helpful to be able to disable it temporarily.

jpoon:
Dammit, I don't know how to git :(

I intended to resolve the conflicts, but ended up just creating a new brand-new branch. https://github.com/VSCodeVim/Vim/tree/toggle_vim

rebornix:
@jpoon i'll rebase and resolve the conflicts

jpoon:
I already did the work of resolving the conflicts here: https://github.com/VSCodeVim/Vim/tree/toggle_vim. We can create a new PR with that and close this one? Of you can try cherry-picking commits from there to your current branch.

Sorry, this was more complex than it needed to be.

PR #1386, opened 2017-03-12
fixes #1382

### LINKED ISSUE #1382: Executing a macro that runs many operators creates many undo points
It should only create one.

xconverge:
seems to work pretty well...


johnfn:
Awesome.

PR #1387, opened 2017-03-12
fixes #1385 % motion with a count

### LINKED ISSUE #1385: The 'jump to n percentage of file' function seems not working

-----

### What did you do?

50%

Jump to the middle of current file

### What happened instead?
The cursor is no where near the middle of file

### Technical details:

* VSCode Version: 1.10.2
* VsCodeVim Version: v0.6.3
* OS: windows 10.0.14393

xconverge:
This is a funky one but o well, I think it is better here than somewhere else... What do you think?

### DIFF
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -5494,6 +5494,17 @@ class MoveToMatchingBracket extends BaseMovement {
       return result.getRight();
     }
   }
+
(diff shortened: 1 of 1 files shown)


johnfn:
Yeah this is exactly where it belongs. LGTM.

Issue #1356, opened 2017-03-03
Not respecting the cusor-style value


harryio:
### What did you do?
Changed custor style to line-thin in user settings.

### What did you expect to happen?
The cursor to be thinned down in insert mode.
### What happened instead?
Cursor width remained the same; upon disabling VSCodeVim entirely, the expected behaviour is exhibited by vscode.
### Technical details:

* VSCode Version: 1.10
* VsCodeVim Version: latest
* OS: Linux

xconverge:
fixed in master

PR #1399, opened 2017-03-15
Allow users to use their own cursor style for insert from editor.cursorStyle

xconverge:


src/configuration/configuration.ts
```diff
@@ -57,6 +57,21 @@ class ConfigurationClass {
...
+  }
+
+  private cursorStyleFromString(cursorStyle: string): vscode.TextEditorCursorStyle {
+    if (cursorStyle === 'line') {
+      return vscode.TextEditorCursorStyle.Line;
+    } else if (cursorStyle === 'block') {
+      return vscode.TextEditorCursorStyle.Block;
+    } else if (cursorStyle === 'underline') {
+      return vscode.TextEditorCursorStyle.Underline;
+    }
```
xconverge: Once we can upgrade to 1.10.0^ for the engine, I can add 2 more cases here for line-thin and underline-thin and it will work right away

xconverge:
@johnfn should be good to go. I tried to use vimState.editor when we had it available (in actions.ts for sure and some commands) this fixes #1356 and updates the minimum version of the latest extension to 1.10.0, now seems like an OK time due to some stability with the latest releases in my opinion

xconverge:
@johnfn can you take a peek at this when you get a chance

johnfn:
I was slightly intimidated by the sheer line count of changes, haha. It looks fine. Feel free to merge when you get to my one comment.

xconverge:
@johnfn might need to click finish review then ;) I dont see a comment

johnfn:
Grah, the number of times I've forgotten to click that button is too high.

src/configuration/configuration.ts
```diff
@@ -57,6 +57,28 @@ class ConfigurationClass {
...
       }
     }
+
+    // Get the cursor type from vscode
+    const cursorStyleString = vscode.workspace.getConfiguration().get("editor.cursorStyle") as string;
+    this.userCursor = this.cursorStyleFromString(cursorStyleString);
+  }
+
+  private cursorStyleFromString(cursorStyle: string): vscode.TextEditorCursorStyle {
+    if (cursorStyle === 'line') {
```
xconverge: I can, I actually copied this code from vscode since I was curious how they did it

johnfn: Hah really? But they don't expose that anywhere? Meh

xconverge: https://github.com/Microsoft/vscode/blob/fa2e9d173e5d5820843830c3da4d7f139e27a09d/src/vs/editor/common/config/commonEditorConfig.ts#L439

johnfn: Haha we should fix it and then submit it as a PR back to them 😁

johnfn: We should also refine their type `string` to be a union type amongst all the different cursor types... Haha, but you dont have to do that, we don't get much of an advantage from it unfortunately.

xconverge: Ok I think I changed it to what you meant...let me know

PR #1413, opened 2017-03-20
fix undo points when moving around in insert with mouse or arrow keys

xconverge:
I think you commented this out when you added multicursor...it is really a drag if you are an arrow/mouse user, so atleast here this will improve it for everyone except for multiCursor AND arrow/mouse users

but multicursor is already messed up if you use the mouse so I don't think we need to fix this now


johnfn:
Yeah this is fine.

Issue #1424, opened 2017-03-24
Allow users to delegate certain keys back to vscode


xconverge:
I am proposing a way to allow users to delegate certain key combinations back to vscode.

This will only include modifier key combos that we map.

I will add a setting like:

```
vim.handleKeys{
    <C-f>:false,
    <C-a>: false
}
```

The default for all of our mappings will be true.

This config option will be checked for every key when we go to register them.

In VsVim the screen looks like this:

![image](https://cloud.githubusercontent.com/assets/4219348/24305837/9af6d970-107b-11e7-9b9e-141fa3eb9999.png)

xconverge:
@johnfn @rebornix Any thoughts?

rebornix:
I really like the idea. Just my 2 cents here

1. About how we modify the `when` clause, as we are adding more to it, we may want to merge all vim related entries to a single one in a keybinding entry. For example
    ```
    {
                "key": "ctrl+k",
                "command": "extension.vim_ctrl+k",
                "when": "editorTextFocus && vim.useCtrlKeys && vim.mode != 'Insert' && !inDebugRepl"
    },
    ```

    Right now we are passing `useCtrlKeys` and `mode` back to Code's context and asking Code to do the math. But we still want to put `disableTemporarily` and your `handleKeys` in the table, so maybe we can merge them into a single `vim.CtrlK` boolean and update it when necessary.

2. One thing I'm not sure about if we want to let ppl know about `scope` as ppl may want to use `ctrl+c` in Insert Mode but not others. The VSVim options don't have a good way to specify the scope/vim-mode/etc, do they?

xconverge:
1. I think that is a great idea. I think my first commit will have it as a separate entry in "When". Then a followup commit can combine all of our logic and whens into a single vim.ctrlK
2. I think we can ignore scope for now, lets  just do an all or nothing

johnfn:
Love it, I just wish we had a better interface than JSON - like VSVim does.

xconverge:
oooooo well :)

johnfn:
What's important is: don't get rid of `useCtrlKeys`. Basically have useCtrlKeys flip all these to true, and then have these settings override individual keys.

xconverge:
no I am not getting rid of that option, I am however getting rid of that from the package.json though

johnfn:
Perfect. And yeah assuming VSCode ever rolls out some sort of gui thing, we'll be ready. :)

[...]

PR #1425, opened 2017-03-25
Allowing user to selectively disable some key combos

xconverge:
I like this. PR for #1424

example config option that may be handy on windows for some

```
 "vim.handleKeys":{
        "<C-a>": false,
        "<C-f>": false
    },
```

I would like to have it all in configuration.ts so that it can happen without a vscode restart so that will be next. Then trying to clean it up some more and make sure all of our keys in package.json are handled (only modifier key combos I think will be part of this)

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,22 +60,22 @@
             {
                 "key": "cmd+left",
                 "command": "extension.vim_cmd+left",
-                "when": "editorTextFocus && !inDebugRepl && vim.mode != 'Insert'"
+                "when": "editorTextFocus && vim.use<D-left> && !inDebugRepl && vim.mode != 'Insert'"
             },
             {
                 "key": "cmd+right",
                 "command": "extension.vim_cmd+right",
-                "when": "editorTextFocus && !inDebugRepl && vim.mode != 'Insert'"
+                "when": "editorTextFocus && vim.use<D-right> &&  !inDebugRepl && vim.mode != 'Insert'"
             },
             {
                 "key": "cmd+d",
                 "command": "extension.vim_cmd+d",
-                "when": "editorTextFocus && vim.use<D-d> !inDebugRepl"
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -119,107 +119,107 @@
...
                 "key": "ctrl+n",
                 "command": "extension.vim_ctrl+n",
-                "when": "suggestWidgetVisible && vim.useCtrlKeys"
+                "when": "suggestWidgetVisible && vim.use<C-n> vim.useCtrlKeys"
             },
             {
                 "key": "ctrl+p",
                 "command": "extension.vim_ctrl+p",
-                "when": "suggestWidgetVisible && vim.useCtrlKeys"
+                "when": "suggestWidgetVisible && vim.use<C-p> vim.useCtrlKeys"
```
xconverge: lol

package.json
```diff
@@ -119,107 +119,107 @@
...
                 "key": "ctrl+a",
                 "command": "extension.vim_ctrl+a",
-                "when": "editorTextFocus && vim.useCtrlKeys && !inDebugRepl"
+                "when": "editorTextFocus && vim.use<C-a> !inDebugRepl"
             },
             {
                 "key": "ctrl+n",
                 "command": "extension.vim_ctrl+n",
-                "when": "suggestWidgetVisible && vim.useCtrlKeys"
+                "when": "suggestWidgetVisible && vim.use<C-n> vim.useCtrlKeys"
```
xconverge: lol

johnfn:
This LGTM. It's really interesting that properties in package.json can include < and > characters.

xconverge:
Yea that felt a bit crazy, but if it works it works. I figured it wasn't any worse than vim.useCtrl+k etc, and then atleast the nomenclature is the same throughout our remappings etc

PR #1428, opened 2017-03-25
gUU and guu

### LINKED ISSUE #1423: gUU and guu doesn't work

-----

### What did you do?

Pressed gUU or guu keys in normal mode. The same for gUgU and gugu.

The whole line should be in uppercase (gUU) or in lowercase (guu).

guu

### What happened instead?

Nothing.

### Technical details:

* VSCode Version: 1.10.2
* VsCodeVim Version: 0.6.11
* OS: Windows 10 Pro

Thank you!

xconverge:
fixes #1423

### DIFF
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -2082,7 +2082,7 @@ export class UpperCaseOperator extends BaseOperator {
       await TextEditor.replace(range, text.toUpperCase());
 
       vimState.currentMode = ModeName.Normal;
-      vimState.cursorPosition = start;
+      vimState.cursorPosition = vimState.cursorPositionJustBeforeAnythingHappened[0];
 
       return vimState;
     }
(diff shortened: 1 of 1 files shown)


src/actions/actions.ts
```diff
@@ -2082,7 +2082,7 @@ export class UpperCaseOperator extends BaseOperator {
       await TextEditor.replace(range, text.toUpperCase());
 
       vimState.currentMode = ModeName.Normal;
-      vimState.cursorPosition = start;
+      vimState.cursorPosition = vimState.cursorPositionJustBeforeAnythingHappened[0];
```
johnfn: Why is this necessary over using start? (Also the [0] won't be multi cursor friendly like start and end are)

src/actions/actions.ts
```diff
@@ -5132,6 +5133,32 @@ class MoveFormat extends BaseMovement {
 }
 
 @RegisterAction
+class MoveUpercaseLine extends BaseMovement {
```
johnfn: Can you leave a comment here saying that these are for guu and gUU (for searchability mostly)

src/actions/actions.ts
xconverge: this entire command is already not multicursor friendly though

start in the case of gUU is start of line and end is end of line

gUU is not supposed to move the cursor so it needs to perform the action and remain in place

johnfn: gUU looks like it just moves the cursor to the beginning of the line or am I wrong?

xconverge: you are right...I am so confused, I wonder if it does the same thing in gvim, I tried yesterday...I will change to start, thats easy

johnfn: haha no worries

src/actions/actions.ts
```diff
@@ -5131,6 +5132,34 @@ class MoveFormat extends BaseMovement {
   }
 }
 
+// Used for gUU
```
johnfn: It'd probably be good to move these next to the other weirdo motions I made for dd yy cc etc.

xconverge: they are....

johnfn: oh, nice! i didn't hit that little fold out button enough.

xconverge: yea I think the > and = are also of the same weirdo type

PR #1431, opened 2017-03-25
Change easymotion decoration colors to use searchHighlight colors

xconverge:
I can get the same result without the library if we want, but it definitely cleans up the code

Also, I spent a while using different functions of it to manipulate the color wheel to try out different things for readability. Complementary font color, etc don't look good.

Black and white and using the light()/dark() luminosity formula seemed to give the best results.

### DIFF
--- a/src/easymotion/easymotion.ts
+++ b/src/easymotion/easymotion.ts
@@ -127,26 +127,24 @@ export class EasyMotion {
   /**
    * Create and cache the SVG data URI for different marker codes and colors
    */
-  private static getSvgDataUri(code: string, backgroundColor: string, font: string, fontColor: string): vscode.Uri {
-      // var cache = this.svgCache[code];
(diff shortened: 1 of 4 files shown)


xconverge:
fixes #1431

src/easymotion/easymotion.ts
```diff
@@ -294,8 +324,6 @@ export class EasyMotion {
       // Get keys after the depth we're at
       var keystroke = marker.name.substr(this.accumulation.length);
 
-      let fontColor = keystroke.length > 1 ? "orange" : "red";
```
xconverge: I am disappointed to take this out but am not sure how else I could accomplish it

xconverge: decided to leave it in, and make fontcolor and background color change with a configuration flag instead

PR #1433, opened 2017-03-26
fixes #1312

### LINKED ISSUE #1312: This extension completely breaks the `Select All Occurrences` command

-----
### What did you do? (Version 1)
`test testit testworks testproperly`
In insert mode selected `test` with the mouse, either click on `Select All Occurrences`  in the menu bar or use the shortcut(`cmd-shift-l` then press backspace. 

### What did you do? (Version 2)
`test testit testworks testproperly`
In insert mode place the cursor at the start of the line, then either click on `Select All Occurrences` in the menu bar or use the shortcut then press backspace. 

`| |it |works |properly`
Cursors are represented as pipes.

### What happened instead?
Undefined (behaves inconsistently and unpredictably random)
All the following have occurred:
`|s test|properly`
`test|works test|properly`
`|works test|properly`
`| test|it test|works test|properly`
`test| test|it test|works test|properly`

However unlike `Add Next Occurrence`, which is also broken, cursors other than the first may or might not be bugged.

### Technical details:

* VSCode Version: 1.9.1
* VsCodeVim Version: 0.5.3
* OS: MacOS 10.12

Needless to say, all of the above work properly and consistently with the extension disabled. 

I am aware that it also does not work in visual mode. But that is a part of this extension and it world require additional implementation to get it to work. 

I am pointing out this issue because an extension should not break functionality present in the stand-alone base product. (BTW I am aware of the `gc` command, and it is one of of the friendliest multi-cursor implementations for modal-editing that I have encountered so far. Unfortunately, it does not do everything that I need it to do).

xconverge:
pretty scared of this function still...

### DIFF
--- a/src/mode/modeHandler.ts
+++ b/src/mode/modeHandler.ts
@@ -602,6 +602,16 @@ export class ModeHandler implements vscode.Disposable {
       return;
     }
 
+    if (e.selections.length !== this.vimState.allCursors.length) {
+      // Number of selections changed, make sure we know about all of them still
+      this.vimState.allCursors = vscode.window.activeTextEditor!.selections.map(x =>
+        new Range(Position.FromVSCodePosition(x.start), Position.FromVSCodePosition(x.end)));
+
+      await this.updateView(this.vimState);
+
+      return;
+    }
+
     if (this.vimState.isMultiCursor) {
       // AAAAAARGGHHHHH - johnfn
 


johnfn:
Could it... could it really be that easy?

Does this also fix other instances of us not handling multi cursor correctly?

That would be awesome!

xconverge:
I am not sure it gets us all the way, but from my initial testing it gets us a lot closer... and it makes 100% sense

Issue #1167, opened 2016-12-16
Cmd+v to paste in normal mode does not create an undo point


xconverge:
### What did you do?

Paste text with cmd+v, enter insert and type something, press u

Cmd+v i 123 esc u

### What did you expect to happen?

Only undo 123, then u again would undo what was pasted

### What happened instead?

The pasted text and inserted text are all undone at once

### Technical details:

* VSCode Version: 1.8.0
* VsCodeVim Version: 0.4.9
* OS: macOS

sophiebits:
Seeing this too. I was hoping I could map `Cmd-V` to `Cmd-V Ctrl-G u` to make a new undo block (http://vimdoc.sourceforge.net/htmldoc/insert.html#i_CTRL-G_u) but it doesn't seem like `Ctrl-G u` is implemented in this plugin (at least, it opens a "Type a line number … to navigate to".

PR #1435, opened 2017-03-26
fix undo points for using arrow keys in insert mode

xconverge:
@johnfn this is really really hacky

that being said it seems to work

Any suggestions on how I can clean it up?

This handles both single and multicursor.

|test
|test
|test  

 Enter insert mode, type something, then use arrowright a few times and type some more. Esc, then undo, and undo. It separates the 2 places you entered stuff

The version in modehandler works the same way, but instead of using<right> click with the mouse somewhere to the right (this does not work in multicursor as well, but isn't too bad)

### DIFF
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -550,6 +550,13 @@ export class CommandInsertInInsertMode extends BaseCommand {
     const char = this.keysPressed[this.keysPressed.length - 1];
     const line = TextEditor.getLineAt(position).text;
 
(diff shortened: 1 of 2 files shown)


xconverge:
this works really well, I just don't like how it is done since I need to check the previous action was arrows AND that its the first key of the current action (so that an undo point is not created at every step in the current action after arrow keys are used)

xconverge:
it seems like there should be something I can put in modehandler to address this AND #1167 at the same time...

Issue #1300, opened 2017-02-15
cit for embedded html tag


octref:
Please *thumbs-up* 👍 this issue if it personally affects you! You can do this by clicking on the emoji-face on the top right of this post. Issues with more thumbs-up will be prioritized.

-----

### What did you do?

```
    <div>
      |
      <div></div>
    </div>
```

`cit`

### What did you expect to happen?

```
    <div>|</div>
```
  

### What happened instead?

Nothing

### Technical details:

* VSCode Version: 1.9.1
* VsCodeVim Version: 0.5.3
* OS: OS X 10.11.6

xconverge:
think this is a dupe of #1108

xlaech:
There are still some things to improve
e.g.:
- The search regex for the tags
- an option for including the tag or not (@xconverge: where is this used?)

I also found another bug which is not caused by the it movement but by the Changecommand "c"
When the end of the selection (cit or visual mode) is at the beginning of line, it will delete the first character of this line.
=> I open a new Issue and fix it in another branch

johnfn:
This has been fixed and will be out in the next version of VSCodeVim. :)

johnfn:
_Now_ this has been fixed and will be released in the next version of VSCodeVim.

Issue #1232, opened 2017-01-19
tag movement commands are single line only


jimmymain:
### What did you do?
Attempted to yank/copy/delete inside a tag. 
```
<outer>
    <tag>data</tag>
</outer>
```
keys pressed: `dit`

### What did you expect to happen?
when the cursor is placed at the closing `>` of `<outer>` in gvim (line 1) the contents of outer are completely removed. In other words the tag movement commands in gvim are **multi line**.

It would appear that the tag commands in vsvim are always single line only.

### What happened instead?
nothing.

### Technical details:

* VSCode Version: 1.8.1
* VsCodeVim Version: 0.4.10
* OS: Windows 10 x64

johnfn:
This will be fixed in our next release!

PR #1454, opened 2017-03-31
fix multi line in 'at' and 'it' commands

### LINKED ISSUE #971: Deleting or changing multi-line tag blocks does not work

jrenton:
Fixes #971, #1108, #1232, and #1300.

Remaining existing issues: `cit` does not enter insert mode when tag is empty, and it also doesn't return to the correct position when you run `cit` in the closing tag.

### DIFF
--- a/src/matching/tagMatcher.ts
+++ b/src/matching/tagMatcher.ts
@@ -1,8 +1,8 @@
 export class TagMatcher {
-  static openTag = "<";
-  static closeTag = ">";
-  static closeSlash = "/";
-  static tagNameRegex = /[^\s>]+/;
+  static TAG_REGEX = /\<(\/)?([^\>\<\s]+)\s*(\/?)\s*\>/g;
+  static OPEN_FORWARD_SLASH = 1;
+  static TAG_NAME = 2;
+  static CLOSE_FORWARD_SLASH = 3;
 
   openStart: number|undefined;
   openEnd: number|undefined;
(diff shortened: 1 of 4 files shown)


src/matching/tagMatcher.ts
```diff
@@ -1,50 +1,88 @@
...
-    const close = corpus.indexOf(TagMatcher.closeTag, opening + 1);
-    const tagName = tagNameMatch[0];
 
-    // We now know where the opening tag is
-    this.openStart = opening;
-    this.openEnd = close + 1;
+    const stack : any = [];
+    const matchedTags : any = [];
+
+    tags.forEach((tag) => {
```
johnfn: Can you just use a simple `for of` loop here?

src/matching/tagMatcher.ts
```diff
@@ -1,50 +1,88 @@
...
+
+      match = TagMatcher.TAG_REGEX.exec(corpus);
     }
-    const close = corpus.indexOf(TagMatcher.closeTag, opening + 1);
-    const tagName = tagNameMatch[0];
 
-    // We now know where the opening tag is
-    this.openStart = opening;
-    this.openEnd = close + 1;
+    const stack : any = [];
```
johnfn: Can you give the proper typings here, rather than using `any`? I think it's `{ name: string, type: "close" | "open", startPos: number, endPos: number }[]`

johnfn: (And our code formatting conventions are `const stack: type`)

johnfn:
Overall looks good to me. My biggest concern is that we have to parse the entire document every time we do a cit. How is the perf of that in a file such as actions.ts? (And you thought that file was huge for no reason. Of course not, it's to test perf concerns! 😉 )

jrenton:
@johnfn Hah... The performance was a concern of mine too, but I did some tests in very large html files (170,000+ characters) and saw no noticeable delay.

xconverge:
@jrenton I think there was one other comment about using a for of loop, it is in our coding style guide

This is so close that I would like to see it in the next release!

theamazingfedex:
Is there anything that I could do to help expedite this fix getting merged?

jrenton:
@theamazingfedex There are some underlying bugs within VSCodeVim itself outside of this issue relating to assumptions made in regards to changing/deleting text (at least the last I checked). They show themselves through various bugs that still exist with this, including:
- `cit` does not enter insert mode when tag is empty
- When you run `cit` in the closing tag, it doesn't return to the correct position.
- If a closing tag is the first character on a newline, it cannot properly close the tag.

Technically this should be able to be merged, but I kind of gave up when I kept running into bugs that were related to underlying issues with the plugin itself.

Chillee:
So just to be clear, the underlying algorithm for finding the corresponding tags works fine, the issues are mainly related to where the cursor ends up after?

jrenton:
@Chillee Yes

Chillee:
Do you think they're fixable (just a massive pain), or is there some fundamental limitation with our current code structure?

johnfn:
There aren't any fundamental limitations to getting tag text objects to work. It's just more work. :)

PR #1465, opened 2017-04-04
Fix type suggestion for handleKeys object

abhiranjankumar00:
It warns that array is expected, though object is the correct value.

Sorry, didn't run the tests. Is short of time.

2017-04-04-180111_5206x1080


xconverge:
Thanks!

PR #1487, opened 2017-04-09
[WIP] change system clipboard library to a newer more maintained library

xconverge:
This library is being worked on consistently and is not stale with potential issues

Still having issues with the character • which I use for testing. Once it works perfectly with this I will also test chinese characters

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,6 @@ os:
 addons:
   apt:
     packages:
-    - xsel
 
 language: node_js
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ os:
 addons:
   apt:
     packages:
-    - xclip
+    - xsel
 
 language: node_js
 
--- a/README.md
+++ b/README.md
@@ -183,7 +183,7 @@ Or bind `<leader>w` to save the current file:
 #### useSystemClipboard
   * Enable yanking to the system clipboard by default
   * Type: Boolean (Default: `false`)
-  * Note: Linux users must have xclip installed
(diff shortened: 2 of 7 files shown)


xconverge:
Tested both this and current master (node-copy-paste) library on Ubuntu and all Chinese characters seem to work great. This new library is also cool because it worked out of the box, no need to install xclip. Back to macOS to continue to try and figure this out...

xconverge:
@rebornix this is starting to look great, but I wonder what should be fixed in the scope of this plugin, the copy paste plugin, or vscode

Since you are the copy paste formatting expert for vscode perhaps you can help me out here

Can you test this a bit? The key is mostly setting the LANG env variable for vscode to a UTF-8 compatible locale...

rebornix:
@xconverge sure, I'm a Chinese expert and I'll give it a try.

xconverge:
arg I can't get it to work for windows 10 :(

rebornix:
@xconverge copying Chinese characters works good.

xconverge:
Too bad it doesn't work on windows yet :/

xconverge:
I opened https://github.com/sindresorhus/clipboardy/issues/11 we will see if it gets any traction

xconverge:
I am going to merge this for now actually since it FIXES the problem on osx and linux, uses a better supported library, and does not make windows worse (doesn't work correctly on master with old library either)

octref:
@xconverge Hey wondering if Windows clipboard is still a standing issue? I want to push for https://github.com/Microsoft/vscode/issues/217.

Issue #1481, opened 2017-04-06
Is there any way to remap the easymotion keys?


edasaki:
In vim I have the "Search Character" set to just s since I never really use the actual s in insert mode.

Is it possible to change the binding in vscodevim? I use the search a lot but pressing the leader twice is a bit tiring sometimes.

And kind of vaguely related, is it possible to change the red/orange font colors? I want to keep the black background but the red/orange is a bit hard to read for me.

johnfn:
> And kind of vaguely related, is it possible to change the red/orange font colors? I want to keep the black background but the red/orange is a bit hard to read for me.

It sounds like you're talking about theming VSCode, which is totally separate from VSCodeVim! Check here https://code.visualstudio.com/docs/getstarted/themes

edasaki:
Oh sorry I realized that totally wasn't clear haha.

I was talking about the popups from the easymotion plugin

![](http://i.imgur.com/nOGu0kA.png)

Is that also handled by VSCode themes? Since it seems like it's something that would be decided by vscodevim

johnfn:
OHH, those themes! My bad! Yeah huh.. those should probably be re-themeable, sure.

To be honest, the unfortunate truth about extensions (e.g. EasyMotion) is that they're mostly community supported, as it's just too much effort for me to maintain both Vim and an extension ecosystem. You _may_ be able to convince @xconverge if you ask really nicely, but that guy is pretty busy too haha. :P

xconverge:
@edasaki we DO allow you to change the color

from here https://github.com/VSCodeVim/Vim#how-to-use-easymotion

"If you set vim.easymotionChangeBackgroundColor to true you can use the searchHightlightColor as the background color for the text decorations, however you then lose the red/orange indicators on whether it is a one key or two key combination since the font color needs to stay readable."

xconverge:
I analyze the colors and then set the font to either white or black based on the search highlight color you use

FelikZ:
> Is it possible to change the binding in vscodevim?

```json
{
        "vim.otherModesKeyBindingsNonRecursive": [
        {
            "before": ["s"],
            "after": ["leader", "leader", "s"]
        }]
}
```

edasaki:
Thank you so much @FelikZ , that worked!

And thank you too @xconverge! That's fine for me

xconverge:
There is 1 related bug that is somewhat funny though :) #1426, feel free to thumbs up it if it impacts you

Thanks

[...]

PR #1493, opened 2017-04-11
Added new style settings (color, size, etc.) for easymotion markers

edasaki:
Made this for my own issue :P #1481 

Here's an example of the settings in use:

```
    "vim.easymotionMarkerBackgroundColor": "#3C6478",
    "vim.easymotionMarkerForegroundColorOneChar": "white",
    "vim.easymotionMarkerForegroundColorTwoChar": "white",
    "vim.easymotionMarkerHeight": 20,
    "vim.easymotionMarkerYOffset": 12,
    "vim.easymotionMarkerWidthPerChar": 10,
```
**I removed the existing easymotion background color option since I feel this encompasses that, but feel free to add it back if you think it's better to keep it!**

Here are some examples of what the styling can look like: (p.s. I happened to use white and black for the font colors but they can be anything, like yellow text on dark blue bg or something)

### DIFF
--- a/src/easymotion/easymotion.ts
+++ b/src/easymotion/easymotion.ts
@@ -27,7 +27,11 @@ export class EasyMotion {
    */
   private static decorationTypeCache: vscode.TextEditorDecorationType[] = [];
   private static svgCache: { [code: string]: vscode.Uri } = {};
-  private static setBackgroundColor: string = "";
+  private static cachedBackgroundColor: string = "";
+  private static cachedOneFontColor: string = "";
+  private static cachedTwoFontColor: string = "";
+  private static cachedWidthPerChar: number = -1;
+  private static cachedHeight: number = -1;
 
   /**
    * The key sequence for marker name generation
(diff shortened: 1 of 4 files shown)


src/easymotion/easymotion.ts
```diff
@@ -299,18 +301,12 @@ export class EasyMotion {
...
-    const fontFamily = vscode.workspace.getConfiguration().get("editor.fontFamily") as string;
-    const fontSize = vscode.workspace.getConfiguration().get("editor.fontSize") as string;
-    const fontWeight = vscode.workspace.getConfiguration().get("editor.fontWeight") as string;
+    const fontFamily = Configuration.easymotionMarkerFontFamily;
+    const fontSize = Configuration.easymotionMarkerFontSize;
+    const fontWeight = Configuration.easymotionMarkerFontWeight;
 
     // Compute font color based on background (remove opacity)
-    var Color = require('color');
-    let backgroundColor = Color.rgb(Configuration.searchHighlightColor).alpha(1.0);
```
xconverge: Can you remove this dependency from package.json since this was the only place it was used

src/easymotion/easymotion.ts
```diff
@@ -61,41 +61,41 @@ export class EasyMotion {
     var totalSteps = 0;
 
     if (length >= keyTable.length) {
-        var totalRemainder = Math.max(length - keyTable.length, 0);
-        totalSteps = Math.floor(totalRemainder / keyTable.length);
+      var totalRemainder = Math.max(length - keyTable.length, 0);
```
xconverge: I noticed all the formatting was wrong before last week, I imagine you just used the formatter on these files right?

src/easymotion/easymotion.ts
```diff
@@ -131,33 +131,35 @@ export class EasyMotion {
   private static getSvgDataUri(code: string, backgroundColor: string, fontFamily: string, fontColor: string,
     fontSize: string, fontWeight: string): vscode.Uri {
 
-      // Clear cache if the backgroundColor has changed
-      if (this.setBackgroundColor !== backgroundColor) {
-        this.svgCache = {};
-        this.setBackgroundColor = backgroundColor;
-      }
+    // Clear cache if the backgroundColor has changed
+    if (this.setBackgroundColor !== backgroundColor) {
```
xconverge: Not a huge deal but the cached resources will be out of date whenever you change a setting BESIDES the background color, so tuning the colors is a bit complicated. A clean restart of vscode handles this. It might be a good idea to check some other things here and clear the cache in the case of any main changes now that they are all configurable at runtime

src/easymotion/easymotion.ts
```diff
@@ -299,18 +301,12 @@ export class EasyMotion {
...
 
     // Compute font color based on background (remove opacity)
-    var Color = require('color');
-    let backgroundColor = Color.rgb(Configuration.searchHighlightColor).alpha(1.0);
-
-    let fontColor = "white";
-    if (backgroundColor.light()) {
-      fontColor = "black";
-    }
+    // var Color = require('color');
```
xconverge: remove

src/easymotion/easymotion.ts
```diff
@@ -328,10 +324,10 @@ export class EasyMotion {
         this.decorations[keystroke.length] = [];
       }
 
-      if (!Configuration.easymotionChangeBackgroundColor) {
-        fontColor = keystroke.length > 1 ? "orange" : "red";
-        backgroundColor = Color('black');
-      }
+      let fontColor = keystroke.length > 1 ?
```
xconverge: const

src/easymotion/easymotion.ts
```diff
@@ -328,10 +324,10 @@ export class EasyMotion {
...
+        Configuration.easymotionMarkerForegroundColorTwoChar
+        : Configuration.easymotionMarkerForegroundColorOneChar;
+      let backgroundColor = Configuration.easymotionMarkerBackgroundColor;
```
xconverge: const

src/easymotion/easymotion.ts
edasaki: Yeah I think I have my vscode set to auto format on save

src/easymotion/easymotion.ts
edasaki: yeah I noticed that but didn't realize it was b/c of cache, I'll see if I can fix it

xconverge: should just be able to add a few more checks right here

edasaki:
I think it should be mostly fixed now, I tested a bit and it seems to update fine now without restarting.

xconverge:
Awesome @edasaki thanks a lot! You will see it in the next release :)

PR #1495, opened 2017-04-11
Add 5i= or 4a- so that the previously inserted text is repeated upon exiting to normal mode

### LINKED ISSUE #1122: repeat inserted character/word (e.g., 5i=[esc]) not supported

--------

### What did you do?

I pressed `5i=[esc]`

what to get =====, what to do a quick input for reStructuredText title tag. XD
### What happened instead?
Nothing
### Technical details:

* VSCode Version: 1.7.2
* VsCodeVim Version: 0.4.8
* OS: Windows 8.1

xconverge:
Fixes #1122

@johnfn what do you think, looking for any ways to clean this up a bit

### DIFF
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -964,13 +964,12 @@ class CommandEscInsertMode extends BaseCommand {
 
     // If we wanted to repeat this insert, now is the time to do it. Insert
     // count amount of these strings before returning back to normal mode
-    if (vimState.insertRepeatCount > 1) {
-      for (let i = 0; i < vimState.insertRepeatCount - 1; i++) {
+    if (vimState.recordedState.count > 1) {
+      for (let i = 0; i < vimState.recordedState.count - 1; i++) {
         vimState.recordedState.transformations.push({
           type: "dot"
         });
       }
-      vimState.insertRepeatCount = 0;
     }
 
     if (vimState.historyTracker.currentContentChanges.length > 0) {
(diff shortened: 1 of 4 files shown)


src/actions/actions.ts
```diff
@@ -962,6 +962,16 @@ class CommandEscInsertMode extends BaseCommand {
 
     vimState.currentMode = ModeName.Normal;
 
+    // If we wanted to repeat this insert, now is the time to do it. Insert
+    // count amount of these strings before returning back to normal mode
+    if (vimState.recordedState.count > 1) {
+      for (let i = 0; i < vimState.recordedState.count - 1; i++) {
+        vimState.recordedState.transformations.push({
```
xconverge: @rebornix instead of me trying to run "dot" here, is there a way to just push the recent content change multiple times?

rebornix: IMHO `vimState.historyTracker.currentContentChanges` is what you need. Each time we enter Insert Mode, `vimState.historyTracker.currentContentChanges` will be refreshed so here at this line, you can try to duplicate this array and ask the transformations to apply the content changes for you, similar to 

```
      type: "macro",
      register: register,
      replay: "contentChange"
    });
```

But I'm afraid we need a new transformation type as right now I don't think we can replay contentChange directly

rebornix: The only thing we need to think about if it confuses our Macro.

xconverge:
@johnfn @rebornix ah crap so we actually do weird stuff with insert mode content changes, we clear out historyTracker.currentContentChange and manage them on our own elsewhere into the recordedstate action run

I got this working a bit, am I on the right track?

xconverge:
I give up for fixing the cursor position for now, I can't find a clean place to put it.

I need "if final content change for now was complete, move left 1 char", yikes

rebornix:
Touching the contentChange part is the proper way to make sure both this command and macro work at this moment. `contentChange` is the major part that we differ from Vim and maybe I should write some doc about it.

Issue #1375, opened 2017-03-09
It would be really awesome if c-w h and c-w l would navigate between other panes


johnfn:
Like if I could navigate to search results, or the terminal in the bottom.

mightyiam:
Because that's how it works in vim. And I don't always wanna have to know the specific key binding of jumping to a specific thing.

xconverge:
https://code.visualstudio.com/updates/v1_11#_navigate-between-views

mightyiam:
Perhaps assign something reasonable to these by default?

PR #1504, opened 2017-04-13
Navigate between view

lyup:
Ref #1375. Leverage new Code api to navigate between viewlets, panels, editors by <Ctrl>-<hjkl>.

This is contributed by rebornix/Peng's personal account on his personal machine in personal time.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "url": "https://github.com/VSCodeVim/Vim/issues"
   },
   "engines": {
-    "vscode": "^1.10.0"
+    "vscode": "^1.11.0"
   },
   "categories": [
     "Other",
--- a/package.json
+++ b/package.json
@@ -195,6 +195,31 @@
         "when": "editorTextFocus && vim.active && vim.use<C-[> && !inDebugRepl"
       },
       {
+        "key": "ctrl+w h",
+        "command": "extension.vim_navigateLeft",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -187,6 +187,31 @@
                 "when": "editorTextFocus && vim.use<C-[> && !inDebugRepl"
             },
             {
+                "key": "ctrl+w h",
+                "command": "extension.vim_navigateLeft",
+                "when": "vim.use<C-w>"
```
xconverge: these are all for when !editorTextFocus right? we should add !editorTextFocus

Because then the actions.ts ones cover when we are focused?

johnfn:
This is awesome. Would love to see it land.

xconverge:
I didn't even read/see who you were :)

mightyiam:
This will save my life.

johnfn:
I am not a huge fan of defining keybindings purely in package.json because then we can't remap them. :/

rebornix:
@johnfn I see your points. Let's see if `vim.remap` command help in this case, if not, I'll tweak the code accordingly :)

Chillee:
I don't think this handles ctrl+w j and ctrl+w k right?

yielder:
@rebornix Any reason why only ctrl+w h/l is supported?
ctrl+w j/k would be super useful when move focus between terminal and editor pane

charliesbox:
`ctrl-w w` should be part of this enhancement too

PR #1506, opened 2017-04-13
Add line comment operator

fiedler:
This PR adds functionality similar to tpope's [vim-commentary](https://github.com/tpope/vim-commentary). As it didn't require a lot of work I thought I would just create a PR instead of opening an issue / feature request.

It can toggle line comment the current line using `gbb` or toggle comments on the target of a motion i.e. `gb5j` to toggle line comments on current line and the following next four lines or `gbap` to toggle line comments in a paragraph etc.

It uses the built-in _Toggle Line Comment_ functionality of VSCode to determine how to apply/unapply comments.

A few questions:
- I would have liked to use `gc` as the original Vim plugin but since this is already used in this extension I settled for `gb`. Thought on this?
- Since VSCode's _Toggle Line Comment_ relies on the current editors language mode to determine how to apply comments I could not find an good way to add tests for this as it does nothing in Plain Text mode. Any advice on the best way to add a test case that runs under another language mode (for example JavaScript)?

Let me know what you think. Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,7 @@ VSCodeVim is a [Visual Studio Code](https://code.visualstudio.com/) extension th
 * Multi-cursor support. Allows multiple simultaneous cursors to receive Vim commands (e.g. allows `/` search, each cursor has independent clipboards, etc.).
 * The [EasyMotion plugin](#how-to-use-easymotion)!
 * The [Surround.vim plugin](#how-to-use-surround)!
+* The [Commentary plugin](#how-to-use-commentary).
 * And much more! Refer to the [roadmap](ROADMAP.md) or everything we support.
 
 Please [report missing features/bugs on GitHub](https://github.com/VSCodeVim/Vim/issues), which will help us get to them faster.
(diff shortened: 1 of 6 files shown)


xconverge:
You could then remap gc to send gb in your settings

Also if we allow remapping particular actions in the future this would be solved, you. Could map gc to CommentLine action

fiedler:
Remapping gc to send gb sounds good to me! Tried it out and it works like you'd expect.

Are you saying that this PR wasn't needed if we could remap to actions? I guess there should need to be some way to handle selection based on motion and cursor position after the action for that to work out...

If you think we should move forward with this PR I can try to look in to how to add a few tests. My initial thoughts are to add the tests to `test/mode/modeNormal.test.ts` and find a way to set the languageId on the textEditor for these specific tests.

xconverge:
No that is not what I was saying, I was saying now that the action atleast exists in the codebase, later on it can be explicitly remapped.

instead of ```before: gc after: gb``` it will be ```before: gc after: CommentOperator ```

I am all for moving forward with this, it has been asked for quite a few times, I would even be interested in you making a CommentBlockOperator using the other native vscode command for commenting blocks. These 2 actions would be very useful to have in the code base, how we remap them and trigger them is secondary as long as there is A WAY right now

fiedler:
Ah I see what you mean now.

Great! I've added some tests for the comment operator. There is no API to change the language of an editor so instead I decided to change the `setupWorkspace` function to accept a file extension for the file created for a test case.

I'd be happy to create the `CommentBlockOperator` as well. I won't have time the next couple of days but I'll get to it some time next week.

xconverge:
I am a bit torn on the gc/gb thing. Anyone who uses this will want gc :/ @johnfn Any ideas here?

johnfn:
Yeah, kind of a namespacing bummer there. 

I suppose we could have an option. By default, let's put this on gb as to not break any existing workflows. Later we can add an option to say which addon gets gc.

johnfn:
I'd say if you can add a little blurb about this into the README, and maybe include your remapper, we should be good to go here.

fiedler:
Yeah it's a bit unfortunate that `gc` was already taken here...

I've added the `BlockCommentOperator` as well as a small section in README.md that describes the usage and how to remap if you want to.

Please take another look at this and let me know if you want me to change anything.

Thanks!

xconverge:
This is really awesome, thanks a lot @fiedler

Chillee:
@fiedler It kinda seems like the features you implemented align more with http://www.vim.org/scripts/script.php?script_id=1173.

Is that right?

fiedler:
Hmm. It's been a while since I used Vim so it might actually have been TComment. I did this implementation based on the behaviour I am use to in other editors. So it is likely that it is indeed a mix of multiple implementations.

Chillee:
Well, it doesn't too much :P Commentary is the most popular plugin (and it's also super simple), so I don't see any problem with that.

Also, in a PR I merged recently, I swapped `gc` and `gb`, so the comment operator has been restored to its true form :)

fiedler:
Nice 🎉

Issue #1513, opened 2017-04-15
Deleting on a whitespace only line


Chillee:
* Click *thumbs-up* 👍 on this issue if you want it!
* Click *confused* 😕 on this issue if not having it makes VSCodeVim unusable. 

The VSCodeVim team prioritizes issues based on reaction count. 

--------

### What did you do?

http://i.imgur.com/JlmuwMM.jpg

With a setup like this, move your cursor to the middle of the top line. Move down a line, enter insert mode, and then press backspace.

### What did you expect to happen?
Delete characters before the cursor.

### What happened instead?
Deletes characters after the cursor.

Note: I've taken a look at the code and I know how to fix this.

### Technical details:

* VSCode Version: 1.11.2
* VsCodeVim Version: 0.6.15
* OS: Arch Linux

Chillee:
Fixed my own bug.

PR #1514, opened 2017-04-15
[WIP] Fixed issue #1513

Chillee:
So I fixed issue #1513. This issue only occurs when Configuration.expandTab is turned on. 
The main behavioral decision I made is that if you press backspace on a whitespace only line, it deletes both to the next "tab" as well as all the whitespace after it.

I'm unsure about a couple of things

1. I'd like to create tests, but from looking through the repo, I don't see how to set Configuration.expandTab to be true while running these tests.

2. When running the tests locally,  I get 56 tests failing, with or without my change. I don't really know what I'm supposed to do there (I thought I would try running the tests on CI, and cool! They passed.).

Cheers,
Horace

### DIFF
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -564,11 +564,11 @@ export class CommandInsertInInsertMode extends BaseCommand {
           // the next lowest level of indentation.
 
           const tabSize = vimState.editor.options.tabSize as number;
-          const desiredLineLength = Math.floor((line.length - 1) / tabSize) * tabSize;
+          const desiredLineLength = Math.floor((position.character - 1) / tabSize) * tabSize;
 
           vimState.recordedState.transformations.push({
             type: "deleteRange",
-            range: new Range(new Position(position.line, desiredLineLength), position)
+            range: new Range(new Position(position.line, desiredLineLength), new Position(position.line, line.length))
           });
         } else {
           vimState.recordedState.transformations.push({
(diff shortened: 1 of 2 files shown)


xconverge:
Hmm yea we are supposed to be ignoring remappings etc when running tests but maybe that is not working so some of your remappings are impacting the tests.

This would be tough to write a test for unless you set the parameter manually and use the old style tests.

Like this one https://github.com/VSCodeVim/Vim/blob/master/test/mode/modeInsert.test.ts#L118

only set Configuration.expandtab = true as the first thing the test does

Chillee:
I can't figure out how to set the tests, but it seems like it's already the necessary setting by default, right?

https://github.com/VSCodeVim/Vim/blob/84d209fcaf4a948ba322a59fcebdd404128eaf08/test/testUtils.ts#L86

xconverge:
ah yea looks like it is set for this case already you are right

Chillee:
I have added two new tests, both of which test the new behavior. However, I was running into some issues, and I believe they are related to issues with the test runner.

If you check the new continuous integration, there are 2 new tests, one of which is the other test but with an extra (non-blank) line. The first test works as expected, but the second test fails. I am unable to reproduce the second test's behavior in a non-test environment, and when viewing the running of the test, it seems to display some erratic behavior with regards to the cursor. I suspect this is a bug with the test runner.

PR #1515, opened 2017-04-15
make surround repeatable with dot

### LINKED ISSUE #1244: surround should be repeatable with .

--------

### What did you do?

`ysw)` then went to another word and pressed `.`

The next word should have be surrounded with parentheses

### What happened instead?

Nothing

### Technical details:

* VSCode Version: 1.8.1
* VsCodeVim Version: 0.5.2
* OS: OSX 10.10.5 / Windows 10

xconverge:
A little funky but I don't think its too bad :)

This is to fix #1244

Any suggestions on a better way would be great... @johnfn

### DIFF
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -6714,6 +6714,11 @@ class CommandSurroundModeStart extends BaseCommand {
 
     if (!operatorString) { return vimState; }
 
+    // Start to record the keys to store for playback of surround using dot
+    vimState.recordedState.surroundKeys.push(vimState.keyHistory[vimState.keyHistory.length - 2]);
+    vimState.recordedState.surroundKeys.push("s");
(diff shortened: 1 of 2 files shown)


johnfn:
I think there should be a better way. Just hold on to this guy: https://github.com/VSCodeVim/Vim/blob/master/src/mode/modeHandler.ts#L105

and then re-execute with him when the user presses `.`. That should work, right?

xconverge:
No the problem is getting the operators and movement for the range. It needs to be recalculated based on new position

xconverge:
Example: ysaw" we need to redo aw

johnfn:
Ah I see. So you'd have to refactor him to change the Range into a Motion or something. Meh, this is fine if you don't want to do that.

xconverge:
it is SUPER handy to have, that's for sure... I am not sure I know how to do the motion and either way there is some sense of replaying it back, but then atleast the surround object is fully contained

PR #1521, opened 2017-04-18
fixes #1519

### LINKED ISSUE #1519: Repeated `s` behaves incorrectly

--------

### What did you do?

`0iHello<Esc>04sfoo<Esc>`

`Hello` is typed. Cursor moves back to column zero.

The first four characters are replaced by `foo`, leaving `fooo`. This is what Vim does.

### What happened instead?

The first four characters are replaced by `foo` _four times_, leaving `foofoofoofooo`.

### Technical details:

* VSCode Version:

1.12.0-insider

* VsCodeVim Version: *[please ensure you are on the latest]*

0.6.16.

* OS: 

Mac 10.12.4.

xconverge:
is this better or worse than adding another flag type @johnfn?

### DIFF
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -962,9 +962,11 @@ class CommandEscInsertMode extends BaseCommand {
 
     vimState.currentMode = ModeName.Normal;
 
-    // If we wanted to repeat this insert, now is the time to do it. Insert
+    // If we wanted to repeat this insert (only for i and a), now is the time to do it. Insert
     // count amount of these strings before returning back to normal mode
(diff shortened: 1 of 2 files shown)


johnfn:
I think this is ok. The thing i'm curious about is the `- 3`. RecordedState actually has a number of convenience methods to e.g. pull out the current operator etc. Can you use one of those?

xconverge:
hmm unfortunately the current operator is escaping to normal, the previous is inserting content, and before that is starting insert/append... I could add some convenience methods but they are pretty obscure

Issue #1528, opened 2017-04-19
"daw" op should delete the world along with the space"


reminia:
* Click *thumbs-up* 👍 on this issue if you want it!
* Click *confused* 😕 on this issue if not having it makes VSCodeVim unusable. 

The VSCodeVim team prioritizes issues based on reaction count. 

--------

### What did you do?
The content is as below: 

> Cursor is in the end

Press daw 
### What did you expect to happen?

The cursor need to be on the end 'e' of 'the' word

### What happened instead?

Cursor on the space before end.

### Technical details:

* VSCode Version: 1.11.1
* VsCodeVim Version: *0.6.16*
* OS: macOS Sierra

Chillee:
My pull request should fix this issue.

rebornix:
`vaw` has the same issue. Whenever implementing word related commands, we always take prefix/suffix whitespace into account as that's when you are surprised with Vim commands. I didn't know how I miss this one ...

xconverge:
Fixed on master

PR #1536, opened 2017-04-20
Fixed Issue #1528

Chillee:
The issue is that if your cursor starts on the end of a word, getCurrentWordEnd should clearly include the current word.

I plan on adding a couple more unit tests for some edge cases.


xconverge:
if it doesn't break another test...makes sense to me :) thanks!

Issue #1534, opened 2017-04-20
Commentary toggle at start of current line comments previous line too


jgoz:
* Click *thumbs-up* 👍 on this issue if you want it!
* Click *confused* 😕 on this issue if not having it makes VSCodeVim unusable. 

The VSCodeVim team prioritizes issues based on reaction count. 

--------

### What did you do?

```
    line 1
    |line 2
```

`^gbb` or `0gbb` on "line 2"

### What did you expect to happen?

Comment line 2:

```
    line 1
|   // line 2
```

### What happened instead?

Commented both lines

```
|   // line 1
    // line 2
```

### Technical details:

* VSCode Version: 1.11.2
* VsCodeVim Version: 0.6.17
* OS: macOS Sierra

Chillee:
So I looked into the code, and it seems like the reason is that when you run the command "gbb", "gb" triggers the commenting mode, and then the second "b" jumps back a word. When the cursor is at the beginning of the word, this leads to highlighting two lines.

Can @fiedler chime in on whether this is intended behavior? I'm inclined to say no due to the README.

Chillee:
I added a simple (not very clean perhaps) fix to this issue.

fiedler:
Oh! It's definitely not the intended behaviour. It should stay on the same line.

Your reasoning behind why this is happening looks to be correct. I didn't experience this since I've remapped gc to gb. We should add a unit test covering this issue.

I think we need someone to look in to your PR. @xconverge can you help?

Chillee:
@fiedler I don't like my solution very much (although I can't think of any alternatives) , as I don't think it's very resistant to keybind changes. It's possible it might work out for this specific case (I can't think of any other movement commands that can move up a line) unintentionally. 

But if, for example, the user bound gj to commenting, pressing gjj would always comment two lines.

PR #1540, opened 2017-04-20
Possible fix to issue 1534

Chillee:
Fixes issue #1534. The problem inherent in issue #1534 is that of gbb. VSCodeVim doesn't have proper support for commands with repeated letters such as gbb, so what's happening in the bug is that "gb" loads the "commentLine" operator, and then applies it with the "b" movement.

The other double letter commands (dd, yy, etc.) [are implemented](https://github.com/VSCodeVim/Vim/blob/master/src/actions/actions.ts#L5107) in an (imo) somewhat fragile way. Although it works fine, it does have the side effect of causing all sorts of commands to exist (dy, d>, d<, yc, etc.) I think it may be worth refactoring this somehow.

However, back to the matter at hand, "b" is not an operator, so it can't be implemented in that manner. One possible solution (presented in this PR), is to have a corner case check inside MoveBeginningWord that checks if the current operator is CommentLine. 

I don't like this fix, and I don't know what kind of possible interactions it might have with other parts of the code. However, it seems like the simplest way of fixing this bug without a large refactoring.


xconverge:
I am not quite sure what the best way of fixing this is, I think most people should use gc and gcc

This could be merged, but I will also leave it open to see if @johnfn has a better idea to address the gcc gbb gc gb weirdness we now have :)

Chillee:
I think I want to spend some time fixing the repeated operator actions, so I'm gonna close this PR.

Issue #1533, opened 2017-04-19
Copy after vi' is incorrect (last character missing)


Valve:
* Click *thumbs-up* 👍 on this issue if you want it!
* Click *confused* 😕 on this issue if not having it makes VSCodeVim unusable. 

The VSCodeVim team prioritizes issues based on reaction count. 

--------

### Tried to copy text using ctrl+c after selecting a string between  quotes.

`s = "hello, world"`

Position cursor anywhere between quotes. Press `vi"`. Then press `ctrl+c`

### What did you expect to happen?
`hello, world` in the clipboard`

### What happened instead?
`hello, worl`  in the clipboard

### Technical details:

* VSCode Version: 1.1.12
* VsCodeVim Version: 0.6.16
* OS: Ubuntu 16.04

m47730:
confirmed on:
VSCode: 1.11.2
VsCodeVim: 0.6.17
OS: Fedora 24

In view mode on double click the selection missing the last char.
ex:
**"helloword"** on double click on **w** the selection is **hellowor**, i'm expected **helloword**

(in insert mode this is not happening)

Chillee:
Just wondering, are you guys setting vim.useCtrlKeys: true?

Valve:
yes, otherwise, how do I copy to the system clipboard? Maybe there is a way to do it, but I don't know how @Chillee

Chillee:
I think it's definitely a bug (that I've run into a bunch as well), but I'm just trying to track it down.

Chillee:
Ok, so I think I found part of the issue.

https://github.com/VSCodeVim/Vim/blob/master/src/mode/modeHandler.ts#L765

That should capture the default ctrl+c and replace it with our "<copy>", which handles all those cases. However, when useCtrlKeys is false, it doesn't get to this code.

I'll root around a bit more for the issue.

However, if you set vim.useCtrlKeys to be true, you can also use cmd+d (for whatever your keyboard is) to copy, and that should give you the correct behavior.

xconverge:
@Chillee fixed it on master

m47730:
i just installed version 0.6.20 and the things getting weird.
In insert mode everything is all right.
In view mode:

if a select a word visually the last character is missing, but if i do copy&paste the selection is complete and the character is not missing.
If a select a word and paste the last character is not overridden.

i create a gif of the problem to help you to understand:

[...]

PR #1542, opened 2017-04-20
Fixed issue #1533

Chillee:
We need to handle ctrl+c separately, as vim.overrideCopy should override vim.useCtrlKeys. 

I plan on adding tests later today.


xconverge:
thanks!

Issue #1350, opened 2017-03-02
Around Word at end of line


Tarmean:
Please *thumbs-up* 👍 this issue if it personally affects you! You can do this by clicking on the emoji-face on the top right of this post. Issues with more thumbs-up will be prioritized.

-----

### What did you do?

Use the `around word` motion on the last word in the line. Example with `daw`:

```
one  t|w|o
  three
```
### What did you expect to happen?

Motion selects the whitespace before the word and the word itself

```
on|e|
  three
```

### What happened instead?
Motion selects the newline, the whitespace at the beginning of the next line and the word itself.

```
  one |t|hree
```

### Technical details:

* VSCode Version: 1.10.1
* VsCodeVim Version: 0.5.3
* OS: Win10

PR #1549, opened 2017-04-22
Fixes #1193, #1350

### LINKED ISSUE #1193: 'caw' at first word on line

-----

### What did you do?

In normal mode with the cursor at the beginning of the first word of a line, pressed 'caw'

The first word would be deleted and insert mode would be entered with my cursor in the same position it started in.

### What happened instead?

All characters between position 1 and the end of the first word on the line are deleted. Insert mode is entered starting on position 1 of the line. See gif below:

### Technical details:

* VSCode Version: 1.8.1
* VsCodeVim Version: 0.4.10
* OS: Mac OSX

Chillee:
As I was working on the "aw" object, I thought I would fix the other related bugs.

It'd be ideal if I didn't need to special case this, but from thinking about it, it does seem like it's a special case. Normally, when you daw with a "." right at the end, you also want to delete all the whitespace on the left of the word until you reach another word. In this scenario, however, you want it to stay in the same place.

I was working on a complete rewrite of the 'aw' object based off "boundaries", but decided that it was overengineering. However, while writing it, I came some more cases where our behavior doesn't align with vim. I'd like some feedback on whether those cases are worth fixing (imo, vim does some very bizarre things in some of these cases).

```
one  
|  
three
```
becomes
`|one`

### DIFF
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -5297,7 +5297,9 @@ class SelectWord extends TextObjectMovement {
         stop = position.getWordRight().getLeftThroughLineBreaks();
         // If we aren't separated from the next word by whitespace(like in "horse ca|t,dog" or at the end of the line)
         // then we delete the spaces to the left of the current word. Otherwise, we delete to the right.
-        if (stop.isEqual(position.getCurrentWordEnd(true))) {
+        if (stop.isEqual(position.getCurrentWordEnd(true)) && !position.getWordLeft(true).isEqual(position.getFirstLineNonBlankChar())) {
+          start = position.getWordLeft(true);
+          start = position.getFirstLineNonBlankChar();
           start = position.getLastWordEnd().getRight();
         } else {
           start = position.getWordLeft(true);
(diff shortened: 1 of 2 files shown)


xconverge:
does this also handle this test:

```
  newTest({
      title: "Can handle 'daW' around word at end of line",
      start: ['one t|wo',
              ' three'],
      keysPressed: 'daW',
      end: ['on|e',
            ' three']
    });
```

This PR looks great! Thanks!

Chillee:
I've only touched the "aw" (lowercase) object, but as the test case for "aW" you posted is the same as the one for "aw" I fixed, I imagine the fix should be similar.

I'll check it out.

xconverge:
Want me to merge this and make that a separate issue?

Chillee:
Nah I'm almost done with the bug fix.

Chillee:
Actually, could you? I found some other bugs with "aW" that seem to be a bit harder to fix. Namely, tests like
```
one |two
```
result in everything being deleted.

I can make the new issue(s).

I'm actually surprised nobody has reported these bugs. I guess people don't really use the "aW" object much?

Issue #1096, opened 2016-11-29
Feature request: custom text objects/motions


timkendrick:
I'm an Atom [vim-mode-plus](https://atom.io/packages/vim-mode-plus) user who's been keeping an eye on VSCode for a while: one of the main things that's holding me back from switching is the inability to define various text objects and motions that I've grown accustomed to.

Some examples of custom text objects in other 'vimplementations':

- [vim-mode-plus-subword-movement](https://atom.io/packages/vim-mode-3plus-subword-movement) (Atom vim-mode-plus)
- [vim-textobj-function](https://github.com/kana/vim-textobj-function) (Vim)
- [vim-textobj-indent](https://github.com/kana/vim-textobj-indent) (Vim)
- [vim-textobj-user](https://github.com/kana/vim-textobj-user) (Vim)
...etc

Is there any chance that this kind of extensibility will be available for text objects and motions within VSCode Vim some day?

johnfn:
Our current policy on extensions is that while we probably won't implement them ourselves, we're completely open to other users implementing them. We're also completely fine with merging them into VSCodeVim core (generally speaking under a flag).

timkendrick:
Sounds great, thanks for the fast response!

johnfn:
Gonna keep this open as it's a valid feature request.

haasDev:
I also would very much like to see this.  Any work going on with this?

sunnyrjuneja:
Hey everyone,

Subword support has been added to VS Code in 1.25 (https://github.com/Microsoft/vscode/pull/48023). Is it possible to enable subword for motion hot keys?

muhajirframe:
Any progress on this?

tsujp:
Still no progress?

skywind3000:
Still no progress ?? at least can we support [textobj-parameter](https://github.com/sgur/vim-textobj-parameter) ??

It is the best textobj plugin which can be used to modify function arguments and any comma separated list items in a fast way. I really miss it in vim:

Usage
-----

-   `i,` to inner parameter object

    ```vim
    function(param_a, param_b, param_c)
             |<--->|  |<--->|  |<--->|
    ```

-   `a,` to a parameter object including whitespaces and comma

    ```vim
    function(param_a, param_b, param_c)
             |<----->|
    function(param_a, param_b, param_c)
                    |<----->|
    function(param_a, param_b, param_c)
                             |<----->|
    ```

[...]

PR #1550, opened 2017-04-22
Add support for indent objects

mikew:
As defined at [michaeljsmith/vim-indent-object](https://github.com/michaeljsmith/vim-indent-object).

Supports `ai`, `ii`, and `aI` with `c`, `d`, and `v` operators.

Refs #1096.


mikew:
Actually, looks like I broke some logic when I refractored before final submission. The tests aren't specific enough

mikew:
Hmmm, that failing test wasn't modified by this PR and passed the first time. I'm guessing a maintainer can retry the build.

I do have some questions though:

1. I made a `class IndentObjectMatch extends BaseMovement` in `actions.ts` but I'm not sure the use of it. I based it off of `MoveTagMatch` with some logic removed.
2. `TextEditor.getLineMaxColumn` errors when dealing with line 0 when it seems like there's nothing stopping it from working. I've worked around this by using `TextEditor.readLineAt(endLineNumber).length`
3. I'm not sure how to implement the `d` operator properly. As currently implemented, `dii` will leave you with a blank line, when really that line should be gone too. I couldn't figure out how to get rid of it without also removing the first character of the next line.

jpoon:
Restarted.

mikew:
Appreciated. Though I'm only just now noticing the existing `TextObjectMovement` class. I can easily refactor all logic into that style.

johnfn:
Wow, impressive work! Thanks for the contribution. 

Would you mind adding a little blurb in the README to explain what's going on here/how to use these guys?

And yeah if you want to refactor to the TextObjectMovement class too, that would be awesome. It's not necessary for the merge though, so just let me know.

mikew:
Already done days ago, just never pushed. Working on the README now but it seems it's changed a lot in master.

johnfn:
Yeah, we had a PR a few days ago that totally rewrote it. :) Should be stable-ish now.

mikew:
I merged master and updated the README. I decided against examples since it seemed to add a lot and instead went with comparing it to `cit/ci{/ci[/ciB`

johnfn:
Looks good to me. It's kind of funny since I added the custom `af` command to VSCodeVim to do the exact same thing. Eh, what are you going to do? 

Thanks again for the awesome work. :)

mikew:
Glad to help. And I was just thinking it would be cool if the count prefix could mean "look *n* levels up". As it stands, `canBePrefixedWithCount` should be `false`. Missed that one.

Issue #1323, opened 2017-02-22
Issue with cursor position using visual line movement (gj and gk)


kevinjalbert:
Please *thumbs-up* 👍 this issue if it personally affects you! You can do this by clicking on the emoji-face on the top right of this post. Issues with more thumbs-up will be prioritized.

-----

### What did you do?

I've set the following as I like to work with word wrap enabled:
```
    "vim.otherModesKeyBindingsNonRecursive": [
        {
            "before": ["j"],
            "after": ["g", "j"]
        },
        {
            "before": ["k"],
            "after": ["g", "k"]
        }
    ]
```

I noticed a discrepancy in behaviour for lines without wrapping. As you navigate between lines of varying length the cursor position doesn't keep its horizontal position.

The following GIFs just show motion using `j` and `k`.

I would expect both GIFs to show the same thing regardless if the `gj` and `gk` mappings are applied.

### What did you expect to happen?

Expected behaviour (the keybinding mappings not applied):
![normal vim movement](https://dl.dropbox.com/s/s5s5q9gfd3p7r4b/2017-02-22_10-21-17.gif?dl=0)

### What happened instead?

Current behaviour (the keybinding mappings applied):
![normal vim movement](https://dl.dropbox.com/s/siloxrd1fqpxanr/2017-02-22_10-24-10.gif?dl=0
)

### Technical details:

* VSCode Version: 1.9.1
* VsCodeVim Version: *0.5.3*
* OS: MacOS 10.12.3

xconverge:
copying comment from dupe ticket linked:

This is a tough one

I couldn't figure out a way to get the start and end column positions of the wrapped line.

I tried this:
```
const posLine = vimState.editor.selection.active.line;

await vscode.commands.executeCommand("cursorMove", {
	to: "wrappedLineStart"
});

const startCol = vimState.editor.selection.active.character;

await vscode.commands.executeCommand("cursorMove", {
	to: "wrappedLineEnd"
});

const endCol = vimState.editor.selection.active.character;
const lineLength = endCol - startCol;

vimState.cursorPosition =
	new Position(posLine, Math.min(lineLength + startCol, vimState.desiredColumn + startCol));
```
but moving the cursor 3 times for every 1 time we want to move it makes no sense from a performance standpoint.

I think that we need to get wrappedLine info via the vscode API but I could not find it currently...

[...]

Issue #1467, opened 2017-04-04
Add next occurence in insert and normal mode is limited to a maximum of 2 cursors


RodrigoRoaRodriguez:
Please *thumbs-up* 👍 this issue if it personally affects you! You can do this by clicking on the emoji-face on the top right of this post. Issues with more thumbs-up will be prioritized.

-----

### What did you do?
Go to ~either normal or~ insert mode and press _Add next occurence_ more than two times. 
For instance press it four times for the following text (the pipe `|` represents a cursor)
```
cursors|[1] = 1
cursors[2] = 2
cursors[3] = 3
cursors[4] = 4
```

### What did you expect to happen?
I expected a new cursor to be added for every press.
```
cursors|[1] = 1
cursors|[2] = 2
cursors|[3] = 3
cursors|[4] = 4
```
### What happened instead?
Only one new cursor is created, and after that _Add next occurrence_ no longer adds cursors or seems to do anything.
```
cursors|[1] = 1
cursors|[2] = 2
cursors[3] = 3
cursors[4] = 4
```
### Technical details:
* VSCode Version: 1.10
* VsCodeVim Version: 0.6.13
* OS: OSX 10.11 (El Capitan)

### Additional notes
Strangely enough it works on visual mode. Curiously, more than two cursors can be added through other commands such as _Select all occurrences_ or `alt-click`

The `Cmd-d` shortcut seems to be disabled by default while in insert mode, however it seems that can be restored by adding `"vim.handleKeys": { "<D-d>":false}` to your configuration (This however breaks normal mode by limiting it to two cursors).

johnfn:
This DOES work without making that change to your settings, right?

RodrigoRoaRodriguez:
@johnfn the change to the configuration is only necessary to trigger _Add next occurrence through the default shortcut (however without it it could still be activated through other means such as the menu item). I will edit the issue to be clearer in this regard so that no further misunderstandings occurs .

johnfn:
cmd-d works perfectly for me on OSX, no configuration changes required! Do you not experience the same?

xconverge:
Yea I would not recommend that configuration change since I think we are handling it better for our usage, I was just saying it could be tried

RodrigoRoaRodriguez:
@johnfn and @xconverge `cmd-d` does not work for me while I am in insert mode without that configuration change. I tried it in two different computers.

xconverge:
[...]

Issue #1004, opened 2016-10-27
Folded code unfolds if you spend some (unknown number) of ms idle in a fold


Phrohdoh:
Please _thumbs-up_ 👍 this issue if it personally affects you! You can do this by clicking on the emoji-face on the top right of this post. Issues with more thumbs-up will be prioritized.

---
### What did you do?

Scrolled through folded code once with the `j` key held down, another time pressing it over and over.
### What did you expect to happen?

Folded lines were skipped.
### What happened instead?

Cursor did not skip the folded lines (watch the line numbers in the gif).
### Technical details:
- VSCode Version: 1.6.1
- VsCodeVim Version: 0.4.0
- OS: macOS Sierra (10.12)

![vscode-vim-skip-folds](https://cloud.githubusercontent.com/assets/4861023/19783950/8b993180-9c59-11e6-89f9-52da941701d3.gif)

johnfn:
This is annoying, but also hard. @aminroosta tried but had a bad time: https://github.com/VSCodeVim/Vim/pull/708

The problem as I understand it is that the only API VSCode allows us is `"cursorMove"` to move the cursor over folded lines, but we would actually like to know the position of the cursor as well (which `"cursorMove"` does not provide).

NikosEfthias:
on ubuntu, i also suffer from this a log I found the solution using `gg` to jump to the last line

Phrohdoh:
`gg` takes you to the first column on the first row, not the last line.

NikosEfthias:
@Phrohdoh  for example `25gg` will take you to 25th line not the beginning i meant end of the block by last line

Phrohdoh:
@NikosEfthias Oh I see, TIL. Thanks!

exhuma:
I can confirm this on VSCode 1.9.1 with Vim Plugin 0.5.3.

This makes working with larger, more complex files very unwieldy!

I've tried disabling the vim plugin and moving with the cursor keys. In that case it works as expected (folds remain closed).

skawaii:
+1 for this. Having my folds open while I'm navigating is quite frustrating. I find myself having to use my mouse to reposition my cursor, which should make all of us sad.

zwing99:
+1 for fixing this asap

ztmdsbt:
+1 for this. It makes me very painful.

Phrohdoh:
+1 comments aren't going to do anything but annoy the repo owners.
[...]

PR #1552, opened 2017-04-23
Fixing the automatic fold expansion (#1004)

Chillee:
One of the most requested issues (and one of the things that bothers me the most) is that j and k automatically open up folds. #1004 

However, this problem is kinda tricky to solve due to VScode having a somewhat limited API when it comes to folds.

The one thing that's exposed is "cursorMove". Although it says it should move "logically", the only command that allowed me to skip folds was "cursorMove" with "wrappedLine" set to true.

Thus, if you do the equivalent of a "gj" in Vim (which skips over folds without opening them), and then check if you've moved more than one line down, you know if you're over a fold or not.

Then, you can do a "gk" to get back* to where you are before if you're not over a fold, and follow up with a regular "j".

* There's one bug with gj that causes us to drop "desiredColumn" information. Check #1323 and Microsoft/vscode#23045 for more information on that. This is probably the biggest issue with this potential fix. I think it's possible to fix it, but it might fairly involved. If you play around with this build, I think you'll find that this is the biggest pain point that this fix introduces. UPDATE: PR #1561 should fix this. I've also integrated PR #1561 into this fix, so this is now a superset of PR #1561.

There's also a couple things that are broken from this change, like up and down arrows, and using j and k in visual mode. However, I think those can be fixed.

This is most definitely not an optimal solution (it's really hackish), but I think the performance hit may be negligible (many of our other operations are a lot more expensive), and it may be good enough while we wait for VSCode to extend their API.

It's also similar to the approach @xconverge wrote about in #1467.

In general, this approach has some other bugs that are fairly hard to fix (mostly related to desiredColumn issues). (UPDATE: I've fixed most of the desiredColumn issues). Due to the myriad issues (and the general hackishness of this approach), I would propose that we add this behind a flag, as a temporary fix for fold related issues. Looking at the issue thread, it seems to be a major pain point (it is for me too).

Overall, I'd just like some feedback on this approach. I realize the code quality isn't great, but think of this as a prototype that'll be rewritten if we decide it's worth it. If it's worth exploring, I can try fixing gj and the other related bugs. (UPDATE: I've fixed the gj bugs).


johnfn:
This is really awesome man. It's really appreciated.

I'm going to be busy this weekend with Ludum Dare (ludumdare.com), so I can't check it out immediately, but I will soon. 

Thanks again!

Chillee:
No rush. Ludum dare is awesome! If only I didn't have a ton of work piled up over the next week :(

Chillee:
I actually have a fix for 'gj' and 'gk' coming up that I think should make this PR a bit better.

Chillee:
I committed the gj/gk fix. To me at least, it feels pretty good.

Chillee:
Alright I changed some things around and added it behind a flag. I think this is ready for merging; the only question is whether the flag should be on by default, or whether it should even be included at all. 

I vote that it should definitely be included, at least under a flag.

Chillee:
I figured out a solution that's pretty solid, if I say so myself. This also fixes #1323 now as well.

Chillee:
Actually, I think I have an even better solution (potentially). It'll need to wait a couple days though.

Chillee:
Nevermind, my proper fix isn't possible right now. I'll keep the version that was working.

Chillee:
Ok, as the current fold fix doesn't break anything else, is behind a flag, and isn't too bad of a fix, I'm planning on merging this at the end of the day (unless somebody objects). @rebornix @johnfn @xconverge

Chillee:
Alright, I'm gonna merge this one now.

ZuBB:
does anyone know when we can expect next release with this change?

ps: thanks a lot @Chillee! you are my hero 🙇

Chillee:
Probably within the next day or two. @ZuBB Just so you don't get your hopes up too high, it's definitely not a perfect fix. It's a hack through and through, and leads to some (relatively) minor sideffects with `j` and `k`, namely that you'll see some flickering on wrapped lines (which is why it's not enabled by default). 

However, it's definitely a huge improvement for folded lines, and I think it'll satisfy the needs of most users who care about folds.

ZuBB:
@Chillee thanks for details on how it will work. Cant wait to try it 😄

Chillee:
You'll also need to enable `vim.foldfix` to get the change, once it's released.

ZuBB:
thanks for hint. I am sure it be usefull for others too

Chillee:
@ZuBB We just pushed out a new release, so you can now try the `vim.foldfix` option!

ZuBB:
thanks for a notification!

elsonwx:
the `M`(jump to the middle of screen),`nj`(down number lines),`nk`(up number lines) seems still very troubled

Chillee:
@elsonwx Unluckily those are things that are very tricky to fix. Folds have always been a tricky issue.

Issue #1553, opened 2017-04-23
daW breaks when the cursor is on the beginning of the word


Chillee:
* Click *thumbs-up* 👍 on this issue if you want it!
* Click *confused* 😕 on this issue if not having it makes VSCodeVim unusable. 

The VSCodeVim team prioritizes issues based on reaction count. 

--------

### What did you do?

Put 
```
one |t|wo
```
and type "daW".

### What did you expect to happen?
For the two to be deleted.

### What happened instead?
The entire line is deleted.

### Technical details:

* VSCode Version: 1.11.2
* VsCodeVim Version: 0.6.17
* OS: Arch Linux

I'm fixing this bug right now. Since the issue is on the aw object and not daw in particular, this also happens for vaw, yaw, caw, etc.

PR #1555, opened 2017-04-23
Fixes #1554, #1553

### LINKED ISSUE #1554: daW breaks around word at end of line

--------

### What did you do?

Start with
```
one two
  three
```
press daW
```
on|e
 three
```
### What happened instead?
```
one |three
```
### Technical details:

* VSCode Version:
* VsCodeVim Version: *[please ensure you are on the latest]*
* OS: 

I'm fixing this.

Chillee:
Fixes those two issues. One potentially bad practice is that I'm accessing vimState.recordedState.count. I don't know how that might play with things like undo.

This commit also fixes a regression I introduced in PR #1549 . In my defense, the reason the regression happened was that buggy behavior was hiding another bug.

I also added a comment explaining code added in PR #1549 .

### DIFF
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -5348,6 +5348,7 @@ class SelectABigWord extends TextObjectMovement {
         start = position.getLastBigWordEnd().getRight();
         stop = position.getCurrentBigWordEnd();
     } else {
+        // Check 'aw' code for much of the reasoning behind this logic.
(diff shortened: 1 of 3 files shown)


xconverge:
LGTM, thanks!

xconverge:
Vim does do something weird here, so it is a special case

PR #1561, opened 2017-04-24
Fixes gj and gk cursor issues.

Chillee:
#1323 is a big issue for me. It makes it extraordinarily painful to use 'gj' full time instead of j, and makes #1552 less ideal for everybody.

VSCode continues to refuse to provide good support for getting their "hidden" desiredColumn, so I handle it myself in a special case for "gj" and "gk". Make no mistake, this is definitely not the best way to handle this. However, the perfect is the enemy of the good, and this solution will allow us to make 'gj' and 'gk' usable.

I think it's a good design principle to optimize for the common use case, and a common use case for gj and gk is rebinding j and k to them and scrolling through files. This fixes it.

Another point in favor of this pull request is that it's basically a strict improvement. The only places it fails are where the old gj also fails, and this pull request works the majority of the time.

It still fails on visual mode I think.

Integrating this PR also allows a workable solution (ehhh) #1552 to folds, which is the biggest pain point towards adoption for VSCodeVim (in my opinion).

It's not perfect, but I think it's good enough. (Worse is Better anyone?)


johnfn:
Can't review this quite yet - just wanted to send a big thanks for all your recent PRs. They are massively appreciated!

Chillee:
Closed in favor of #1552 .

PR #1572, opened 2017-04-25
Fix logo src so logo displays inside VSCode

cobbweb:
The relative URL that was being used only works when viewing the README
from GitHub, and leaves a big blank space if you open VSCodeVim's page
inside VSCodeVim.

This uses a full URL, which should fix the issue. This is how the Chrome
Debugger plugin loads their logo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 <h1 align="center">
-    <img src="https://cdn.rawgit.com/VSCodeVim/Vim/master/images/icon.png" height="128">
+    <img src="https://raw.githubusercontent.com/VSCodeVim/Vim/master/images/icon.png" height="128">
     <br>
     VSCodeVim
 </h1>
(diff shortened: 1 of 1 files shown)


jpoon:
Wasn't relative file path working previously? What changed?

jpoon:
I see, it's because we moved it to HTML

jpoon:
@cobbweb what is rawgit and why should we use it over directly linking to GH?

cobbweb:
[rawgit](http://rawgit.com) provide a free CDN wrapper around files in GitHub repos. I only used it because that's how the Chrome Debugger reference theirs.

IIRC correctly, rawgit's main feature is that it serves JS files with proper headers and maybe doesn't provide any value for us here. We can just link directly to: https://raw.githubusercontent.com/VSCodeVim/Vim/master/images/icon.png

Let me know if you want it changed.

jpoon:
Yeah, let's go with `raw.githubusercontent.com`. Don't want to rely on an external dependency if we don't have to. rawgit sounds like an awesome project, but don't know how long a free project/hosting will last.

jpoon:
Whoa, that's crazy. I have commit rights to your fork 😮

jpoon:
Thanks @cobbweb.

PR #1595, opened 2017-04-28
Fixes #1599

### LINKED ISSUE #1216: Using "." in a macro causes the macro to execute forever
Create a macro that uses `.` to repeat the last command.

`qadd.q`

Execute it.

`@a`

Macro deletes all lines in the file. It seems like the `.` doesn't repeat the `dd` command, but the `@a` command, so we have an infinite recursive call.

### Technical details:

* VSCode Version: 1.8.1
* VsCodeVim Version: 0.4.10
* OS: Windows 10

### LINKED ISSUE #1309: Limit infinite recursion in macros
### What did you do?

Accidentally wrote a macro with an infinite loop

qqiexampletextEscq@q

Expected the macro definition to fail and @q to remain undefined

### What happened instead?

Macro got into an infinite loop and required a vscode restart

### Technical details:

* VSCode Version: 1.9.1
* VsCodeVim Version: 0.5.2
* OS: 10.12.3

### LINKED ISSUE #1599: . doesn't work in a macro

--------

### What did you do?

qqdd.q
@q @q 
For the macro to delete 2 lines every time.

### What happened instead?
The macro deletes 2 lines the first time, but doesn't delete 2 lines afterwards.

### Technical details:

* VSCode Version: 0.6.20
* VsCodeVim Version: 1.11.2
* OS: Arch

Chillee:
Pretty simple. We don't want to prevent us from creating an undo point by setting "runRepeatableAction" to false, we want to prevent us from creating an undo point by checking if we're replaying a macro.

On a related note, I found this issue while trying to fix #1216, which doesn't seem to be an issue anymore.

### DIFF
--- a/extension.ts
+++ b/extension.ts
@@ -16,6 +16,8 @@ import { Globals } from './src/globals';
 import { AngleBracketNotation } from './src/notation';
 import { ModeName } from './src/mode/mode';
 import { Configuration } from './src/configuration/configuration'
+import { ICodeKeybinding } from './src/mode/remapper';
+import { runCmdLine } from './src/cmd_line/main';
 
 interface VSCodeKeybinding {
   key: string;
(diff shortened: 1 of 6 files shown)


johnfn:
I'm a bit confused though. I thought that the problem with #1309 was that macros could cause infinite recursion by calling themselves. How is that related to undo points? Were you referring to a different issue?

Chillee:
Oh sorry. I guess I'm not really fixing #1309. When trying to fix that issue, I found that it seems to have been fixed at some point (I'm not able to replicate it). However, the dot command still doesn't work properly, because it doesn't do anything. 

This fixes that. I'll make a new issue and change my commit or something

Chillee:
@johnfn I added a test that tests the thing I fixed.

That should explain the issue I'm trying to fix a bit more.

johnfn:
Alright, got it! Thanks for the PR. :)

PR #1596, opened 2017-04-28
Fixes #1251

### LINKED ISSUE #1251: Reformatting with "gq" always adds 1 column of whitespace before formatted block

-----

### What did you do?

`vipgq
gq
gq`

or

`Vgq
gq
gq`

Applying gq should not be shift the formatted block to the right (add indentation), ie, normal gq behaviour.

### What happened instead?

Repeatedly applying `gq` would repeatedly add a space to each row of formatted text.

### Technical details:

* VSCode Version: 1.8.1
* VsCodeVim Version: 0.5.2
* OS: Windows 10

Chillee:
I think the issue here is that when we go word by word adding it in, we always add " {$word}". However, the first word shouldn't have an extra space in front of it.

This problem should also be fixable here, but I thought my current solution was fine. 
https://github.com/VSCodeVim/Vim/blob/master/src/actions/actions.ts#L4514,

### DIFF
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -4513,7 +4513,7 @@ class ActionVisualReflowParagraph extends BaseCommand {
           if (lines[lines.length - 1].length + word.length + 1 < maximumLineLength) {
             lines[lines.length - 1] += ` ${ word }`;
           } else {
-            lines.push(` ${ word }`);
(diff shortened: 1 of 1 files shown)


johnfn:
LGTM. Thanks for the fix! :)

PR #1598, opened 2017-04-28
add tag text object matcher

vilisper:
modify the tagMatcher.ts to support vit dit comments, it works in some case now. nested tags, multi-line, self-closing tag.  but now if the cursor is in closing tag, when cit delete the inner text, the cursor is not move to the right place.

### DIFF
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -6256,7 +6256,7 @@ abstract class MoveTagMatch extends BaseMovement {
 
   public async execAction(position: Position, vimState: VimState): Promise<IMovement> {
     const text = TextEditor.getLineAt(position).text;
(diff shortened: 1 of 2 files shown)


vilisper:
I run the tests locally, event the unchanged main branch print the fails. 
I just changed the tagMatcher.ts and the actions.ts where call then tagMatcher.

johnfn:
Hey @vilisper - I'm really sorry, but we actually had a bug fix for this bug already in the PR queue that just got merged. 😞  I know it is very disappointing to realize that work has been duplicated. In the future, I'd really encourage you to hop on our Slack channel or take a peek at the other active PRs in the queue.

Again, sorry man.

Issue #1377, opened 2017-03-10
Issue with visual line select mode and 'gj'


rufusroflpunch:
* Click *thumbs-up* 👍 on this issue if you want it!
* Click *confused* 😕 on this issue if not having it makes VSCodeVim unusable. 

The VSCodeVim team prioritizes issues based on reaction count. 

--------
### What did you do?

1. Enter visual line selection mode: `V`
2. Use `gj`  to move between wrapped lines.

### What did you expect to happen?

The selection to extend as many lines as required.

### What happened instead?

The selection doesn't extend past three lines.

![example gif of bug](http://i.giphy.com/xUPGcKBfdlzYeuoyDC.gif)

### Technical details:

* VSCode Version: 1.10.2
* VsCodeVim Version: 0.6.2
* OS: macOS Sierra

johnfn:
As a temporary solution, you can remap `j` to `gjo`. 😉  But I think @xconverge has a better solution coming soon.

xconverge:
related https://github.com/Microsoft/vscode/issues/23045

I should be able to work through this with what we have so far right now though, but it might not be pretty

xconverge:
Also we need to make sure the when we expand and shrink the selection, it works similar to how V with j and k work.

Currently, it seems like if you do V then gk gk gk gk, when you do gj, it will expand from the bottom, instead of shrinking the selection by 1 line

PR #1600, opened 2017-04-28
Fixes #890, #1377

### LINKED ISSUE #890: Visual Movement does very weird things with visual selection

--------

### What did you do?

Similar setup to #889, I have bound `0`, `$`, `j`, and `k` to their `g`-prefixed counterparts. I then pressed `v` and started to use `j`, `k`, `l`, and `h` to select text.

Visual mode works as normal.
### What happened instead?

The selection seems to follow my cursor. GIF's are worth a thousand words: https://gyazo.com/a25cf4cd86d2b8f45d4a27913476c9a8. Note that _all_ I'm doing is starting with one cursor, then pressing `v`, followed by `j` and `k`.
### Technical details:
- VSCode Version: 1.6.0
- VsCodeVim Version: 0.3.5
- OS: Windows

Chillee:
I think this fixes all the visual selection related bugs I could find.

### DIFF
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -3673,6 +3673,11 @@ abstract class MoveByScreenLine extends BaseMovement {
   value: number = 1;
 
   public async execAction(position: Position, vimState: VimState): Promise<Position | IMovement> {
+    // Because we can't support moving by screen line when in visualLine mode,
+    // we change to moving by regular line in visualLine mode. We can't move by
+    // screen line is that our ranges only support a start and stop attribute,
+    // and moving by screen line just snaps us back to the original position.
+    // Check PR #1600 for discussion.
     if (vimState.currentMode === ModeName.VisualLine) {
       if (this.movementType === "up" || this.movementType === "down") {
         this.by = "line";
@@ -3692,7 +3697,6 @@ abstract class MoveByScreenLine extends BaseMovement {
        * cursorMove command is handling the selection for us.
        * So we are not following our design principal (do no real movement inside an action) here.
        */
-
       let start = Position.FromVSCodePosition(vimState.editor.selection.start);
       let stop = Position.FromVSCodePosition(vimState.editor.selection.end);
       let curPos = Position.FromVSCodePosition(vimState.editor.selection.active);
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -3666,6 +3666,7 @@ abstract class MoveByScreenLine extends BaseMovement {
   value: number = 1;
 
   public async execAction(position: Position, vimState: VimState): Promise<Position | IMovement> {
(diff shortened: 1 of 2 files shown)


src/actions/actions.ts
```diff
@@ -3673,6 +3673,11 @@ abstract class MoveByScreenLine extends BaseMovement {
   value: number = 1;
 
   public async execAction(position: Position, vimState: VimState): Promise<Position | IMovement> {
+    if (vimState.currentMode === ModeName.VisualLine) {
```
johnfn: Could we have a quick comment here to say why this is necessary (or which issue it is fixing)?

johnfn:
Again, thank you so much for the bevy of PRs recently. Much appreciated. :)

src/actions/actions.ts
Chillee: So the main issue is that we don't handle visualLine selections, due to the fact that our cursor must be at either start or stop (which as far as I can tell, works for every case but this one?).

However, in vim in visual line mode, doing gj on a wrapped line causes you to maintain the same highlighted area, however, your cursor is moved up one wrapped line.

However, we always snap the cursors at the beginning and the end of the selection to the end of the line (to maintain visual line mode). Thus, if we do a "gj" on a wrapped line, we are snapped back to our initial position, and we can't move down.

I don't think there's a good solution for this, so the workaround I came up with was to move by a "line" instead of a "wrappedLine" if you're doing visualLine movement.

This behavior doesn't exactly match vim, but I really doubt (I hope) that anybody relied on that feature of vim.

Chillee: Hmm, actually.

Maybe it is possible? The vscode.selection objects have a start, stop and active attribute. The problem (I think) is that our Range objects only have a start and stop.

src/actions/actions.ts
```diff
@@ -3673,6 +3673,16 @@ abstract class MoveByScreenLine extends BaseMovement {
   value: number = 1;
 
   public async execAction(position: Position, vimState: VimState): Promise<Position | IMovement> {
+    // Because we can't support moving by screen line when in visualLine mode,
+    // we change to moving by regular line in visualLine mode. We can't move by
+    // screen line is that our ranges only support a start and stop attribute,
+    // and moving by screen line just snaps us back to the original position.
+    // Check PR #1600 for discussion.
+    if (vimState.currentMode === ModeName.VisualLine) {
+      if (this.movementType === "up" || this.movementType === "down") {
```
rebornix: I'm not a fan of putting this filter here. Each non-abstract extended MoveByScreenLine defines its `movementType`, if the logic is only related to it, then we may want to put it in that extended class.

Chillee: Do you think it's better to define another class "MoveUpByScreenLineVisualLine"?

rebornix: oh that would be better.

Chillee: The new commit should have done that.

PR #1601, opened 2017-04-28
Made command line persistent when switching windows

Chillee:
I think this is an improvement, and in addition, it follows what vim does.


johnfn:
Very cool. Thanks! 😁

PR #1602, opened 2017-04-28
Fix status bar color when change mode

zelphir:
Update to the new and stable api for `workbench.colorCustomizations`


johnfn:
Thanks @zelphir! Much appreciated.

xconverge:
This means don't make a release until next vscode version

I would do a 0.7.0 at that time and also include @rebornix scroll select stuff

PR #1603, opened 2017-04-29
Fixes #1575 (and also adds general newline support for searches)

### LINKED ISSUE #1575: Search doesn't match new lines

--------

### What did you do?

/\n

New lines in the document should have been highlighted

### What happened instead?

Nothing. 

Searching for new lines using \n works in standard VS Code search with regex turned on.

### Technical details:

* VSCode Version: 1.11.2
* VsCodeVim Version: 0.6.18
* OS: macOS 10.12.3

Chillee:
Basically, the general approach is that we count how many newlines there are in their search string, and expand our search beyond line by line accordingly.

Thus, if their string contains two newline characters, then we check two lines at a time, and so on.

Accordingly, if there are no newline characters, this shouldn't affect performance at all, and performance should then degrade linearly with the amount of newlines (which makes sense).

One thing that it's missing is that Vscode doesn't seem to support highlighting the "newline" character. That makes newlines pretty invisible.

I've also added a test.

PS: You'll notice the commit that made the command line persistent, PR #1601. I accidentally checked it out and don't know how to delete a commit in the past, so I simply reverted it and then squashed the commit. I hope that's not an issue if both PR's are merged...

### DIFF
--- a/src/state/searchState.ts
+++ b/src/state/searchState.ts
@@ -48,14 +48,12 @@ export class SearchState {
 
   private _recalculateSearchRanges({ forceRecalc }: { forceRecalc?: boolean } = {}): void {
     const search = this.searchString;
-    const selection = vscode.window.activeTextEditor!.selection;
-    const lineToStartAt = Math.min(selection.start.line, selection.end.line);
-
     if (search === "") { return; }
 
     // checking if the tab that is worked on has changed, or the file version has changed
     const shouldRecalculate = (this._cachedDocumentName !== TextEditor.getDocumentName()) ||
       (this._cachedDocumentVersion !== TextEditor.getDocumentVersion()) || forceRecalc;
+
     if (shouldRecalculate) {
       // Calculate and store all matching ranges
       this._cachedDocumentVersion = TextEditor.getDocumentVersion();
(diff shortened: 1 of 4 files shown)


src/state/searchState.ts
```diff
@@ -92,20 +92,38 @@ export class SearchState {
 
       let lineIdx = lineToStartAt;
 
+      let numNewLines = (searchRE.match(/\\n/g) || []).length;
       // Start at the current line and wrap the document if we hit the end.
       outer:
       do {
-        const line = TextEditor.getLineAt(new Position(lineIdx, 0)).text;
+        let rangeStart = new Position(lineIdx, 0);
```
johnfn: Minor I know, but can we have a convention of using `const` when the variable doesn't change?

src/state/searchState.ts
```diff
@@ -92,20 +92,38 @@ export class SearchState {
...
+        let rangeEnd = rangeStart.getLineEnd();
+        let lineLengths = [];
+        for (var i = 0; i < numNewLines; i++) {
```
johnfn: ..and `let` when it does?

src/state/searchState.ts
```diff
@@ -92,20 +92,38 @@ export class SearchState {
...
+          rangeEnd = rangeEnd.getDown(1).getLineEndIncludingEOL();
+        }
+        const line = TextEditor.getText(new vscode.Range(rangeStart, rangeEnd));
         let result = regex.exec(line);
 
         while (result) {
           if (this._matchRanges.length >= SearchState.MAX_SEARCH_RANGES) {
             break outer;
           }
+          let resultToPosition = function(resultIdx: number, lineIdx: number, lineLengths: Array<number>, result: RegExpExecArray){
```
johnfn: As a further style suggestion, it's good to prefer `=>` over `function` in the general case.

src/state/searchState.ts
```diff
@@ -92,20 +92,38 @@ export class SearchState {
 
       let lineIdx = lineToStartAt;
 
+      let numNewLines = (searchRE.match(/\\n/g) || []).length;
```
johnfn: regex-ception

src/state/searchState.ts
```diff
@@ -92,20 +92,38 @@ export class SearchState {
...
+        }
+        const line = TextEditor.getText(new vscode.Range(rangeStart, rangeEnd));
         let result = regex.exec(line);
 
         while (result) {
           if (this._matchRanges.length >= SearchState.MAX_SEARCH_RANGES) {
             break outer;
           }
+          let resultToPosition = function(resultIdx: number, lineIdx: number, lineLengths: Array<number>, result: RegExpExecArray){
+            let cur = resultIdx;
```
johnfn: Could we have a slightly better name than `cur`, like `positionInAggregateLines` or something?

Chillee:
So there is a definitely a noticeable delay. I'll try to work on the performance.

Chillee:
Alright, so I implemented a different model that searches the whole text instead of line by line. This is way faster (there's pretty much no drop in speed so far as I can tell, with an arbitrarily high number of newlines).

Chillee:
Alright. I think the current one is in a mergeable state. The performance is very good, and I've ironed out all the other miscellaneous bugs.

I think the failing test is just the test server screwing up. The previous commit passed all tests (only failed due to linting issue), and the final commit shouldn't have changed anything related to that.

PR #1605, opened 2017-04-29
Fixes #1532

### LINKED ISSUE #1532: column not set properly after goto definition

--------

### What did you do?

Find a symbol. Press `gd` to jump to the definition. Hit `j` and then `k`. Notice that the cursor is now at column 0.

The cursor remains at the original column of the definition.

### What happened instead?

The cursor column got set to 0.

### Technical details:

* VSCode Version: 1.11.2
* VsCodeVim Version: 0.6.16
* OS: Ubuntu-like

Chillee:
this._vimState.editor.selection.start doesn't get updated until we call updateView. 

Why, you ask?

No clue. But it works!

(Well, we know that updateView updates the vscode object a bunch of times, so that probably has something to do with it.)

### DIFF
--- a/src/mode/modeHandler.ts
+++ b/src/mode/modeHandler.ts
@@ -39,6 +39,7 @@ import { PairMatcher } from './../matching/matcher';
 import { Globals } from '../../src/globals';
 import { ReplaceState } from './../state/replaceState';
 import { GlobalState } from './../state/globalState';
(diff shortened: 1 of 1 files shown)


johnfn:
I just realized that waiting on updateView makes no sense because updateView should actually change the cursor position... It's super weird that this would work.

PR #1612, opened 2017-04-30
Fixes 1196

Chillee:
The getCurrentParagraphEnd() object keeps on going down until the current line is empty. At that point, we should be getting the end of the line above, not the current line. However, going leftthroughlinebreaks works just as well on an empty line.


mikew:
This breaks regular `}` movement, and still has issues with `d}`.

Chillee:
Oh whups I didn't commit in my other fix. Hold up.

PR #1616, opened 2017-04-30
Visualstar

mikew:
The default behaviour of `*` and `#` isn't really intuitive, and Vim itself has [many plugins to handle this](https://www.google.com/search?q=vim+visual+star).

This PR adds a new option to treat `*` and `#` as a new search when pressed in visual mode.

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,6 +180,10 @@ These settings are specific to VSCodeVim.
     }
 ```
 
+#### `"vim.visualstar"`
+* When pressing `*` or `#` in visual mode treat it as a new search
+* Type: Boolean (Default: `false`)
+
 ### Key remapping
 
 There's several different settings you can use to define custom remappings. Also related are the [`useCtrlKeys`](#vimusectrlkeys) and [`handleKeys`](#vimhandlekeys) settings.
--- a/README.md
+++ b/README.md
@@ -181,7 +181,7 @@ These settings are specific to VSCodeVim.
 ```
 
 #### `"vim.visualstar"`
-* When pressing `*` or `#` in visual mode treat it as a new search
+* In visual mode, start a search with * or # using the current selection
 * Type: Boolean (Default: `false`)
 
 ### Key remapping
--- a/package.json
+++ b/package.json
@@ -441,6 +441,11 @@
         "vim.statusBarColors": {
           "type": "object",
           "description": "Customize colors per mode when VSCodeVim controls status bar colors"
+        },
+        "vim.visualstar": {
+          "type": "boolean",
+          "description": "When pressing * or # in visual mode treat it as a new search",
+          "default": true
         }
       }
     }
--- a/package.json
+++ b/package.json
@@ -445,7 +445,7 @@
         "vim.visualstar": {
           "type": "boolean",
           "description": "When pressing * or # in visual mode treat it as a new search",
-          "default": true
+          "default": false
         }
       }
     }
--- a/package.json
+++ b/package.json
@@ -444,7 +444,7 @@
         },
         "vim.visualstar": {
           "type": "boolean",
-          "description": "When pressing * or # in visual mode treat it as a new search",
+          "description": "In visual mode, start a search with * or # using the current selection",
           "default": false
         }
       }
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -1702,14 +1702,29 @@ class CommandCmdA extends BaseCommand {
 
 function searchCurrentWord(position: Position, vimState: VimState, direction: SearchDirection, isExact: boolean) {
     const currentWord = TextEditor.getWord(position);
-    if (currentWord === undefined) {
+
+    return performSearchMovement(currentWord, vimState, direction, isExact);
+}
+
+function searchCurrentSelection (vimState: VimState, direction: SearchDirection, isExact: boolean) {
+    const selection = TextEditor.getSelection();
+    const end = new Position(selection.end.line, selection.end.character + 1);
+    const currentSelection = TextEditor.getText(selection.with(selection.start, end));
+
+    // Go back to Normal mode, otherwise the selection grows to the next match.
+    vimState.currentMode = ModeName.Normal;
+
+    return performSearchMovement(currentSelection, vimState, direction, isExact);
+}
+
+function performSearchMovement (needle: string | undefined, vimState: VimState, direction: SearchDirection, isExact: boolean) {
+    if (needle === undefined || needle === null || needle.length === 0) {
       return vimState;
     }
 
-    // For an exact search we need to use a regex with word bounds.
     const searchString = isExact
-      ? `\\b${currentWord}\\b`
-      : currentWord;
+      ? `\\b${needle}\\b`
+      : needle;
 
     // Start a search for the given term.
     vimState.globalState.searchState = new SearchState(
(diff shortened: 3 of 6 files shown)


package.json
```diff
@@ -441,6 +441,11 @@
         "vim.statusBarColors": {
           "type": "object",
           "description": "Customize colors per mode when VSCodeVim controls status bar colors"
+        },
+        "vim.visualstar": {
+          "type": "boolean",
+          "description": "When pressing * or # in visual mode treat it as a new search",
+          "default": true
```
Chillee: Should this be "default": true?

mikew: Good catch, that may be why the tests were inexplicably passing ...

johnfn: I'm fine with toggling bonus features like this on, actually.

README.md
```diff
@@ -180,6 +180,10 @@ These settings are specific to VSCodeVim.
     }
 ```
 
+#### `"vim.visualstar"`
+* When pressing `*` or `#` in visual mode treat it as a new search
```
johnfn: Could we rephrase this slightly:

"In visual mode, start a search with * or # using the current selection."

package.json
Chillee: I think the difference is that this one affects default behavior.

src/actions/actions.ts
```diff
@@ -1739,6 +1781,10 @@ class CommandSearchCurrentWordExactForward extends BaseCommand {
   runsOnceForEachCountPrefix = true;
 
   public async exec(position: Position, vimState: VimState): Promise<VimState> {
+    if (Configuration.visualstar && vimState.currentMode !== ModeName.Normal) {
```
johnfn: Instead of a custom dispatch on current mode, can we have two separate commands, one for ModeName.Normal, and one for ModeName.Visual/Line?

src/actions/actions.ts
```diff
@@ -1723,12 +1710,67 @@ function searchCurrentWord(position: Position, vimState: VimState, direction: Se
...
+    return performSearchMovement({
+      needle: currentSelection,
+      vimState,
+      direction,
+      isExact,
+      searchStartCursorPosition
+    });
+}
+
+interface IPerformSearchMovementArgs {
```
johnfn: I'm fine with inlining this in `performSearchMovement` arguments.

(When you extract it to an interface, I subconsciously think that we're using it elsewhere in the code.)

src/actions/actions.ts
```diff
@@ -1723,12 +1710,67 @@ function searchCurrentWord(position: Position, vimState: VimState, direction: Se
...
+    needle: string | undefined;
```
johnfn: Better written as `needle?: string`

src/actions/actions.ts
```diff
@@ -1723,12 +1710,67 @@ function searchCurrentWord(position: Position, vimState: VimState, direction: Se
...
+    vimState: VimState;
+    direction: SearchDirection;
+    isExact: boolean;
+    searchStartCursorPosition: Position;
+}
+
+function performSearchMovement(args: IPerformSearchMovementArgs) {
```
johnfn: Suggestion to destructure `args` here e.g.

`const { needle, isExact, vimState } = args`

Removes a lot of `args.` throughout this function and should improve readability ever so slightly.

src/actions/actions.ts
```diff
@@ -1723,12 +1710,67 @@ function searchCurrentWord(position: Position, vimState: VimState, direction: Se
...
+    if (args.needle === undefined || args.needle === null || args.needle.length === 0) {
```
johnfn: No need to check against null because you said args is `string | undefined`

src/actions/actions.ts
johnfn: Can you give this function a better name? Possibly `moveCursorToNextSearchMatch`.

src/actions/actions.ts
mikew: Yeah I asked about this on slack, because the inlined arguments made the linter complain about the line length. So the question was numerous arguments over a new line versus an interface.

README.md
mikew: Gladly, I was never sure what the wording should be.

src/actions/actions.ts
mikew: Ha, that's how I had things originally.

src/actions/actions.ts
mikew: I'm fine with making the name more clear if you think it's ambiguous, but it's really doing two things: performing the search and making the movement.

src/actions/actions.ts
johnfn: Oh I see. If you indent every line of the object you should be good right?

src/actions/actions.ts
mikew: I'll just go back to inlining them

johnfn: It's just that "performing the search" feels a bit general to me. You mean we create the current search state right? 

Maybe `createSearchStateAndMoveToMatch`?

mikew:
All those changes are in, let me know what you think

src/actions/actions.ts
```diff
@@ -1702,27 +1702,55 @@ class CommandCmdA extends BaseCommand {
...
+    // If going right then use `getRight()` on the selection end. This ensures
+    // that any matches happen outside of the currently selected word.
+    const searchStartCursorPosition = direction === SearchDirection.Backward
+      ? vimState.lastVisualSelectionStart.getLeft()
+      : vimState.lastVisualSelectionEnd.getRight();
+
+    return createSearchStateAndMoveToMatch(currentSelection, vimState, direction, false, searchStartCursorPosition);
+}
+
+function createSearchStateAndMoveToMatch(
```
johnfn: Ohh wait a sec. I mean like this:
```
function createSearchStateAndMoveToMatch(args: {
 needle: string | undefined,
 vimState: VimState,
 direction: SearchDirection,
 isExact: boolean,
 searchStartCursorPosition: Position,
}
```

IMO it's important to use an object here so we associate names with the arguments. Otherwise when we pass in the 5 different args to the function it's hard to remember what each one is.

johnfn:
One last thing - sorry about that, I could have been more clear! Then we should be good to go. :)

src/actions/actions.ts
mikew: Ohhhhh, I didn't know that was possible in typescript.

johnfn: Yeah! :) Fun fact, if you can specify a good default for each, you can even auto-destructure the arguments:

`function doStuff({ a = 1, b = "foo" })`

johnfn: (That's not possible here - you can't really get a default VimState - so don't worry about it)

mikew:
Heck of a lot of work for a long line.

And in the end, a line was added that's just on the cusp of the limit.

src/actions/actions.ts
mikew: I knew object destructuring defaults from ES6, wasn't sure how to use them in typescript with type information as well.

johnfn:
To be fair, it's less about the length of the line, and more about trying to move this codebase in the direction of explicitly naming all arguments when argument length gets past a certain point. I find `foo(bar, baz, undefined, true, 5)` to be a lot less readable than (say) `foo({ name: bar, type: baz, deleteLine: true })`. 

Thank you for making those changes, and thank you again for the great PR. I (and all VSCodeVim users!) will really appreciate it. :)

mikew:
I'm with you on the number of arguments, any more than 2 or 3 really.

PR #1617, opened 2017-05-01
Fixes #1588

### LINKED ISSUE #1588: Ctrl+A and Ctrl+X when cursor placed after only number

--------

### What did you do?

Editing a JSON file with the following content
```Javscript
{
   id: 1,
   someOtherId: 1
}
```
Press `j` + `$` to bring the cursor to end of the first line, then press `Ctrl` + `A`

Nothing, the cursor was not in the correct position (before or over the last number in the current line).

### What happened instead?

```Javascript
{
    id: 1,
    id: 2,: 1, 
}
```

### Technical details:

* VSCode Version: 1.11.2
* VsCodeVim Version: 0.6.20
* OS: Windows 10

Chillee:
Comes from a bug in the IterateWords helper function used. The function should stop at the end of the line, but since wordEnd() isn't at lineEnd(), it continues onto the next line and takes characters from there.

Fixes another small bug as well. Namely, 
```
|-5
abcd
```
does not do right things.

### DIFF
--- a/src/motion/position.ts
+++ b/src/motion/position.ts
@@ -264,6 +264,7 @@ export class Position extends vscode.Position {
     }
   }
 
+  // Iterates through words on the same line, starting from the current position.
   public static *IterateWords(start: Position): Iterable<{ start: Position, end: Position, word: string }> {
     const text = TextEditor.getLineAt(start).text;
     let wordEnd = start.getCurrentWordEnd(true);
@@ -279,7 +280,7 @@ export class Position extends vscode.Position {
         return;
       }
       start = start.getWordRight();
-      wordEnd = start.getCurrentWordEnd();
+      wordEnd = start.getCurrentWordEnd(true);
     } while (true);
   }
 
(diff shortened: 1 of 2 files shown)


src/motion/position.ts
```diff
@@ -275,11 +276,11 @@ export class Position extends vscode.Position {
         word: word,
       };
 
-      if (wordEnd.isLineEnd()) {
+      if (wordEnd.getRight().isLineEnd()) {
         return;
       }
       start = start.getWordRight();
-      wordEnd = start.getCurrentWordEnd();
+      wordEnd = start.getCurrentWordEnd(true);
```
johnfn: Gah - we need to use more keyword arguments. e.g. i have no idea what this `true` means, but if it was `start.getCurrentWordEnd({ exclusive: true })`, that would be much better.

Not pertinent to this PR, but it's been on my mind.

johnfn:
LGTM. Thanks for the PR! :)

Chillee:
I think we also just need to comment and explain all of our functions haha.

johnfn:
That would help too. :) I like object arguments because they are like enforced comments.

PR #1620, opened 2017-05-01
Further work on tag matching (based off of #1454)

Chillee:
The only bug that isn't fixed right now (that I can tell), is dit on a tag that's empty (I added a jank experimental fix for cit for empty tags). Otherwise, I think this PR is in a pretty good place, and I'd love to hear about any more bugs. Specifically, I fixed the latter two bugs linked [here](https://github.com/VSCodeVim/Vim/pull/1454#issuecomment-296725399)

EDIT: I have now fixed all of the bugs I know of.

This is building off of work done by @jrenton in #1454.


johnfn:
Big thanks to @jrenton for kicking this off and @Chillee for finishing it up. Props to both of you.

PR #1621, opened 2017-05-01
Fixes #1196, #1197: d}/y} not working correctly

### LINKED ISSUE #1196: d} deletes extra (empty) line

-----

### What did you do?
Given
```
|foo

bar
```
where `|` marks the cursor position (and isn't actually a character),
press `d}`

In (my) terminal vim -
```
|
bar
```

### What happened instead?
```
|bar
```
Notice that it also deleted the empty line between them.

### Technical details:

* VSCode Version: 1.8.1
* VsCodeVim Version: 0.4.10
* OS: OS X El Capitan Version 10.11.6

### LINKED ISSUE #1197: y}p handles newlines incorrectly

-----

I'm going to give two variations that are handled incorrectly, but that I also expect might require separate logic.

### What did you do?
```
|foo
fun

bar
```
where `|` marks the cursor position (and isn't actually a character),
press `y}p`

```
foo
|foo
fun
fun

bar
```
Notice that when cursor is at the beginning of the line, it treats it like a line copy and starts a new line when you paste. It also includes the newline after the second (final) line.

### What happened instead?
```
f|foo
fun
oo
fun

bar
```
Here it starts pasting after the first character instead.

### What did you do?
```
f|oo
fun

bar
```
press `y}p`

```
fo|oo
funo
fun

bar
```
Notice that this does _not_ copy the newline after the second (final) line.

### What happened instead?
```
fo|oo
fun
o
fun

bar
```
Here it _does_ copy the newline after the second (final) line.

I don't think I've ever personally used variation 2, but I checked how it was handled while comparing the first one, which I use fairly often.

### Technical details:

* VSCode Version: 1.8.1
* VsCodeVim Version: 0.4.10
* OS: OS X El Capitan Version 10.11.6

Chillee:
This fixes #1196, which was a lot harder than it seemed. I also refactored a tiny part of modeHandler that made it very difficult/impossible to fix this.

I also fixed #1197 

However, the fix also actually broke a test, but I think it actually broke the test so it actually has correct behavior now! I've committed the test I changed as well.

### DIFF
--- a/src/mode/modeHandler.ts
+++ b/src/mode/modeHandler.ts
@@ -1209,16 +1209,7 @@ export class ModeHandler implements vscode.Disposable {
     vimState.recordedState.count = 0;
 
     // Keep the cursor within bounds
-
-    if (vimState.currentMode === ModeName.Normal && !recordedState.operator) {
-      for (const { stop, i } of Range.IterateRanges(vimState.allCursors)) {
-        if (stop.character >= Position.getLineLength(stop.line)) {
-          vimState.allCursors[i].withNewStop(
-            stop.getLineEnd().getLeft()
-          );
-        }
-      }
-    } else {
+    if (vimState.currentMode !== ModeName.Normal || recordedState.operator) {
       let stop = vimState.cursorPosition;
 
       // Vim does this weird thing where it allows you to select and delete
@@ -1258,12 +1249,8 @@ export class ModeHandler implements vscode.Disposable {
       }
 
       if (!cachedMode.isVisualMode && cachedRegister !== RegisterMode.LineWise) {
-        if (Position.EarlierOf(start, stop) === start) {
-          stop = stop.getLeft();
-        } else {
-          stop = stop.getRight();
-        }
-      }
+        stop = stop.getLeftThroughLineBreaks(true);
+     }
 
       if (this.currentModeName === ModeName.VisualLine) {
         start = start.getLineBegin();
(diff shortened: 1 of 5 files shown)


src/mode/modeHandler.ts
```diff
@@ -1256,11 +1256,7 @@ export class ModeHandler implements vscode.Disposable {
       }
 
       if (!cachedMode.isVisualMode && cachedRegister !== RegisterMode.LineWise) {
-        if (Position.EarlierOf(start, stop) === start) {
```
Chillee: This is a bizarre if condition, considering that we set start to the earlier of [start, stop], unless start is equal to stop.

src/mode/modeHandler.ts
```diff
@@ -1211,7 +1211,7 @@ export class ModeHandler implements vscode.Disposable {
     if (vimState.currentMode === ModeName.Normal && !recordedState.operator) {
       for (const { stop, i } of Range.IterateRanges(vimState.allCursors)) {
         if (stop.character >= Position.getLineLength(stop.line)) {
-          vimState.allCursors[i].withNewStop(
+          vimState.allCursors[i] = vimState.allCursors[i].withNewStop(
```
Chillee: hahaha there's the cause of so many of our test errors. Is there not some kind of lint setting for this?

johnfn:
If you say this is good to go then I am good with merging it. :)

Chillee:
Lemme add a couple more integration tests, and then it should be good to go. Thanks!

Chillee:
Alright repushed my tests (that I didn't push before). This should be good to go!

PR #1629, opened 2017-05-02
Fixes #1503: Undo history isn't kept when switching tabs

### LINKED ISSUE #1503: 'u' cannot undo changes if the foucs on the current tab has been switched out and back.

--------

### What did you do?
Open several files in different tabs.
Edit in one tab.
Switch focus to any other tab.
Switch focus back to the modified file.
Press 'u'.

Undo the changes.

### What happened instead?
Nothing happened.

### Technical details:

* VSCode Version: 1.11.1
* VsCodeVim Version: 0.6.15
* OS: Windows 7

Chillee:
It seems that whenever we switch editors, we don't have access to cursorPosition until after the stack clears?

Whatever it is, it's similar to this PR: https://github.com/VSCodeVim/Vim/pull/1605

### DIFF
--- a/extension.ts
+++ b/extension.ts
@@ -73,27 +73,23 @@ let modeHandlerToEditorIdentity: { [key: string]: ModeHandler } = {};
 let previousActiveEditorId: EditorIdentity = new EditorIdentity();
 
 export async function getAndUpdateModeHandler(): Promise<ModeHandler> {
-  const oldHandler = modeHandlerToEditorIdentity[previousActiveEditorId.toString()];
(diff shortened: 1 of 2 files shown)


johnfn:
Yeah that is weird and probably should be investigated further.

This is good to go though? No bad conflict with e.g. gd I assume? :)

Chillee:
It all seems to work for me. gd works properly, undo history seems to work even when switching files, whether it's from gd or ctrl+tab.

After playing around with it for 5 minutes or so, I couldn't get it to break, so I think it's good to go.

PR #1633, opened 2017-05-03
Fixes #1309: Limiting infinite recursion in macros

### LINKED ISSUE #1309: Limit infinite recursion in macros

Chillee:
After fixing this and testing it out, I'm not actually sure it's worth fixing. Our macros are slow enough that a value like 1000 takes a long time to execute anyways, and I feel like if we're just gonna limit a recursive macro to a depth of 100, why even allow them at all?

Considering that you can just interrupt the recursive macro anyways by switching tabs, I'm not sure this is worth fixing.


johnfn:
I was gonna suggest a value more like... 5.... haha. Which could be a setting maybe?

src/mode/modeHandler.ts
```diff
@@ -186,7 +186,7 @@ export class VimState {
   public cursorPositionJustBeforeAnythingHappened = [ new Position(0, 0) ];
 
   public isRecordingMacro: boolean = false;
-  public isReplayingMacro: boolean = false;
+  public isReplayingMacro: number = 0;
```
jpoon: Can you rename this variable to something along the lines of `replayingMacroCount`

src/mode/modeHandler.ts
```diff
@@ -1446,7 +1448,7 @@ export class ModeHandler implements vscode.Disposable {
             await this.handleMultipleKeyEvents(keyStrokes);
           }
 
-          vimState.isReplayingMacro = false;
+          vimState.isReplayingMacro --;
```
jpoon: Doing this, isn't the count 0 at the end of every iteration? So effectively isReplayingMacro will always be either 0 or 1?

Chillee: The idea is that macros can be recursive, so we might call this code again inside `             vimState = await this.runMacro(vimState, recordedMacro);
`
vimState is a sort of global state.

jpoon: That makes sense. Only comment is the variable name. LGTM.

Chillee: Wait, I'm not sure I want to merge this yet. Although this does fix the bug mentioned, I think the other PR I submitted for this issue(https://github.com/VSCodeVim/Vim/pull/1635) may be a better long term solution. I have a lot of work coming up this week so I'll take another look in a week or two.

jpoon:
As this PR has gone stale, I'm going to close it. Feel free to re-open when ready for review/merge.

PR #1635, opened 2017-05-03
WIP: Alternative solution to fixing #1309: hash based solution

Chillee:
Upon fixing the other one with an enforced recursive depth, it seemed to me like a pretty worthless fix. If the limit is set at something like 1000, then it'll take like 20 seconds to reach that limit, and by then the better solution would have been to switch tabs and switch back.

On the other hand, if you set it at something like 100, then why even allow recursive macros at all? Just disallow them explicitly.

This solution hashes the state at each step (the document text and the cursor positions), and only stops when the hashes are equal. I think this is a better solution. We can also take this approach for count commands.

Both of these don't work as is btw. They both have a bunch of bugs. I'm mostly posting this one for feedback.

PS: Oh yeah. I do use a nodejs library for obtaining object hashes, but there's no reason they're needed.


johnfn:
Whoa, interesting, but seems very subtle to get correct. For example how does it work with setting marks? `ma` doesn't move any cursors or change any text but could be followed by `'a` later.

johnfn:
This however is a really fun solution, and I feel it could also be applied to our problems with say `1000w` where we'd apply the word motion 1000 times but really they hit the end of document after like 200 and we just wasted 800x cycles for some large value of x. 

Dunno if anyone runs into that problem but it's nice not to crash VSCodeVim when users do stuff like that.

johnfn:
...we just need to choose a hash big enough that the birthday paradox over how many thousands of DAU still has a very low chance of causing a collision 😄

Chillee:
I'm having a hard time imagining any kind of recursive macro you could write that would break. 

Since the hash is only computed every time you call a new macro, you would need a macro that needs to be called repeatedly that only has a mark action. 

Besides, if it ever becomes an issue, we can just just use that to compute the hash too. 

The one thing I'm somewhat concerned about right now is whether or not I'm actually using the document to hash. I guess loading the document is a lot faster than I thought? 

Performance wise, it definitely does make a difference (about 23 to 26 seconds for deleting first 1000 lines in actions.ts, so I think it's  probably a good idea to only enable it once the recursive depth is more than 1.

Chillee:
I'm not interested in this solution anymore. Long term, I think the right thing to do is offload macros to neovim.

PR #1642, opened 2017-05-03
Refactor

### LINKED ISSUE #1630: diw on trailing whitespace deletes newline
- **VSCode Version**: 1.11
- **VsCodeVim Version**: 0.6.20
- **OS**: Arch

**What happened**:
Given 
```
foo bar |
fun
```
I pressed `diw` and got
```
foo bar|un
<!--
Exact keys you pressed.
-->
```

```
foo ba|r
```

**Note**:

`ciw` gives

```
foo bar|fun
```

But after PR #1621, it will also the same thing as `diw` (not intended behavior)

rebornix:
* [x] Move things out of actions.
* [x] Fix #1630


rebornix:
Ping @johnfn @xconverge , jumping inside `actions.ts` now is a real pain so I'm thinking about separating actions to `motion`, `operator`, `textobject` and plugins.

xconverge:
works for me, it seemed to bother people, I don't have strong opinions on it but looks like a good idea!

johnfn:
Ooh it's happening!!! I am 100% in favor of this.

Chillee:
I actually started doing this myself, so I could finish it if none of you guys have any particular opinions about how it's done?

I was pretty much just planning on separating it similarly to how rebornix was suggesting.

johnfn:
Basically I think @rebornix's solution of subdividing it into different files within actions.ts is exactly correct. We already have very natural subdivisions for the most part - motions, commands, operators. Motions subdivide into normal motions and text objects. Commands could subdivide into normal mode commands and special insert mode commands. After that we could think about going further if necessary.

rebornix:
1. Moving motion/operator/textobject out as they are basic elements
2. Moving plugins out
3. Moving insert command out
4. Now actions only contain commands which are composed by above elements.
5. Moving things to `src/common` if they are common logic code
6. As we have more *layers*, I enable absolute path module resolution, so now if you want to reference an operator, do `src/actions/operator` instead of `../../xyz/operator`

PR #1644, opened 2017-05-04
Fixes #1256 and #394: Fixes delete key and adds functionality

### LINKED ISSUE #1256: Inconsistencies with visual mode deletion

-----

### What did you do?

1. While in normal mode, use mouse to select any amount of text, triggering visual mode selection.
2. Press `delete` key (not backspace).

I'd expect the same as if I had used `d` to delete the text after selecting it:

Selected text is deleted, and editor returns to normal mode.
Cursor is displayed as block cursor, indicating normal mode.

### What happened instead?

Selected text is deleted, and editor remains in visual mode.
Cursor is bar cursor, indicating visual mode.

### What did you do?

1. Select text as before.
2. a) Press `delete` or `d`, then press a vertical movement key: `arrow up`, `arrow down`, `j`, or `k`.
    b) Press `delete`, then press a horizontal movement key. (Horizontal movement works as expected for deletion with `d`.)

Moving the cursor originates movement from its currently displayed location.

### What happened instead?

Movement occurs relative to the character position of what was the end of the selection before deleting it. This leads to a horizontal jump when moving the cursor vertically (after either `delete` or `d`), or a sudden visual mode selection when moving the cursor horizontally (only after `delete`).

### Technical details:

* VSCode Version: 1.8.1
* VsCodeVim Version: 0.5.2
* OS: Ubuntu 16.10

Chillee:
Otherwise, this PR fixes the first part of #1256 through just adding more key capturing, and fixes the second part by setting desiredColumn in visual mode.

### DIFF
--- a/package.json
+++ b/package.json
@@ -117,7 +117,7 @@
       {
         "key": "Delete",
         "command": "extension.vim_delete",
-        "when": "editorTextFocus && vim.active && vim.mode == 'Normal' && !inDebugRepl"
+        "when": "editorTextFocus && vim.active && !inDebugRepl"
       },
       {
         "key": "tab",
--- a/src/actions/commands/actions.ts
+++ b/src/actions/commands/actions.ts
@@ -2204,6 +2204,9 @@ class ActionDeleteCharWithDeleteKey extends BaseCommand {
   canBeRepeatedWithDot = true;
 
   public async execCount(position: Position, vimState: VimState): Promise<VimState> {
+      // If <del> has a count in front of it, then <del> deletes a character
+      // off the count. Therefore, 100<del>x, would apply 'x' 10 times.
+      // http://vimdoc.sourceforge.net/htmldoc/change.html#<Del>
       if (vimState.recordedState.count !== 0) {
         vimState.recordedState.count = Math.floor(vimState.recordedState.count / 10);
         vimState.recordedState.actionKeys = vimState.recordedState.count.toString().split("");
--- a/src/actions/commands/actions.ts
+++ b/src/actions/commands/actions.ts
@@ -2203,17 +2203,15 @@ class ActionDeleteCharWithDeleteKey extends BaseCommand {
   runsOnceForEachCountPrefix = true;
   canBeRepeatedWithDot = true;
 
-  public async exec(position: Position, vimState: VimState): Promise<VimState> {
-    // N<del> is a no-op in Vim
-    if (vimState.recordedState.count !== 0) {
-      return vimState;
-    }
-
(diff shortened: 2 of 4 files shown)


src/actions/actions.ts
```diff
@@ -2154,7 +2151,7 @@ export class ShiftYankOperatorVisual extends BaseOperator {
 
 @RegisterAction
 export class DeleteOperatorXVisual extends BaseOperator {
```
Chillee: Decided not to change the name because <Del> is just a synonym in visual mode. Come to think of it, can't we combine `d` in visual mode with this as well?

package.json
```diff
@@ -117,7 +117,7 @@
       {
         "key": "Delete",
         "command": "extension.vim_delete",
-        "when": "editorTextFocus && vim.active && vim.mode == 'Normal' && !inDebugRepl"
+        "when": "editorTextFocus && vim.active && !inDebugRepl"
```
xconverge: does delete work in insert mode now?

Chillee: I think it still passes through to VSCode since none of our actions capture `<Del>` in InsertMode.

rebornix:
@Chillee sorry I can't simply rebase this one as it somehow leads to circular dependency.

src/actions/commands/actions.ts
```diff
@@ -2203,17 +2203,15 @@ class ActionDeleteCharWithDeleteKey extends BaseCommand {
...
-    }
-
-    const state = await new operator.DeleteOperator(this.multicursorIndex).run(vimState, position, position);
-
-    state.currentMode = ModeName.Normal;
-
-    return state;
+  public async execCount(position: Position, vimState: VimState): Promise<VimState> {
+      if (vimState.recordedState.count !== 0) {
+        vimState.recordedState.count = Math.floor(vimState.recordedState.count / 10);
```
xconverge: What is going on here, not sure I follow

Chillee: http://vimdoc.sourceforge.net/htmldoc/change.html#<Del>

What the <Del> key does if you have a count in front of it is to remove the last number from the count. So if you typed 100<Del>x, then you would only delete 10 characters.

So I simply divide the count by 10 (which removes the last number).

Chillee: Come to think of it though, this might not work so well if `<Del>` is pressed when it shouldn't be pressed (in the middle of another operator for example).

johnfn: Yeah could we leave a comment explaining what's going on here?

Chillee: Just did :+1:

PR #1653, opened 2017-05-05
Removing failing tests to make Travis more meaningful.

Chillee:
We have a bunch of tests that fail randomly, so I think it's time we remove them to make travis tests more meaningful.

I'm just gonna start commenting out tests that seem to fail very often on travis (but work fine locally).

### DIFF
--- a/src/actions/commands/actions.ts
+++ b/src/actions/commands/actions.ts
@@ -1348,14 +1348,20 @@ export class PutCommandVisual extends BaseCommand {
   public async exec(position: Position, vimState: VimState, after: boolean = false): Promise<VimState> {
     let start = vimState.cursorStartPosition;
     let end = vimState.cursorPosition;
-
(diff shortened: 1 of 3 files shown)


src/actions/commands/actions.ts
```diff
@@ -1348,14 +1348,20 @@ export class PutCommandVisual extends BaseCommand {
   public async exec(position: Position, vimState: VimState, after: boolean = false): Promise<VimState> {
     let start = vimState.cursorStartPosition;
     let end = vimState.cursorPosition;
-
+    const isLineWise = vimState.currentMode === ModeName.VisualLine;
     if (start.isAfter(end)) {
       [start, end] = [end, start];
     }
+    // The reason weneed to handle Delete and Yank separately is because of
```
jpoon: > weneed
typo

PR #1665, opened 2017-05-06
Add :close support based on :quit

mspaulding06:
This pull request implements the close command.  It is largely based on the existing quit command.  The difference is that it will refuse to close the last window as stated in the vim documentation.

http://vimdoc.sourceforge.net/htmldoc/windows.html#:close


johnfn:
Looks good to me. Thanks for the PR, @mspaulding06! Much appreciated! 😄

Issue #1395, opened 2017-03-13
Change log link to release text.


jesterKing:
Please add change log links to the (future) release texts as well, i.e.

https://github.com/VSCodeVim/Vim/compare/v0.6.4...v0.6.5, or with link text [Change log](https://github.com/VSCodeVim/Vim/compare/v0.6.4...v0.6.5)

That makes it so much easier to find out what has changed between the releases.

johnfn:
Which release texts are you referring to?

jesterKing:
The release descriptions here: https://github.com/VSCodeVim/Vim/releases

These are the release texts linked to from the vs code extension changelog section in vs code.

johnfn:
You mean my little descriptions aren't helpful??? 😛 

The whole point of me writing them up was so we couldn't have to refer you to the v0.6.4...v0.6.5 thing which is just a barrage of commit messages that no one except maybe me, @rebornix and @xconverge can read.

jesterKing:
You can add me to the list :-)

Those not interested can just not click the link. I prefer to have easy access to the commit history.

jpoon:
This isn't currently supported by Travis (https://github.com/travis-ci/travis-ci/issues/6423) so we'll have to do it manually every time we do a release :(

jpoon:
We're doing this manually. Closing this issue.

PR #1670, opened 2017-05-08
Actually add stuff into CHANGELOG.MD

Chillee:
I just basically copy pasted the release notes into there, but if other people have suggestions for better formatting, I'd gladly take them.


jpoon:
If we could auto-sync release-notes or commit history -> changelog that would be great: https://github.com/VSCodeVim/Vim/issues/1395

Chillee:
I'm not sure that's actually that bad, since we have to write up release notes anyways. We just copy paste it into CHANGELOG.MD every time.

PR #1677, opened 2017-05-08
Changes tabs to navigate inside the same split

vinicio:
Commands `gt` and `gT` are currently navigating through all opened tabs, instead of navigating inside a split. This change let the behavior more like the default `CTRL-Tab` shortcut, that navigates in files inside the same split/tab group.


xconverge:
I think this makes a lot of sense, thanks!

jpoon:
Thanks @vinicio

PR #1680, opened 2017-05-10
feat(surround): support complex tags surround

admosity:
Fixes issue with complex tag surround by only including the actual tag name instead of the whole contents of the tag.

Before:
image

After:
image

Couldn't find any tests written for surround... So no tests. If someone lets me know where I should put tests for this; I'll update the PR.


xconverge:
let me add a few surround tests real quick, then you can add yours there! :)

xconverge:
There you go, could you add one or two to test/plugins/surround.test.ts

Thanks!

admosity:
Thanks! Updated.

PR #1707, opened 2017-05-17
Fixes #1436: extension prevents 'find all references' pop-up from closing through <esc> if it's empty.

### LINKED ISSUE #1436: This extension prevents 'find all reference' pop-up from closing using ESC key

-----
Original issue: [link](https://github.com/Microsoft/vscode/issues/23135)

### What did you do?
Activate 'find all reference' on something that is not a symbol
Press ESC to close the popup and nothing happens

shift + f12
ESC

The pop-up should disappear

### What happened instead?
The pop-up still exists

### Technical details:

* VSCode Version: 1.10.2
* VsCodeVim Version:  0.6.11
* OS: windows 10.0.14393

Chillee:
So pretty much the issue is that here, because nothing pops up, vscode doesn't think we're in the "find all references" context, and lets vscodevim capture keystrokes again.

Thus, all we need to is call API calls that close stuff when the user presses `<Esc>` in normal mode. Doesn't seem to have any performance impact, and fixes an annoyance.

### DIFF
--- a/src/actions/commands/actions.ts
+++ b/src/actions/commands/actions.ts
@@ -431,8 +431,9 @@ class CommandEsc extends BaseCommand {
 
   public async exec(position: Position, vimState: VimState): Promise<VimState> {
     if (vimState.currentMode === ModeName.Normal && !vimState.isMultiCursor) {
-      // If there's nothing to do on the vim side, we might as well call vscode's default "close notification" actions.
-      await vscode.commands.executeCommand('workbench.action.closeMessages');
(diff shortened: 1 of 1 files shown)


xconverge:
do we need to await even? up to others, I think this is fine with me

Chillee:
I honestly don't know. In general, the executeCommand is asynchronous, and it just seemed safer to me to just await the results.

Chillee:
Alright I think this is fine.

PR #1708, opened 2017-05-17
Fixes #1655: Extends gf to line numbers

### LINKED ISSUE #1655: gf doesn't work with filepath:linenumber paths

--------
**BUG REPORT:**

**Environment**:

- **VSCode Version**: 1.12.1
- **VsCodeVim Version**: 0.7.0
- **OS**: macOS Sierra 10.12.4

**What happened**:

When placed in NORMAL mode ontop of a filepath and trying to open another file using `gf` it doesn't work if the filepath is followed by :linenumber as one might see in a build or test error-log.

The file opened includes the :linenumber and you thus get an "Unable to open" error from VS Code.

It works in VIM.

It should open the file, preferably on the line number in question though that's not the behaviour of VIM (VIM just opens on last location in file)

**How to reproduce it**:
In NORMAL mode place cursor on top of valid file path with `:32` afterward and press gf.

Example:
`my/path/to/a/file.txt:32`

**Additional**:

I've tracked down the problem to the Position motion implementation which has a regex defined for filenames [here](https://github.com/VSCodeVim/Vim/blob/b1a9f8483bc91abd9ceba31768429d665b1ab333/src/common/motion/position.ts#L104). It doesn't include colon due to windows support obviously. Perhaps the solution is to have a special handling in the [CommandOpenFile](https://github.com/VSCodeVim/Vim/blob/5e49c3917774da8679b4dcfa5f90461bf884bccc/src/actions/commands/actions.ts#L1844-L1869) class or event better in [getFilePathRight](https://github.com/VSCodeVim/Vim/blob/b1a9f8483bc91abd9ceba31768429d665b1ab333/src/common/motion/position.ts#L503-L505)?

Chillee:
Pretty simple, mostly. One thing to note is that I'm not sure what's the best way of setting cursor position after we change files. Any ideas?

[Link to the regex used](https://regex101.com/r/o1MJt5/2)

### DIFF
--- a/src/cmd_line/commands/file.ts
+++ b/src/cmd_line/commands/file.ts
@@ -87,18 +87,9 @@ export class FileCommand extends node.CommandBase {
       await vscode.commands.executeCommand("vscode.open", folder,
         this._arguments.position === FilePosition.NewWindow ? this.getViewColumnToRight() : this.getActiveViewColumn());
 
-      await vscode.commands.executeCommand("cursorMove", {
-          to: "up",
-          by: "line",
-          value: 100000000,
-          select: false
-      });
-      await vscode.commands.executeCommand("cursorMove", {
-          to: "down",
-          by: "line",
-          value: this.arguments.lineNumber ? this.arguments.lineNumber - 1 : 0,
-          select: false
-      });
+      if (this.arguments.lineNumber) {
+        vscode.window.activeTextEditor!.revealRange(new vscode.Range(new vscode.Position(this.arguments.lineNumber, 0), new vscode.Position(this.arguments.lineNumber, 0)));
+      }
     }
   }
 
--- a/src/cmd_line/commands/file.ts
+++ b/src/cmd_line/commands/file.ts
@@ -13,6 +13,7 @@ export enum FilePosition {
 export interface IFileCommandArguments extends node.ICommandArgs {
   name?: string;
   position?: FilePosition;
+  lineNumber?: number;
 }
 
 export class FileCommand extends node.CommandBase {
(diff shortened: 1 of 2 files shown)


src/cmd_line/commands/file.ts
```diff
@@ -85,6 +86,19 @@ export class FileCommand extends node.CommandBase {
       let folder = vscode.Uri.file(newFilePath);
       await vscode.commands.executeCommand("vscode.open", folder,
         this._arguments.position === FilePosition.NewWindow ? this.getViewColumnToRight() : this.getActiveViewColumn());
+
+      await vscode.commands.executeCommand("cursorMove", {
```
xconverge: How do we handle the :50 command? This should be done the same way, this hack should not be necessary for sure... There should also be a vscode command to go to line

Chillee: Can we do `:50` to open a new file?

xconverge: No, go to line. The file action looks good to me, then jumping to the right line needs to be cleaned up

Chillee: It's no problem to do `:50` in the same file, as we can just move the cursor ourselves. The problem is that once we call the API to open a new file, we don't have access to the new modeHandler yet, in this file.

Come to think of it though, there was another part of the code where we might have needed to do something similar. I'll take a look at that later.

xconverge: What about a vscode command to go to line?

src/cmd_line/commands/file.ts
```diff
@@ -85,6 +86,10 @@ export class FileCommand extends node.CommandBase {
       let folder = vscode.Uri.file(newFilePath);
       await vscode.commands.executeCommand("vscode.open", folder,
         this._arguments.position === FilePosition.NewWindow ? this.getViewColumnToRight() : this.getActiveViewColumn());
+
+      if (this.arguments.lineNumber) {
+        vscode.window.activeTextEditor!.revealRange(new vscode.Range(new vscode.Position(this.arguments.lineNumber, 0), new vscode.Position(this.arguments.lineNumber, 0)));
+      }
```
Chillee: I guess this is probably better?

xconverge: it sure is

xconverge:
Thanks!

PR #1718, opened 2017-05-19
WIP Fixes #754: Adds j,k,o,<Enter>, gg, G, ctrl+d, and ctrl+u commands for navigating inside the file explorer

### LINKED ISSUE #754: Using j,k,l,i to navigate inside explorer list

--------

That's great if I can forget arrow keys when using VIM mode. That feature already have on Sublime & Atom but I can't find it on VSCode with VSCodeVim.

### LINKED ISSUE #1351: Revert file with `:e!`

-----

### What did you do?

`:e!`

Revert file

### What happened instead?

> Unable to open '!'

### Technical details:

* VSCode Version: 1.10.0
* VsCodeVim Version: 0.5.4
* OS: OSX Sierra

Chillee:
I just copied @rebornix 's keybindings into package.json.

Fixes #754 and fixes #1351

### DIFF
--- a/extension.ts
+++ b/extension.ts
@@ -334,8 +334,10 @@ export async function activate(context: vscode.ExtensionContext) {
   Configuration.boundKeyCombinations = [];
 
   for (let keybinding of packagejson.contributes.keybindings) {
+    if (keybinding.when.indexOf("editorTextFocus") === -1) {
+      continue;
+    }
     let keyToBeBound = "";
-
     /**
      * On OSX, handle mac keybindings if we specified one.
      */
--- a/extension.ts
+++ b/extension.ts
@@ -334,7 +334,7 @@ export async function activate(context: vscode.ExtensionContext) {
   Configuration.boundKeyCombinations = [];
 
   for (let keybinding of packagejson.contributes.keybindings) {
-    if (keybinding.when.indexOf("editorTextFocus") === -1) {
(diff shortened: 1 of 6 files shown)


Chillee:
I just wanted to note that these keybindings don't actually seem to work when I run it in debug mode. I'm not sure why.

Chillee:
Oh, I just realized why we can't do this right now ( I think).

Chillee:
Just kidding.

Chillee:
Actuall

xconverge:
does it work? I dont think it does because of duplicate bindings

Chillee:
No it doesn't. My internship's been sucking up a lot of my time so I've had less time to work on vscodevim :(

xconverge:
The problem is we still think we have an active text editor I think, even when we are focused on the list

Chillee:
@xconverge So the problem is this line (I think): https://github.com/VSCodeVim/Vim/blob/master/extension.ts#L355
`    registerCommand(context, keybinding.command, () => handleKeyEvent(`${ bracketedKey }`));
`

It binds all of our commands to `handleKeyEvent` pretty much. I'm not really sure this is the right approach, but I'd need to think more about how to handle it better (if we even can).

For now, we can just check the `keybinding.when` to see whether we should be handling it.

Chillee:
I also removed the duplicate keybindings test. I don't really think there's much benefit to it.

PR #1722, opened 2017-05-19
Fixes #1720: Removed unused <c- > bindings from package.json

### LINKED ISSUE #1720: VSCode Zen Mode Shortcut not working
In the newer versions of VSCode they added a shortcut for the Zen-Mode.
If you press ctrl+k z, the zen-mode is toggled.

However with VSCodeVim enabled, the shortcut does not work.
I do not know any ctrl+k vim shortcuts, but maybe there are. Is this behavior expected?

**Environment**:

- **VSCode Version**: 0.7.1
- **VsCodeVim Version**: 1.12.2
- **OS**: Linux Mint 18 (But also, Windows, Mac)

**What happened**:
Nothing

Zen-Mode to toggle. If you remap to something else it works.

**How to reproduce it**:
1. open vscode
2. press ctrl+k z

Chillee:
Just to be a good extension citizen, we should only be capturing as much as we use. We can re-add them back in if we need them.

Fixes #1720

### DIFF
--- a/package.json
+++ b/package.json
@@ -140,16 +140,6 @@
         "when": "editorTextFocus && vim.active && vim.use<C-b> && vim.mode != 'Insert' && !inDebugRepl"
       },
       {
-        "key": "ctrl+j",
-        "command": "extension.vim_ctrl+j",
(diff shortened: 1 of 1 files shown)


jpoon:
Thanks @Chillee

Issue #1451, opened 2017-03-30
Imitate default vim key bindings


mightyiam:
Hey. Shouldn't the default vim keybindings be imitated by default?

I'm coming from a few years of vim and recently neovim and I have expectations. The first, is that the default set of keybindings will be implemented. For example, I couldn't make splits with bindings. At first I thought that it must be because there is no implementation of splits. But I've found out that there is.

xconverge:
We really need some more info here. Can you use the template for issues, is. What did you try, what was expected, what actually happened.

What key binding are you trying for splits.

If you look at ROADMAP.md you can see what is currently supported.

FabioAntunes:
I guess he's talking about this:

* `<C-w>n `- :new horizontal split (editing a new empty buffer)
* `<C-w>s` - :split window horizontally (editing current buffer)
* `<C-w>v` - :vsplit window vertically (editing current buffer)
* `<C-w>c` - :close window
* `<C-w>o` - close all windows, leaving :only the current window open
* `<C-w>w` - go to next window
* `<C-w>p` - go to previous window
* `<C-w><Up>` - go to window above
* `<C-w><Down>` - go to window below
* `<C-w><Left>` - go to window on left
* `<C-w><Right>` - go to window on right

mightyiam:
@FabioAntunes I was hit by some of these, specifically. In general, I would like to see a big table, in the most visible place in the readme. This table would look like something like

default vim keybinding | implementation
-------|---
`<C-w>n`- :new horizontal split (editing a new empty buffer) | :white_check_mark: opens a new editor to the right
`<C-w>s` - :split window horizontally (editing current buffer) | :construction: (link to discussion issue)

Now, this is the way I see things. I don't know how other users will see them.

I think that this kind of table would help me a lot.

Or maybe this is not a table, but a listing of issues with a certain label and a link to that issue search of both open and closed issues with this label.

johnfn:
@mightyiam you mean... something sorta like this? https://github.com/VSCodeVim/Vim/blob/master/ROADMAP.md 😉 

(tho embarrassingly we dont have the split stuff on there. huh, not sure how we missed that!)

mightyiam:
Yup. Exactly like that. The sooner the better :man_facepalming:.

So please feel free to change the title of this issue to whatever you know is missing exactly. Maybe splits.

johnfn:
[...]

PR #1795, opened 2017-06-01
Vertical split shortcut keys

beefsack:
This PR binds `C-w v` to `workbench.action.splitEditor`, achieving similar functionality to what happens in Vim.

I've previously manually maintained a shortcut to match the Vim vertical split shortcut, but thought it might be helpful to add it for everyone else. Please close if it's not a complete fit, I'm happy to maintain the shortcut configuration myself.

At the moment it's just the binding, had a look at the automated tests and not entirely sure how I'd go about testing it. If this is required, some rough direction would be greatly appreciated.

Keep up the great work, this is one of the best Vim modes I've ever used, if not the best.

Relates to #1451.

### DIFF
--- a/src/actions/commands/actions.ts
+++ b/src/actions/commands/actions.ts
@@ -2242,6 +2242,21 @@ class BaseTabCommand extends BaseCommand {
 }
 
 @RegisterAction
+class VerticalSplit extends BaseCommand {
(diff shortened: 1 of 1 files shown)


Chillee:
These look pretty good to me!

You're right that we don't have automated tests for these... It seems pretty difficult and not very helpful to do something like that.

Thanks for the PR! And thanks for the praise!

Issue #1589, opened 2017-04-27
Undo don't work after full search and replace


Matheus-de-Souza:
* Click *thumbs-up* 👍 on this issue if you want it!
* Click *confused* 😕 on this issue if not having it makes VSCodeVim unusable. 

The VSCodeVim team prioritizes issues based on reaction count. 

--------

### What did you do?

[On Normal Mode]
:%s/pedido/perfil/gi [enter] u

I did a full search and replace, but I typed bad the last part of the expression. I should type 'gI' not 'gi'

Then I tried to undo, but it did not work.

### What did you expect to happen?

I expect a "undo"

### What happened instead?

Nothing

### Technical details:

* VSCode Version: 1.11.1
* VsCodeVim Version: Lastest
* OS: Windows

Chillee:
I can't replicate this exact issue. Could you explain it a bit more on what you're doing? (Did you just open up a file, etc.)

Matheus-de-Souza:
It's wierd because sometimes it works and sometimes doesn't...

Chillee:
@Matheus-de-Souza Does it have to do with how many matches you have? Like, does it work for small files but not large ones?

Chillee:
I can replicate this consistently with any find and replace that matches more than about 23 words.

Chillee:
This is fixed with the new "enableNeovim" option (on master).

Chillee:
@johnfn I have a question. This issue is still present, but I know the cause (but not the reasoning). The issue is that when you undo, only the last 50 changes are kept, due to this line: https://github.com/VSCodeVim/Vim/blob/master/package.json#L448 combined with this line: https://github.com/VSCodeVim/Vim/blob/master/src/history/historyTracker.ts#L414

Is there any reason we have that line? Setting that line to something super high fixes all of these issues.

johnfn:
@Chillee sorry for slow response. notification spam makes it hard to find important things to respond to.

I honestly have no clue why that was added. It appears to be an incorrect overloading of the vim.history setting, which i think was initially added to track the number of SEARCH items (and maybe ex commands?) to store. I can't imagine why you'd want to truncate the undo history, that seems like a recipe for sadness. 

I'm fine with changing it back!

PR #1833, opened 2017-06-13
Fixes #1794: Undo not undoing all changes

### LINKED ISSUE #1794: Undo does not restore all changes
If you make some kind of huge change (like a large search and replace or something), and then undo, vscodevim will not undo all of the changes.

Chillee:
For whatever reason, we were cutting off the history kept.

https://github.com/VSCodeVim/Vim/issues/1589#issuecomment-305946543

This simply removes that line!


xconverge:
I think this was a misunderstanding by me by the vim history setting.

Lets just remove it and be done with it, no one wants this :)

PR #1840, opened 2017-06-15
Support Neovim plugins (and basic .vimrc support)

Chillee:
A basic proof of concept. Many things need to be added and made more general before this is ready to push, but I think it's pretty cool.

Refer to https://github.com/Chillee/Vim/blob/neovimPlugins/README.md#plugins for the plugins I've verified work so far (ie: just targets.vim so far).

### DIFF
--- a/src/neovim/nvimUtil.ts
+++ b/src/neovim/nvimUtil.ts
@@ -10,10 +10,12 @@ import { attach } from "promised-neovim-client";
 import { Register, RegisterMode } from "../register/register";
 import { ModeName } from "../mode/mode";
 
+type UndoTree = {entries: Array<{seq: number, time: number}>, save_cur: number, save_last: number, seq_cur: number, seq_last: number, synced: number, time_cur: number};
+
 export class Neovim {
 
   static async initNvim(vimState: VimState) {
-    const proc = spawn(Configuration.neovimPath, ['-u', 'NONE', '-N', '--embed'], { cwd: __dirname });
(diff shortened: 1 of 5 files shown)


src/neovim/nvimUtil.ts
```diff
@@ -9,11 +9,16 @@ import { spawn } from "child_process";
...
+import * as os from "os";
+
+type UndoTree = {entries: Array<{seq: number, time: number}>, save_cur: number, save_last: number, seq_cur: number, seq_last: number, synced: number, time_cur: number};
 
 export class Neovim {
 
   static async initNvim(vimState: VimState) {
-    const proc = spawn(Configuration.neovimPath, ['-u', 'NONE', '-N', '--embed'], { cwd: __dirname });
+    this.syncVSSettingsToVim();
+    const proc = spawn(Configuration.neovimPath, ['-u', '~/.config/nvim/init.vim', '-N', '--embed'], { cwd: __dirname });
```
justinmk: could omit the `-u ...` entirely so that Nvim looks in the usual places (on Windows also)

Chillee: I haven't added the logic yet, but I think the neovim embedding in VSCode should use a custom init.vim file. We're going to need to do a lot of autogeneration of the init.vim file for various setting sync reasons, and I'd rather not be overwriting users' original settings.

Chillee:
This will probably be a no go, in favor of the full neovim integration.

Chillee:
As I'm nearly certain this is the way to go for neovim integration: https://github.com/VSCodeVim/Vim/pull/1897, I will finally be closing this PR.

PR #1842, opened 2017-06-15
Removed solid block cursor

Chillee:
As per some investigative work by @xconverge, this setting actually makes the extension a piece of crap.

We apologize if you were using this setting and wondered how this extension was so slow.


xconverge:
LGTM

PR #1846, opened 2017-06-16
fixes #1843 A and I preceded by count

### LINKED ISSUE #1843: Append text at the end of the line (N times)

--------

**Environment**:

- **VSCode Version**: 1.14.0-insider
- **VsCodeVim Version**: 0.8.6
- **OS**: Windows 10

**What happened**: Appending for example 3afoo appends foo three times correctly. One would expect similar behavior with 3Afoo when appending to end of line. But the count part is not doing anything. It works in Vim. Same issue with insert

**What did you expect to happen**: 3Afoo to append foofoofoo to end of line

**How to reproduce it**: 3Afoo

xconverge:
Still a bit weird but works for me

### DIFF
--- a/src/actions/commands/actions.ts
+++ b/src/actions/commands/actions.ts
@@ -1974,7 +1974,7 @@ class CommandGoForwardInChangelist extends BaseCommand {
 }
 
 @RegisterAction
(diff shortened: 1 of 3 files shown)


Chillee:
LGTM

PR #1851, opened 2017-06-16
Update README.md (minor)

BlueDrink9:
Updated key for adding new cursor from gc (now comment) to gb.


Chillee:
Nice catch! Thanks!

PR #1859, opened 2017-06-21
Select in visual mode when scrolling

### LINKED ISSUE #907: Ctrl+D / Ctrl+U (page down/up) moves selection when in Visual Line mode

--------
### What did you do?

Make sure `vim.useCtrlKeys` setting is set to `true`.
Open a file that has more content than can fit on the screen.
Place cursor on the first line.
Press `Shift+V` to enter visual line mode (bug is present in visual mode as well)
Press `Ctrl+D` for a page down movement
Notice that the selection is now gone or starts in a different place

Text selection should still start on the line which user started Visual mode in.
### What happened instead?

Text selection moves to the current cursor location.
### Technical details:
- VSCode Version: 1.6.0
- VsCodeVim Version: 0.3.7
- OS: macOS Sierra

Chillee:
Based off @rebornix's changes.

We just need to make sure we wait for the right cursor after doing the view change. Otherwise, the cursor change that's a result of `revealCursor: true` isn't propagated, and we get stuck once we're at the edge.

Fixes #907,

### DIFF
--- a/src/actions/commands/actions.ts
+++ b/src/actions/commands/actions.ts
@@ -522,7 +522,7 @@ class CommandCtrlOpenBracket extends CommandEsc {
 }
 
 abstract class CommandEditorScroll extends BaseCommand {
-  modes = [ModeName.Normal];
+  modes = [ModeName.Normal, ModeName.Visual, ModeName.VisualLine, ModeName.VisualBlock];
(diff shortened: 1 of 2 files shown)


xconverge:
Ah nice!

PR #1879, opened 2017-06-28
Formattted everything with prettier

Chillee:
Before we merge this, I think it'd be a good idea to put some kind of commit hook so new contributors wouldn't have to use prettier.

However, it wouldn't be the end of the world to run prettier manually once in a while.

The settings I ran it with were
```
prettier --write --print-width 100 --single-quote --trailing-comma es5
```

Opinions on these settings?


jpoon:
My 2cents. Add a gulp task and do it manually once in awhile. As we talked about on Slack, 15seconds is a looong time to wait.

Chillee:
@jpoon What we can do is just do only prettier on the changed files. That'd reduce the time down to <1s total.

Chillee:
@jpoon For example, on the most recent commit, prettier only took 300 ms to run.

Chillee:
The only thing left to do is to fail the CI if not all the code is prettified.

Chillee:
Cool! I think everything should work. @jpoon I dunno if I did something horrendous to the gulp or travis files, so I'd appreciate a quick review on that.

gulpfile.js
```diff
@@ -1,56 +1,78 @@
...
-  var tests = gulp.src(paths.tests_ts)
+  var tests = gulp
+    .src(paths.tests_ts)
     .pipe(tslint({ formatter: 'verbose' }))
     .pipe(tslint.report(tslintOptions));
   return merge(srcs, tests);
 });
 
-gulp.task('default', ['tslint', 'compile']);
+gulp.task(
```
jpoon: Why not use https://www.npmjs.com/package/gulp-prettier?

Chillee: Mainly because that package isn't maintained, and it's missing the "es5" trailing commas option.

.travis.yml
```diff
@@ -28,6 +28,8 @@ install:
 - npm install;
 
 script:
+- git ls-tree -r master --name-only | grep ".*.[t|j]s$" | xargs ./node_modules/prettier/bin/prettier.js --write --print-width 100 --single-quote --trailing-comma es5
```
jpoon: You could have called the `gulp` task here?

Chillee: I was thinking that we should be checking all the files to see if any of them haven't been prettified.

I think I set up `gulp prettier` to only prettify files that have changed since the last commit.

Come to think of it, I think this workflow can be improved...

Issue #1875, opened 2017-06-26
Badges are all sorts of messed up


Chillee:
I had to do a bunch of weird changes to get the new vscode marketplace to accept our badges, and now the badges are all sorts of messed up :(

Chillee:
Badges look good enough to me! We still probably wanna add a slack link.

PR #1880, opened 2017-06-28
Fix badge maybe?

jpoon:
I wonder if it's the extra space? https://github.com/VSCodeVim/Vim/issues/1875


Chillee:
@jpoon The CI fail is due to an update to typescript, I believe.

Chillee:
@jpoon I fixed it on master. Those are markdown images which are alt. SO you need to wrap them in a markdown link.

PR #1909, opened 2017-07-08
Substitute global flag (like Vim's `gdefault`)

philipmat:
#### `"vim.substituteGlobalFlag"`
* Similar to Vim's `gdefault` setting.
* `/g` flag in a substitute command replaces all occurrences in the line.
  Without this argument, replacement occurs only for the first occurrence in each line.
* When `"vim.substituteGlobalFlag"` is `true`, the 'g' is default on.
  This means that all matches in a line are substituted instead of one.
  When a 'g' flag is given to a ":substitute" command, this will toggle the substitution
  of all or one match.

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,6 +146,15 @@ These settings are specific to VSCodeVim.
 #### `"vim.useSolidBlockCursor"`
 We have removed this option, due to it making VSCodeVim's performance suffer immensely.
 
+#### `"vim.substituteGlobalFlag"`
+* Similar to Vim's `gdefault` setting.
+* `/g` flag in a substitute command replaces all occurrences in the line.
+  Without this argument, replacement occurs only for the first occurrence in each line.
+* When `"vim.substituteGlobalFlag"` is `true`, the 'g' is default on.
+  This means that all matches in a line are substituted instead of one.
+  When a 'g' flag is given to a ":substitute" command, this will toggle the substitution
+  of all or one match.
+
 #### `"vim.useCtrlKeys"`
 * Enable Vim ctrl keys overriding common VS Code operations (eg. copy, paste, find, etc). Enabling this setting will:
     * `ctrl+c`, `ctrl+[` => `<Esc>`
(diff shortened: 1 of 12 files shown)


Chillee:
Seems like there are relevant tests failing?

philipmat:
@Chillee - the tests that are failing are actually the tests that I've added for this feature. To my fault, I did test with the neovim flag set to false.

Let me see install neovim and see if I can make it work with true.

Chillee:
@philipmat Do the tests run fine locally?

To be clear:
1. The code you've added only affects `:%s` for non-neovim Ex mode.
2. The current tests for Ex-mode are run for neovim Ex mode.
3. If you are having difficulty getting the tests to work properly, I am fine with having no tests for this configuration option specifically. The logic is simple and I'm almost certain there will be no bugs related to it.

philipmat:
The tests work for `enableNeovim=false` because the `/g` flag is being set in `substitute.ts according to the proposed `substituteGlobalFlag`.

I need to find a way to pass this configuration flag to Neovim so that tests with `enableNeovim=true` work just as well.

If you see more value in having this flag in right now, I'd be happy to take the offending tests out; however, if it's all the same to you I'd rather have the tests and make them work right.

Although, it just occurred to me that due to `testUtils.ts` enabling *neovim* for all tests, that effectively bypasses testing the `SubstituteCommand` from `substitute.ts`.

Chillee:
Yeah the testutils part is what I was talking about. Ex mode without neovim integration is in somewhat of a "maintenance" mode, where it's unlikely that we'll add features to it that can be covered by neovim.

Chillee:
Looks good to me! Could you add a note that it only works when not using the neovim integration for Ex-commands?

Thanks!

philipmat:
@Chillee - with [b70942e](https://github.com/VSCodeVim/Vim/pull/1909/commits/b70942ee08a32f76d8427cf7b2884eb50abc0069) it actually works with neovim too - which is why it passes the tests.

Chillee:
Ah nice. Looks good to me!

Issue #1896, opened 2017-07-02
Ctrl-C in Visual Mode should copy to clipboard


philipmat:
* Click *thumbs-up* 👍 on this issue if you want it!
* Click *confused* 😕 on this issue if not having it makes VSCodeVim unusable. 

The VSCodeVim team prioritizes issues based on reaction count. 

--------

The default behavior of gVim on Windows is to copy the selected text when using `Ctrl-C`. Effectively:

    vnoremap <C-C> "+y

Could we please get the same behavior in VSCodeVim?

Note that:
1. `Ctrl-C` behavior in insert mode is correct (takes you out of insert mode)
2. On MacOS `Cmd-C` behaves as intended, that is it copies the selected text to `+`.

**Environment**:

- **VSCode Version**: 1.13.1
- **VsCodeVim Version**: 0.9.0
- **OS**: Windows 10

**What happened**:

Selected text in visual mode then pressed `Ctrl-C`.

**What did you expect to happen**: 

Selected text to be copied to the `+` register and with `"vim.useSystemClipboard": true` to be in the system clipboard as well.

**How to reproduce it**:

- Select text in visual mode 
- Press `Ctrl-C`
- Execute `:reg` and notice that the previously selected text is not in the `+` register.

Compa0:
Agree, I can't paste to another program when using shortcut of copy or cut.

jpoon:
Apparently fixed as of v0.10.0. Please re-open if that is not the case.

yifanchen:
Anyone is aware of #3229?

PR #1912, opened 2017-07-09
Ctrl-C should copy to clipboard in visual mode - fix for #1896

philipmat:
When yanking to clipboard, the selection should also reset to normal

### DIFF
--- a/src/actions/commands/actions.ts
+++ b/src/actions/commands/actions.ts
@@ -931,6 +931,8 @@ class CommandOverrideCopy extends BaseCommand {
     }
 
     util.clipboardCopy(text);
+    // all vim yank operations return to normal mode.
+    vimState.currentMode = ModeName.Normal;
 
     return vimState;
   }
--- a/src/mode/modeHandler.ts
+++ b/src/mode/modeHandler.ts
@@ -799,8 +799,13 @@ export class ModeHandler implements vscode.Disposable {
         key = '<copy>';
       }
 
-      if (process.platform !== 'darwin' && key === '<C-c>' && !Configuration.useCtrlKeys) {
-        key = '<copy>';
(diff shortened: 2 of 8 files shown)


philipmat:
I'll look into the failing tests shortly.

Chillee:
You might already know this (I didn't know this when I first started contributing), but you can run only a select few tests using `newTestOnly`.

If you have any questions about anything, feel free to drop by our slack :)

Chillee:
Ah I've been wondering what that failing test was caused by (which was why I wasn't merging). It's caused by an upstream change (more specifically, this PR: https://github.com/Microsoft/vscode/pull/28238).

I'm fine with merging this and then fixing that upstream change (eventually).

philipmat:
Submitted PR #1926 to fix the tests.
Hopefully, if that passes and gets merged to master before all these other PRs, it should fix them too.

ferrao:
@philipmat Why do you say that `When yanking to clipboard, the selection should also reset to normal`?

All other IDEs that I use to not clear the visual selection mode when I press CTRL-C. 
This behaviour is actually very annoying to me and gets in the way of my workflow.
Is there any option to control this?

ferrao:
@philipmat Just found that `vim.overrideCopy: false` does the trick! I'm happy now!

philipmat:
@ferrao - in Vim itself - i.e. canon, if you select text and yank it (by either `y` or `Ctrl-C`) the mode switches back to Normal from Visual (or Select).

ferrao:
@philipmat Yes, but I am actually on a Mac and  `Cmd-C`/`Cmd-v` does nothing on Vim, allowing me to have two different ways of copying and pasting text around.

With `vim.overrideCopy: true` I loose that.

PR #1926, opened 2017-07-13
Update tests due to VSCode PR 28238

philipmat:
This PR: Microsoft/vscode#28238 inserts spaces inside block comments, which affects VSCodeVim tests.


Chillee:
Thanks!

PR #1932, opened 2017-07-17
Clean up easymotion code

MaxfieldWalker:
Cleaned up codes related to the easymotion plugin before adding some new features to it!

Main changes are:
- Use `const`, `let` keyword, instead of `var`
- Optimize code flow
- remove duplication of code


Chillee:
Greatly appreciated. That code hasn't been touched in a long time :)

Even with the neovim rewrite, we will still likely need to grandfather in a lot of the easymotion code; It seems difficult to port easymotion with a neovim embedding.

MaxfieldWalker:
I see you @Chillee is working on neovim integration. I'm not familiar with neovim, but neovim seems to support external operation through its API, right? As you say, it is difficult, or even impossible to achieve easymotion functionality using neovim APIs.

I was going to add some improvements, like (1) two-char move `<leader><leader><s>{char}{char}` (2) jump immediately when a single candidate found, to easymotion. Can I move forward in the old way, not the neovim way?

xconverge:
2 char moves would be great!

Chillee:
@MaxfieldWalker Ideally, we'd able to use the neovim API to simply allow us to use the entirety of the easymotion extension within vscode.

I asked a question here: https://github.com/easymotion/vim-easymotion/issues/331 but it seems that it may be tricky to support it.

Thus, we will probably have to use our own implementation anyways, as many people rely on easymotion. Feel free to do it in the old way, it's not work that will be wasted.

If you're familiar with vimscript, it may be worth seeing if it's possible to modify the easymotion code somehow so we can reuse code.

I'm going to hold off on merging things until I can fix our Travis issues.

xconverge:
It is close enough in implementation I would just add 2 char support and go from there if more stuff is requested

MaxfieldWalker:
@Chillee Yes, it should be avoided to reinvent the weel, but unfortunately, no one seems to be familiar with the original easymotion implementation(vimscript) and Neovim API to integrate easymotion features to the VSCode Vim extension. I will work on in the old way:+1:

Things I'm planning to implement are:

- [x] 2-char move support `<leader><leader><s>{char}{char}`
- [x] Jump immediately when a single candidate found
- [ ] Highlight characters that match the typed key

And more goals (if possible)

- [ ] n-character search [n-character search](https://github.com/easymotion/vim-easymotion#n-character-search-motion)

Chillee:
@MaxfieldWalker Hey could you run prettier (you can do it just by running gulp before you commit, I believe)?

MaxfieldWalker:
@Chillee Yes. I ran prettier using Bash on Window.

Chillee:
I'm a little bit worried about merging this kind of rewrite because we don't have unit tests covering this behavior.

I'll try to spend some time soon going through it, but how have you been verifying your behavior is still correct?

MaxfieldWalker:
As you say, there seems to be no tests for easymotion. Actually, I have already implemented further easymotion actions like n-char move. Let me make some tests to verify all easymotion behavior including n-char move.

Chillee:
I wouldn't be too surprised if our current testing infrastructure had issues with easymotion. If so, I'd be fine with just an assurance that you've tried out using your refactor, and that it doesn't break horribly :)

Chillee:
I think I'm going to close this PR since the other PR is built off of this one and has the tests as well.

PR #1952, opened 2017-08-05
Fixes #1950: counter for $

### LINKED ISSUE #1950: {counter}$ is not work
the counter value is overlooked

tagniam:
Implements counter for line end actions(`$`, `<End>`, `<D-right>`). Fixes #1950


Chillee:
@tagniam Could you write some unit tests covering this case?

tagniam:
@Chillee done!

src/actions/motion.ts
```diff
@@ -740,12 +740,16 @@ class MoveRepeatReversed extends BaseMovement {
 }
 
 @RegisterAction
-class MoveLineEnd extends BaseMovement {
+class MoveDownLineEnd extends BaseMovement {
```
Chillee: I think it's better to keep this as `MoveLineEnd`. That's the functionality that most people associate with this key.

Chillee:
Thanks for your work!

Issue #1935, opened 2017-07-18
Fix the effect of "ciW" when starting from the beginning of a word


uvatbc:
* Click *thumbs-up* 👍 on this issue if you want it!
* Click *confused* 😕 on this issue if not having it makes VSCodeVim unusable. 

The VSCodeVim team prioritizes issues based on reaction count. 

--------

- **VSCode Version**:
Version 1.14.1
Commit 2648980a697a4c8fb5777dcfb2ab110cec8a2f58
Date 2017-07-13T19:12:01.887Z
Shell 1.6.6
Renderer 56.0.2924.87
Node 7.4.0

- **VsCodeVim Version**: 
v0.9.0

- **OS**: 
Kubuntu 17.04

**What happened**:
Lets take the following line as an example:
"Navigate to the main page"
My cursor starts off on the "m" of "main".
I type in "ciW": edit word that I'm currently in by removing only that word.

**What did you expect to happen**: 
"Navigate to the _page", with the cursor represented by the underscore. This is what vim does.    

**What actually happened**: 
"Navigate to _ page". Yes, the word "the" is also eaten up in addition to the word "main".

**How to reproduce it**:
See above.

Chillee:
I see. This was a bug fixed for ciw that wasn't fixed for ciW, I believe.

Chillee:
This was fixed by https://github.com/VSCodeVim/Vim/pull/1977

PR #1977, opened 2017-08-22
Fixed iW on beginning of word (#1935)

Ghust1995:


Chillee:
Thanks! Tests look irrelevant to what's failing, so don't worry about it.

Ghust1995:
There was an equivalent test for lowercase w, and it was failing before, so I just added it to guarantee using diW in the beginning of the word behaved correctly.

Chillee:
Yay! I just removed the flaky tests. Thanks for the contribution!

PR #1990, opened 2017-08-31
Fixed <C-e>/<C-y> being very slow

### LINKED ISSUE #1943: Scrolling with CTRL-E is very slow

--------

**BUG REPORT**:

**Environment**:

- **VSCode Version**: 1.15.0-insider
- **OS**: Windows 10

**What happened**: Scrolling down with `CTRL-E` by more than two lines makes the scroll very slow. While the scrolling itself is very slow, you have to wait until it catches up to where it was supposed to scroll based on the time the keys were pressed. For example, pressing `CTRL-DOWN` for 2 seconds makes the scroll 2 seconds long, scrolling down by 100 lines. Pressing `CTRL-E` for 2 seconds makes the scroll 6 seconds long, scrolling down by 100 lines. 

**What did you expect to happen**: I expect the `CTRL-E` and `CTRL-DOWN` to scroll with equal speed.

**How to reproduce it**: Open a file with enough lines to scroll through and compare `CTRL-E` and `CTRL-DOWN`.

Chillee:
This is the line causing all of the issues. In general, we should try *very* hard to avoid calling `allowVSCodeToPropagateCursorUpdatesAndReturnThem()`. It's very slow.

Fixes #1943, #1985,

Ah I see why I added it. If you don't do this, then the selection will block you from scrolling down more. Sigh...

### DIFF
--- a/src/mode/modeHandler.ts
+++ b/src/mode/modeHandler.ts
@@ -43,7 +43,6 @@ import { Globals } from '../../src/globals';
 import { ReplaceState } from './../state/replaceState';
 import { GlobalState } from './../state/globalState';
 import { Nvim } from 'promised-neovim-client';
-import { allowVSCodeToPropagateCursorUpdatesAndReturnThem } from '../util';
 
 export class ViewChange {
   public command: string;
@@ -1913,7 +1912,6 @@ export class ModeHandler implements vscode.Disposable {
     for (let i = 0; i < this.vimState.postponedCodeViewChanges.length; i++) {
       let viewChange = this.vimState.postponedCodeViewChanges[i];
       await vscode.commands.executeCommand(viewChange.command, viewChange.args);
-      vimState.allCursors = await allowVSCodeToPropagateCursorUpdatesAndReturnThem();
     }
 
     // If user wants to change status bar color based on mode


Chillee:
Agh this is very frustrating. Both issues are complained about a lot: https://github.com/VSCodeVim/Vim/issues/907, and I'm not sure it's possible for us to fix both.

xahon:
Looks like prs here merged very rare. I'm waiting to fix this issue a long time

Chillee:
@xahon no. There just isn't a good way to fix this issue without breaking other behavior.

TyOverby:
@Chillee: the "cursor must be visible at all times" behavior of vim is due to legacy reasons and I would argue that it should be done away with.  

I'd understand if you wanted to keep it for compat reasons, but please give us a way to turn it off.  I've been using vim for 10 years and it's always annoyed me.  Now it seems like it's the cause of serious performance issues in vscode.

TyOverby:
@Chillee: reading a bit more into this, is there a way to get vscode-vim to replace ctrl-e/ctrl-y behavior with ctrl-up/ctrl-down behavior?

TyOverby:
@Chillee: is there any update on this?  

I use ctrl-e / ctrl-y for scrolling in vim, and this is keeping me from using vscode.  I'm at the point where I'm planning on forking the extension just for this PR.

Chillee:
@TyOverby I'm still not really sure what to do about this. However, as a temporary fix, I'll add a flag (ugh) that toggles between either breaking #907 or breaking this issue.

addepar-andy:
@Chillee: did you end up adding the flag?  if so, what is it?

Chillee:
I investigated a bit more into the issue, and I think I have a solution that resolves both issues.

Closing this PR in favor of https://github.com/VSCodeVim/Vim/pull/1990

PR #1993, opened 2017-09-02
Implement '' and ``.

### LINKED ISSUE #1089: Marks: go to the position before the last jump

---

I'm opening this issue as per your instructions on the roadmap. Hoping to bump this feature up a bit. 

It might be because I'm a Vim newb who hasn't unlocked the full potential of marks, but I find myself constantly pressing `' '` and nothing happens.

brandonbloom:
Fixes #1089.

Dependency: https://github.com/Microsoft/vscode/pull/33715

### DIFF
--- a/src/actions/commands/actions.ts
+++ b/src/actions/commands/actions.ts
@@ -2311,6 +2311,49 @@ class CommandNavigateForward extends BaseCommand {
 }
 
 @RegisterAction
+class CommandNavigateLast extends BaseCommand {
+  modes = [ModeName.Normal];
+  keys = ['`', '`'];
+  runsOnceForEveryCursor() {
+    return false;
+  }
+
+  public async exec(position: Position, vimState: VimState): Promise<VimState> {
+    const oldActiveEditor = vimState.editor;
+
+    await vscode.commands.executeCommand('workbench.action.navigateLast');
+
+    if (oldActiveEditor === vimState.editor) {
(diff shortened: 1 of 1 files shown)


brandonbloom:
This is currently untested, pending acceptance of upstream patch.

Chillee:
@brandonbloom Sorry, what's the difference between `<C-o>` and the actions you've implemented?

brandonbloom:
The difference is what happens f you execute the command twice in a row. Instead of going back, then back again, it goes back, then forward, alternating back and forth. Very useful for comparing things at distinct points the same file quickly: just hit tick or single-quote over and over again to flip back and forth.

xconverge:
Nice job on the two repo attack here! It is pretty cool when your change gets merged into vscode and here and then the next release makes you happier :)

Chillee:
@brandonbloom Awesome! Nice job! Could you add some tests for this feature?

brandonbloom:
Are there any tests for CommandNavigateBack?

Chillee:
Actually, I doubt we need tests for this; it's just a wrapper of the corresponding VSCode command.

Chillee:
Actually, this doesn't seem to work on the current release. I'm guessing it's an issue with the keybindings.

brandonbloom:
Check to see if the "Go Last" command is available in VS Code. My guess is not yet.

PR #1996, opened 2017-09-02
Improvements to paragraph text objects.

brandonbloom:
See individual commit messages and associated issues for details.

### DIFF
--- a/src/actions/textobject.ts
+++ b/src/actions/textobject.ts
@@ -408,9 +408,18 @@ export class SelectParagraph extends TextObjectMovement {
       }
     }
 
+    // Include additional blank lines.
+    let stop = position.getCurrentParagraphEnd();
+    while (
+      TextEditor.getLineAt(stop.getDown(0)).text.trim() === '' &&
+      stop.line < TextEditor.getLineCount() - 1
+    ) {
+      stop = stop.getDown(0);
+    }
+
     return {
       start: start,
-      stop: position.getCurrentParagraphEnd(),
+      stop: stop,
     };
   }
 }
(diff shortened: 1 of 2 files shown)


Chillee:
It's not clear to me that the test is actually a failure and not just a result of you fixing the bug. Feel free to replace that test with a `newTest`.

brandonbloom:
Nope, I definitely broke sentences, which treat whitespace-only lines differently. Will fix it and add some tests.

brandonbloom:
For my own reference as I fix this, this is in the vim help:

```
							*paragraph*
A paragraph begins after each empty line, and also at each of a set of
paragraph macros, specified by the pairs of characters in the 'paragraphs'
option.  The default is "IPLPPPQPP TPHPLIPpLpItpplpipbp", which corresponds to
the macros ".IP", ".LP", etc.  (These are nroff macros, so the dot must be in
the first column).  A section boundary is also a paragraph boundary.
Note that a blank line (only containing white space) is NOT a paragraph
boundary.
Also note that this does not include a '{' or '}' in the first column.  When
the '{' flag is in 'cpoptions' then '{' in the first column is used as a
paragraph boundary |posix|.
```

And:

```
ap			"a paragraph", select [count] paragraphs (see
			|paragraph|).
			Exception: a blank line (only containing white space)
			is also a paragraph boundary.
			When used in Visual mode it is made linewise.

							*v_ip* *ip*
ip			"inner paragraph", select [count] paragraphs (see
			|paragraph|).
			Exception: a blank line (only containing white space)
			is also a paragraph boundary.
			When used in Visual mode it is made linewise.
```

brandonbloom:
OK, I've fixed sentence handling and made even more fixes to "inner" paragraph handling. This time, I used "find all references" to ensure I didn't affect any other code paths.

Chillee:
Looks good to me!

Issue #1683, opened 2017-05-10
Easymotion "Start of word Forwards/backwards" doesn't work for large files


atiqorin:
* Click *thumbs-up* 👍 on this issue if you want it!
* Click *confused* 😕 on this issue if not having it makes VSCodeVim unusable. 

The VSCodeVim team prioritizes issues based on reaction count. 

--------

**BUG REPORT** :

**Environment**:

- **VSCode Version**: 1.12.1
- **VsCodeVim Version**: 0.7.1
- **OS**: Windows 10

**What happened**:
For lagrge files Start of word backwards < leader > < leader > b  and Start of word forwards < leader > < leader > w does not work properly. See screenshots.

**What did you expect to happen**: 
I expect to go to a word backwards.

**How to reproduce it**:
copy the following line more then 82 times:
< add key="GuId" value="{436B2A9D-06C7-4627-ACE3-E0C6D0D4451F}"  / >

Go to last line and press < leader > < leader > b with easy motion on.
The option to jump back to near words from end of line higher then 81 (eg. 88)  is not present. I can only go back to words in line 81. See attached screenshot:

![image](https://cloud.githubusercontent.com/assets/19419058/25907515/4867ffb0-356d-11e7-9065-26370550109b.png)

Same happens for search of words forwards. Cannot go forward immediately for the same file from line 2. Only able to go to line 7.

![image](https://cloud.githubusercontent.com/assets/19419058/25908546/136a4a4a-3570-11e7-8a4d-40b1f10f749c.png)

xconverge:
the way we fixed #1627 causes this

I think it is an improvement from the previous behavior you reported, and is now a feature request

Thanks for reporting it! What does easymotion in vim do? does it change to 3 characters ever?

@rebornix

rebornix:
We switched from showing undefined to nothing. I'm not familiar with the behavior of easymotion :(

Chillee:
Isn't the problem that we're starting from the wrong place? I don't think it's a problem that we don't show the entire file, so long as we show stuff near the cursor.

atiqorin:
@xconverge nope I've never seen it change to 3 characters ever.

Chillee:
Actually, I can't reproduce this issue.

atiqorin:
I initially reported it for Windows. It happens in Mac also. @Chillee , use the following line for reproducing the issue. Paste it more than 81 times and it should be reproduced.

[...]

PR #1997, opened 2017-09-03
Fix#1981

### LINKED ISSUE #1981: easymotion: ability to limit the characters that appear

--------

id like to use only (or at least prioritize) the home row keys for easy motion

```vimrc
let g:EasyMotion_keys="asdfghjkl;'"
```
Source: https://github.com/easymotion/vim-easymotion/blob/e4d71c7ba45baf860fdaaf8c06cd9faebdccbd50/doc/easymotion.txt#L405

MaxfieldWalker:
This PR fixes #1981.

## Changes
- [x] Make it possible to use custom characters used for marker name
- [x] Fix no more than 26 characters appear as marker name


Chillee:
Perhaps this is the same issue as the second one: https://github.com/VSCodeVim/Vim/issues/1683?

MaxfieldWalker:
@Chillee Strictly speaking, #1683 is not the same as the second checkbox, but #1683 seems to be already fixed without noticing it.

PR #2001, opened 2017-09-03
Use Go To Def & history absent a tag stack.

### LINKED ISSUE #1999: Cltr-] could be mapped to Go To Definition action

--------

In the original vim <C-]> is mapped to go to the tag action. It would be productive
to map it to the Go To Definition action. The mapping for Ctrl+] in VSCode is used to
indent a line - functionality which is already covered by VSCodeVim.

brandonbloom:
Fixes #1999


Chillee:
LGTM.

PR #2002, opened 2017-09-03
Fix deref of undefined race on startup.

brandonbloom:
This can happen if you interact with the editor while vim mode is still
loading.


Chillee:
Ah this has been fairly annoying sometimes. Thanks for tracking it down and fixing it!

Issue #1741, opened 2017-05-23
Support `s` mapping for EasyMotion to any surrounding word


fourcolors:
* Click *thumbs-up* 👍 on this issue if you want it!
* Click *confused* 😕 on this issue if not having it makes VSCodeVim unusable. 

The VSCodeVim team prioritizes issues based on reaction count. 

--------

**FEATURE REQUEST** (choose one):

**Environment**:

- **VSCode Version**:  1.12.2
- **VsCodeVim Version**: .7.1
- **OS**: MacOS
**Request**:

I would REALLY like to map my `s` key to search forward AND backwards. I should be able to jump to any word and LINE BREAKS (please start including \n like normal EasyMotion does) with two-three characters.

I am currently using easymotion with VIM and this is a huge deal for me and a lot of people I know.

Chillee:
I'm thinking that Easymotion might be another good target for neovim integration. I'd rather not reimplement the entirety of EasyMotion again... I've opened up an issue here to ask about it: https://github.com/easymotion/vim-easymotion/issues/331

fourcolors:
@Chillee right now you can't seem to get Neovim to do the easymotion thing, can you? Unless I'm mistaken.

Chillee:
@fourcolors Correct. EasyMotion does a bunch of hacks that make it difficult to integrate. @MaxfieldWalker has been on a tear implementing new EasyMotion features, so you might want to check some of the new features he's implemented out and see if they cover some of your use cases.

MaxfieldWalker:
@fourcolors You can now map a single character to a easymotion action in the latest version of the extension with configuration like below.


```json
    "vim.otherModesKeyBindingsNonRecursive": [
        {
            "before": [
                "s"
            ],
            "after": [
                "<leader>",
                "<leader>",
                "2",
                "s"
            ]
        }
    ]
```

fourcolors:
@MaxfieldWalker Is it possible to make `s` do a search for the entire file I'm working in, basically adding a label for every word both ways? This is how I have it set up in VIM.

Basically I just type `s` and it will auto search and label each word in the file, no need to search.

fourcolors:
basically i'm looking for this mapping `nmap s <Plug>(easymotion-bd-w)`

MaxfieldWalker:
@fourcolors Unfortunately, bidirectional word motion is not supported so far, but it is easy to implement this, so I will work it on 👍

[...]

PR #2017, opened 2017-09-13
Easymotion improvements

MaxfieldWalker:
Make easymotion work even better.

## Improvements

- [x] Use vscode theme color for markers
- [x] Support bi-directional motions #1741
- [x] Tweak status bar text

## Fixes

- [x] Fix EASYMOTION MODE text is not rendered on the status bar after switching texteditors

### DIFF
--- a/README.md
+++ b/README.md
@@ -402,8 +402,12 @@ Motion Command | Description
 `<leader><leader> j`|Start of line forwards
 `<leader><leader> k`|Start of line backwards
 `<leader><leader> / <char>... <CR>`|Search n-character
+`<leader><leader><leader> bdt`|Til character
+`<leader><leader><leader> bdw`|Start of word
+`<leader><leader><leader> bde`|End of word
+`<leader><leader><leader> bdjk`|Start of line
 
-`<leader><leader> (2s|2f|2F|2t|2T) <char><char>` are also available.
+`<leader><leader> (2s|2f|2F|2t|2T) <char><char>` and `<leader><leader><leader> bd2t <char>char>` are also available.
 The difference is character count required for search.
 For example, `<leader><leader> 2s <char><char>` requires two characters, and search by two characters.
 This mapping is not a standard mapping, so it is recommended to use your custom mapping.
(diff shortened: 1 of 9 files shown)


MaxfieldWalker:
I don't know what color is good for markers' background and foreground.
For now, I used `activityBarBadge.background` and ``activityBarBadge.foreground`.

https://github.com/MaxfieldWalker/Vim/blob/abffbce06dc2f2a16cbfdcaccf1f63f3ff612145/src/actions/plugins/easymotion/easymotion.ts#L213-L229

#### Examples

[Spirited Away Theme](https://marketplace.visualstudio.com/items?itemName=MaxfieldWalker.vscode-color-theme-spirited-away)

image

Solarized Dark Theme

image

Chillee:
Nice! That looks awesome (way better than our current default of red and black...). However, it seems like you've gotten rid of the current method of choosing colors. Would it be possible to only use the background color if the user has not manually chosen his own colors? I am worried that it does limit customizability for those that want it.

We're also having some trouble with Travis. I'll take a look tonight.

MaxfieldWalker:
This is the function to select color. 

https://github.com/MaxfieldWalker/Vim/blob/abffbce06dc2f2a16cbfdcaccf1f63f3ff612145/src/actions/plugins/easymotion/easymotion.ts#L193-L207

`vim.easymotionMarkerBackgroundColor`, `vim.easymotionMarkerBackgroundColor`, `vim.easymotionMarkerBackgroundColor` options are still alive.

This is a customize example with the configuration below.

```json
{
    "vim.easymotionMarkerForegroundColorOneChar": "#000",
    "vim.easymotionMarkerForegroundColorTwoChar": "#000"
}
```
image

Chillee:
@MaxfieldWalker Awesome. So it only uses the theme's settings if those settings aren't manually set?

MaxfieldWalker:
@Chillee Yes, exactly.

Chillee:
Great! I've fixed the Travis issues, so I'll merge this (and some other PR's) and then issue a release tonight!

Chillee:
@MaxfieldWalker Actually, it seems like an easymotion test is failing, so can you check that?

Issue #1860, opened 2017-06-22
Selecting left while in visual multi cursor mode causes jumping cursors


jsonMartin:
* Click *thumbs-up* 👍 on this issue if you want it!
* Click *confused* 😕 on this issue if not having it makes VSCodeVim unusable. 

The VSCodeVim team prioritizes issues based on reaction count. 

--------

**Is this a BUG REPORT or FEATURE REQUEST?** (choose one): Bug Report

**Environment**: MacOS

- **VSCode Version**: 1.13.1
- **VsCodeVim Version**: 0.8.6
- **OS**: 10.12.5

**What happened**: Selecting left while in visual multi cursor mode causes jumping cursors

**What did you expect to happen**: Multiple cursors in visual mode should be able to select left as well as right

**How to reproduce it**: Add multiple cursors in visual mode, and try to select left

![large gif 640x377](https://user-images.githubusercontent.com/4033621/27412913-01f544dc-56ad-11e7-868c-83cf6f83ba9e.gif)

westim:
Hello! New contributor here :)

I believe I may have a fix for this issue. Mind if I take it on?

Chillee:
@westim Of course!

westim:
In the first commit on my fork, I simply changed the `Position.FromVSCodeSelection()` method to use the `Selection.anchor` and `Selection.active` properties instead of the `Selection.start` and `Selection.end` properties, respectively. This appears to have eliminated the cursor jitter we see above.

![Demo](https://thumbs.gfycat.com/CircularAcclaimedIndigobunting-size_restricted.gif)

As you can see in the GIF above, this change has introduced a new problem: the Selection gets extend to the EOL to the right. I'm not too familiar with the workflow of this extension, but it appears that the action isn't considered "done" after the first move as expected and instead keeps calling the `updateView()` method until the `cursorStart` position is bumped all the way to the EOL by the following code in modeHandler.ts:

```Typescript
      } else {
        // MultiCursor mode is active.

        if (vimState.currentMode === ModeName.Visual) {
          selections = [];

          for (let { start: cursorStart, stop: cursorStop } of vimState.allCursors) {
            if (cursorStart.compareTo(cursorStop) > 0) {
              cursorStart = cursorStart.getRight();

            selections.push(new vscode.Selection(cursorStart, cursorStop));
          }
```
Removing the `if` statement comparing the cursor positions results in the following behavior:

[...]

PR #2029, opened 2017-09-18
Fix for issue #1860.

westim:
This PR fixes issue #1860: Selecting left while in visual multi cursor mode causes jumping cursors. See comments on that issue for details, including the problem definition & testing.

No unit tests have been added in association with this PR because all behavior is multi-cursor, which does not appear to be testable. See #1860 comments for details on manual testing performed.

### DIFF
--- a/src/mode/modeHandler.ts
+++ b/src/mode/modeHandler.ts
@@ -764,9 +764,9 @@ export class ModeHandler implements vscode.Disposable {
         }
 
         if (
+          Configuration.mouseSelectionGoesIntoVisualMode &&
           !this._vimState.getModeObject(this).isVisualMode &&
-          (this._vimState.getModeObject(this).name !== ModeName.Insert ||
-            Configuration.mouseSelectionGoesIntoVisualMode)
+          this._vimState.getModeObject(this).name !== ModeName.Insert
         ) {
           this._vimState.currentMode = ModeName.Visual;
           this.setCurrentModeByName(this._vimState);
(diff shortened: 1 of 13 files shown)


xconverge:
Just a word of warning, this function has the highest chance of causing a regression in the entire extensions codebase IMO

did you read the comment and do a few of the tests in the gist?

My ears perk up whenever a change to this function is made, not because the function is right, but because of the impact it has on many other areas of the extension

src/mode/modeHandler.ts
```diff
@@ -650,9 +650,17 @@ export class ModeHandler implements vscode.Disposable {
       this.vimState.currentMode !== ModeName.VisualBlock
     ) {
       // Number of selections changed, make sure we know about all of them still
-      this.vimState.allCursors = e.textEditor.selections.map(
-        x => new Range(Position.FromVSCodePosition(x.start), Position.FromVSCodePosition(x.end))
-      );
+      let cursors: Range[];
+      cursors = [];
+      for (let sel of e.textEditor.selections) {
```
nguymin4: @westim Any reason why you chose `for ... of` instead of the `map`

westim: I wasn't sure if it was still possible to capture the `if` statement in a call to `map`, and writing it out this way made the logic more clear to me. Otherwise, there's no semantic justification for using `for...of`.

nguymin4: I would prefer to keep the `map` and just change the logic inside

westim:
I've managed to make a mess out of this pull request thanks to my inexperience with Git. I'm going to close this Pull request and re-create a cleaner Pull request as well as implement the suggested changes.

PR #2033, opened 2017-09-20
Pass and receive buffer to/from nvim without indentation conversion

nguymin4:
```
[
    {
         "_id": "0008",
         "startDate": "2009-07-21",
         "endDate": null,
    },
    ....
]
```
I execute command  `g/endDate/d`. But if I want to undo it, what i expected is I just need to press `u` only once to revert back. In fact, I have to press `u` twice. The reason is, at the moment, we convert tabs and spaces before passing and after receiving buffer to/from nvim when `insertSpaces: true` (or equivalent to `expandtab: true`). So the first `u` actually does nothing and the second `u` is to undo.

In more general case, I think we shouldn't modify the buffer because when the user set `insertSpaces: true`, he/she expect both vscode and nvim should share the same view instead of indentation conversion.


Chillee:
@nguymin4 The entire problem is that as an user, we would expect the vscode setting to transfer over to vim as well. However, if you remove that piece of code, then doing something like `'<'> >` would cause you to indent by tabs.

nguymin4:
@Chillee That's strange. I tried `>>` and then the tab is still converted to spaces as I expected.
I'm using Linux so cannot test for other platforms, but if you meet that problem then I think we can close this one. That's double `u` issue is not a real big one after all.

PR #2037, opened 2017-09-24
Fixes #1248: support for '., `., and gi commands.

### LINKED ISSUE #1248: Add support for gi, '., and `.
Could you add support for `gi`/`'.`/`` `. ``?
See [this](http://stackoverflow.com/questions/7749159/how-to-jump-to-the-exact-last-change-point-in-vim).

westim:
This is a PR to resolve issue #1248.

gif

I'm not sure whether the `historyTracker` class can be properly unit-tested, as I couldn't find any related tests. If it's possible, I'd love to get some suggestions & backfill those tests.

### DIFF
--- a/src/history/historyTracker.ts
+++ b/src/history/historyTracker.ts
@@ -558,7 +558,7 @@ export class HistoryTracker {
     if (this.currentHistoryStepIndex === 0) {
       return undefined;
     }
-    let lastChangeIndex = this.historySteps[this.currentHistoryStepIndex].changes.length;
+    const lastChangeIndex = this.historySteps[this.currentHistoryStepIndex].changes.length;
     if (lastChangeIndex === 0) {
       return undefined;
     }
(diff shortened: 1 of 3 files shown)


xconverge:
Can you update the roadmap as well please thanks

https://github.com/VSCodeVim/Vim/blob/master/ROADMAP.md

src/history/historyTracker.ts
```diff
@@ -548,6 +548,31 @@ export class HistoryTracker {
...
+   * Gets the ending cursor position of the last Change of the last Step.
+   *
+   * In practice, this sets the cursor position to the end of
+   * the most recent text change.
+   */
+  getLastChangeEndPosition(): Position | undefined {
+    if (this.currentHistoryStepIndex === 0) {
+      return undefined;
+    }
+    let lastChangeIndex = this.historySteps[this.currentHistoryStepIndex].changes.length;
```
xconverge: can we use const instead of let here

xconverge:
thanks, LGTM, just one small comment!

PR #2040, opened 2017-09-27
Fixes #1824: g; and g, commands.

### LINKED ISSUE #1824: g; and g, lag, off by one
g; and g, allow you to go backwards and forwards in the changelist. They seem to almost work correctly but have a kind of delayed response. 

When using either command for the first time nothing happens. Then when typing either command again, the previous command will take affect.

**Example**
g;  (nothing happens)
g, (goes back, as if g; was pressed)
g, (goes forward, as if g, was pressed)
g; (goes forward, as if g, was pressed)
and so on...

westim:
This is a fix for the g, and g; commands, which currently have offset issues per #1824.

gif


xconverge:
I like it, I think I wrote these late at night and @johnfn was like, I dont understand your random preincrementing (neither does present day me)

thanks!

PR #2055, opened 2017-10-06
[WIP] Fixes #1992 Batch replay text insertion for dot command

### LINKED ISSUE #1992: Repeating typing with `.` is extremely slow.

--------

**What happened**:

Typing a bunch of stuff, exiting insert mode, perform a motion, then press `.` to repeat typing. The previously typed characters will be inserted one at a time. If you typed a whole word or two, it can take up to a second to repeat the typing action.

It should instantaneously insert all of the typed characters in one go.

**How to reproduce it**:

1. Open a new file and insert several blank lines.
2. Go to the first blank line, type `i` to enter insert mode, and type random giberish on that line.
3. Press `esc` to exit insert mode.
4. Press `j` to go to the next line.
5. Press `.` to repeat the typing from step 2.

Notice how long it takes for the gibberish insert to be reproduced.

You may repeat steps 4 and 5 for as many blank lines as you have to experience the issue again.

Depending on timing factors, you may also notice that about halfway through the insertion, the auto-completion dropdown momentarily appears, indicating that this assert is not being performed atomically.

xconverge:
Need to do some more testing on here, I think it is the right idea but need to make sure it doesn't break anything.

Regular use shows it is nearly instantaneous, but I am worried I am breaking some other content change replay besides text insertion

### DIFF
--- a/src/actions/commands/actions.ts
+++ b/src/actions/commands/actions.ts
@@ -32,23 +32,6 @@ export class DocumentContentChangeAction extends BaseAction {
       return vimState;
     }
 
-    // Batch the content change on replay
-    if (vimState.isRunningDotCommand) {
-      const changesArray = this.contentChanges;
-      let stringToInsert = '';
-      for (let i = 0; i < changesArray.length; i++) {
-        stringToInsert += changesArray[i].textDiff.text;
-      }
-
-      // Add a transform containing the change
-      vimState.recordedState.transformations.push({
-        type: 'insertTextVSCode',
-        text: stringToInsert,
-      });
-
-      return vimState;
-    }
-
     let originalLeftBoundary: vscode.Position;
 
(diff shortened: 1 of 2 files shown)


xconverge:
Also need to handle the case of autocomplete counting as a content change that overlaps other content changes. Basically I need to loop through all content changes, batch together the ones the overlap and push individual chunk inserts for each batch

Chillee:
Awesome!

xconverge:
Arg this is a lot harder than it seems.

We already do batching/look for overlapping transforms with the insertText transform, the problem is we tend to use insertTextVSCode which allows vscode to add closing parenthesis etc.

I might need to give up for a bit on this, will think about it some more..

xconverge:
Closing this for now

PR #2064, opened 2017-10-11
An option to show the colon at the start of the command line box

gadkadosh:
Just a simple option to show the colon character at the start of the command line box.
Something I've been missing from vim itself, maybe others would find this useful too. If set to false the behavior should be the same like now.
I find it helps distinguish the vim command line with the command palette and the file search boxes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -550,11 +550,7 @@
         "vim.cmdLineInitialColon": {
           "type": "boolean",
           "description": "When typing a command show the initial colon ':' character",
-<<<<<<< HEAD
           "default": false
-=======
-          "default": "true"
->>>>>>> dc7602fc44b873ee881c60155fd6065cc27c29be
         }
       }
     }
--- a/package.json
+++ b/package.json
@@ -550,7 +550,7 @@
         "vim.cmdLineInitialColon": {
           "type": "boolean",
           "description": "When typing a command show the initial colon ':' character",
-          "default": "true"
+          "default": false
         }
       }
     }
--- a/package.json
+++ b/package.json
@@ -541,7 +541,7 @@
         "vim.substituteGlobalFlag": {
           "type": "boolean",
           "description": "Automatically apply the global flag, /g, to substitute commands. When set to true, use /g to mean only first match should be replaced.",
-          "default": "false"
+          "default": false
         },
         "vim.cursorStylePerMode": {
           "type": "object",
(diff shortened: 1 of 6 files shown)


xconverge:
I would like to have it set to false by default as to not confuse our current userbase. Also can you add this to the README.md settings area, thanks!

gadkadosh:
Sorry about that, was being stupid. Or just tired.
Fixed also a typo in package.json related to vim.substituteGlobalFlag option. 
I would suggest setting ignoreFocusOut to false for the command line box. The way it is now is often quite a bit in the way, and forces you to use the mouse again to gain focus back, in a very non-vim way.
Should I squash commits together?

xconverge:
Hmm the merge still doesn't look right...can you try again please? Github lets me squash and merge which I can do at the end once it is ready to go

gadkadosh:
What do you mean? It merges fine for me now. Have I messed something up?

package.json
```diff
@@ -550,7 +550,11 @@
         "vim.cmdLineInitialColon": {
           "type": "boolean",
           "description": "When typing a command show the initial colon ':' character",
+<<<<<<< HEAD
```
jpoon: Can you fix these merge conflicts?

src/configuration/configuration.ts
```diff
@@ -328,7 +328,11 @@ class ConfigurationClass {
   /**
    * When typing a command show the initial colon ':' character
    */
+<<<<<<< HEAD
```
jpoon: Unresolved merge conflicts

xconverge:
you didn't modify extension.ts yet it shows up in the diff...

gadkadosh:
I'm sorry, I'm not following. Those snippets seem to me to be from an older commit, not the most recent one. The most recent one really doesn't conflict for me with the master. So at the moment I'm not sure what it is that I should fix. 
Mind you, I'm quite new to this, so maybe I'm just doing something really wrong. I would appreciate if you would point me in the right direction. Thanks!

xconverge:
squashed it and it looks good, thanks for the contribution!

https://github.com/VSCodeVim/Vim/commit/68137e91eb5f8783c08b6f2c7517410e95d7994e

gadkadosh:
Thanks!
I would still like to add ignoreFocusOut false, which seems to be the same like the builtin input boxes of VSCode, and is also more convenient for a vim-like mouse-less usage. What do you think?

xconverge:
sure that sounds good to me!

PR #2062, opened 2017-10-10
Bugfix #1951: text selection in insert mode

DanEEStar:
Use the normal selection in insert mode.
This fixes #1951. Selection in insert mode seems to be
specific to this plugin and is not easily possible in normal Vim.

PR #2065, opened 2017-10-11
removed code which is not needed anymore due to #2062

DanEEStar:
Removed code which does not get executed anymore due to change in #2062.

It removed this code before but I got a bit confused with the pull-request handling so I did not commit it in #2062. Sorry for the additional work.


xconverge:
You are just trying to get your PR count up for the hacktoberfest tshirt ;)

Thanks a lot for the help! Keep them coming!

PR #2067, opened 2017-10-11
added `showOpenDialog` when typing emtpy e

### LINKED ISSUE #1877: Make :e open up a file dialog if no file name is passed in
As @xelra said on the neovim gitter, 

> Well, when I specified a filename it did try to open a file based on the current working directory and offered to create it. For that I'd actually expect it to directly create it. And for without specifying a filename I expected a dialogue to browse for a file.

DanEEStar:
Fixes #1877

I could test this commit on my local machine and it works.
But the `gulp watch` task always complains:
```
src/cmd_line/commands/file.ts(75,44): error TS2339: Property 'showOpenDialog' does not exist on type 'typeof window'.
```

Do I have to update the typings somehow?
How could I do that?

### DIFF
--- a/src/cmd_line/commands/file.ts
+++ b/src/cmd_line/commands/file.ts
@@ -73,13 +73,10 @@ export class FileCommand extends node.CommandBase {
       return;
     } else if (this._arguments.name === '') {
       const fileList = await vscode.window.showOpenDialog({});
-      await vscode.commands.executeCommand(
-        'vscode.open',
-        fileList[0],
-        this._arguments.position === FilePosition.NewWindow
-          ? this.getViewColumnToRight()
-          : this.getActiveViewColumn()
-      );
+      if (fileList) {
+        const doc = await vscode.workspace.openTextDocument(fileList[0]);
+        vscode.window.showTextDocument(doc);
+      }
       return;
     }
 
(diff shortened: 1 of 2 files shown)


xconverge:
when you do ```npm install``` does anything happen

xconverge:
Ah yes you can update the typings in package.json

I think here

https://github.com/VSCodeVim/Vim/blob/master/package.json#L27

src/cmd_line/commands/file.ts
```diff
@@ -72,9 +72,14 @@ export class FileCommand extends node.CommandBase {
 
       return;
     } else if (this._arguments.name === '') {
-      if (this._arguments.position === FilePosition.NewWindow) {
-        await vscode.commands.executeCommand('workbench.action.splitEditor');
-      }
+      const fileList = await vscode.window.showOpenDialog({});
+      await vscode.commands.executeCommand(
+        'vscode.open',
```
jpoon: can we use https://code.visualstudio.com/docs/extensionAPI/vscode-api#openTextDocument here?

DanEEStar:
update after @jpoon comment

Chillee:
@DanEEStar Looks like it's good to go?

xconverge:
Thanks again @DanEEStar !

PR #2072, opened 2017-10-14
Change ignoreFocusOut to false for the command line

gadkadosh:
This makes the command line behave like the builtin VSCode command palette and fuzzy file opening box. It also makes a lot of sense to not block the user until he reaches the mouse and refocuses on that box just to be able to close it. Vim users don't like to leave their home row :-)


xconverge:
Thanks! Makes sense to me

PR #2081, opened 2017-10-19
Fixes #1974: U command

### LINKED ISSUE #1974: Capital 'U' doesn't work
I can't set the line to its original state by  typing a capital U.
VsCode: July 2017 (version 1.15) Widows10
Vim: v0.9.0 Hundredth Hazelnut (Jun 24, 2017)

westim:
This PR adds the `U` command, which performs an undo on all changes which occurred on the line of the most recent change.

This feature was relatively complicated to implement. The first challenge is that, unlike the `u` command, the `U` command is _itself a change_. This means that you can undo and redo a `U` command. Therefore, `U` performs the following steps:

1. Identify & collect all consecutive changes to undo by looping through all steps & changes in reverse.
2. Run `undo()` on all the changes in the list.
3. Reverse the undone changes (additions to deletions and vice versa).
4. Remove the undone history steps from `HistoryTracker.historySteps`.
5. Re-add the undone changes as a single history step.

The second challenge was the presence of newlines in some changes. For example, if a change has the text `'\nhello, world!'`, we want to undo only the text `'hello, world!'` and not the newline character, so the newline characters must be removed. This must also account for how newline is `'\n'` on MacOS/Linux and `'\r\n'` on Windows. This special case also has a starting cursor position offset that must be compensated.

I chose to implement workarounds for these challenges rather than modify the architecture elsewhere. The benefit is that any bugs introduced are isolated to this command. The downside is that the code is not as elegant as it could be.

I have added 7 unit tests to cover the major functionality of this command. I have also updated the ROADMAP accordingly.


Chillee:
Oh wow! To be honest, I didn't think anybody would fix this issue: undo is a pretty tricky thing to deal with. 

This deserves a more thorough review, so I'll do it over the weekend.

src/history/historyTracker.ts
```diff
@@ -100,7 +100,7 @@ class HistoryStep {
     cursorEnd?: Position[] | undefined;
     marks?: IMark[];
   }) {
-    this.changes = init.changes = [];
+    this.changes = init.changes || [];
```
Chillee: Was this just a bug? Could this be the cause for the undo issues we've been experiencing?

src/history/historyTracker.ts
```diff
@@ -501,10 +501,10 @@ export class HistoryTracker {
 
     if (this.currentHistoryStep.changes.length === 0) {
       this.currentHistoryStepIndex--;
-    }
 
-    if (this.currentHistoryStepIndex === 0) {
-      return undefined;
+      if (this.currentHistoryStepIndex === 0) {
+        return undefined;
+      }
```
Chillee: What's the reasoning behind putting this in the if statement?

src/history/historyTracker.ts
westim: Yes, this was a bug that was preventing me from creating the new `HistoryStep` near the end of the function. Any list of changes passed in would always be overwritten as an empty array. I'm not aware of the existing undo issues, but it's possible.

src/history/historyTracker.ts
westim: It's not completely necessary, I simply thought it looked odd to have the exact same `if` statement twice in the same block. Nesting this statement makes the relationship between the second and third `if` statements more obvious (to me). I did the same change in the new function.

Chillee: Ah I didn't see the if statement above.

jpoon:
Is this good to merge @Chillee?

jpoon:
@westim Your PR is failing 2 UTs. Can you take a look? Build on `master` looks good, but I'm not all too confident in that...

westim:
Looks like a small bug with the `g;` and `g,` commands. I tweaked these commands in PR #2040 and added the unit tests in PR #2088.

In Vim, when the jumplist is at the most recent state, the first invocation of `g;` moves the cursor to the _most recent change_. In VsCodeVim, we currently move the cursor to the _second most recent change_ in this scenario.

I can work on a fix for this bug. For this PR, should we simply disable these tests?

Chillee:
@westim Yeah that sounds fine to me.

westim:
As indicated in https://github.com/VSCodeVim/Vim/pull/2081/commits/a6fc6b6bc37a97426b8feb04bd26fb3f38674bab, I have decided to revert the small change which caused regressions & failing tests for the `g;` and `g,` commands. I have left a comment to indicate that the line may be a bug.

It looks like I have a Prettier version difference that other PRs are seeing that I need to resolve. Also, it looks like the entire `actions.ts` file believes it was changed after I resolved a merge conflict. I'll see if there is some way to fix that.

westim:
@jpoon this PR is ready for review.

Chillee:
I reviewed this earlier and it all looks good to me! Thanks for the contribution, awesome work!

PR #2082, opened 2017-10-21
Fix #2080

### LINKED ISSUE #2080: cs{[ over multiple lines

--------

BUG REPORt

- **VSCode Version**: 1.17.0
- **OS**: Ubuntu 16.04

**What happened**:
Try to change brackets using surround
```
int main() { 

}
```

cs{[ between the two braces causes nothing to happen. The braces stay the same.
```
int main() [ 

]
```
**How to reproduce it**:
Put cursor between the brackets and type cs{[ in normal mode.

Strafos:
Surround finds second range differently if the braces are on at the beginning of a new line.


Chillee:
Fix the commit log and we can merge.

PR #2104, opened 2017-10-31
Use 'editor.unfold' with direction: 'down'

### LINKED ISSUE #2103: CommandOpenFold: Don't set `direction: 'up'`
Until now, the VSCode command 'editor.unfold' did not have a `direction` parameter.
I added it recently for consistency purposes.
It looks like VIM sets it to 'up', but actually means 'down'

https://github.com/VSCodeVim/Vim/blob/11be4a4a9a2bc2d569c6fa8f9ef67448a7d19920/src/actions/commands/actions.ts#L1655

Please also check the use of 'editor.fold'. Direction 'up' means folding the current region and all parent regions. 'Down' means current region and all child regions.

aeschli:
Fixes #2103.

The VSCode command `editor.unfold` didn't support `direction` yet (`editor.fold` did).
I added it to VSCode for consistency reasons, but that breaks the VIM extension as unfolding with multiple levels is expected to work down, not up.

### DIFF
--- a/src/actions/commands/actions.ts
+++ b/src/actions/commands/actions.ts
@@ -1652,7 +1652,7 @@ class CommandOpenFold extends CommandFold {
     let timesToRepeat = vimState.recordedState.count || 1;
     await vscode.commands.executeCommand('editor.unfold', {
       levels: timesToRepeat,
-      direction: 'down'
+      direction: 'down',
     });
 
     return vimState;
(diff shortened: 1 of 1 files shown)


jpoon:
Thanks @aeschli. Also noticed the docs were a little out of date for fold/unfold. Microsoft/vscode-docs#1230

PR #2111, opened 2017-11-02
fix(line-endings): change all files to lf

jpoon:
Change all line endings to `lf`


xconverge:
LGTM

jpoon:
Thanks @xconverge

PR #2115, opened 2017-11-03
Fixes #2113: Start in Disabled mode configuration.

### LINKED ISSUE #2113: Start in Disabled mode

--------

**FEATURE REQUEST** :

**Environment**:

- **VSCode Version**: 1.17.2
- **VsCodeVim Version**: 0.12.2
- **OS**: Windows

I wish it was possible to start in disabled mode.
For exemple by adding a parameter `vim.startDisabled` false by default.

westim:
This PR is to add the configuration to `package.json` to start Vim in the `Disabled` mode by calling the `toggleVim` command.

This configuration is `false` by default, and therefore has no effect on default configuration behavior of VSCodeVim.


jpoon:
If we are adding this configuration, we can get rid of `Global.active` and maintain the enable/disable state of the extension with the configuration.

westim:
Thanks for the great feedback, @jpoon !

extension.ts
```diff
@@ -334,19 +327,27 @@ export async function activate(context: vscode.ExtensionContext) {
...
+      modeHandlerToEditorIdentity = {};
+      let mh = await getAndUpdateModeHandler();
+      mh.updateView(mh.vimState, { drawSelection: false, revealRange: false });
     }
-  });
+  }
 
-  vscode.commands.executeCommand('setContext', 'vim.active', Globals.active);
+  registerCommand(context, 'toggleVim', async () => {
+    Configuration.disableExtension = !Configuration.disableExtension;
```
jpoon: I think this change is only in-memory and doesn't actually change the settings to the configuration file. I don't think we ever call the `update` fx in the [api docs](https://code.visualstudio.com/docs/extensionAPI/vscode-api#_a-nameworkspaceconfigurationaspan-classcodeitem-id943workspaceconfigurationspan).

It seems like we should convert `Configuration` class to use getters/setters such that if you set any of the configurations, it'll call the aforementioned update api.

README.md
```diff
@@ -197,6 +197,11 @@ We have removed this option, due to it making VSCodeVim's performance suffer imm
         "replace": "block-outline"
     }
 ```
+#### `"vim.disableExtension"`
+* VSCodeVim will be in "Disabled" mode
+* This can be changed at any time using the `toggleVim` command in the Command Palette
+* Note that this is not the same as disabling the VSCodeVim extension
```
jpoon: Nit: probably best to clarify 
> ...disabling the VSCodeVim extension through VS Code.

extension.ts
```diff
@@ -334,19 +327,27 @@ export async function activate(context: vscode.ExtensionContext) {
...
             options.cursorStyle = vscode.TextEditorCursorStyle.Underline;
             break;
-
           default:
             break;
         }
         editor.options = options;
       });
-      await vscode.commands.executeCommand('setContext', 'vim.active', Globals.active);
+      await vscode.commands.executeCommand('setContext', 'vim.active', false);
```
jpoon: Do we need to call this in two different spots (line 335, line 339)? You can probably consolidate it to outside the `if` to use `Configuration.disableExtension`

```
await vscode.commands.executeCommand('setContext', 'vim.active', Configuration.disableExtension)
```

extension.ts
```diff
@@ -310,15 +310,8 @@ export async function activate(context: vscode.ExtensionContext) {
...
-  registerCommand(context, 'toggleVim', async () => {
-    Globals.active = !Globals.active;
-    if (Globals.active) {
-      await vscode.commands.executeCommand('setContext', 'vim.active', Globals.active);
-      compositionState = new CompositionState();
-      modeHandlerToEditorIdentity = {};
-      let mh = await getAndUpdateModeHandler();
-      mh.updateView(mh.vimState, { drawSelection: false, revealRange: false });
-    } else {
+  async function toggleDisabledMode() {
```
jpoon: Hmm, I'm trying to think of a better name as it's not actually toggling the mode (ie. if you call this twice, it doesn't actually toggle and go from disable->enable). 

Maybe this signature:

`toggleExtension(bool isDisabled)`

extension.ts
westim: For now, for the sake of limiting the scope of these changes, I've only implemented a getter/setter for the `Configuration.disableExtension` property.

jpoon:
@westim let me know when this is ready for review.

westim:
Hi @jpoon ,
I believe all 4 comments from the second review were addressed in f68b16f. This currently ready for review.

src/configuration/configuration.ts
```diff
@@ -307,6 +307,27 @@ class ConfigurationClass {
    */
   foldfix = false;
 
+  /**
+   * Determines whether VSCodeVim is in Disabled mode or not.
+   */
+  private _disableExtension: boolean | undefined;
```
jpoon: Does it make sense to set this to `false` to start? Is there a scenario where it'll be `undefined`?

westim: I have some weird behavior I need to work out, and this will be involved in some further changes I need to make. I'll let you know when it's ready. Sorry about all the confusion / edits!

westim:
Hello @jpoon,

Another update on this. I was able to get the configuration to change correctly after fixing two issues:
1. The private backing field needed to match the name in config, not the getter/setter.
2. The `vscode.WorkspaceConfiguration.update()` API needed a `ConfigurationTarget` argument. When that argument was `undefined`, the `update()` method actually attempts to remove the section from config.

I'm not sure exactly how to test this. I confirmed the configuration value was changing using primitive techniques:
```Typescript
console.log(vscode.workspace.getConfiguration('vim').get('disableExtension'));
```
I confirmed that the configuration change carries forward to new [Editor Groups](https://code.visualstudio.com/docs/getstarted/userinterface) in the same Window. I'm not sure I can properly test whether the configuration affects new Windows while using the Extension Development Host. If there is, please let me know.

jpoon:
Thanks @westim

PR #2117, opened 2017-11-05
Fix remapping keys to actions with "mustBeFirstKey", fixes #2216

### LINKED ISSUE #2116: Cannot remap another key to "u", "C" or "gI" via vim.otherModesKeyBindingsNonRecursive

--------

BUG REPORT

Using very latest vscodevim and vscode the following config doesn't work:

```json
        {                                                                                                                                          
            "before": ["l"],                                                                                                                       
            "after": ["u"]                                                                                                                         
        }
    ]                                                                                                                                  
```

After this pushing "l" does nothing rather than undo. It's the same with any key other than "l" also. Users of COLEMAK (a QWERTY alternative) rebind `hjkl` to `snei` to restore the familiar movement keys, this necessitates moving other keys around. A popular colemak vim rebinding loses undo support because of this issue.

### LINKED ISSUE #2216: VSCodeVIM is unusable (incredibly slow) on larger files, especially in INSERT MODE

--------

**Is this a BUG REPORT or FEATURE REQUEST?** (choose one): **_BUG REPORT_**

**Environment**:

- **VSCode Version**: **_1.18.1_** (I have tried Insiders build too, and didn't work there either)
- **VsCodeVim Version**: **_0.15.5_**
- **OS**: **_MacOS High Siera_** 

**What happened**:
When editing longer files, VSCodeVim becomes so slow as to be completely unusable. Typing in insert mode, the keyboard response lags so hard it is impossible to use for anything other than reading files (I've had to switch back to Jetbrains IDE to work on this project in the meantime, since that's the only other IDE I have with a decent VIM plugin). 

I've included some GIFs below to show this bug in action. Note that I type 100+ WPM, and these videos are in real-time, running at 30FPS. 

On a <100 line file, typing doesn't seem to have much of a delay:
100_line

On a 1-2K line file, typing is already noticeably so slow that it is completely impossible to use:
1k_line

On a 10K line file, typing is even slower and more unusable:
10k_line

However, doing VIM commands in normal mode like navigating words seems to work at normal speed, and commands such as copy/pasting etc works (yank and paste is still a little slower but 
 nowhere near as slow as insert mode): 
command_mode

Disabling the VIM extension allows typing in real time, even on the largest 10K+ files: 
vscode_vim_disabled

Key presses type in real time on files, without an unusable delay. It should behave at the same or similar speed to how editing performs with VIM disabled. 

**How to reproduce it**:

Type in insert mode on large files with the newest version of VSCodeVim and VSCode.

ohjames:
Fix #2116

### DIFF
--- a/src/actions/base.ts
+++ b/src/actions/base.ts
@@ -124,9 +124,9 @@ export class BaseAction {
       return false;
     }
     if (
+      this.mustBeFirstKey &&
       vimState.recordedState.getCurrentCommandWithoutCountPrefix().length - keysPressed.length >
-        0 &&
-      this.mustBeFirstKey
+        0
     ) {
       return false;
     }
@@ -145,9 +145,9 @@ export class BaseAction {
       return false;
     }
     if (
+      this.mustBeFirstKey &&
       vimState.recordedState.getCurrentCommandWithoutCountPrefix().length - keysPressed.length >
-        0 &&
-      this.mustBeFirstKey
+        0
     ) {
       return false;
     }
--- a/src/actions/base.ts
+++ b/src/actions/base.ts
@@ -125,8 +125,7 @@ export class BaseAction {
     }
     if (
       this.mustBeFirstKey &&
-      vimState.recordedState.getCurrentCommandWithoutCountPrefix().length - keysPressed.length >
-        0
+      vimState.recordedState.numberOfKeysInCommandWithoutCountPrefix - keysPressed.length > 0
     ) {
       return false;
     }
@@ -146,8 +145,7 @@ export class BaseAction {
     }
     if (
       this.mustBeFirstKey &&
-      vimState.recordedState.getCurrentCommandWithoutCountPrefix().length - keysPressed.length >
-        0
+      vimState.recordedState.numberOfKeysInCommandWithoutCountPrefix - keysPressed.length > 0
     ) {
       return false;
     }
--- a/src/mode/modeHandler.ts
+++ b/src/mode/modeHandler.ts
@@ -321,6 +321,11 @@ export class RecordedState {
   public commandList: string[] = [];
 
   /**
+   * The number of keys the user has pressed that have been remapped.
+   */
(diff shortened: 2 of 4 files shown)


src/actions/base.ts
```diff
@@ -124,9 +124,9 @@ export class BaseAction {
       return false;
     }
     if (
+      this.mustBeFirstKey &&
       vimState.recordedState.getCurrentCommandWithoutCountPrefix().length - keysPressed.length >
```
ohjames: `this.mustBeFirstKey` is false for all but 3 actions, makes sense to take advantage of short-circuit evaluation here to avoid work.

src/actions/base.ts
```diff
@@ -124,9 +124,8 @@ export class BaseAction {
       return false;
     }
     if (
-      vimState.recordedState.getCurrentCommandWithoutCountPrefix().length - keysPressed.length >
-        0 &&
-      this.mustBeFirstKey
+      this.mustBeFirstKey &&
```
ohjames: Should take advantage of short-circuit evaluation to avoid work, only 3 commands have `mustBeFirstKey` set.

ohjames:
Simplified things a little.

src/mode/modeHandler.ts
```diff
@@ -901,8 +921,6 @@ export class ModeHandler implements vscode.Disposable {
 
     if (result === KeypressState.NoPossibleMatch && !isPotentialRemapping) {
       vimState.recordedState = new RecordedState();
-      vimState.recordedState.commandList = [];
```
ohjames: No point in repeating the work of the constructor here.

ohjames:
It might be simpler just to not add the remapped keys to `recordedState.commandList` but I don't know if that will have other implications.

xconverge:
looks good to me!

ohjames:
So good you want to merge it 🌴 🛩 ✈️

xconverge:
ship it!

PR #2121, opened 2017-11-08
fix vscode launch/tasks

jpoon:
* Enable running all `gulp` tasks through VS Code
* Run "gulp: default" to build.


xconverge:
lgtm

Issue #1654, opened 2017-05-05
workbench.colorCustomizations override in main settings.json


zelphir:
Great work, but after updating to the latest release I've noticed that every time that you switch mode the whole `workbench.colorCustomizations` in the main `settings.json` is overridden, that means that if you have custom settings (like me) they are gone!

jpoon:
@xconverge any ideas?

xconverge:
Yea, we should read the existing colorCustomizations in configuration.ts at initialization, then we can only change the entries we want if the setting is active and use the stored configuration from configuration.ts

zelphir:
@xconverge @jpoon any update on that?

jpoon:
I don't believe either of us are working on it at the moment, but we are happy to take PRs.

zelphir:
@jpoon what's wrong with saving the settings in the workspace instead of the user settings?

Chillee:
@zelphir Haha. Funny that you ask.

https://github.com/VSCodeVim/Vim/issues/1565

zelphir:
@Chillee did you have any chance to take a look at it? I think this setting should stay in the `.vscode` folder and not in the user settings. Removing the `true` from [this line](https://github.com/VSCodeVim/Vim/blob/eae183596104d820ca2f87a1934d7a03097dd32c/src/mode/modeHandler.ts#L1984) will fix it (it will move the colors back in the `.vscode` settings). Or alternatively the `modeHandler` should be able to update only this block

```javascript
{
   'statusBar.background': `${color}`,
   'statusBar.noFolderBackground': `${color}`,
   'statusBar.debuggingBackground': `${color}`,
}
```

Without overriding all the other settings.

autoferrit:
Any updates on this fix. This issue alone is making it so I can use it as i have many colors customized. Is there a workaround?

zelphir:
@Chillee Any update on this one?

jpoon:
Fixed with https://github.com/VSCodeVim/Vim/pull/2122

PR #2122, opened 2017-11-08
keep workbench color customizations when using status bar color

rodrigo-garcia-leon:
When the status bar color is controlled by the current mode, the workbench color customizations get overwritten by the status bar customizations, removing any previous existing customization. This change merges both existing and status bar customizations to avoid this behavior.


jpoon:
I believe this fixes https://github.com/VSCodeVim/Vim/issues/1654.

src/mode/modeHandler.ts
```diff
@@ -2065,11 +2065,11 @@ export class ModeHandler implements vscode.Disposable {
   setStatusBarColor(color: string): void {
     vscode.workspace.getConfiguration('workbench').update(
```
jpoon: Consider separating the get and the update into two separate lines for readability.

jpoon:
@rodrigo-garcia-leon mind addressing that one comment and then we can get this in?

thousand:
@jpoon looks good to me! I would expect that the plugin's behavior could overwrite or replace the behavior of the related configs.

src/mode/modeHandler.ts
```diff
@@ -2063,13 +2063,16 @@ export class ModeHandler implements vscode.Disposable {
   }
 
   setStatusBarColor(color: string): void {
+    const currentColorCustomizations = vscode.workspace
+      .getConfiguration('workbench')
+      .get('colorCustomizations');
     vscode.workspace.getConfiguration('workbench').update(
       'colorCustomizations',
-      {
+      Object.assign({}, currentColorCustomizations, {
```
jpoon: I've never used `Object.assign` before, is there a difference between:

1) `Object.assign({}, currentColor....)` and
2) `Object.assign(currentColor..., `

ie. why do you need the initial `{}`?

jpoon:
Actually, I misinterpreted the behaviour of `object.assign`, I thought it would only set the value if the key is null, but it's more of an upsert.

That being said, that means we'll be overwriting some settings at the `workbench` level. I'm OK with this change, but want to double-check with your opinions @thousand @rodrigo-garcia-leon @zelphir  on that's what you guys expect.

src/mode/modeHandler.ts
rodrigo-garcia-leon: The first parameter of the `Object.assign` function is the target object. Like you wrote below, it works like an upsert, where from left to right properties get merged.

This is the behavior that I intend, since I want to get a copy of the color customizations and then overwrite its properties regarding the status bar.

The target object gets modified by the `Object.assign` call, that's why I pass an empty object at first. But indeed, in this case, this is not needed. I will remove it.

jpoon:
Thanks @rodrigo-garcia-leon

rodrigo-garcia-leon:
No problem! 😄

Issue #2101, opened 2017-10-27
Can't search using caret ^ (Regex starts-with)


kitsu:
* Click *thumbs-up* 👍 on this issue if you want it!
* Click *confused* 😕 on this issue if not having it makes VSCodeVim unusable. 

The VSCodeVim team prioritizes issues based on reaction count. 

--------

**Environment**:

VSCode 1.17.2 VSCVim 0.10.2 on Win64

**What happened**:

Trying to search with matches starting at the beginning of line using `/^text` results in no matches.
My specific use case was searching for the next level-two header in a Markdown file using `/^##\s`.

My workaround of `/[^#]##\s` does work, but leaves the cursor before the match causing multiple matches of the same text. Searching with the cursor-to-end feature `/[^#]##\s/e` doesn't work either.

geordanr:
It looks like the `^` and `$` anchors are being applied to the entire file, not on a line-by-line basis.

parkovski:
They are. I have a fix for this, just need to add tests.

J-Fields:
Fixed via #2123

Issue #2100, opened 2017-10-27
Case sensitive search option - \c / \C


parkovski:
* Click *thumbs-up* 👍 on this issue if you want it!
* Click *confused* 😕 on this issue if not having it makes VSCodeVim unusable. 

The VSCodeVim team prioritizes issues based on reaction count. 

--------

**Is this a BUG REPORT or FEATURE REQUEST?** (choose one):
Feature request

In Vim, putting `\c` anywhere in the search term makes it case insensitive, and `\C` forces case sensitivity. I know about smartcase, but I prefer defaulting to case sensitive searches and using `\c` to make them insensitive, and miss having this feature in VSCode.

rishabh-bhatnagar-xebia:
How was the case sensitivity implemented?
By default, if I use * or f or / or ?
it is searching in case insensitive mode.

tgs:
You can trigger case sensitivity by putting `\C` anywhere in your regex (although I do wish I could make it the default)

rickgm:
If you set vim.ignorecase to false (defaults to true) in your settings.json then your searches will default to case sensitive.

rickgm:
> You can trigger case sensitivity by putting `\C` anywhere in your regex (although I do wish I could make it the default)

If you set vim.ignorecase to false (defaults to true) in your settings.json then your searches will default to case sensitive.

samos123:
Thanks @rickgm ! I still think it would be helpful to default to what's the default in vim.

danawoodman:
@rickgm you made my day thank you 🙇‍♂️

PR #2123, opened 2017-11-08
Fix ^, $, add case sensitivity override in search

### LINKED ISSUE #2086: Search for end of line ($) returns end of file

--------

**Environment**:

- **VSCode Version**: 1.17.2
- **VsCodeVim Version**: 0.10.2
- **OS**: Win10

**What happened**: When I try searching for end of line anchor (`/$`), vscode goes straight to the end of the file, and pressing `n` to go to the next occurance does nothing (indicating only one occurance). 

**What did you expect to happen**: In vim, *all* end of lines positions are found. This is especially visible with the `hlsearch` option.

**How to reproduce it**: Open any file, type out a few lines, hit `/$`.

parkovski:
- Fixes #2086 and #2101 by adding the 'm' flag to the search regex, which causes `^` and `$` to match per-line instead of per-file.
- Implements #2100 by stripping all `\c` and `\C` from the search pattern and changing `ignorecase` for that search based on the first one found - this mimics vim's behavior.

Substitute is not affected by either of these issues - `^` and `$` already work there, and `/i` to override case is already implemented - vim doesn't support `\c` there.

### DIFF
--- a/src/state/searchState.ts
+++ b/src/state/searchState.ts
@@ -15,6 +15,7 @@ export enum SearchDirection {
 export class SearchState {
   private static readonly MAX_SEARCH_RANGES = 1000;
   private static specialCharactersRegex: RegExp = /[\-\[\]{}()*+?.,\\\^$|#\s]/g;
+  private static caseOverrideRegex: RegExp = /\\[Cc]/g;
   public previousMode = ModeName.Normal;
 
   private _matchRanges: vscode.Range[] = [];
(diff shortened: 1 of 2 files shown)


jpoon:
LGTM. Fix the build error and then we can merge. 

```
Compile with gulp first (to format things properly with prettier) before pushing!"; sleep 1; exit 1; }; else echo "Prettier passed."; fi
```

parkovski:
When I run gulp locally nothing changes - it's giving the same error for all the recent PRs, it's a diff that none of them created.

parkovski:
I see what's wrong, I had an older version of prettier than travis. Should I add the fixed files to this PR, and if so should I update `package.json`? Since the version of prettier listed may not work.

PR #2161, opened 2017-11-19
Fixed incorrect styling of 'fake' cursors

Chillee:
Our decorations were stupidly done. I just copied over the relevant CSS from the actual cursor in VSCode. Now they should be indistinguishable.

Before:

After:

### DIFF
--- a/src/mode/modeHandler.ts
+++ b/src/mode/modeHandler.ts
@@ -512,15 +512,14 @@ export class ModeHandler implements vscode.Disposable {
   public identity: EditorIdentity;
 
   private _caretDecoration = vscode.window.createTextEditorDecorationType({
+    backgroundColor: new vscode.ThemeColor('editorCursor.foreground'),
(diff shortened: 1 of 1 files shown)


src/mode/modeHandler.ts
```diff
@@ -512,15 +512,14 @@ export class ModeHandler implements vscode.Disposable {
   public identity: EditorIdentity;
 
   private _caretDecoration = vscode.window.createTextEditorDecorationType({
+    backgroundColor: new vscode.ThemeColor('editorCursor.foreground'),
+    borderColor: new vscode.ThemeColor('editorCursor.foreground'),
     dark: {
-      // used for dark colored themes
-      backgroundColor: 'rgba(224, 224, 224, 0.4)',
-      borderColor: 'rgba(0, 0, 0, 1.0)',
+      color: '#515052',
```
jpoon: To conform with the `light` color, put rgb value. => `rgb(81,80,82)`

PR #2171, opened 2017-11-22
Fix #2168: Surround offset

### LINKED ISSUE #2168: Spacing bug using surround until end of word (ysw)
**BUG REPORT**

**Environment**:

- **VSCode Version**: 1.17.12
- **VsCodeVim Version**: 0.10.5
- **OS**: macOS Sierra 10.12.6 (16G29)

**What happened**:
The surround command does not space correctly around a word:

for example: `word word word`

is changed to: `word [ word  ]word`
With 2 space before the bracket and zero after

**Exact keys you pressed**
Cmd + n *(new file)*
iword word word *(some text)*
<Esc>bb *(locate cursor)*
ysw[

word [ word ] word

**How to reproduce it**:
As described above

**Notes**

- Also goes wrong for other delimiters `()[] ...`
- The same bug impacts a word (eg. `ysaw[`)

- inner word works correctly (eg. `ysiw[` )
- I love this plugin, keep up the good work

westim:
This fixes the extra whitespace created when the `ysaw` and `ysw` commands are used.

As explained in  https://github.com/VSCodeVim/Vim/commit/56371b63933b0489331cc7fcc634acc3e65cd829, this occurs because the `yaw` and `yw` commands normally yank the trailing whitespace after the current word. In this surround case, we want to ignore that extra character and modify the ending cursor position accordingly.

I have added 6 new tests to cover these scenarios.


Chillee:
Aghhhh!!! All of these have a space/don't have a space/have a new line at the beginning/end permutations are so annoying to deal with.

Thanks!

Issue #2172, opened 2017-11-23
:edit! triggers an open file dialog box


squgeim:
* Click *thumbs-up* 👍 on this issue if you want it!
* Click *confused* 😕 on this issue if not having it makes VSCodeVim unusable. 

The VSCodeVim team prioritizes issues based on reaction count. 

--------

**Is this a BUG REPORT or FEATURE REQUEST?**:

BUG REPORT

**Environment**:

- **VSCode Version**: 1.18.1
- **VsCodeVim Version**: 0.10.5
- **OS**: MacOS 10.13.1

**What happened**:

`:e!` or `:edit!`

This command reverts the file to the last saved state as expected, but then opens the open file dialog and I have to close that dialog to continue working.

**What did you expect to happen**:

On vim, `:e!` clears the current file without any kind of prompt. This was my expected behavior and the behavior this extension had until some versions back. 

I understand that "clear this file and open another file" is a valid use case, but how one would do that on vim is `:e! .` or `:e! /path/to/file`. `:e!` should not have any additional prompt.

**How to reproduce it**:
Make some changes to a file and then do `:e!`.

jpoon:
`edit` is not impl'd. For some reason, the cmd-parser thinks it's a "file" command (likely due to the bang), and since there are no arguments, opens the file prompt

PR #2215, opened 2017-12-12
Do not open open file dialog when calling `:e!`

squgeim:
Changes:

- When using a hash bang, return after the file is reverted. This prevents opening an open-file-dialog.

Fixes:  #2172


jpoon:
Thanks

PR #2223, opened 2017-12-14
Add C-[ to Replace Mode escape

### LINKED ISSUE #2221: Escape replace mode with <C-[>
https://github.com/VSCodeVim/Vim/blob/113dd77978b739ec27e34fe8fcdb87706a0be2e1/src/actions/commands/actions.ts#L529

Hey there, is it reasonable to add `<C-[>` to the `CommandEscReplaceMode.keys` array? Don't know if it's just an oversight of if there's a Vim or VS Code reason not to.

deybhayden:
Closes #2221 - simple change, just adds `C-[` as a key that can escape replace mode.


jpoon:
Thanks @deybhayden

PR #2243, opened 2017-12-24
Misc Bug Fixes and Refactoring

jpoon:
Bug Fixes:
* history-tracker: handle when editor is null
* fix bug where search decorator would reset following initial search
* show save details `:w` in status bar
* remove generator-changelog scripts. we'll stop using GH releases and generate using npm-generate-changelog
* #2247. displays error on cmd_line errors instead of silently failing.

Refactor:
* moving logic away from modeHandler into separate classes (e.g. remappers, statusbar)
* less spaghetti code with vim state removing dependency on modeHandler


jpoon:
Any review comments before I merge?

Issue #1787, opened 2017-05-31
Request: Add ability to toggle between INSERT and REPLACE modes with Insert key


benjamin-rood:
* Click *thumbs-up* 👍 on this issue if you want it!
* Click *confused* 😕 on this issue if not having it makes VSCodeVim unusable. 

The VSCodeVim team prioritizes issues based on reaction count. 

--------

It would be great if there was an (option?) to enable [Vim-like behaviour](http://vimdoc.sourceforge.net/htmldoc/insert.html) of the <kbd>Insert</kbd> key in VSCodeVim.

That is, pressing the <kbd>Ins</kbd> or <kbd>Insert</kbd> key would:

1) Switch to INSERT MODE if currently in NORMAL MODE
2) Toggle/switch between INSERT MODE and REPLACE MODE

PR #2252, opened 2017-12-30
Provide an InsertMode command to be able to handle Insert key

gibranrosa:
**What this PR does / why we need it**: It addresses partially #1787 by creating an command vim.insertMode that can be mapped to Insert key, and it simply remap to 'i' when not in insert mode.

**Which issue(s) this PR fixes**: #1787

### DIFF
--- a/extension.ts
+++ b/extension.ts
@@ -19,6 +19,7 @@ import { Neovim } from './src/neovim/nvimUtil';
 import { AngleBracketNotation } from './src/notation';
 import { StatusBar } from './src/statusBar';
 import { taskQueue } from './src/taskQueue';
+import { InsertMode } from './src/mode/modeInsert';
 
 interface VSCodeKeybinding {
   key: string;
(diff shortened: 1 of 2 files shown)


jpoon:
@gibranrosa thanks for the PR. It's difficult to ascertain the changes you are making to the `package.json`. I think a better approach may be to add a keybinding for insert and handle it as an action.

gibranrosa:
I think the editor has screwed it up the package.json, I have only made what you said hehe, I create an action exported in the package.json, in which it handles as a "i" to enter the Insert mode. The action is vim.insertMode, if you prefer I can restore the package.json and redo.

gibranrosa:
@jpoon , as @xconverge  have said, with the ?w=1 parameter the diff shows correctly. The code of the extension is a bit complicated, so this was the simplest way I've found to implement it. If you think it really not fits well, I will keep in my fork only, until someone makes it another way or I have some time to digg in.

Thanks!

jpoon:
The proposed fix is a bit hacky, the proper way of doing this would be to implement this as an action. From your last message @gibranrosa, it seems like you are abandoning this PR so I'm going to close it.

gibranrosa:
@jpoon I think I see now, it seems the most simple way would be to add a key in the line below, threating the Insert directly as a "i", do you think it would be appropriated? https://github.com/VSCodeVim/Vim/blob/79ec895622e859b1cb683b766d05344bcaf2f549/src/actions/commands/actions.ts#L623

gibranrosa:
@jpoon  Could you only take a look at my suggestion?

jpoon:
Yes. I think that'll be a better pattern

Issue #2162, opened 2017-11-20
Keys specified in vim.handleKeys aren't being handled


kmckelvin:
* Click *thumbs-up* 👍 on this issue if you want it!
* Click *confused* 😕 on this issue if not having it makes VSCodeVim unusable. 

The VSCodeVim team prioritizes issues based on reaction count. 

--------

**Is this a BUG REPORT or FEATURE REQUEST?** (choose one):
BUG REPORT

**Environment**:

- **VSCode Version**: 1.18.1
- **VsCodeVim Version**: 0.10.4
- **OS**: Linux (Ubuntu 17.10). Also experiencing it on macOS 10.13.1

**What happened**:

Using these configuration options:

```
    "vim.useCtrlKeys": false,
    "vim.handleKeys": {
        "<C-o>": true,
        "<C-i>": true,
        "<C-r>": true
    },
```

When using `<C-o>`, `<C-i>` or `<C-r>`, the keys are handled by VSCode. E.g. `<C-o>` displays the File Open dialog.

**What did you expect to happen**: 

Using keys marked as `true` in the `vim.handleKeys` section should be handled as vim bindings. `<C-o>` and `<C-i>` should navigate the jump list. `<C-r>` should redo.

**How to reproduce it**:

Set `useCtrlKeys` to false, and handle specific key bindings

```
    "vim.useCtrlKeys": false,
    "vim.handleKeys": {
        "<C-o>": true,
        "<C-i>": true,
        "<C-r>": true
    },
```

Press `<C-o>`. The File Open dialog is shown.

I've confirmed this last worked in 0.10.2 by installing the vsix file. The behaviour breaks in 0.10.3.

Arxzin:
same issue here.

**Environment**
- **VSCode Version**: 1.19.1
- **VsCodeVim Version**: 0.10.7
- **OS**: Windows 10, Version 1709, Build 16299.192

jpoon:
Ah, we're only handling the negation of `handleKeys`. Should be an easy fix.

jpoon:
Proposal: Remove `useCtrlKeys` altogether. The functionality of `useCtrlKeys` and `handleKeys` overlaps.

This will be a breaking change as people will need to update their configs, but the configuration of the keys will be more obvious.

gwojan:
[...]

PR #2269, opened 2018-01-06
fix: <C-d> remapping disabled by default. functionality controlled by "handleKeys"

Arxzin:
This PR make ```<C-d>``` disabled by default, unless users explicitly defined ```<C-d>:true``` in ```handleKeys```.

> handleKeys config trumps usectrlkeys conifg 

This configuration
```
"vim.useCtrlKeys": false,
"vim.handleKeys": {
    "<C-d>": true
}
```
It should have ```<C-d>``` working as **Vim move page half down**, now it behaviors as ```add selection to next find match```.
 
This PR modified [this line](https://github.com/VSCodeVim/Vim/blob/master/src/mode/modeHandler.ts#L289), eventually results in this logic:

|                             | useCtrlKeys = true                     | useCtrlKeys  = false                |
|--------------------- | ------------------------------|-----------------------------|
| handleKeys = true | VSCodeVim move page half down | VSCodeVim move page half down
| handleKeys = false | remap to ```<D-d>``` | remap to ```<D-d>```
| undefined               | VS Code default behavior | VS Code default behavior

#2162 

```<D-d>``` perform as **VS Code default behavior**, which is ```add selection to next find match```.

### DIFF
--- a/src/mode/modeHandler.ts
+++ b/src/mode/modeHandler.ts
@@ -286,17 +286,10 @@ export class ModeHandler implements vscode.Disposable {
       }
     }
 
-    // Now keep the remapping but check if <C-d> is explicitly defined
-    // within the handleKeys scope firstly
-    if (key === '<C-d>') {
-      const useKeyCtrlD = Configuration.handleKeys['<C-d>'];
-      if (useKeyCtrlD !== undefined) {
-        if (!useKeyCtrlD) {
-          key = '<D-d>';
-        }
-      } else if (!Configuration.useCtrlKeys) {
-        key = '<D-d>';
-      }
+    // <C-d> triggers "add selection to next find match" by default,
+    // unless users explicity make <C-d>: true
+    if (key === '<C-d>' && !(Configuration.handleKeys['<C-d>'] === true)) {
+      key = '<D-d>';
     }
 
     this.vimState.cursorPositionJustBeforeAnythingHappened = this.vimState.allCursors.map(
--- a/src/mode/modeHandler.ts
+++ b/src/mode/modeHandler.ts
@@ -285,9 +285,21 @@ export class ModeHandler implements vscode.Disposable {
         }
       }
     }
-    if (key === '<C-d>' && !Configuration.useCtrlKeys) {
-      key = '<D-d>';
+
+    // #2162 fixed by 5cc821e except <C-d> key due to this remapping
+    // Now keep the remapping but check if <C-d> is explicitly defined
+    // within the handleKeys scope firstly
+    if (key === '<C-d>') {
+      const useKeyCtrlD = Configuration.handleKeys['<C-d>'];
+      if (useKeyCtrlD !== undefined) {
+        if (!useKeyCtrlD) {
+          key = '<D-d>';
+        }
+      } else if (!Configuration.useCtrlKeys) {
+        key = '<D-d>';
+      }
     }
+
     this.vimState.cursorPositionJustBeforeAnythingHappened = this.vimState.allCursors.map(
       x => x.stop
     );
--- a/src/mode/modeHandler.ts
+++ b/src/mode/modeHandler.ts
@@ -286,7 +286,6 @@ export class ModeHandler implements vscode.Disposable {
       }
     }
 
-    // #2162 fixed by 5cc821e except <C-d> key due to this remapping
     // Now keep the remapping but check if <C-d> is explicitly defined
     // within the handleKeys scope firstly
     if (key === '<C-d>') {

### CODE REVIEW

**Comment by jpoon** [comment]
As mentioned in the #2162, I think we should delete the <d-d> key entirely.

Your proposed fix would not fix windows/linux.

**Comment by Arxzin** [comment]
This "fix" is for windows/linux, well, if that remapping line is for mac, which I assumed, not sure what it does originally.

**Comment by jpoon** [comment]
This is the original PR adding the change. https://github.com/VSCodeVim/Vim/pull/1631

**Comment by Arxzin** [comment]
So after reexamined #1631, #1535, #1467, #1705.
It's far more complicated than I thought... not even a configuration thing, that line is a hack.

How about  
```
    let useKeyCtrlD = Configuration.handleKeys['<C-d>'];
    if (useKeyCtrlD !== undefined) {
      if (!useKeyCtrlD && key === '<C-d>' && !Configuration.useCtrlKeys) {
        key = '<D-d>';
    }
```

if users explicitly disabled <C-d> means they want default <C-d> multi-cursor feature, else handle <C-d> as the rest.

File: src/mode/modeHandler.ts [hunk 1]
```diff
@@ -285,9 +285,21 @@ export class ModeHandler implements vscode.Disposable {
         }
       }
     }
-    if (key === '<C-d>' && !Configuration.useCtrlKeys) {
-      key = '<D-d>';
+
+    // #2162 fixed by 5cc821e except <C-d> key due to this remapping
```
jpoon: remove the comment. that information is in `git blame`

src/mode/modeHandler.ts
```diff
@@ -285,9 +285,21 @@ export class ModeHandler implements vscode.Disposable {
...
+    // Now keep the remapping but check if <C-d> is explicitly defined
+    // within the handleKeys scope firstly
+    if (key === '<C-d>') {
+      const useKeyCtrlD = Configuration.handleKeys['<C-d>'];
```
jpoon: This can be all simplified to one statement. 
```
if (Configuration.handleKeys['<C-d>'] ===false  || !Configuration.useCtrlKeys) {
...
}
```

Arxzin: remap \<C-d\> to \<D-d\> within this statement?
what about
```
"vim.useCtrlKeys": false,
"vim.handleKeys": {
    "<C-d>": true
}
```
, should vim <C-d> enable?

jpoon:
Let's draw this out cause it's kind of confusing.

Legend: Top = `useCtrlKeys`. Left = `handleKeys`.

|  | true | false |
|---|---|---|
| true|   remap to `<D-d>`| remap to `<D-d>`  | 
| false|  VS Code default behaviour |   VS Code default behaviour | 
| undefined| VSCodeVim move page half down | VS Code default behaviour |

where VS Code default behaviour for `<C-d>` is 'add selection to next find match'

Arxzin:
That's the confusing part, in my mind it's

|                             | useCtrlKeys = true                     | useCtrlKeys  = false                |
|--------------------- | ------------------------------|-----------------------------|
| handleKeys = true | VSCodeVim move page half down | VSCodeVim move page half down
| handleKeys = false | remap to ```<D-d>``` | remap to ```<D-d>```
| undefined               | VSCodeVim move page half down | VS Code default behavior

where 
```<D-d>``` = VS Code default behavior = add selection to next find match
```<C-d>``` = VSCodeVim move page half down

I thought **remap to \<D-d\>** does the same thing as **VS Code default behaviour**, ```add selection to next find match```, from these line, [1](https://github.com/VSCodeVim/Vim/blob/master/src/actions/commands/actions.ts#L3448), [2](https://github.com/VSCodeVim/Vim/blob/master/src/actions/commands/actions.ts#L3471).

From user's perspective, they want ```<C-d>``` behavior as
```Vim move page half down```, write ```<C-d>: true```, 
```Add selection to next find match```, write ```<C-d>: false```

**Comment by jpoon** [comment]
I want to eventually remove `useCtrlKeys` -- this setting is making everything so complicated. If that is the case, how about:

|                             | useCtrlKeys = true                     | useCtrlKeys  = false                |
|--------------------- | ------------------------------|-----------------------------|
| handleKeys = true | VSCodeVim move page half down | VSCodeVim move page half down
| handleKeys = false | remap to ```<D-d>``` | remap to ```<D-d>```
| undefined               | *VS Code default behavior* | VS Code default behavior

**Comment by Arxzin** [comment]
Yep, it either further confuses users and developers, or makes people suffer from writing long configs to decide every Ctrl key. lol.

Maybe remove it is a better idea. The "consistence"  is more important after all.

**Comment by jpoon** [comment]
> makes people suffer from writing long configs to decide every Ctrl key

When I get rid of `useCtrlKeys`, I'll include a list of default things so *most* people don't need to reconfigure. But it will be a breaking change :(.

Not sure when I'll get around to doing that. As such, do you want to update this PR to match the latest discussed table then we can merge?

**Comment by Arxzin** [comment]
Yes, already made Vim ```<C-d>``` disable by default, unless user defined ```<C-d>: true```, it will be remapped to ```<D-d>```, which perform  **add selection to next find match**.
```
    if (key === '<C-d>' && !Configuration.handleKeys['<C-d>'] === true) {
      key = '<D-d>';
    }
```

Arxzin:
Updated PR, thanks Github for showing all these editing.

jpoon:
Thanks @Arxzin

GuoQichen:
@jpoon hi, How could i set handleKeys to undefined ? I want use VS Code default behavior in C-d

lzfee0227:
Try to use the trick instead before resolved:
1. ```useCtrlKeys": false```
2. Move cursor to target text
3. Set vim to normal mode with esc
4. Select text with your mouse / touch pad
5. Use cmd / ctrl + d of vscode
6. Set vim to insert mode with i
7. move your cursor, it may works with multiple cursors (not multiple selections)

onezhaoyn:
1.(Use ⌘+K  ⌘+S) Open Keyboard Shortcuts, 
2.Then open the JSON File, 
3.Search for 'cmd+d' keyword, comment out the code block:
`//{`
    `// "key": "cmd+d",`
     `//"command": "-extension.vim_cmd+d",`
    ` //"when": "editorTextFocus && vim.active && vim.use<D-d> && !inDebugRepl"`
 `//}`

this works for me perfectly, hope anyone see this solve the problem i have.
p.s. I'm on a Mac.

PR #2288, opened 2018-01-13
Fix BaseAction.couldActionApply to work with two-dimensional keys array

jpotterm:
When `BaseAction.couldActionApply` slices the `keys` array it assumes it is one-dimensional (e.g. `['a', 'b']`). This pull request fixes it so that it also works when the `keys` array is two-dimensional (e.g. `[['a', 'b'], ['c', 'd']]`).

### DIFF
--- a/src/actions/base.ts
+++ b/src/actions/base.ts
@@ -140,9 +140,13 @@ export class BaseAction {
     if (this.modes.indexOf(vimState.currentMode) === -1) {
       return false;
     }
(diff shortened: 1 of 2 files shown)


jpoon:
Hmm, the code change seems to make sense. I'm surprised we didn't catch this before. Any chance you can add some tests?

jpotterm:
@jpoon I've added tests!

Issue #2299, opened 2018-01-16
'<G' and '<xG' shift left commands off by one line


jessewmc:
**Is this a BUG REPORT or FEATURE REQUEST?** (choose one):
BUG REPORT
**What happened**:
`    zxcv
    zxcv
    zxcv`

Entering `<G`  results in
`zxcv
zxcv
    zxcv`
instead of
`zxcv
zxcv
zxcv`

Entering `<2G` on the original 3 indented lines results in
`zxcv
    zxcv
    zxcv`
instead of 
`zxcv
zxcv
    zxcv`
**What did you expect to happen**:

**How to reproduce it (as minimally and precisely as possible)**:
Any new file will do with a few lines of text. Can indent all with `>G` which works correctly, then try the above sequences.

**Environment**:

- Extension (VsCodeVim) version: 0.10.9
- VSCode version: 1.19.2
- OS version: Linux x64 4.9.0-4-amd64

jessewmc:
I believe I have a fix for this, just struggling a bit with tests.
```
import { getAndUpdateModeHandler } from '../../extension';
import { ModeHandler } from '../../src/mode/modeHandler';
import { getTestingFunctions } from '../testSimplifier';
import { cleanUpWorkspace, setupWorkspace } from '../testUtils';

suite('shift operator', () => {
  let modeHandler: ModeHandler;

  let { newTest, newTestOnly } = getTestingFunctions();

  setup(async () => {
    await setupWorkspace();
    modeHandler = await getAndUpdateModeHandler();
  });

  teardown(cleanUpWorkspace);

  newTest({
    title: 'basic shift left test',
    start: ['    |zxcv', '    zxcv', '    zxcv'],
    keysPressed: '<<',
    end: ['|zxcv', '    zxcv', '    zxcv'],
  });
});
```
This test should pass I believe, but doesn't. For sanity, I tried something simple like this in addition to the above:
```
  newTest({
    title: 'sanity check',
    start: ['    |zxcv', '    zxcv', '    zxcv'],
    keysPressed: 'dd',
    end: ['|    zxcv', '    zxcv'],
  });
```
This passes as I expect.

One of the actual tests I would like is:
```
  newTest({
    title: 'shift left goto end test',
    start: ['    |zxcv', '    zxcv', '    zxcv'],
    keysPressed: '<G',
    end: ['|zxcv', 'zxcv', 'zxcv'],
  });
```
[...]

PR #2300, opened 2018-01-16
Left shift fix 2299

jessewmc:
Fixes [https://github.com/VSCodeVim/Vim/issues/2299](https://github.com/VSCodeVim/Vim/issues/2299)

Left shift operator with G goto line/end command was not indenting last line. Seems to be because editor.action.outdentLines ignores last line of vscode.Selection when the end position is at the beginning of the last line. Taking end.getLineEnd() in the selection fixes this.
#2299

### DIFF
--- a/test/testSimplifier.ts
+++ b/test/testSimplifier.ts
@@ -171,11 +171,23 @@ function tokenizeKeySequence(sequence: string): string[] {
   let key = '';
   let result: string[] = [];
 
+  // no close bracket, probably trying to do a left shift, take literal
+  // char sequence
+  function rawTokenize(): void {
+    for (const char of key) {
+      result.push(char);
+    }
+  }
+
   for (const char of sequence) {
     key += char;
 
     if (char === '<') {
-      isBracketedKey = true;
+      if (isBracketedKey) {
+        rawTokenize();
+      } else {
+        isBracketedKey = true;
+      }
     }
 
     if (char === '>') {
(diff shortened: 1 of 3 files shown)


test/testSimplifier.ts
```diff
@@ -171,11 +171,23 @@ function tokenizeKeySequence(sequence: string): string[] {
   let key = '';
   let result: string[] = [];
 
+  // no close bracket, probably trying to do a left shift, take literal
```
jpoon: I tried walking through this a couple of times, it's probably cause it's the end of the day, but I can't grok this code. 

The code might be simpler if instead we walk the string, if we see `<`, find the matching `>`. If there's no matching `<`, then it's a left shift.

jessewmc: So what I'm doing is sort of the opposite, I'm looking for another `<` following a `<`, and then working back instead of looking ahead.

 If I see a second `<` (i.e. `isBracketedKey` is already truthy when I hit a `<`) I know whatever was previous was a left shift and subsequent literal character inputs, so I take the currently built `key` and tokenize it one character per token.

After the tokenizing loop has run, if `isBracketedKey` is truthy I know we also saw an unclosed `<` so again I tokenize the currently built `key` as one character per token.

I don't mind looking ahead for closing `>` instead if you still prefer that. I didn't do that because it's a bit more complicated in the case of `keysPressed` like `<<<ESC>` where I would have to start counting open `<`s and looking for matches or recursing which seemed overcomplicated for a convenience test helper function. That would be starting to get into serious parsing territory.

jessewmc: Although now that I've explained this I've found a bug, I will update.

jpoon: Ah, thanks for the explanation.

Issue #1348, opened 2017-03-02
Ctrl-D and Ctrl-U do not update cursor correctly


tiehuis:
* Click *thumbs-up* 👍 on this issue if you want it!
* Click *confused* 😕 on this issue if not having it makes VSCodeVim unusable. 

The VSCodeVim team prioritizes issues based on reaction count. 

--------

### What did you do?

Use Ctrl-D and Ctrl-U as movement keys. Also affects Ctrl-f and Ctrl-b.

### What did you expect to happen?

The cursor should retain its relative position on the screen (disregard the edge of file case) during Ctrl-D and Ctrl-U movement.

### What happened instead?

The cursor position is not updated within the file and drops off the visible portion of the screen. On the next successive movement there appears to be some check for an out of bounds cursor which places it to the bottom of the screen regardless of its initial position.

This makes movement with these keys unusable.

### Technical details:

Pretty certain this a regression since VSCode 1.9 which did not have this problem. Doesn't appear to be any recent issues on VSCode itself relating to movement right now so not sure who the cause of this is.

* VSCode Version: 1.10
* VsCodeVim Version: 0.5.3
* OS: Arch Linux + Windows 10

johnfn:
Interesting - are you saying that this begun happening after the latest vscode release? 

Until we solve this correctly, you can use the workaround found here https://github.com/VSCodeVim/Vim/issues/907#issuecomment-264738452 to redefine c-u and c-d. A lot of people rely on these motions, so we'll try to get there quickly!

JessThrysoee:
Upgrading to Visual Studio Code Version 1.10.1 did not fix this for me.

gwojan:
@johnfn "Interesting - are you saying that this begun happening after the latest vscode release?"

Yes, this behavior just started with 1.10.1. I was using the Insiders release and noticed something funky but it didn't really hit me until I updated to 1.10.1 non-Insiders earlier today.

rhys-vdw:
FWIW I have this problem on `1.10.0`. I haven't yet restarted to update.

gwojan:
I guess I should've mentioned I did not see the problem on 1.9.1. 😉

johnfn:
It sounds a lot like this is a regression in VSCode itself. We've got our best men (aka @rebornix) on it. :)

mrtoniliu:
It seems this problem is resolved in VSCode 1.10.2.

gwojan:
[...]

PR #2314, opened 2018-01-19
Default to vim behaviour for Ctrl+D

### LINKED ISSUE #2312: extension.vim_ctrl+d changed?
BUG REPORT

**What happened**:
extension.vim_ctrl+d, which used to scroll half page down, doesn't work anymore. Instead, some kind of find selection feature is working.

Scroll half page down.

Add some keymap to "extension.vim_ctrl+d". (e.g. "cmd+9")

It used to work like "ctrl+d" in vim (scroll half page down), but it is now triggering the find popup.

**Environment**:

- Extension (VsCodeVim) version: 0.10.11
- VSCode version: 1.19.2
- OS version: macOS 10.12.6(16G1114)

Graham42:
Users installing a vim plugin will probably expect Ctrl+D to work like
vim. So this changes the default settings to do that.

Fixes #2312

I tried adding tests, but then realized that <kbd>Ctrl+D</kbd> / <kbd>Cmd+D</kbd> doesn't actually work as per https://github.com/VSCodeVim/Vim/issues/1348 and I've been using the workaround posted there.

### DIFF
--- a/package.json
+++ b/package.json
@@ -514,7 +514,10 @@
         },
         "vim.handleKeys": {
           "type": "object",
-          "description": "Option to delegate certain key combinations back to VSCode to be handled natively"
+          "description": "Option to delegate certain key combinations back to VSCode to be handled natively",
+          "default": {
+            "<C-d>": true
+          }
         },
         "vim.statusBarColorControl": {
           "type": "boolean",


intijk:
@jpoon Thank you for your hard work :+1: 

Could you trigger a release?

Issue #2315, opened 2018-01-20
Key repeat still doesn't work on Mac


puradox:
**Is this a BUG REPORT or FEATURE REQUEST?** (choose one):
Bug report

**What happened**:
Key repeat isn't working on Mac, even after executing the commands in the README.

**What did you expect to happen**:
Motion keys {h,j,k,l} to repeat after holding down.

**How to reproduce it (as minimally and precisely as possible)**:
Install VSCodeVim then execute: 
```
defaults write com.microsoft.VSCode ApplePressAndHoldEnabled -bool false         # For VS Code
defaults write com.microsoft.VSCodeInsiders ApplePressAndHoldEnabled -bool false # For VS Code Insider
```

**Environment**:
MacBook Pro

- Extension (VsCodeVim) version: 0.10.11
- VSCode version: 1.19.2
- OS version: macOS 10.13.2

PR #2316, opened 2018-01-20
Update README for Mac key repeat

puradox:
I was having trouble getting my Mac to key repeat, I did some quick research and found https://github.com/Microsoft/vscode/issues/31919. The issue was that I had a global default that was true, which was overriding the application level `com.microsoft.VSCode` setting.

Fixes key repeat for Mac users that already have a global default set for `ApplePressAndHoldEnabled`.

https://github.com/VSCodeVim/Vim/issues/2315

There might be a better way than resetting the global default.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,7 @@ All common Vim commands are supported. For a detailed list of supported features
 If key repeating isn't working for you, execute this in your Terminal.
 
 ```sh
+defaults delete -g ApplePressAndHoldEnabled                                      # Reset global default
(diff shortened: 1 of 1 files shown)


jpoon:
Not a fan of recommending folks reset a global default. Does it not work with: https://github.com/Microsoft/vscode/issues/31919#issuecomment-351951174?

puradox:
@jpoon That solution didn't work for me

muhajirframe:
Doesn't work for me too

PR #2347, opened 2018-01-31
Fixes #2346

### LINKED ISSUE #2346: New feature request:  Global command
**Is this a BUG REPORT or FEATURE REQUEST?** (choose one):  Feature Request

**What happened**: 
Upon typing ":g/vpcid/ d" and hitting Enter/Return key, shows the message "E492: Not an editor command. g/vpcid/ d"

I expect all the lines that match the pattern specified between the slashes, to be deleted.  I would like the same functionality as the global command g and v (refer [http://vim.wikia.com/wiki/VimTip227](Power of g)).  This feature is especially useful when dealing with data files.

Enter the above command for a file with some text pattern that is present in the file.  And hit Enter/Return key.

**Environment**:

- VSCode version: 1.19.3
- OS version: macOS Sierra (v10.12.6)

Chillee:
https://github.com/VSCodeVim/Vim/issues/2346#issuecomment-361826197

explains the problem and the fix.

### DIFF
--- a/src/cmd_line/commandLine.ts
+++ b/src/cmd_line/commandLine.ts
@@ -39,7 +39,7 @@ export class CommandLine {
       }
     } catch (e) {
       if (e instanceof VimError) {
-        if (e.code === ErrorCode.E492 && Configuration.enableNeovim) {
+        if (e.code === ErrorCode.E492 && configuration.enableNeovim) {
           await Neovim.command(vimState, command);
         } else {
           StatusBar.SetText(
--- a/src/cmd_line/commandLine.ts
+++ b/src/cmd_line/commandLine.ts
@@ -38,14 +38,17 @@ export class CommandLine {
         await cmd.execute(vimState.editor, vimState);
       }
     } catch (e) {
-      console.log(e);
       if (e instanceof VimError) {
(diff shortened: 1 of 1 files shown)


src/cmd_line/commandLine.ts
```diff
@@ -38,14 +38,17 @@ export class CommandLine {
...
-      console.log(e);
       if (e instanceof VimError) {
-        StatusBar.SetText(
-          `${e.toString()}. ${command}`,
-          vimState.currentMode,
-          vimState.isRecordingMacro,
-          true
-        );
+        if (e.code === ErrorCode.E492) {
+          await Neovim.command(vimState, command);
```
jpoon: Even for users that haven't configured to use neovim, you want to use neovim as a fallback

Chillee: My bad. You're right, there should be another check here.

Chillee:
Finally, it works. I didn't realize Travis built with some version of the main branch merged with the PR.

jpoon:
>  I didn't realize Travis built with some version of the main branch merged with the PR.

@Chillee what do you mean? Travis will build this particular PR.

PR #2102, opened 2017-10-29
Extend statusBarColors setting to be able to set foreground color of status bar based on current mode

xahon:
You're able to define your status bar customizations like:

``` JSON
  "vim.statusBarColors": {
    "normal": ["#007BFF", "#000"],
    "insert": ["#FF004E", "#F00"],
    "visual": "#FFC200",
    "visualline": "#FFC200",
    "visualblock": "#FFC200",
    "replace": "#FF00AE"
  }
```

Where `[backgroundColor, foregroundColor]`
You can also use default string form to change only background

PR #2350, opened 2018-01-31
Possibility to set statusBar foreground color

mgor:
Basically a reimplementation of #2102. Updated documentation to include instructions on how colors are set.

Having the possibility to set the foreground color makes it easier to select a background color and get a good contrast between the two.

N/A.

N/A.


src/statusBar.ts
```diff
@@ -31,15 +31,22 @@ class StatusBarClass implements vscode.Disposable {
     }
   }
 
-  public SetColor(color: string) {
+  public SetColor(background: string, foreground?: string) {
     const currentColorCustomizations = vscode.workspace
```
jpoon: Can you call `Configuration.getConfiguration` here?

src/statusBar.ts
```diff
@@ -31,15 +31,22 @@ class StatusBarClass implements vscode.Disposable {
...
-    });
+
+    const newColorCustomizations = {
+      'statusBar.background': `${background}`,
+      'statusBar.noFolderBackground': `${background}`,
+      'statusBar.debuggingBackground': `${background}`,
+    };
+    if (foreground !== undefined) {
+      newColorCustomizations['statusBar.foreground'] = `${foreground}`;
+    } else if (currentColorCustomizations !== undefined && 'statusBar.foreground' in currentColorCustomizations) {
```
jpoon: Should be sufficient to only have an `else`, and run `delete currentColorCustomizations['statusBar.foreground'];` if foreground is undefined.

mgor: That would result in `[ts] Object is possibly 'undefined'`.

src/statusBar.ts
mgor: You mean instead of calling `vscode.workspace.getConfiguration('workbench').get('colorCustomizations')` I should use something like `Configuration.getConfiguration('workspace').get('colorCustomizations')`? (not able to look up the exact syntax right now :)). If so, then it should be used [when updating](https://github.com/mgor/Vim/blob/15ace4f2355ed1c624ed8868980263c5a73b3140/src/statusBar.ts#L52) the configuration as well?

mgor:
Sorry for late response, I has been a hectic week. I'll fix it during the weekend, when my additional questions has been cleared out.

src/statusBar.ts
jpoon: Actually scratch that, I realize this fx is not exposed in `IConfiguration`.

src/statusBar.ts
jpoon: Actually, this `delete` doesn't work. As you are modifying the object, but not actually updating it with VSCode.

This should be better:
```
  public SetColor(background: string, foreground?: string) {
    const currentColorCustomizations = vscode.workspace
      .getConfiguration('workbench')
      .get('colorCustomizations');

    let colorCustomizations = Object.assign(currentColorCustomizations, {
      'statusBar.background': `${background}`,
      'statusBar.noFolderBackground': `${background}`,
      'statusBar.debuggingBackground': `${background}`,
    });

    if (foreground === undefined) {
      delete colorCustomizations['statusBar.foreground'];
    } else {
      colorCustomizations['statusBar.foreground'] = `${foreground}`;
    }

    vscode.workspace
      .getConfiguration('workbench')
      .update('colorCustomizations', colorCustomizations, true);
  }
```

jpoon:
Sorry for taking so long to merge this in. Still on vacation. I'll merge as soon as this new build passes CI.

PR #2351, opened 2018-02-01
Build Improvements

jpoon:

* Move to npm for typings. DefinitelyTyped is deprecated.
* Adds `gulp test` command which builds a Docker container and runs tests inside that container. Makes it easier to run tests as it doesn't take over existing VS Code instance.


Chillee:
LGTM

PR #2366, opened 2018-02-07
`jj` cursor position fix for #1418

### LINKED ISSUE #1418: Wrong cursor position entering normal mode
### What did you do?

Add this to config:
```
        {
            "before": [
                "d",
                "f"
            ],
            "after": [
                "<Esc>"
            ]
        }
    ]
```

to work keys same as in vim and same as <Esc>

### What happened instead?

put cursor in the middle of text and press df, it enters normal mode, but places cursor one position further than <Esc>

### Technical details:

* VSCode Version: 1.10.2
* VsCodeVim Version: 0.6.11
* OS: macOS

prog666:
fixes #1418

**Special notes for your reviewer**: tried to make test, but tests are made in the way that `jj` not actually passed to editor.

### DIFF
--- a/src/common/motion/position.ts
+++ b/src/common/motion/position.ts
@@ -373,7 +373,7 @@ export class Position extends vscode.Position {
   }
 
   /**
-   * Gets the position one to the left of this position. Does not go up line
+   * Gets the position one or more to the left of this position. Does not go up line
    * breaks.
    */
   public getLeft(count: number = 1): Position {
--- a/src/common/motion/position.ts
+++ b/src/common/motion/position.ts
@@ -376,9 +376,9 @@ export class Position extends vscode.Position {
    * Gets the position one to the left of this position. Does not go up line
    * breaks.
    */
-  public getLeft(): Position {
+  public getLeft(count: number = 1): Position {
     if (!this.isLineBeginning()) {
-      return new Position(this.line, this.character - 1);
+      return new Position(this.line, this.character - count);
     }
 
     return this;
--- a/src/common/motion/position.ts
+++ b/src/common/motion/position.ts
@@ -377,8 +377,9 @@ export class Position extends vscode.Position {
    * breaks.
    */
   public getLeft(count: number = 1): Position {
-    if (!this.isLineBeginning()) {
-      return new Position(this.line, this.character - count);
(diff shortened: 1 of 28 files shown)


Chillee:
I think @jpoon recently added functionality for testing remappings. Have you tried the method [here](https://github.com/VSCodeVim/Vim/blob/master/test/configuration/configuration.test.ts)?

prog666:
it tests that keys are replaced with correct keys, but that is not the case

prog666:
anyway, im using vim with my fix for 3 days now, everything is ok

prog666:
What is stopping us from merging this pr?

AlexArendsen:
+1 I'd love to see this merged in too.

Chillee:
@prog666 Sorry, I didn't immediately merge this because I was confused what the error was (I think partially because the repro in the issue linked isn't related to the title of this PR). After playing around with it a bit though, I see what the issue was.

LGTM!

prog666:
what are we waiting for?

src/common/motion/position.ts
```diff
@@ -376,9 +376,9 @@ export class Position extends vscode.Position {
    * Gets the position one to the left of this position. Does not go up line
    * breaks.
    */
-  public getLeft(): Position {
+  public getLeft(count: number = 1): Position {
     if (!this.isLineBeginning()) {
```
jpoon: Can you update this check to ensure that we don't go past the beginning of the line?

src/common/motion/position.ts
```diff
@@ -376,9 +376,9 @@ export class Position extends vscode.Position {
    * Gets the position one to the left of this position. Does not go up line
```
jpoon: Please update this comment.

src/common/motion/position.ts
prog666: there's no check for getRight()

prog666:
@jpoon done

jpoon:
On vacation in Africa at the moment. We'll be back to hotel tomorrow and I can take a look then. Otherwise @chillee can merge assuming my comments were addressed.

Chillee:
Looks like both of jpoon's comments have been addressed. I'm gonna merge this. Thanks!

PR #2395, opened 2018-02-20
fix: clean-up neovim processes. closes #2038

### LINKED ISSUE #2038: Multiple instances of nvim.exe process

--------

**Maybe BUG REPORT**:

It seems that nvim.exe proces is spawn for each tab I edit. These processes are still running even I close all tabs (VS Code is still running but without any opened tab). This is not a big problem ... but after 8 hours of work my Task Manager looks like this

image

**Environment**:

- **VSCode Version**: 1.16.1
- **OS**:  Windows 10

Some cap on number of active nvim processes. e.g max 10 is kept running. Oldest are closed

**How to reproduce it**:

Open multiple files. Then close them. Keep your editor running

jpoon:
**What this PR does / why we need it**: Properly kills neovim processes on exit. Refactoring of neovim to a class. 

**Which issue(s) this PR fixes** #2038

### DIFF
--- a/src/cmd_line/commandLine.ts
+++ b/src/cmd_line/commandLine.ts
@@ -1,7 +1,7 @@
 import * as vscode from 'vscode';
 
 import { configuration } from '../configuration/configuration';
-import { Neovim } from '../neovim/nvimUtil';
+import { Neovim } from '../neovim/neovim';
 import { VimState } from '../state/vimState';
 import { StatusBar } from '../statusBar';
 import * as parser from './parser';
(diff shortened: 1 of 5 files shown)


src/cmd_line/commandLine.ts
```diff
@@ -33,14 +33,14 @@ export class CommandLine {
...
-        await Neovim.command(vimState, command);
+        await vimState.nvim.run(vimState, command);
       } else {
         await cmd.execute(vimState.editor, vimState);
       }
     } catch (e) {
       if (e instanceof VimError) {
         if (e.code === ErrorCode.E492 && configuration.enableNeovim) {
-          await Neovim.command(vimState, command);
+          await vimState.nvim.run(vimState, command);
```
jpoon: @chillee, I don't quite understand this bit. If it's `E492` we want to retry running neovim again?

Chillee: When we pass the command into our own command line command/parser, there's lots of cases in which the parser will error (since it doesn't support a lot of commands).

If the parser can't parse it, then we try sending it to the neovim parser (which may support the command).

Chillee: So basically, the above code that detects if the command is neovim capable and passes it to Neovim only works for commands that we **know** are neovim capable. There's lots of commands that we don't implement, and the parser errors out in those cases.

jplew:
just want to say that v11.0 has restored neo-vim functionality for me. Neo-vim features were broken for me in 10.3 but this patch seems to have fixed it.

Issue #1162, opened 2016-12-15
:tabe {file} is relative to current file


gugahoa:
Please *thumbs-up* 👍 this issue if it personally affects you! You can do this by clicking on the emoji-face on the top right of this post. Issues with more thumbs-up will be prioritized.

-----

### What did you do?

In a folder with `file1` in the workspace root and `file2` inside `{workspaceDir}/dir1` open `file2` and press :tabe file1

### What did you expect to happen?
`file1` to be opened in another tab

### What happened instead?
Plugin looked for `file1` inside `dir1` instead of `{worspaceDir}`

### Technical details:

* VSCode Version: 1.8.0
* VsCodeVim Version: 0.4.9
* OS: Arch Linux 64bits

jpoon:
Yes, `:tab` should be operating at the "workshape" level which will correspond to `vim` implementatin.

PR #2400, opened 2018-02-25
Fix :tabe {file} only relative to current file (#1162)

arussellk:
Fixes https://github.com/VSCodeVim/Vim/issues/1162

This PR extends the logic for `:tabe` to better match user expectation. `:tabe` now handles the following situations:
1. `:tabe {file}` when you have a workspace open will try to open `{file}` relative to the root of your first workspace folder
2. `:tabe {file}` when you do not have a workspace will fall back to open `{file}` relative to your current document
3. `:tabe` will open an untitled document
4. `:tabe {absolutePath}` will open `{absolutePath}`

Fixes https://github.com/VSCodeVim/Vim/issues/1162

**Special notes for your reviewer**

1. I was unable to use `fs.mkdir` to make a folder because I kept getting a `EACCES: permission denied, mkdir` error. Because of that, I did not add tests for when a user is in a workspace. I did add tests for the other cases that `:tabe` handles.
2. I am now `export`ing `testUtils.ts function createRandomFile` for use in my tests.

### DIFF
--- a/src/cmd_line/commands/tab.ts
+++ b/src/cmd_line/commands/tab.ts
@@ -43,7 +43,7 @@ export class TabCommand extends node.CommandBase {
     }
   }
 
-  execute(): void {
+  async execute(): Promise<void> {
     switch (this._arguments.tab) {
       case Tab.Next:
         if (this._arguments.count /** not undefined or 0 */) {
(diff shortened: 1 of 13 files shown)


src/cmd_line/commands/tab.ts
```diff
@@ -43,7 +43,7 @@ export class TabCommand extends node.CommandBase {
     }
   }
 
-  execute(): void {
+  async execute(): Promise<void> {
```
jpoon: Agreed that we should make this `async`. However, can you update the other return paths to also return a promise?

arussellk: @jpoon I `await`ed everything I could with the following two exceptions:
1. `await vscode.commands.executeCommand('workbench.action.openEditorAtIndex1');` never resolves. I think it may be a bug with VSCode, so I should reproduce it in a small extension and open an issue on the VSCode repo.
2. `vscode.commands.executeCommand('activeEditorMove')` needs to be updated to `vscode.commands.executeCommand('moveActiveEditor')`. It isn't quite a simple name replacement, so it should probably be done in a different PR.

jpoon: Filed https://github.com/VSCodeVim/Vim/issues/2401 for the refactoring.

jpoon:
Thanks @arussellk

PR #2418, opened 2018-03-04
fix: selection in search in visual mode #2406

### LINKED ISSUE #2406: Selection range is strange when using incremental search in visual mode
BUG REPORT

**What happened**:
In VSCodeVim, when you use incremental search `/` in visual mode, selection starts from the position you've pressed `/` key, instead of the position you've pressed `v` key.

Same behavior as vim,
that is, selection should start from the position you've pressed `v` key.

In the following text:
```
foo bar vim
```
You are at the character `f`, the first character on the line.
When you presses `vw/v<Enter>`, then, `bar v` is selected, while standard vim would select `foo bar v`.

**Environment**:

- Extension Version: 0.11.0
- OS Version: Windows_NT x64 10.0.16299
- VSCode version: 1.20.1

shortheron:
Fix the selection range in search in visual mode.
When you start search by `/` or `?` in visual mode, the start of selection remains the position you started the visual mode in vim, but it reset to the position you started the search in VSCodeVim without this fix.
See the issue #2406 for detail.

fixes #2406 

This is my first pull request in github.
Although I read the contributing guidelines, I am sorry if something is missing.

### DIFF
--- a/src/mode/modeHandler.ts
+++ b/src/mode/modeHandler.ts
@@ -1140,12 +1140,16 @@ export class ModeHandler implements vscode.Disposable {
 
     if (args.drawSelection) {
       let selections: vscode.Selection[];
+      const selectionMode: ModeName =
+        vimState.currentMode === ModeName.SearchInProgressMode
+          ? vimState.globalState.searchState!.previousMode
+          : vimState.currentMode;
 
       if (!vimState.isMultiCursor) {
         let start = vimState.cursorStartPosition;
         let stop = vimState.cursorPosition;
 
-        if (vimState.currentMode === ModeName.Visual) {
+        if (selectionMode === ModeName.Visual) {
           /**
            * Always select the letter that we started visual mode on, no matter
            * if we are in front or behind it. Imagine that we started visual mode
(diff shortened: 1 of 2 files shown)


shortheron:
Thank you for your reply.

I can write test like the following, based on the existing test.
This test checks if selection starts correctly, although I have some concern.

```javascript
    newTest({
      title: 'Works with /, and selection\'s start is correct',
      start: ['f|oo', 'bar', 'fun', 'baz'],
      keysPressed: 'vj/baz\nx',
      end: ['f|az'],
    });
```

This test would pass *even without the current fix*, because the current problem is caused by `handleSelectionChange()` (see the [comment](https://github.com/VSCodeVim/Vim/issues/2406#issuecomment-370209538)), but [`handleSelectionChange()` is disabled in test](https://github.com/VSCodeVim/Vim/blob/ab2c18e959a1178f173c7e57ec2780acd9f32df0/src/mode/modeHandler.ts#L113).
I'm afraid that this test cannot detect the problem. (Even so, it is better than nothing?)

I would appreciate it if you would give me some advice.

jpoon:
We wouldn't be able to use `newTest` as that doesn't deal with selection properly. We'd need to write one like...https://github.com/VSCodeVim/Vim/blob/113dd77978b739ec27e34fe8fcdb87706a0be2e1/test/mode/modeVisualLine.test.ts#L112, but also adding an assert to ensure selection start+end is correct.

shortheron:
Thank you for your advice.
I added a test to check selection range during search, by `test()`.
Also, I rebased to the latest master.

Chillee:
@shortheron Thanks for the PR! It seems like @jpoon 's comment has been resolved with an unit test, so I'll be merging this.

PR #2433, opened 2018-03-08
Readds vimState.lastClickWasPastEOL. Fixes #2404

### LINKED ISSUE #2404: Backward text selection using mouse does not include the last character of the selection area.
Probably bug report

**What happened**:
After I update to 0.11.0, when I use mouse to backward select an area of text, the last character is not included.

The last character should be selected.

e.g., `abcd`
If I want to select the whole word `abcd`, I can put my mouse at the end of line or on the character `d`, backward move to `a`. But when I do it in Vim mode, `d` is not selected. However, if I use keyboard by moving cursor to `d` and pressing `vB`, the whole word is selected.

**Environment**:

- Extension (VsCodeVim) version: 0.11.0
- VSCode version: 1.20.1
- OS version: macOS 10.13.3

Chillee:
We need it to be part of vimstate because we need state that persists between multiple calls to handleSelectionChange.

### DIFF
--- a/src/mode/modeHandler.ts
+++ b/src/mode/modeHandler.ts
@@ -234,6 +234,7 @@ export class ModeHandler implements vscode.Disposable {
           const newStart = new Position(selection.anchor.line, selection.anchor.character + 1);
           this.vimState.editor.selection = new vscode.Selection(newStart, selection.end);
           this.vimState.cursorStartPosition = selectionStart;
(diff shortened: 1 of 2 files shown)


jpoon:
Hm, maybe I missed something in my earlier change -- I didn't see a case this needed to be persisted between multiple calls.

xconverge:
This code still makes me shudder

Issue #2434, opened 2018-03-08
[Feature Request] Support easymotion-jumptoanywhere


xkxx:
[FEATURE REQUEST]

Hello,

I'm happy to report that I'm a daily user of vscodevim and really happy with the integration of easymotion so far!

It would be cool to support `easymotion-jumptoanywhere`, which is basically a combination of `<leader><leader> w/W` and `<leader><leader> b/B` - it searches for start and end of word on the entire viewport. [Documentation](https://github.com/easymotion/vim-easymotion/blob/master/doc/easymotion.txt#L340).

It basically gives us the same functionality as [jumpy](https://marketplace.visualstudio.com/items?itemName=wmaurer.vscode-jumpy), which has 16k installs despite the fact that it has never worked 🤣. 

How easy / hard would it be to implement this feature? I'd happy to send a PR over the weekend if you could give me a hand on how you think it should be implemented.

Thanks!

xkxx

jsonMartin:
@xkxx - Have you tried using easymotion's "BDW" command?

For example, I have the following bound to space bar, which seems like it can perform the behavior you're looking for:
```
"vim.otherModesKeyBindingsNonRecursive": [{
    "before": [" "],
    "after": ["<leader>", "<leader>", "<leader>", "b", "d", "w"]
  }],
```

jsonMartin:
Hey @xkxx ,

I ended up liking this feature and missing it from VSCode so much that I just hacked together a solution to add this functionality. It's my first time ever working on the VSCodeVim codebase so the implementation is currently quite rough around the edges, but it definitely seems to get the job done!

![2018-03-15 21 47 54](https://user-images.githubusercontent.com/4033621/37504327-db6f3b72-289a-11e8-8ee4-2a3caf9cd0b7.gif)

^ There's an example of it in action - note that my spacebar is mapped to `<Leader><Leader><Leader>bdj`, which is the new command I created for `jumptoanywhere`. 

I just pushed up the new functionality to my forked Repo on the master branch, you can feel free to install it on your end and give it a try if you'd like. It's still very much a work in progress but jumptoanywhere-like functionality should work! Here's the link: https://github.com/jsonMartin/Vim

xkxx:
Hi @jsonMartin,

[...]

PR #2454, opened 2018-03-20
Add jumptoanywhere command for easymotion

jsonMartin:
This PR adds the missing `jumptoanywhere` command to the `easymotion` plugin. `jumptoanywhere` allows quick text navigation, and is basically a combination of `bd-w` and `bd-j` with some added behavior.

According to the easymotion VIM docs, the default behavior for this command is: `Beginning and End of word, camelCase, after '_', and after '#'`. The current implementation matches this expected behavior.

Here's a demo of what it looks like in action (_note my space bar is mapped to <Leader><Leader><Leader>j_):

The user can also configure custom jumptoanywhere regex to fine tune the behavior. For example, this is my current setting in VSCode, which also adds the first/last char of the line and `//` comments to the normal behavior (note the double escaping necessary in the config file):
```
"vim.easymotionJumptoanywhereRegex": "^\\s*.|\\b[A-Za-z0-9]|[A-Za-z0-9]\\b|_.|\\#.|[a-z][A-Z]|//"
```
https://github.com/VSCodeVim/Vim/issues/2434#issuecomment-373471794

This PR has been updated to also include custom regex settings via user preferences to match the behavior of `Easymotion_re_anywhere` that can be set in .vimrc


jsonMartin:
Just added another commit allowing the addition of custom regex for jumptoanywhere in user preferences; this allows the jumptoanywhere integration to allow custom matching options a la the `Easymotion_re_anywhere` in `.vimrc` setting.

I'll update the PR description with the new information

src/actions/plugins/easymotion/easymotion.cmd.ts
```diff
@@ -299,8 +299,14 @@ export class EasyMotionWordMoveCommandBase extends BaseEasyMotionCommand {
     vimState: VimState,
     options?: EasyMotion.SearchOptions
   ): EasyMotion.Match[] {
-    // Search for the beginning of all words after the cursor
-    return vimState.easyMotion.sortedSearch(position, new RegExp('\\w{1,}', 'g'), options);
+    const regex = this._options.jumptoanywhere
+      ? configuration.easymotionJumptoanywhereRegex
+        ? new RegExp(configuration.easymotionJumptoanywhereRegex, 'g')
```
jpoon: Can you capitalize the `Anywhere`?

src/actions/plugins/easymotion/easymotion.cmd.ts
```diff
@@ -299,8 +299,14 @@ export class EasyMotionWordMoveCommandBase extends BaseEasyMotionCommand {
...
+        : // Default behavior matches beginning & ending of word, camelCase, after _ and after #
+          new RegExp('\b[A-Za-z0-9]|[A-Za-z0-9]\b|_.|#.|[a-z][A-Z]', 'g')
```
jpoon: Is this match intended:

For `MaybeThisWillDo`, the above regex will match the M, e, T, s, W, D

jpoon: This is a better visualization of what I mean: https://regexr.com/3mo29

src/actions/plugins/easymotion/types.ts
```diff
@@ -11,4 +12,5 @@ export interface EasyMotionCharMoveOpions extends EasyMotionMoveOptionsBase {
 
 export interface EasyMotionWordMoveOpions extends EasyMotionMoveOptionsBase {
   labelPosition?: LabelPosition;
+  jumptoanywhere?: jumptoanywhere;
```
jpoon: Please camel case. Unless is it generally known as `jumptoanywhere` all lowercase?

src/actions/plugins/easymotion/easymotion.cmd.ts
jpoon: Ah, I see your comment about camelCase, this is probably expected behaviour, but just want to confirm.

jsonMartin: Hey @jpoon - Yeah, this is definitely intended behavior to match CamelCase. Since the VIM version of `jumptoanywhere` always places the cursor at the end of the match, this behavior pairs perfectly with the `labelPosition: after` setting that was already present in the easymotion API 😎

Here's a sample of what that particular string looks like in action with the jumptoanywhere motion & current default regex:

src/actions/plugins/easymotion/types.ts
jsonMartin: So all lowercase `jumptoanywhere` is how the original easymotion creators describe it in their vim documentation, so I was just trying to be consistent with the original implementation:
screen shot 2018-03-23 at 12 09 10 pm

However, I agree that the camelcase naming convention is much easier to read and more JS-like. I have went ahead and updated all JS code references to be camelcased named, and only left the original `jumptoanywhere` text as documentation to describe the original matching motion found in standard VIM.

jsonMartin:
Ok @jpoon , I have implemented the requested changes with camelCase naming convention.

Also while preparing that commit, I fixed another bug I just discovered that arose when I refactored the default Regex from a Regex Literal to use `new Regex()` instead (forgot to double escape `\\b` when converting to a string). I might not have noticed it TBH unless you brought up regex, so thanks for that! Didn't originally notice this issue on my end because my custom jumpToAnywhere regex in preferences was working as intended! 😆

Tested it again, and it's working perfectly for me using the default behavior or custom regex. 

Thanks again for the feedback!

README.md
```diff
@@ -422,6 +422,7 @@ Motion Command | Description
 `<leader><leader><leader> bdw`|Start of word
 `<leader><leader><leader> bde`|End of word
 `<leader><leader><leader> bdjk`|Start of line
+`<leader><leader><leader> j`|Jumptoanywhere motion; default behavior matches beginning & ending of word, camelCase, after _ and after #
```
jpoon: `Jumptoanywhere ` Casing :)

src/configuration/configuration.ts
```diff
@@ -203,6 +203,7 @@ class Configuration implements IConfiguration {
   easymotionMarkerFontWeight = 'normal';
   easymotionMarkerYOffset = 0;
   easymotionKeys = 'hklyuiopnm,qwertzxcvbasdgjf;';
+  easymotionJumpToAnywhereRegex = '';
```
jpoon: Let's set the default here: `\\b[A-Za-z0-9]|[A-Za-z0-9]\\b|_.|#.|[a-z][A-Z]`.

jpoon: You'll also need to update the `package.json` to add this value. Check out the comment at the top of this class.

jsonMartin: @jpoon Great idea. Much cleaner logic in `easymotion.cmd.ts` now :)

jsonMartin:
Hey @jpoon, finally had a chance to catch up on things today, have been under the weather these past few days.

I have just finished implementing the requested changes and pushed them up 😎

xconverge:
Hope you are feeling better :) Thanks for getting back into it!

src/actions/plugins/easymotion/types.ts
```diff
@@ -1,4 +1,5 @@
 export type LabelPosition = 'after' | 'before';
+export type jumpToAnywhere = true | false;
```
jpoon: Capital here please as we're exposing the type :)

jsonMartin: OK, updated 👍

PR #2522, opened 2018-04-09
Fixes #2482: Neovim not spawned in appropriate directory

### LINKED ISSUE #2482: ! ex commands are run in the wrong directory
**BUG REPORT**

**What happened**:

`!` ex commands are run in the wrong directory, eg.:

    :r !pwd
    /home/dark/.vscode-insiders/extensions/vscodevim.vim-0.11.3/out/src/neovim

    :!notify-send $(pwd)
    /home/dark/.vscode-insiders/extensions/vscodevim.vim-0.11.3/out/src/neovim
    (shows in a system notification)

Such commands to run in the project directory

    :r !pwd

**Environment**:

- VSCode version: 1.22.0-insider
- OS version: Void Linux 4.13.6_1 x86_64

Chillee:
This makes the neovim process spawn in the directory of the file it's opened in.


src/neovim/neovim.ts
```diff
@@ -14,8 +15,9 @@ export class Neovim implements vscode.Disposable {
   private nvim: Nvim;
 
   async initialize() {
+    const dir = dirname(vscode.window.activeTextEditor!.document.fileName);
```
jpoon: I think what you want is this `vscode.window.activeTextEditor!.document.uri.fsPath`

Chillee: In the documentation, it said that this is 
```
		/**
		 * The file system path of the associated resource. Shorthand
		 * notation for [TextDocument.uri.fsPath](#TextDocument.uri). Independent of the uri scheme.
		 */
```
so I didn't think there was a difference.

jpoon: Hm, that's confusing. Generally, I would understand things named as `fileName` to be just the file name (`source.ts`), whereas `filePath` would be a file path (`/my/dir/source.ts`)

Issue #2545, opened 2018-04-18
Undo replaces file content


Diadlo:
**This a BUG REPORT**

**How to reproduce it**:
1. Open document A
2. Open document B
4. Type `:wq` (or just close and save file)
5. Open B again
6. You can do some changes or do not (most annoying if you do)
7. Press `u`

**What happened**:
Content of document B will be replaced on content of A

**What did you expect to happen**:
Nothing

**Environment**:
- Extension (VsCodeVim) version: 0.11.4
- VSCode version: 1.22.2
- OS version: Arch Linux

doudou:
This is a longstanding bug (see https://github.com/VSCodeVim/Vim/issues/2007). Thank you SO MUCH for figuring out a way to reproduce it. As you can see, a lot of people tried without success.

For the record, I needed to *really open* the documents, i.e. either Ctrl+S or double-click the header. Temporarily opening A did not trigger the bug, a.k.a

1. Open document A
1. CTRL+S
1. Open document B
1. modify B
1. `:wq`
1. Open B again
1. Press `u`

Diadlo:
@doudou You are right. Thanks for thr record :+1:

Chillee:
Awesome! Thanks for the repro. I did some digging a week ago, and I'm pretty sure the commit that caused this issue was https://github.com/VSCodeVim/Vim/issues/1503.

Having a repro so I can make sure both cases still work is great! Hopefully I'll be able to nail down this issue once and for all soon.

dagadbm:
oh nice! I ended up just using visual studio undo and redo commands overriding vscode undos because of this random problem. also vscodevim would also "lowerify" all my text so that was awesome -_-

Chillee:
This particular issue is fixed in https://github.com/VSCodeVim/Vim/pull/2547

Issue #2007, opened 2017-09-06
Pressing 'u' will undo all the stack.


yacinehmito:
* Click *thumbs-up* 👍 on this issue if you want it!
* Click *confused* 😕 on this issue if not having it makes VSCodeVim unusable. 

The VSCodeVim team prioritizes issues based on reaction count. 

--------

**Is this a BUG REPORT or FEATURE REQUEST?** (choose one):
This is a bug report

**Environment**:

- **VSCode Version**: 1.15.1
- **VsCodeVim Version**: 0.10.0
- **OS**: macos 10.12.6

**What happened**:

I press `u`in Normal mode in an editor window. Every single action in the undo stack was undone. Here's a GIF:

![bug vscode](https://user-images.githubusercontent.com/6893840/30132151-b8ea3678-934f-11e7-9580-b51bb3211030.gif)

**What did you expect to happen**: 

I expected the last action on the stack to be undone.

**How to reproduce it**:

I have no idea. I was reliably triggering the bug with the exact buffer I had but at some point it stopped doing it. Then it did it again.

yacinehmito:
And now it's doing it again in another project. When I reloaded the project the bug stopped.

The only similarity between the two instances of the bug is that it happened both times on an HTML-like file (first time on a `vue` file; second time on a proper `html` file).

This is the list of the extensions that I have installed and their versions:

- Auto Close Tag 0.4.3
- Auto Rename Tag 0.0.14
- Code Runner 0.7.1
- Debugger for Chrome 3.2.1
- EditorConfig 0.10.1
- ESLint 1.2.11
- Git History (git log) 0.2.3
- Git Lens 4.4.3
- HTML Snippets 0.1.0
- JavaScript (ES6) code snippets 1.4.1
- Nix 1.0.1
- npm 0.2.1
- npm Intellisense 1.3.0
- Path Intellisense 1.4.2
- Search node_modules 1.1.1
- stylelint 0.29.0
- Vetur 0.9.7
- Vim 0.10.0
- vscode-icons 7.12.0

ryanpage42:
Having same issue

Extension List

- Beautify css/sass/scss/less 1.7.2
- Color Picker 0.4.5
- HTML CSS support 0.1.8
- HTML Snippets 0.1.0
- IntelliSense for CSS class names 1.10.0
- Javascript ES6 Code Snippets 1.4.1
- Live Server 2.1.1
- Node.js Modules IntelliSense 1.4.0
[...]

Issue #928, opened 2016-10-17
u restores multiple edits


octref:
* Click *thumbs-up* 👍 on this issue if you want it!
* Click *confused* 😕 on this issue if not having it makes VSCodeVim unusable. 

The VSCodeVim team prioritizes issues based on reaction count. 

--------

F2 on a word and rename it. Then make one more edit. Press u.

Both the F2 rename and the "one more edit" get undo-ed.
- VSCode Version: 1.7 Insider
- VsCodeVim Version: 0.3.7
- OS: OS X 10.11.6

johnfn:
Interesting. Seems like we should be listening inside didUpdateSelectionEvent or whatever for a text editor change and if so trigger an undo point.

xconverge:
we have onDidChangeTextDocument, it is probably something in there..also this happens during normal mode which is interesting, we probably have logic in there for snippets/insert mode only

offero:
I am seeing similar behavior. The undo is actually replaying history from more than 1 change back. It's really confusing. It's actually bad enough that I have no choice but to switch away from this extension at the moment.

jvbianchi:
I also suffered from this bug

Chillee:
When you guys say you suffer from this bug, do you mean this specific bug, or undo behavior in general?

offero:
In general for me. I didn't press F2 or use rename.

Chillee:
@offero I do think our current undo behavior is fairly wonky, but do you have a specific repro case?

jvbianchi:
In my case is the specific bug (using F2 and rename)

johnfn:
Another great way to repro this is to use TypeScript, use the helper tool to auto insert an import, then undo. 

It seems like cases where text is inserted not through VSCodeVim that we have a problem with.

jleclanche:
Here's another issue with multiple edits

Prerequisite: Create a file with the following contents:
```
One
Two
Three
```

1. Position your cursor on the beginning of the second line (`T`).
2. Hit <kbd>dd</kbd>; the file now looks like `One\nThree`
3. Hit <kbd>↑</kbd>; cursor is now at the beginning of file
4. hit <kbd>P</kbd>; the file now looks like `Two\nOne\nThree`
5. Hit <kbd>↓</kbd>; the cursor is now on the second line
6. Hit <kbd>dd</kbd>; the file now looks like `Two\nThree`

[...]

PR #2547, opened 2018-04-18
Fixed undo issue given in #2545

Chillee:
Thanks to @Diadlo for this repro. https://github.com/VSCodeVim/Vim/issues/2545.

Basically, we don't always want to create a `prevHandler` even if it doesn't exist. To be honest, this code is messy and I'd like to rewrite it entirely, if it wasn't for the fact that this kind of frail code is hard to test.

This should fix many of the reports given in https://github.com/VSCodeVim/Vim/issues/2007. However, the first reports of the issue started before this repro was introduced. That suggests to me that there's actually 2 separate bugs. One where the file gets replaced by another file (issue 1), and one where the file actually drops back to an earlier undo state (issue 2).

Issue 1 was introduced around January 12th in this seemingly innocuous commit: c05006620562db74c43aa73f7a9adb14073ebcb6 .

The first report of this (replaced with another file is here): https://github.com/VSCodeVim/Vim/issues/2007#issuecomment-361799959 Many of the following comments were also expressing this issue.

However, the first report of issue 2 were given here: https://github.com/VSCodeVim/Vim/issues/928 The issue is slightly confused by the fact that the initial bug that @octref reports is a much more innocuous "Certain VSCode actions aren't given their own undo stop", while issue 2 is really "VSCodeVim is wiping out a ton of work". The first real report of issue 2 appears to be https://github.com/VSCodeVim/Vim/issues/928#issuecomment-317941397, on July 26th. I believe that this bug was likely introduced by this commit: https://github.com/VSCodeVim/Vim/pull/1629 on May 3rd.

This commit fixes issue 1. More work is needed to locate the cause of issue 2.

@jpoon @johnfn @xconverge @rebornix


jpoon:
> Basically, we don't always want to create a prevHandler even if it doesn't exist. 

Why?

Chillee:
@jpoon When we close a file, we delete its handler from `ModeHandlerMap`. When we were calling `ModeHandlerMap.getOrCreate` in `getAndUpdateModeHandler`, `prevHandler` did not exist at that step. So, it would create a new handler for the file we just closed. If you take a look at the code for creating a new mode handler, it creates it out of the current file.

This caused us to initialize the undo history of the closed file with the text of the current file. Then, when you reopened that file, pressing undo would set you to the text of the other file.

So, in this repro: https://github.com/VSCodeVim/Vim/issues/2545
1. We have the correct state for document A.
2. We have the correct state for document A & B.
3. Closing document B first deletes the current `modeHandler` we have, but `getOrCreate` creates another modeHandler initialized with the vimState of document A. So, we have the correct state for document A but the wrong one for document B. Only document A is currently open.
4. Opening this file and changing the contents is treated as an undoable change.
5-6. Undoing overwrites the contents of document B with document A.

extension.ts
```diff
@@ -30,7 +30,7 @@ let extensionContext: vscode.ExtensionContext;
 let previousActiveEditorId: EditorIdentity = new EditorIdentity();
 
 export async function getAndUpdateModeHandler(): Promise<ModeHandler> {
-  const [prevHandler] = await ModeHandlerMap.getOrCreate(previousActiveEditorId.toString());
```
jpoon: Maybe we can change this to a `create` function, but I can do that refactor in a later commit.

src/mode/modeHandlerMap.ts
```diff
@@ -16,6 +16,11 @@ class ModeHandlerMapImpl {
     return [modeHandler, isNew];
   }
 
+  async get(key: string): Promise<ModeHandler | null> {
+    let modeHandler = this.modeHandlerMap[key];
```
jpoon: Can probably do this all in one line.

johnfn:
Oh man what a hero!

johnfn:
I agree with your conclusion, too. The ModeHandler stuff is a mess and has been responsible for more bugs than I could count. Worst part is, the biggest benefit of ModeHandler is per-file state, and I can't even remember why we need that. Nearly everything is per-project.

jcjolley:
@Chillee, you're amazing.

jpoon:
@Chillee ok to merge?

Chillee:
Yep!

jpoon:
I'll cut a release for it.

rifotu:
great guys!
thanks a lot :)

PR #2568, opened 2018-04-25
Use extension dir as cwd for neovim process

BinaryKhaos:
Not everything that is being edited has a file backing it. Thus, ``vscode.window.activeTextEditor!.document.uri.fsPath`` might not return something that resolves to an actual existing file/path on disk. We might end up with an invalid cwd for the neovim process which will
lead to a failure to start the neovim process and thus an inoperable plugin state (needs fixing as well).

Opening the vscode settings is one such case where vscode returns an imaginary path "/0/settings.json".

For now, always use the extension directory to be on the safe side. But if further commands are passed through to neovim that might rely on the correct working directory of the file being edited, this will need fixing.

### DIFF
--- a/src/neovim/neovim.ts
+++ b/src/neovim/neovim.ts
@@ -15,9 +15,8 @@ export class Neovim implements vscode.Disposable {
   private nvim: Nvim;
 
   async initialize() {
-    const dir = dirname(vscode.window.activeTextEditor!.document.uri.fsPath);
     this.process = spawn(configuration.neovimPath, ['-u', 'NONE', '-N', '--embed'], {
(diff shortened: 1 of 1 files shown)


BinaryKhaos:
LGTM. I didn't know that the read command was also marked neovimCapable -- I hadn't checked actually, sorry. But nevertheless, I should have gone for your solution straight away -- much more future-/error-proof... don't know what I was thinking. :-) Thanks!

xconverge:
@BinaryKhaos no problem at all, you got the issue fixed, and it will be in the next release!

PR #2570, opened 2018-04-25
Fixed neovim spawning in invalid directories

Chillee:
We can currently spawn neovim in invalid directories, as https://github.com/VSCodeVim/Vim/pull/2568 notes.

This checks to make sure the directory is a real path, otherwise falls back to `__DIR`.


xconverge:
Yes I like the fallback approach

BinaryKhaos:
LGTM. Thanks!

PR #2572, opened 2018-04-26
Fixes #2569. Fix vi{ for nested braces.

### LINKED ISSUE #2569: vi{ does not work correctly in specific nested cases
BUG REPORT

**What happened**:
vi{ selected from the starting brace to the line of the brace *before* the matching brace.

vi{ should select inside matching braces.

Start:
````
{
    te|st
    {
        test
    }
}
````

Keys pressed: vi{d

Expected:
````
{
|}
````
Actual:
````
{
|    }
}
````

**Investigation**:
It looks to me like this is caused by the call to 'getLeft' on line 1443 of motion.ts. In this specific case, where the final brace is the only character on the line (no leading whitespace) and the line above it also has only a single character (with leading whitespace), it ends up treating the character on the line above as the closing brace.

I've tested removing the 'getLeft' call, and it seems to fix the issue without breaking any tests. I'd like to do a bit more manual testing and maybe clean up that code a little (I think that whole block can be merged into the else clause above it), and I should be able to have a PR ready this evening.

**Environment**:

- Extension (VsCodeVim) version: v0.11.5
- VSCode version: 1.22.2
- OS version: Windows 10 (10.0.16299 Build 16299)

Shadaraman:
Fixes the logic for selecting within pairs of braces in certain specific cases:
* Nested braces where the outer closing brace has no leading whitespace and the brace above does
* Closing braces indented by only one space

#2569

### DIFF
--- a/src/actions/motion.ts
+++ b/src/actions/motion.ts
@@ -1433,18 +1433,14 @@ export abstract class MoveInsideCharacter extends BaseMovement {
       }
     } else {
       startPos = startPlusOne;
+
+      // If the closing character is the first on the line, don't swallow it.
+      if (endPos.isInLeadingWhitespace()) {
+        endPos = endPos.getLineBegin();
+      }
+
       if (vimState.currentMode === ModeName.Visual) {
         endPos = endPos.getLeftThroughLineBreaks();
-      }
-    }
-
-    // If the closing character is the first on the line, don't swallow it.
-    if (!this.includeSurrounding) {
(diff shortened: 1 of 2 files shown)


xconverge:
Looks good to me, the code here looks a bit finnicky but as long as we keep adding tests and they keep passing I am ok with it!

Thanks!

xconverge:
Really appreciate the contribution!

Issue #1991, opened 2017-09-01
Pasting in visual mode doesn't update the default buffer's contents


nikklassen:
* Click *thumbs-up* 👍 on this issue if you want it!
* Click *confused* 😕 on this issue if not having it makes VSCodeVim unusable. 

The VSCodeVim team prioritizes issues based on reaction count. 

--------

**This a BUG REPORT**:

**Environment**:

- **VSCode Version**: 1.15.1
- **VsCodeVim Version**: 0.10.0
- **OS**: Mac OS X 10.12.6

**What happened**:

When "pasting over" a selection in visual (or visual line) mode, the text that is removed by that paste isn't put in the default buffer.

**What did you expect to happen**: 

The removed text should now be the contents of the default buffer.

**How to reproduce it**:

Starting with the text
```
hello
world
```
Do
1. dd
2. V
3. p
4. P

This results in

```

hello
hello

```

In Vim the results are

```
world
hello
```

Note: there are extraneous newlines in the VSCode output, but I believe this is a separate issue.

Chillee:
@nikklassen I believe this is an issue specific to visual line mode, not visual mode, right?

nikklassen:
I observed this behaviour in both visual line and visual mode.

Chillee:
Ok, to be more exact, this is an issue specific to having the text being pasted be linewise.

nikklassen:
Sorry, yes. That is correct. The expected behaviour is observed if the text was yanked using a visual selection or a motion like `ye`. This issue only occurs with `yy` or yanks done from visual line mode.

innerdaze:
I'm seeing the same result in vim as we're getting in VSCode. I get this in both:
```
hello
hello
```

Chillee:
@innerdaze I can't reproduce what you're saying in vim. I get the expected
```
world
hello
```

PR #2579, opened 2018-04-28
[Fix] Don't add beginning newline of linewise put in visual-mode

tyru:
Thanks for great work.
I'm using Vim for primary text editor but I have to use VSCode at work.
But I found this extension is awesome because Vim like keybinding helps me a lot to concentrate :)

In original Vim behavior, `p` command in visual-mode replaces current text.
VSCodeVim also does it but it is slightly different.
Unnecessary newline is inserted before the text when the text is yanked as linewise.

1. Yank current line's text by `yy`
1. Replace current line's text by `Vp`

In original Vim, it does not change the text (the buffer is marked modified, though).
But VSCodeVim inserts a newline before the current line.

No exact issue yet.
But the problem is touched by #1991 comment.

> Note: there are extraneous newlines in the VSCode output, but I believe this is a separate issue

### DIFF
--- a/src/actions/commands/actions.ts
+++ b/src/actions/commands/actions.ts
@@ -1289,12 +1289,12 @@ export class PutCommand extends BaseCommand {
       textToAdd = text;
       whereToAddText = dest;
     } else if (
-      vimState.currentMode === ModeName.Visual &&
+      (vimState.currentMode === ModeName.Visual || vimState.currentMode === ModeName.VisualLine) &&
       register.registerMode === RegisterMode.LineWise
     ) {
       // in the specific case of linewise register data during visual mode,
       // we need extra newline feeds
-      textToAdd = '\n' + text + '\n';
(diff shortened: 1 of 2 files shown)


tyru:
oops, okay, we should fix `p` to see if current mode is linewise,
But `vimState.currentMode` seems `2 (Visual)` not `4 (VisualLine)` when `p` is invoked... who changed that? :(

tyru:
Fixed the problem and I added the test case too.
But I don't know why the test case seems to pass also in the master branch... :(

Btw, is there a way to run only certain suites/tests?

tyru:
> But I don't know why the test case seems to pass also in the master branch... :(

It seems just I forgot to transpile test code... ;)
Everything seems to work fine.

Chillee:
@tyru To only run certain test cases, write `newTestOnly` instead of `newTest`.

tyru:
@Chillee Ah, thanks! it saves a lot of time! 😄

tyru:
Thanks all!

PR #2581, opened 2018-04-29
[Fix] Transition between v,V,<C-v> is different with original Vim behavior

tyru:

I expect the same behavior as original Vim :)

* `vv` -> normal mode
* `VV` -> normal mode
* `<C-v><C-v>` -> normal mode

* `vV` -> visual line mode
* `v<C-v>` -> visual block mode
* `Vv` -> visual (char) mode
* `V<C-v>` -> visual block mode
* `<C-v>v` -> visual (char) mode
* `<C-v>V` -> visual line mode

I could not find the issue.

### DIFF
--- a/test/mode/modeVisual.test.ts
+++ b/test/mode/modeVisual.test.ts
@@ -874,4 +874,51 @@ suite('Mode Visual', () => {
       end: ['    func() {', '        hi;', '        alw;', '|        hi;', '        alw;', '    }'],
     });
   });
+
+  suite('Transition between visual mode', () => {
+    test('vv will back to normal mode', async () => {
+      await modeHandler.handleMultipleKeyEvents(['v', 'v']);
+      assertEqual(modeHandler.currentMode.name, ModeName.Normal);
+    });
+
+    test('vV will transit to visual line mode', async () => {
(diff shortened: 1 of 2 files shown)


test/mode/modeVisual.test.ts
```diff
@@ -874,4 +874,51 @@ suite('Mode Visual', () => {
...
+      await modeHandler.handleMultipleKeyEvents(['<C-v>', 'v']);
+      assertEqual(modeHandler.currentMode.name, ModeName.Visual);
+    });
+
+    test('<C-v>V will back to visual line mode', async () => {
+      await modeHandler.handleMultipleKeyEvents(['<C-v>', 'V']);
+      assertEqual(modeHandler.currentMode.name, ModeName.VisualLine);
+    });
+
+    test('<C-v><C-v> will back to normal mode', async () => {
```
xconverge: any chance you can switch these to asset that we got into the correct mode as well? Not a big deal but might be a good idea

C-v...assert in visual block, C-v...assert back in normal etc

tyru:
Fixed test and previous commit message a bit :)

xconverge:
Thank you! This one is a no brainer!

tyru:
Thanks! :smile:

PR #2588, opened 2018-05-01
[Fix] p in visual-mode should save last selection

### LINKED ISSUE #2587: gv does not select the last pasted text with p in visual-mode
# Is this a BUG REPORT or FEATURE REQUEST?

Bug report

# What happened

After `p` command is used in visual-mode, `gv` behavior is wrong.
According to Vim's help (:help gv):

> After using "p" or "P" in Visual mode the text
> that was put will be selected.

# What did you expect to happen

Select the last replaced text range, not the last selected range in visual-mode.
In other word, `p` command in visual-mode should update the last selected range.

# How to reproduce it (as minimally and precisely as possible)

1. Yank `with this text` in `v` (characterwise)

```
replace this
with this text
```

1. Select `<` ~ `>` text at the first line with `v` (characterwise)

```
<replace this>
with this text
```

1. Type `p` to replace it

```
with this text
with this text
```

1. Typing `gv` selects below `<` ~ `>` range

```
<with this te>xt
with this text
```

But I expect

```
<with this text>
with this text
```

# Environment

- Extension (VsCodeVim) version: 71f41af8af0ecfb32d4bb980acb9493d50577dce
- VSCode version: 1.22.2
- OS version: Windows 10 Pro 64bit

tyru:

Please see #2587

This PR covers characterwise and linewise.
Blockwise put seems unimplemented (also in normal-mode), so this PR does not care.

Fixes #2587

### DIFF
--- a/src/actions/commands/actions.ts
+++ b/src/actions/commands/actions.ts
@@ -1325,6 +1325,26 @@ export class PutCommand extends BaseCommand {
       }
     }
 
+    // After using "p" or "P" in Visual mode the text that was put will be
+    // selected (from Vim's ":help gv").
+    if (
+      vimState.currentMode === ModeName.Visual ||
+      vimState.currentMode === ModeName.VisualLine ||
+      vimState.currentMode === ModeName.VisualBlock
+    ) {
+      vimState.lastVisualMode = vimState.currentMode;
+      vimState.lastVisualSelectionStart = whereToAddText;
+      let textToEnd = textToAdd;
+      if (
+        vimState.currentMode === ModeName.VisualLine &&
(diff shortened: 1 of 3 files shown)


tyru:
All done.
Squashed commits in a single commit.

tyru:
Pushed change again because the previous change did not fix the problem when a file is only 1 line.
And sorry, this contains massive refactoring of `PutCommand.exec()`.
I separated visual-mode handling and normal-mode handling into two methods.
Please tell me if you think that is not good change.

tyru:
Ah sorry, I was mistaken.
I pushed wrong change as this branch.

PR #2596, opened 2018-05-03
add easymotion-lineforward and easymotion-linebackward

hy950831:

add two new easymotion commands

`easymotion-lineforward` and `easymotion-linebackward`

The behaviour of these two commands are slightly different from the original easymotion because the current easymotion implementation doesn't allow me to limit the search within a line.

**Special notes for your reviewer**: Sorry for the messy commit logs, i dk why the last PR's(which is already merged) commit logs are still there.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,10 +35,9 @@ before_install:
     sleep 3;
   fi
 
-
 script:
-- gulp forceprettier
-- if [[ $(git diff-index HEAD --) ]]; then
+- npm run forceprettier
+- if [[ $(git diff-index HEAD -- *.js *.ts) ]]; then
     git diff;
     echo "Prettier Failed. Run `gulp` or `gulp forceprettier`";
     exit 1;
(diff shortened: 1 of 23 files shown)


xconverge:
Is this exactly the same as regular easymotion? Do you have some configs in your vimrc for easymotion?

I am looking here and it looks like a configurable option but only uses single leader presses not double

https://github.com/easymotion/vim-easymotion#hjkl-motions

hy950831:
There is no default mapping for lineforward and linebackward command in vim-easymotion. So i followed both the current vscodevim convention(which uses double leaders for word and line moving) and the configuration in the readme of vim-easymotion(which uses l and h as lineforward and backward).
btw, Actually all the default mappings for vim-easymotion use single leader.

xconverge:
Thank you very much!

hy950831:
Thank you for merging~

longdeqidao:
I think `easymotion-lineforward` should be moving with the line [[0]]. How about adding an option like `vim.easymotionWithinLineForward` with default `false`. I think it is more effective than sneak in moving, currently it's bad experience to type 2 chars to jump within line.
@hy950831 @xconverge 

[0]: https://github.com/easymotion/vim-easymotion#hjkl-motions.

Issue #2467, opened 2018-03-22
Feature request: close split window via Ctrl-W, C


nzbart:
I opened a vertical split window with Ctrl-W, V, and then attempted to close it with Ctrl-W, C. The split window didn't close as expected.

Steps to reproduce:
1. Open a file for editing.
2. Split the window using Ctrl-W, V.
3. Attempt to close the newly opened split window with Ctrl-W, C. It doesn't work.

**Environment**:

VsCode 1.21.1
Vim emulation 0.11.2
Windows 10

yogan:
I agree that it would be nice if VSCodeVim would offer this out of the box.
However, you can bind this yourself. Edit `keybindings.json` and add this snippet:

```
    {
        "key":     "ctrl+w c",
        "command": "workbench.action.closeActiveEditor",
        "when":    "vim.mode != 'Insert'"
    }
```

jpoon:
This would be super easy to add. As an example see: https://github.com/VSCodeVim/Vim/blob/master/src/actions/commands/actions.ts#L2639.

tyru:
Fixed in #2600 (sorry `Fixes #xxx` comment was wrong...)

PR #2600, opened 2018-05-03
Add missing window keys (<C-w><C-[hjklovq]>)

### LINKED ISSUE #2582: <C-w><C-l> doesn't work (but <C-w>l works)
# Is this a BUG REPORT or FEATURE REQUEST?

Bug report

# What happened

Nothing happened.

The following keymappings does not work:

* `<C-w><C-l>`
* `<C-w><C-k>`
* `<C-w><C-j>`
* `<C-w><C-h>`

# What did you expect to happen

Same as `<C-w>l`, `<C-w>k`, `<C-w>j`, `<C-w>h`, jump to the specified window.

# How to reproduce it (as minimally and precisely as possible)

Always happened when the above 4 key sequences are typed

# Environment

- Extension (VsCodeVim) version: 71f41af8af0ecfb32d4bb980acb9493d50577dce
- VSCode version: 1.22.2
- OS version: Windows 10 Pro 64bit

tyru:

Some commands are even implemented but only keymappings are missing.
This PR makes it to work just to add the keymappings to invoke the implementation.

Fixes #2582, #2467

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,53 +72,9 @@
         "when": "editorTextFocus && vim.active && !inDebugRepl"
       },
       {
-        "key": "cmd+left",
-        "command": "extension.vim_cmd+left",
-        "when": "editorTextFocus && vim.active && vim.use<D-left> && !inDebugRepl && vim.mode != 'Insert'"
-      },
-      {
-        "key": "cmd+right",
-        "command": "extension.vim_cmd+right",
-        "when": "editorTextFocus && vim.active && vim.use<D-right> &&  !inDebugRepl && vim.mode != 'Insert'"
-      },
-      {
-        "key": "cmd+d",
-        "command": "extension.vim_cmd+d",
-        "when": "editorTextFocus && vim.active && vim.use<D-d> && !inDebugRepl"
-      },
-      {
-        "key": "cmd+a",
-        "command": "extension.vim_cmd+a",
-        "when": "editorTextFocus && vim.active && vim.use<D-a> && !inDebugRepl && vim.mode != 'Insert'"
-      },
-      {
-        "key": "ctrl+d",
-        "command": "extension.vim_ctrl+d",
-        "when": "editorTextFocus && vim.active && !inDebugRepl"
-      },
-      {
-        "key": "ctrl+alt+down",
-        "linux": "shift+alt+down",
-        "mac": "cmd+alt+down",
-        "command": "extension.vim_cmd+alt+down",
-        "when": "editorTextFocus && vim.active && !inDebugRepl"
-      },
-      {
-        "key": "ctrl+alt+up",
-        "linux": "shift+alt+up",
-        "mac": "cmd+alt+up",
-        "command": "extension.vim_cmd+alt+up",
-        "when": "editorTextFocus && vim.active && !inDebugRepl"
-      },
-      {
         "key": "Backspace",
         "command": "extension.vim_backspace",
         "when": "editorTextFocus && vim.active && !inDebugRepl"
-      },
-      {
-        "key": "shift+backspace",
-        "command": "extension.vim_shift+backspace",
-        "when": "editorTextFocus && vim.active && vim.use<shift+BS> && vim.mode == 'SearchInProgressMode' && !inDebugRepl"
       },
       {
         "key": "Delete",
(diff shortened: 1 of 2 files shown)


jpoon:
If you could do all the `ctrl` keys that would be most appreciated. However, I would also understand if it's too much work.

package.json
```diff
@@ -330,11 +335,6 @@
         "command": "list.collapse",
         "when": "listFocus && !inputFocus"
       },
-      {
-        "key": "g g",
```
jpoon: We need this one. Mind putting it back?

tyru: After removing this, `gg` seems to work.

I'm new to VSCode extension development and I'm not used to package.json and what should I write to `contributes.commands`. so please correct me if I'm wrong.
But unlike other VSCode extensions, I guess it's enough to lists only one key sequences here to send to VSCodeVim's state machine.
The state machine processes keys one by one, and fire the command if key sequence is matched.

tyru:
> If you could do all the ctrl keys that would be most appreciated.

Okay, I'll try it :)

tyru:
Sorted `contributes.commands` alphabetically, but #2601 and another PR(s) have broken the master build.
Please wait merging this PR until the fix.

tyru:
> #2601 and another PR(s) have broken the master build.

Sent #2609.
It fixes broken master.

But I found another problem.
The problem is that the test codes depend on user settings (settings.json).
After I changed [my config](https://gist.github.com/tyru/8f4d7bdfac08d3d83e859091ab99e93c) to `"editor.autoClosingBrackets": false`, the below test case failed.
because it assumes inputting `(` makes buffer content to `()`.

https://github.com/VSCodeVim/Vim/blob/16a234863b33635f688f8df796d7d1d6e31a1fd7/test/mode/modeInsert.test.ts#L209

Removing this line passes the test, however.

package.json
```diff
@@ -72,209 +72,220 @@
...
-        "when": "editorTextFocus && vim.active && vim.use<D-v> && vim.mode == 'SearchInProgressMode' && !inDebugRepl"
+        "key": "ctrl+j",
+        "command": "extension.vim_navigateCtrlJ",
+        "when": "editorTextFocus && vim.active && vim.use<C-j> && !inDebugRepl"
       },
       {
-        "key": "ctrl+[",
-        "command": "extension.vim_ctrl+[",
-        "when": "editorTextFocus && vim.active && vim.use<C-[> && !inDebugRepl"
+        "key": "ctrl+k",
```
jpoon: I think we can merge the two `ctrl+k` to 

```
      {
        "key": "ctrl+k",
        "command": "extension.vim_ctrl+k",
        "when": "editorTextFocus && vim.active && vim.use<C-k> && !inDebugRepl"
      },
```

package.json
```diff
@@ -72,209 +72,220 @@
...
-        "when": "editorTextFocus && vim.active && vim.use<D-c> && vim.overrideCopy && !inDebugRepl"
+        "key": "ctrl+u",
+        "command": "list.focusPageUp",
+        "when": "listFocus && !inputFocus"
       },
       {
-        "key": "ctrl+a",
-        "command": "extension.vim_ctrl+a",
-        "when": "editorTextFocus && vim.active && vim.use<C-a> && !inDebugRepl"
+        "key": "ctrl+v",
```
jpoon: I think we can merge these two `ctrl+v` as well:
```
      {
        "key": "ctrl+v",
        "command": "extension.vim_ctrl+v",
        "when": "editorTextFocus && vim.active && vim.use<C-v> && !inDebugRepl"
      },
```

package.json
```diff
@@ -286,69 +297,58 @@
...
       {
-        "key": "h",
-        "command": "list.collapse",
-        "when": "listFocus && !inputFocus"
+        "key": "cmd+c",
+        "command": "extension.vim_cmd+c",
+        "when": "editorTextFocus && vim.active && vim.use<D-c> && vim.overrideCopy && !inDebugRepl"
       },
       {
-        "key": "g g",
```
jpoon: I'm fairly certain we need this. When the explorer window is selected, pressing `gg` will take you to the first item in the list.

jpoon:
Thanks for sorting the `package.json`. Makes it easier to read/review.

tyru:
> Test should pass on CI though? We need that test as we have has special logic to delete that matching ).
> 
> The tests already have a mechanism to override the "vim" specific configurations (https://github.com/VSCodeVim/Vim/blob/master/test/testConfiguration.ts), maybe we need to do something similar for the global settings.

Ah, okay, forgot about that 😅

package.json
tyru: Oooops I'm sorry... entirely I didn't care the explorer window and the `when` condition`.
Will restore it later.

package.json
tyru: Thanks. how about duplicate `ctrl+h` and `ctrl+j`?
In the same way, I should merge like this?

```
      {
        "key": "ctrl+h",
        "command": "extension.vim_ctrl+h",
        "when": "editorTextFocus && vim.active && vim.use<C-h> && !inDebugRepl"
      },
      {
        "key": "ctrl+j",
        "command": "extension.vim_ctrl+j",
        "when": "editorTextFocus && vim.active && vim.use<C-j> && !inDebugRepl"
      },
```

jpoon: Oh yeah, I didn't spot those duplicates. That looks good.

tyru:
Okay, fixed problems 🎉 
Failing tests should pass if #2609 is merged.

tyru:
Thanks @jpoon for merging, now this branch is all green.

All my branches are left except refactoring `p` implementation in visual-mode.
The refactoring said here https://github.com/VSCodeVim/Vim/pull/2609#issuecomment-386592394 is bit difficult, so please feel free to ship the next release :)

PR #2602, opened 2018-05-03
[Fix] p in visual-mode should update register content

### LINKED ISSUE #1991: Pasting in visual mode doesn't update the default buffer's contents

tyru:

This PR fixes `p` command to update register content, when it is executed in visual-mode.
Please see #1991 for the original Vim's behavior.

NOTE: this PR only fix to update register content.
`p` command is yet not Vim compatible behavior.
For example, the buffer content is:

```
hello
world
```

And cursor position is line 1, and type `ddVpP` results in:

```
world
hello

```

Unnecessary newline at the end.
Expected is:

```
world
hello
```

But this is another problem.
I will fix it later at the another PR.

Fixes #1991

### DIFF
--- a/test/mode/modeVisualLine.test.ts
+++ b/test/mode/modeVisualLine.test.ts
@@ -320,6 +320,8 @@ suite('Mode Visual Line', () => {
     title: 'Vp updates register content',
     start: ['|hello', 'world'],
     keysPressed: 'ddVpP',
+    // TODO: this is not the same behavior as original Vim.
+    // But currently unnecessary line is left at the end (see #2602).
     end: ['|world', 'hello', ''],
   });
 
--- a/test/mode/modeVisualLine.test.ts
+++ b/test/mode/modeVisualLine.test.ts
@@ -316,6 +316,13 @@ suite('Mode Visual Line', () => {
     });
   });
 
+  newTest({
+    title: 'Vp updates register content',
+    start: ['|hello', 'world'],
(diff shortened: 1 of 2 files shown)


test/mode/modeVisualLine.test.ts
```diff
@@ -316,6 +316,13 @@ suite('Mode Visual Line', () => {
     });
   });
 
+  newTest({
+    title: 'Vp updates register content',
+    start: ['|hello', 'world'],
+    keysPressed: 'ddVpP',
+    end: ['|world', 'hello', ''],
```
xconverge: can you add a comment here mentioning that this test is actually not truly "correct"

tyru: done :)

Issue #2617, opened 2018-05-06
[FEATURE REQUEST] "q:" command


KamikazeZirou:
**FEATURE REQUEST**:
I would like to often run command from command-line history. 
So I request "q:" command that opens command-line window.
"q:" command may modestly relieve issue #779.

xconverge:
Here is the reference material for this command:

http://vim.wikia.com/wiki/Using_command-line_history

xconverge:
This was implemented in #2618

PR #2618, opened 2018-05-06
Implement "q:" command

KamikazeZirou:
I implemented "q:" command.

#2617

This "q:" command slightly differ from original.
Command is not immediately executed when the command is selected from the history.
Instead of executing command, display it on command-line.
Currently, there is no way to edit past commands, so I thought that this kind of action was better.

### DIFF
--- a/extension.ts
+++ b/extension.ts
@@ -265,6 +265,8 @@ export async function activate(context: vscode.ExtensionContext) {
     }
   }
 
+  CommandLine.SetHistoryDirPath(context.extensionPath);
+
   registerCommand(context, 'toggleVim', async () => {
     configuration.disableExt = !configuration.disableExt;
     toggleExtension(configuration.disableExt);
--- a/src/actions/base.ts
+++ b/src/actions/base.ts
@@ -38,6 +38,14 @@ export let compareKeypressSequence = function(one: string[] | string[][], two: s
     );
   };
 
+  const isSingleAlpha = (s: string): boolean => {
+    if (s.length === 1 && (('a' <= s && s <= 'z') || ('A' <= s && s <= 'Z'))) {
+      return true;
+    } else {
+      return false;
+    }
+  };
+
   for (let i = 0, j = 0; i < one.length; i++, j++) {
     const left = one[i],
       right = two[j];
@@ -53,6 +61,13 @@ export let compareKeypressSequence = function(one: string[] | string[][], two: s
       continue;
     }
     if (right === '<number>' && isSingleNumber(left)) {
+      continue;
+    }
+
+    if (left === '<alpha>' && isSingleAlpha(right)) {
+      continue;
+    }
+    if (right === '<alpha>' && isSingleAlpha(left)) {
       continue;
     }
 
--- a/src/actions/commands/actions.ts
+++ b/src/actions/commands/actions.ts
@@ -371,7 +371,7 @@ class CommandInsertRegisterContentInSearchMode extends BaseCommand {
 @RegisterAction
 class CommandRecordMacro extends BaseCommand {
   modes = [ModeName.Normal, ModeName.Visual, ModeName.VisualLine];
-  keys = ['q', '<character>'];
+  keys = [['q', '<alpha>'], ['q', '<number>']];
 
   public async exec(position: Position, vimState: VimState): Promise<VimState> {
     const register = this.keysPressed[1];
@@ -1701,6 +1701,31 @@ class CommandShowCommandLine extends BaseCommand {
   public async exec(position: Position, vimState: VimState): Promise<VimState> {
     vimState.recordedState.transformations.push({
       type: 'showCommandLine',
+    });
+
+    if (vimState.currentMode === ModeName.Normal) {
+      vimState.commandInitialText = '';
+    } else {
+      vimState.commandInitialText = "'<,'>";
+    }
+    vimState.currentMode = ModeName.Normal;
+
+    return vimState;
+  }
+}
+
+@RegisterAction
+class CommandShowCommandHistory extends BaseCommand {
+  modes = [ModeName.Normal, ModeName.Visual, ModeName.VisualLine, ModeName.VisualBlock];
+  keys = ['q', ':'];
+
+  runsOnceForEveryCursor() {
+    return false;
+  }
+
+  public async exec(position: Position, vimState: VimState): Promise<VimState> {
+    vimState.recordedState.transformations.push({
+      type: 'showCommandHistory',
     });
 
     if (vimState.currentMode === ModeName.Normal) {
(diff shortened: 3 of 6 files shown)


KamikazeZirou:
I'm sorry I fixed code before receiving a comment.
I noticed a bug with the first code.

src/actions/commands/actions.ts
```diff
@@ -371,7 +371,7 @@ class CommandInsertRegisterContentInSearchMode extends BaseCommand {
 @RegisterAction
 class CommandRecordMacro extends BaseCommand {
   modes = [ModeName.Normal, ModeName.Visual, ModeName.VisualLine];
-  keys = ['q', '<character>'];
+  keys = [['q', '<alpha>'], ['q', '<number>']];
```
Chillee: I supposed we should also allow `q"` here.

src/cmd_line/commandLine.ts
```diff
@@ -8,7 +8,75 @@ import * as parser from './parser';
...
+        console.log('CommandLine: Failed to load history.');
+      }
+    } catch (e) {
+      console.error(e);
+    }
+  }
+
+  public save(): void {
+    const fs = require('fs');
+    fs.writeFileSync(this._filePath, JSON.stringify(this._history), 'utf-8');
```
Chillee: Should this be asynchronous? I'd prefer to avoid synchronous file calls in general, but in this case, I don't see any advantage to do it synchronously.

Perhaps the other file calls should also be synchronous too, but I don't have a strong opinion on that.

Chillee:
Oh this is dope! I've thought it was irritating for a while that we couldn't implement `<up>`/`<down>` in the command line, but I didn't even know this was a thing.

extension.ts
```diff
@@ -265,6 +265,8 @@ export async function activate(context: vscode.ExtensionContext) {
     }
   }
 
+  CommandLine.SetHistoryDirPath(context.extensionPath);
```
Chillee: Also, I'm not entirely sure how I feel about this, if only for the selfish reason that it clutters up the development folder.

Nothing wrong with adding it to the .gitignore, but is there a better place to put this?

src/cmd_line/commandLine.ts
parkovski: I have an external HDD that takes a while to spin up. It looks like right now history is being written to the extension folder, but if it was ever made per-project, I'm worried these sync calls could cause a ton of lag. No way around it for loading history, but at least for saving, can this be run async at the same time as the command (e.g. don't await it)?

Similarly for load, there's no way around a slow disk read, but will the sync call cause the editor to freeze if loading takes a long time?

Chillee: Well, if it's made asynchronous, it shouldn't lock up the editor. It won't be able to display until it's done loading, but I don't think that's avoidable. It should only need to load once per VSCode session though.

I'm leaning towards making all the file operations asynchronous.

src/actions/base.ts
```diff
@@ -38,6 +38,14 @@ export let compareKeypressSequence = function(one: string[] | string[][], two: s
     );
   };
 
+  const isSingleAlpha = (s: string): boolean => {
+    if (s.length === 1 && (('a' <= s && s <= 'z') || ('A' <= s && s <= 'Z'))) {
```
jpoon: Regex? ^[a-zA-Z]$

src/cmd_line/commandLine.ts
```diff
@@ -8,7 +8,75 @@ import * as parser from './parser';
 import * as util from '../util';
 import { VimError, ErrorCode } from '../error';
 
+class CommandLineHistory {
```
jpoon: Can you put this in it's own separate file?

extension.ts
jpoon: As it's stored using the extension path which should be shared across vscode instances. we would effectively have one global history across workspaces/projects. do we want that?

Chillee: I think so. That's what vim seems to do, at least.

src/cmd_line/commandLine.ts
jpoon: Async please. Especially considering you are reading it from memory on the `gets` so there's no reason to block on the file write.

jpoon:
Thanks @KamikazeZirou! Also, please add tests.

src/actions/commands/actions.ts
KamikazeZirou: Thanks for pointing out bugs.
I fix this problem(0a635c35a225c39ed1dc2a8b6d1e5e6010183368).

src/cmd_line/commandLine.ts
KamikazeZirou: I made file operations asynchronous.
0a635c35a225c39ed1dc2a8b6d1e5e6010183368

Is it too overworked to save the command executed while loading the file?

src/actions/base.ts
KamikazeZirou: Yeah!.
Should I use regex here?
isSingleNumber does not use regex, so I adapted it.

src/cmd_line/commandLine.ts
KamikazeZirou: Yes.
e43686d36bbf17158c96c20533b8fe1cf771aaf1

extension.ts
KamikazeZirou: vim puts. viminfo in the home directory.
If so, is it better to put this history file in your home directory?
e.g. "~/.cmdline_history", "~/.VSCodeVim/.cmdline_history"

src/actions/base.ts
Chillee: `isSingleNumber` could be regex too, but it's a little bit simpler so I'm more ambivalent.

extension.ts
jpoon: My vote is `~/.VSCodeVim/.cmdline_history`.

KamikazeZirou: For now, I set the file location to \~/.VSCodeVim/.cmdline_history.
a59b00aae406dcb6758d9841f433fb877714d4a0

src/actions/base.ts
KamikazeZirou: I fix it.
1cb61f1f724e058e51774d25acfaf18b4525ef29

KamikazeZirou:
@jpoon . I added tests.

If there is no further indication, I think that this work is completed

src/cmd_line/commandLineHistory.ts
```diff
@@ -5,7 +5,7 @@ import { Logger } from '../util/logger';
 
 export class CommandLineHistory {
   private _history: string[] = [];
-  private _is_loaded: boolean = false;
+  private _is_loading: boolean = false;
```
parkovski: Can you turn this into a promise chain/queue instead? Basically, each time save or load is called, chain it onto the last promise with `finally` and set the member to the new promise.
- You won't have to fail in save because it will never run at the same time as a load
- You won't have to check for new history in load because when it runs it will be the latest operation
- Saves will always write in order, which is more important if this ever gets changed to just appending to the history file or for slow disks.

PR #2622, opened 2018-05-07
Stay cursor option

### LINKED ISSUE #2397: Stay on (yank, indent, outdent, delete, etc) feature
**Is this a BUG REPORT or FEATURE REQUEST?** (choose one): Feature Request

One of the nice things I like about Atom's popular vim plugin, vim-mode-plus, is the option to enable `stayOn{Yank,Delete,etc}`. These options "try to keep original cursor position as much as posible after operation". It is an intentional difference from pure vim, and is behind a configuration flag.

I've got a local copy of this plugin that mimics that behavior - would you be open to a PR that adds similar options to this plugin? Just wondering before I go through the effort to package up a neat PR for it.

davewasmer:
Adds options to keep the cursor in it's original position after an operation. See #2397 for more context.

Fixes #2397

### DIFF
--- a/src/configuration/iconfiguration.ts
+++ b/src/configuration/iconfiguration.ts
@@ -205,4 +205,14 @@ export interface IConfiguration {
   insertModeKeyBindingsNonRecursive: IKeyRemapping[];
   otherModesKeyBindings: IKeyRemapping[];
   otherModesKeyBindingsNonRecursive: IKeyRemapping[];
+
+  /**
+   * Cursor movement options
+   */
+  stayOnYank: boolean;
+  stayOnUpperCase: boolean;
+  stayOnLowerCase: boolean;
+  stayOnToggleCase: boolean;
+  stayOnIndent: boolean;
+  stayOnOutdent: boolean;
 }
--- a/src/configuration/iconfiguration.ts
+++ b/src/configuration/iconfiguration.ts
@@ -214,5 +214,7 @@ export interface IConfiguration {
   stayOnLowerCase: boolean;
   stayOnToggleCase: boolean;
   stayOnIndent: boolean;
+  stayOnVisualBlockIndent: boolean;
   stayOnOutdent: boolean;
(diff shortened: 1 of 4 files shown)


Chillee:
You might have an old version of the `vscode` dependency installed. At least, that's where I recall running into this Travis failure.

Thanks for the contribution! 

Overall, I think I'd prefer it if we put the settings in an array with a schema for autocomplete/ensuring the user has the right settings. This would also be an easy way of allowing the user to enable all the settings.

So the user might be able to do
`vim.stayOn: ['yank', 'uppercase', 'lowercase']`
or 
`vim.stayOn: 'all'`
or
`vim.stayOn: 'none'`

jpoon:
+1 on @Chillee suggestion on making the config an array of items.

src/configuration/iconfiguration.ts
```diff
@@ -205,4 +205,16 @@ export interface IConfiguration {
   insertModeKeyBindingsNonRecursive: IKeyRemapping[];
   otherModesKeyBindings: IKeyRemapping[];
   otherModesKeyBindingsNonRecursive: IKeyRemapping[];
+
+  /**
+   * Cursor movement options
+   */
+  stayOnYank: boolean;
```
jpoon: Please also add these configurations to https://github.com/VSCodeVim/Vim/blob/c8d6d34a381949c5bbaf4ce7330fcc1497a12560/test/testConfiguration.ts

jpoon:
Argh. Our build is still *not* failing on transpiling failures.

jpoon:
Alrighty, fixed our builds to actual fail if there are any transpile errors (https://github.com/VSCodeVim/Vim/pull/2625). Makes it a bit more obvious to read why Travis failed.

jpoon:
Doing a bit of spring cleaning. Closing this PR as it's getting stale. Feel free to re-open after addressing the test failures and CR comments.

PR #2623, opened 2018-05-07
feat: log to outputChannel

jpoon:

* Initial implementation of logger which logs messages to the output channel
* Hopefully, as we start logging more output messages, users can get more insight on what's going on and if there are any issues, help with debugging. 


src/mode/modeHandler.ts
```diff
@@ -71,16 +70,31 @@ export class ModeHandler implements vscode.Disposable {
...
-    }, 0);
+    this.syncCursors();
 
     // Handle scenarios where mouse used to change current position.
     const disposable = vscode.window.onDidChangeTextEditorSelection(e => {
       if (configuration.disableExt) {
         return;
       }
 
+      if (Globals.isTesting) {
```
xconverge: Why do you think this is better here instead of in the actual handler?

Do you know that the handler was only called from here?

jpoon: Tbh, I moved it because I didn't see a reason for it to be in the actual handler. This was also the only spot the handler was being called and so decided to move it up here to save us from queuing something that was just going to exit shortly thereafter.

xconverge: ok, we just don't test this function at all so it terrifies me

xconverge:
This is really cool though!

Issue #935, opened 2016-10-18
Disable Input Method when exiting Insert Mode


kwopera:
---

Please _thumbs-up_ 👍 this issue if it personally affects you! You can do this by clicking on the emoji-face on the top right of this post. Issues with more thumbs-up will be prioritized.

---
### What did you do?

Typing in Chinese in 'Insert Mode' with system Input Method in Mac OSX then exit 'Insert Mode' with 'ESC' key.
### What did you expect to happen?

In MacVim, the Chinese Input Method will be disabled (switched to English mode) or if ':set noimdisable' is used, the Input Method will be switched to 'U.S English' or whatever default Input Method is set in the system. This is to ensure once one exit the Insert Mode, Normal Mode commands can be used immediately without the Input Method intercepting the keystrokes.
### What happened instead?

In VS Code with Vim mode plugin, the editor would just exit to Normal Mode without switching off the Input Method, and one would need to manually disable the Input Method to be able to execute commands such as 'h, j, k, l'
### Technical details:
- VSCode Version: 1.6.1
- VsCodeVim Version: 0.3.7
- OS: Mac OSX El Captain 10.11.6

rebornix:
It's the same behavior as vanilla Vim. But basically I like this idea.

kwopera:
Thank you! It would be very helpful. I'll try to work out a way to implement this function in the meantime.

rebornix:
@kwopera go ahead! This feature includes two pieces of code change, this first one is adding a config at https://github.com/VSCodeVim/Vim/blob/master/src/configuration/configuration.ts#L52 and the second is switching IME when Vim switches mode https://github.com/VSCodeVim/Vim/blob/master/src/mode/modeHandler.ts#L753 .

kwopera:
Great! Thank you so much for your help!! :)

kyuwoo-choi:
@kwopera @rebornix  any progress on this issue?

haolly:
@rebornix when vim compiled whit +multi_byte_ime, it will have the behaviour
Missing this feature too

hh9527:
any progress on this issue? very expecting on this feature.

realwangyibo:
@rebornix This feature is provided by plugin [fcitx.vim](https://github.com/lilydjwg/fcitx.vim), on macOS using https://github.com/CodeFalling/fcitx-remote-for-osx

Hope this helps:smile:

heyeshuang:
[...]

PR #2643, opened 2018-05-10
Integrate SmartIM to VSCodeVim

daipeihust:
This PR integrate vim plugin [SmartIM](https://github.com/ybian/smartim) which can change input method to english when you exit insert mode to normal mode

[#935 Disable Input Method when exiting Insert Mode](https://github.com/VSCodeVim/Vim/issues/935) 

- I only make it work on macOS on this PR
- I add three configuration to control this function
- the default configuration for this function is disabled, so it won't influence VSCodeVim's main function
- this function depend on [im-select](https://github.com/ybian/smartim/blob/master/plugin/im-select),  i would add a guide about how to make this function work to README.md
- I have run the test and pass it
- this function works fine in my VSCode
- It's my first PR, I wish it can be merged


src/actions/plugins/smartim.ts
```diff
@@ -0,0 +1,93 @@
...
+        oldMode !== ModeName.Replace
+      ) {
+        this.enableOriginInputMethod();
+      }
+    }
+  }
+
+  // save origin input method and set input method to default
+  public async disableOriginInputMethod() {
+    if (process.platform === 'darwin') {
```
xconverge: Does this need to be platform dependent? I know you only tested it on osx, but to I don't see anything that is platform dependent here

xconverge: Woops nevermind, https://github.com/ybian/smartim#why-is-it-mac-only

xconverge:
Can you add the README.md changes to this PR too please! Thanks for the contribution, you did a great job of encapsulating it, having it disabled by default, etc, so thanks a lot!

jpoon:
Is there a way to make this work x-plat? I noticed the reliant exe is also MacOS only.

xconverge:
Yea, I came to the same conclusion, I think it can be expanded for cross platform use but I know NOTHING about what IM clients are common

tyru:
I also want this, but I think more generic way is better.

For example, current implementation depends `im-select`.
and including `SmartIM` in options and source codes is not apporopriate, because maybe users and developers don't know SmartIM.

This is just an **example** for more generic option.

```json
{
  "vim.disableIMCmd": {
    "type": "invoke-command",
    "args": ["/usr/local/bin/im-select", "com.apple.keylayout.US"]
}
```

daipeihust:
@tyru Your idea is great, I'll think about it and make some changes

daipeihust:
@xconverge Hey, I have added the guide to README.md and make some optimization in my code.

daipeihust:
@tyru I deleted the SmartIM in my code and changed it to InputMethodSwitcher

the configuration is:

```json
"vim.autoSwitchInputMethod": true,
"vim.autoSwitchInputMethodConfig": {
    "dependencyPath": "/usr/local/bin/im-select",
    "defaultInputMethodKey": "com.apple.keylayout.US"
}
```
How do you like it?

daipeihust:
@jpoon I have read the implementation of im-select, the source code less than 30 lines. So I believe it's easy to implement im-select on windows and linux. I will try my best to make it cross-platform.

tyru:
@daipeihust Thanks! This is my thoughts. I'm not an author nor a maintainer :)

# Argument format

Your implementation requires the executable (`im-select`) to recognize the certain format of arguments.

* Disable: `im-select {layout}`
* Get status: `im-select`

But another program may not fit into that. for example, `xkb-switch` is:

* Disable: `xkb-switch -s ARG`
* Get status: `xkb-switch -p`

However, it seems `xkb-switch` also returns current layout. But other programs may not.
So it should provide the ways to disable/get status separatedly like this.

```json
{
  "vim.autoSwitchIM": true,
  "vim.autoSwitchIMConfig": {
    "disable": {
      "command": "/usr/local/bin/im-select",
      "args": ["com.apple.keylayout.US"]
    },
    "getStatus": {
      "command": "/usr/local/bin/im-select",
      "args": []
    }
  }
}
```

# Keyboard layout != IM status

This is not a review comment.
But just-for-your-information comment :)

On macOS, it is treated as same (or more precisely, there is no "IM status").
But for other OSes, it has different meaning.

* Changing keyboard layout to `en_US`
* Disable input method

It is slightly different.

macOS IM only **switches** direct input and other input method engines, such as Japanese input method engine "ことえり" (Kotoeri).
(I heard ibus (input method framework of *nix) also handles like this)
Input method engine handles all characters what user typed, and converts it to multibyte characters.
Direct input just pass-through all characters as-is.

But on Windows, it **activates** input method when typing multibyte text, and **disables** input method when typing ascii text (when writing program or english comment).

IME is complicated, so I'm not confident, but maybe your implementation seems good.
It's just comparing the output of program and the output when keyboard layout is direct input or IM is disabled (= `defaultInputMethodKey`).

On Windows, we need to just create a program that disables IM, or gets current status like `off`, `on`.

Or, as @t9md [says](https://github.com/VSCodeVim/Vim/issues/935#issuecomment-369502635), if there is a multi-platform way to achieve this, it is the best way I think...

tyru:
> So it should provide the ways to disable/get status separatedly like this.
> 
> ```json
> {
>   "vim.autoSwitchIM": true,
>   "vim.autoSwitchIMConfig": {
>     "disable": {
>       "command": "/usr/local/bin/im-select",
>       "args": ["com.apple.keylayout.US"]
>     },
>     "getStatus": {
>       "command": "/usr/local/bin/im-select",
>       "args": []
>     }
>   }
> }
> ```

Or, after this PR, `vim.autoSwitchIMConfig` should be an array if we will implement `when` property like keybindings in `package.json`.
The following config can select `im-select` on macOS, and select `xkb-switch` on Linux :)
So if we provide **good default values** here, users don't have to change `vim.autoSwitchIMConfig`.

@jpoon @xconverge What do you think?

```json
{
  "vim.autoSwitchIM": true,
  "vim.autoSwitchIMConfig": [
    {
      "disable": {
        "command": "im-select",
        "args": ["com.apple.keylayout.US"]
      },
      "getStatus": {
        "command": "im-select",
        "args": []
      },
      "when": "OS == 'macOS' && executable('im-select')"
    },
    {
      "disable": {
        "command": "xkb-switch",
        "args": ["-s"]
      },
      "getStatus": {
        "command": "xkb-switch",
        "args": ["-p"]
      },
      "when": "OS == 'Linux' && executable('xkb-switch')"
    }
  ]
}
```

daipeihust:
@tyru I kown your suggestion is generic, but I think it's too complicated. Users shouldn't know the implementation of it, just like the disable and getStatus operation. And in fact, we didn't disable the input engine, we just switch input method to English. I believe there is a key to identify a input method or a keyboard layout on any platform, so the setting can be simple. We only need the command(im-select or something else) and the default key of English IM from setting.

so the setting for mac users can be:

```
"vim.autoSwitchIM": true,
"vim.autoSwitchIMConfig": {
    "command": "/usr/local/bin/im-select",
    "defaultIMKey": "com.apple.keylayout.US"
}
```

for windows:

t9md:
> And I have read @t9md 's say, I agree that his implementation is better, He disabled IM engine when exiting Insert Mode. And I have cloned the VSCode's source code. I'm working on it to see if we can make it work on VSCode.

Though I didn't read whole thread,  I at least can say, my implementation is NOT good, its's just simple to implement.
My vim-mode-plus's way is a bit hacky and have limitation such as IM is still enabled in command-palette in normal-mode.
Since my approach actually doesn't have IM disable, it's just mimic as if IM disabled.

Atom have [vim-mode-plus-auto-ime](https://github.com/xream/atom-vim-mode-plus-auto-ime) package which just invoke external command when switching vim's operation modes(insert/normal).
I believe this is better approach and your PR is also taking this approach.

tyru:
@daipeihust 

> ```
> if (macOS) {
>   execute(command + defaultIMKey)
> } else if (linux) {
>   execute(command + '-s' + defaultIMKey)
> } else {
>   ...
> }
> ```
> 
> We should choose specific command on each platform and make our code work on it

Hmm, I thought it locks in the config with specific command, but certainly it may not need to be so generic.

And you don't mention it though, now I think `when` is too generic... it is hard to implement correctly and the advantage is not so big. sorry for confusion.

daipeihust:
@tyru I've browsed some stuff about IM on linux, I found that there are so many different IM framework on linux. Now I think you are right. We should provide the ability to execute different command to change IM.

We just put command and arguments together and call them command. What do you think the following config?

On macOS:
```
"vim.autoSwitchIM": true,
"vim.autoSwitchIMCommand": {
    "getInputMethod": "/usr/local/bin/im-select",
    "switchToEnglish": "/usr/local/bin/im-select com.apple.layout.US"
}
```

On Windows:
```
"vim.autoSwitchIM": true,
"vim.autoSwitchIMCommand": {
    "getInputMethod": "/path/to/xkb-switch -p",
    "switchToEnglish": "/path/to/xkb-switch -s com.windows.xxx"
}
```

tyru:
@daipeihust Thanks for investigation!

> We just put command and arguments together and call them command. What do you think about the following config?

Yes, this PR currently supports only macOS though, it should support other OSes too.

---

FYI, from @t9md's comment, I found https://github.com/heyeshuang/ime_helper that changes current input locale by sending message `WM_INPUTLANGCHANGEREQUEST`.
Users may use this if python is installed.

However, I can't determine sending `WM_INPUTLANGCHANGEREQUEST` message is the right way to *disable IM* (it changes input locale, and indirectly keyboard layout?) due to lack of my experience of Windows API application and IM-aware application...
Changing (input) locale is a bit overwhelmed me...
I don't know this is the common way to do though.

daipeihust:
@tyru I tested the ime_helper from [https://github.com/heyeshuang/ime_helper](https://github.com/heyeshuang/ime_helper) on windows10 but it's failed. I'm looking for a tool that can work on windows. And I've found some IM switch tools provided by linux: [im-switch](https://launchpad.net/ubuntu/trusty/+package/im-switch), [im-config](https://launchpad.net/ubuntu/+source/im-config), [fcitx-remote](https://www.unix.com/man-page/all/1/fcitx-remote/)


And I made some change to this PR, the config is like this:
for macOS:
```json
"vim.autoSwitchIM": {
        "enable": true,
        "defaultIM": "com.apple.keylayout.US",
        "switchIMCmd": "/usr/local/bin/im-select",
        "obtainIMCmd": "/usr/local/bin/im-select"
}
```

for windows:
```json
"vim.autoSwitchIM": {
        "enable": true,
        "defaultIM": "com.xxx.xxx",
        "switchIMCmd": "/path/to/xkb-switch -p",
        "obtainIMCmd": "/path/to/xkb-switch -s"
}
```
What do you think?

tyru:
hmm, I think `switchIMCmd` should include placeholder for argument?
Because program's argument format may be different with those tested programs.

```
"vim.autoSwitchIM": {
        "enable": true,
        "defaultIM": "com.xxx.xxx",
        "obtainIMCmd": "/path/to/xkb-switch -p",
        "switchIMCmd": "/path/to/xkb-switch -s {im}"
}
```

If that is fixed, it appears OK to merge this PR for me :)

daipeihust:
@tyru Your idea is perfect, I couldn't agree more!

daipeihust:
@xconverge @jpoon @tyru 
Hey! I've implemented the windows version im-select, tested it on win7 and win10. I put two version im-select in one repository: [im-select](https://github.com/daipeihust/im-select), and there is a installation guide in this repository's readme.So this function can work on macOS and windows. I'm trying to implement the linux version im-select or find a proper program on linux.

I have added the config guide for autoSwitchIM to README.md, but I don't know if it's proper. I wish you can point out mistakes.

src/actions/plugins/imswitcher.ts
```diff
@@ -0,0 +1,135 @@
...
+    const defaultIMKey = configuration.autoSwitchIM.defaultIM;
+    if (this.savedIMKey !== defaultIMKey) {
+      this.switchToIM(this.savedIMKey);
+    }
+  }
+
+  private async switchToIM(imKey: string) {
+    let switchIMCmd = configuration.autoSwitchIM.switchIMCmd;
+    if (!switchIMCmd.includes('{im}')) {
+      this.showWrongSwitchIMCmdErrorMessage();
```
jpoon: In these error states, can we toggle the configuration to disable this plugin? Doesn't make any sense to retry again on the next command if it'll never work. 

Same comment applies to the other error cases below

src/actions/plugins/imswitcher.ts
```diff
@@ -0,0 +1,135 @@
...
+  private getRawCmd(cmd: string): string {
+    const cmds = cmd.split(' ');
+    let rawCmd = '';
+    if (cmds.length > 0) {
+      rawCmd = cmds[0];
+    }
+    return rawCmd;
+  }
+
+  private showCmdNotFoundErrorMessage(cmd: string) {
```
jpoon: Can you use https://github.com/VSCodeVim/Vim/blob/36bf0a95f80654f47e0ec16794dd245a831fcc85/src/util.ts#L12?

src/actions/plugins/imswitcher.ts
```diff
@@ -0,0 +1,135 @@
...
+    vscode.window.showErrorMessage(
+      'Unable to find ' +
+        cmd +
+        '. check your "vim.autoSwitchIMCommand" in VSCode setting. \
+      Or you can disable "vim.autoSwitchIM" to dismiss this error message'
+    );
+  }
+
+  private showWrongSwitchIMCmdErrorMessage() {
+    vscode.window.showErrorMessage(
```
jpoon: Can you use https://github.com/VSCodeVim/Vim/blob/36bf0a95f80654f47e0ec16794dd245a831fcc85/src/util.ts#L12?

src/actions/plugins/imswitcher.ts
```diff
@@ -0,0 +1,135 @@
...
+      'switchIMCmd config in vim.autoSwitchIM is incorrect, \
+      it should contain the placeholder {im}'
+    );
+  }
+
+  private execShell(cmd: string): Promise<string> {
```
jpoon: Can you move this to `util.ts`?

src/state/vimState.ts
```diff
@@ -42,6 +43,8 @@ export class VimState implements vscode.Disposable {
 
   public editor: vscode.TextEditor;
 
+  public inputMethodSwitcher: InputMethodSwitcher;
```
jpoon: Does this need to be public?

README.md
```diff
@@ -513,6 +514,98 @@ Motion Command | Description
 `<operator>z<char><char>`|Perform `<operator>` forward to the first occurence of `<char><char>`
 `<operator>Z<char><char>`|Perform `<operator>` backward to the first occurence of `<char><char>`
 
+## Switch Input Method Automatically
+(Support macOS and windows for now, linux will come soon)
```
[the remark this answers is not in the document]
daipeihust: I'm not sure the exact reason. I found someone's explaination [here](https://stackoverflow.com/questions/29217501/getkeyboardlayout-doesnt-work-properly-in-some-cases) 
> GetKeyboardLayout works with applications running in the WINDOWS subsystem (it's a user32.dll export). Command line applications run in the CONSOLE subsystem.

src/state/vimState.ts
daipeihust: I will get all thoese problems fixed. Thanks for your review!

jpoon: Let me know if you have any questions on my comments or it doesn't make any sense . Thanks again for your contribution!

README.md
```diff
@@ -36,6 +36,7 @@ Please report missing features/bugs on [GitHub](https://github.com/VSCodeVim/Vim
     * [vim-commentary](#vim-commentary)
     * [vim-indent-object](#vim-indent-object)
     * [vim-sneak](#vim-sneak)
+* [Switch Input Method Automatically](#switch-input-method-automatically)
```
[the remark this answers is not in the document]
daipeihust: The title of each part under 'plugins' is vim-xxx, should we pick a name for this function?

src/actions/plugins/imswitcher.ts
daipeihust: In this `showCmdNotFoundErrorMessage` function, I called `showError` of util.ts. This function is called in two places, I think this encapsulation is necessary.

README.md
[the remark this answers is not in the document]
jpoon: Nah, I think it's ok without the `vim-xxx`. Can you rename the section to `Input Method`?

daipeihust:
@jpoon I think It's ready to review again. But I have no idea how to add tests. This depends on shell command, we can't test it without the shell command. Or maybe we can use code to simulate the shell command.

package.json
```diff
@@ -583,6 +583,16 @@
           "type": "boolean",
           "description": "When typing a command show the initial colon ':' character",
           "default": false
+        },
```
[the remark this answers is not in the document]
daipeihust: For `vim.autoSwitchInputMethod.enable`, the default value is false, but `vim.autoSwitchInputMethod.defaultIM` and the others didn't have a same default value on different platform, so which value should I use for these configurations' default property? empty?

package.json
```diff
@@ -574,6 +574,26 @@
...
+        "vim.autoSwitchInputMethod.enable": {
+          "type": "boolean",
+          "description": "Switch input method automatically when mode changed",
+          "default": false
+        },
+        "vim.autoSwitchInputMethod.defaultIM": {
+          "type": "string",
+          "description": "The input method for your normal mode",
+          "default": "com.apple.keylayout.US"
```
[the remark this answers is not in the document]
daipeihust: I agree with you

daipeihust:
@jpoon I have trouble adding tests. I need to hook `execShell ` method when testing, and use the hooked `execShell` method to validate the test results, but I can't do it in typescript, I got a error that `execShell` is a readonly property. I can't replace `execShell` out of the file which define this method.

I have google it many times, but I got nothing. I wonder if you know some solutions

jpoon:
What I'm suggesting is `strategy pattern`. Something like...

```
export class InputMethodSwitcher {
    constructor(exec: ( cmd: string ) => Promise<string> = None ) {
        if (!exec) {
            exec = util.execShell
        }
    }
}
```

So that in your tests, you can pass in your own dummy `exec` function.

src/actions/plugins/imswitcher.ts
```diff
@@ -0,0 +1,122 @@
+import { exec } from 'child_process';
```
jpoon: unused import

src/actions/plugins/imswitcher.ts
```diff
@@ -0,0 +1,122 @@
...
+import { ModeName, Mode } from '../../mode/mode';
+import { existsSync, exists } from 'fs';
```
jpoon: unused import of `exists`

src/actions/plugins/imswitcher.ts
```diff
@@ -0,0 +1,122 @@
...
+import { configuration } from '../../configuration/configuration';
+import * as vscode from 'vscode';
```
jpoon: unused import.

src/actions/plugins/imswitcher.ts
```diff
@@ -0,0 +1,122 @@
...
+      ) {
+        this.resumeIM();
+      }
+    }
+  }
+
+  // save origin input method and set input method to default
+  private async switchToDefaultIM() {
+    const obtainIMCmd = configuration.autoSwitchInputMethod.obtainIMCmd;
+    const rawObtainIMCmd = this.getRawCmd(obtainIMCmd);
```
jpoon: Can you move this to line 59 when it is actually used?

src/actions/plugins/imswitcher.ts
```diff
@@ -0,0 +1,122 @@
...
+    if (obtainIMCmd !== '') {
+      if (existsSync(rawObtainIMCmd)) {
+        const insertIMKey = await util.executeShell(obtainIMCmd);
```
jpoon: What happens when promise is rejected? At minimum we should log

src/actions/plugins/imswitcher.ts
```diff
@@ -0,0 +1,122 @@
...
+      );
+      this.disableIMSwitch();
+      return;
+    }
+    const rawSwitchIMCmd = this.getRawCmd(switchIMCmd);
+    if (switchIMCmd !== '') {
+      if (existsSync(rawSwitchIMCmd)) {
+        if (imKey !== '' && imKey !== undefined) {
+          switchIMCmd = switchIMCmd.replace('{im}', imKey);
+          await util.executeShell(switchIMCmd);
```
jpoon: At minimum, we should log on a rejected promise.

src/actions/plugins/imswitcher.ts
```diff
@@ -0,0 +1,122 @@
...
+  // resume origin inputmethod
+  private async resumeIM() {
+    if (this.savedIMKey !== configuration.autoSwitchInputMethod.defaultIM) {
+      this.switchToIM(this.savedIMKey);
+    }
+  }
+
+  private async switchToIM(imKey: string) {
+    let switchIMCmd = configuration.autoSwitchInputMethod.switchIMCmd;
+    if (!switchIMCmd.includes('{im}')) {
```
jpoon: Validation of the configuration happens in various functions. Should we consolidate them all and put them all in one spot?

src/actions/plugins/imswitcher.ts
```diff
@@ -0,0 +1,122 @@
...
+          switchIMCmd = switchIMCmd.replace('{im}', imKey);
+          await util.executeShell(switchIMCmd);
+        }
+      } else {
+        this.showCmdNotFoundErrorMessage(rawSwitchIMCmd, 'vim.autoSwitchInputMethod.switchIMCmd');
+      }
+    }
+  }
+
+  private getRawCmd(cmd: string): string {
```
jpoon: I think this can be simplified to `cmd.split(' ')[0]` unless there's something I'm missing...

README.md
```diff
@@ -569,6 +570,91 @@ Motion Command | Description
 `<operator>z<char><char>`|Perform `<operator>` forward to the first occurence of `<char><char>`
 `<operator>Z<char><char>`|Perform `<operator>` backward to the first occurence of `<char><char>`
 
+### Input Method
+(Support macOS and windows for now, linux will come soon)
+
+#### Use our recommended program
+
+##### 1.Install im-select
```
jpoon: Instead of using headings, can you just have a numbered list here?

README.md
```diff
@@ -569,6 +570,91 @@ Motion Command | Description
 `<operator>z<char><char>`|Perform `<operator>` forward to the first occurence of `<char><char>`
 `<operator>Z<char><char>`|Perform `<operator>` backward to the first occurence of `<char><char>`
 
+### Input Method
+(Support macOS and windows for now, linux will come soon)
+
```
jpoon: Can you add a small section here saying something along the lines of 'there are two ways you can configure input method, using `im-select` (with a link to your GH), or your own program'

src/state/vimState.ts
```diff
@@ -41,6 +42,8 @@ export class VimState implements vscode.Disposable {
 
   public editor: vscode.TextEditor;
 
+  private inputMethodSwitcher: InputMethodSwitcher;
```
jpoon: Can you rename this to `_inputMethodSwitcher`

src/actions/plugins/imswitcher.ts
```diff
@@ -0,0 +1,137 @@
...
+    const obtainIMCmd = configuration.autoSwitchInputMethod.obtainIMCmd;
+    const rawObtainIMCmd = this.getRawCmd(obtainIMCmd);
+    if (existsSync(rawObtainIMCmd)) {
+      try {
+        const insertIMKey = await util.executeShell(obtainIMCmd);
+        if (insertIMKey !== undefined) {
+          this.savedIMKey = insertIMKey.trim();
+        }
+      } catch (error) {
+        console.log(error);
```
jpoon: Can you use the logger? https://github.com/VSCodeVim/Vim/blob/master/src/mode/modeHandler.ts#L335

daipeihust: OK

daipeihust:
Hi @jpoon, I've added tests

package.json
```diff
@@ -583,6 +583,26 @@
...
           "description": "Customize cursor style per mode"
+        },
+        "vim.autoSwitchInputMethod.enable": {
+          "type": "boolean",
+          "description": "Switch input method automatically when mode changed",
+          "default": false
+        },
+        "vim.autoSwitchInputMethod.defaultIM": {
+          "type": "string",
+          "description": "The input method for your normal mode, find more information here:https://github.com/daipeihust/Vim#2find-your-default-input-method-key",
```
jpoon: Can you fix the link here and also have it point to VSCodeVim repo?

src/actions/plugins/imswitcher.ts
```diff
@@ -0,0 +1,144 @@
...
+      this.showCmdNotFoundErrorMessage(rawSwitchIMCmd, 'vim.autoSwitchInputMethod.switchIMCmd');
+    }
+  }
+
+  private getRawCmd(cmd: string): string {
+    return cmd.split(' ')[0];
+  }
+
+  private showCmdNotFoundErrorMessage(cmd: string, config: string) {
+    Message.ShowError('Unable to find ' + cmd + '. check your ' + config + ' in VSCode setting.');
```
jpoon: Can you use template strings? https://basarat.gitbooks.io/typescript/docs/template-strings.html. I think it'll look a lot neater.

src/actions/plugins/imswitcher.ts
```diff
@@ -0,0 +1,144 @@
...
+    const originConfig = configuration.autoSwitchInputMethod;
+    configuration.autoSwitchInputMethod = {
+      enable: false,
+      defaultIM: originConfig.defaultIM,
+      switchIMCmd: originConfig.switchIMCmd,
+      obtainIMCmd: originConfig.obtainIMCmd,
+    };
+  }
+
+  private validateConfiguration(): boolean {
```
jpoon: Nitpick: as we are returning a boolean, can you rename this method to `isValidateConfiguration()`

test/plugins/imswitcher.test.ts
```diff
@@ -0,0 +1,70 @@
...
+import { setupWorkspace, cleanUpWorkspace } from '../testUtils';
+import { InputMethodSwitcher } from '../../src/actions/plugins/imswitcher';
+import { ModeName } from '../../src/mode/mode';
+
+
+suite('Input method plugin', () => {
+
+  let savedCmd = '';
+
+  function fakeExecute1(cmd: string): Promise<string> {
```
jpoon: can you rename this to `fakeExecuteChinese`?

test/plugins/imswitcher.test.ts
```diff
@@ -0,0 +1,70 @@
...
+      if (cmd === 'im-select') {
+        resolve('chinese');
+      } else {
+        savedCmd = cmd;
+        resolve('');
+      }
+    });
+  }
+
+  function fakeExecute2(cmd: string): Promise<string> {
```
jpoon: can you rename this `fakeExecuteDefault`?

src/actions/plugins/imswitcher.ts
```diff
@@ -0,0 +1,144 @@
...
+      this.disableIMSwitch();
+      return;
+    }
+    // when you exit from insert-like mode, save origin input method and set it to default
+    if (
+      prevMode === ModeName.Insert ||
+      prevMode === ModeName.SurroundInputMode ||
+      prevMode === ModeName.Replace
+    ) {
+      if (
```
jpoon: I think this can be simplified to `newMode !== prevMode`

src/actions/plugins/imswitcher.ts
```diff
@@ -0,0 +1,144 @@
...
+        return;
+      }
+    }
+    // when you enter insert-like mode, resume origin input method
+    if (
+      newMode === ModeName.Insert ||
+      newMode === ModeName.SurroundInputMode ||
+      newMode === ModeName.Replace
+    ) {
+      if (
```
jpoon: similarly, I think this can be simplified to `newMode !== prevMode`

src/actions/plugins/imswitcher.ts
```diff
@@ -0,0 +1,144 @@
...
+    if (configuration.autoSwitchInputMethod.enable !== true) {
+      return;
+    }
+    if (!this.validateConfiguration()) {
+      this.disableIMSwitch();
+      return;
+    }
+    // when you exit from insert-like mode, save origin input method and set it to default
+    if (
+      prevMode === ModeName.Insert ||
```
jpoon: Can you put this put `ModeName.Insert, ModeName.SurroundInsertMode, ModeName.Replace` into a hashset? Then this check can be simplified to...

```
if (newMode !== prevMode) {
   if (insertLikeMode.has(newMode)) {
   // do whatever
   } else if (insertLikeMode.has(oldMode)) {
   // do whatever
   }
}
```

Is that right? Hopefully my logic is correct there.

package.json
daipeihust: Oh! I'm sorry. I forgot to change it to VSCodeVim repo.

src/actions/plugins/imswitcher.ts
daipeihust: Ok

src/actions/plugins/imswitcher.ts
daipeihust: What if prevMode is `Insert` and newMode is `SurroundInputMode`?
In your logic:
```
if (newMode !== prevMode) {
   if (insertLikeMode.has(newMode)) {
      //It will hit this spot
   } else if (insertLikeMode.has(oldMode)) {

   }
}
```
In fact, It should do noting in this situation!

daipeihust: I use a new concept: ModeCluster to simplify this check, How do you like it?

src/actions/plugins/imswitcher.ts
```diff
@@ -0,0 +1,142 @@
...
+        } catch (e) {
+          logger.error(`IMSwitcher: promise is rejected. err=${e}`);
+        }
+      }
+    } else {
+      this.showCmdNotFoundErrorMessage(rawSwitchIMCmd, 'vim.autoSwitchInputMethod.switchIMCmd');
+    }
+  }
+
+  private getModeCluster(mode: ModeName): ModeCluster {
```
jpoon: This is me nitpicking, but there will only ever be two values in this enum correct? If that is the case, I think you can get rid of the enum and instead return a boolean:

```
private isInsertLikeMode(mode: ModeName): boolean
```

That should simplify the number of lines of code:

```
const isPrevModeInsertLike = this.isInsertLikeMode(prevMode);
const isNewModeInsertLike = this.isInsertLikeMode(newMode);
if (isPrevModeInsertLike !== isNewModeInsertLike ) {
    if (isNewModeInsertLike ) {
       this.resumeIM();
     } else {
        this.switchToDefaultIM();
     }
```

Is that right? Hopefully this suggestion makes sense.

src/actions/plugins/imswitcher.ts
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  private showCmdNotFoundErrorMessage(cmd: string, config: string) {
+    Message.ShowError(`Unable to find ${cmd}. check your ${config} in VSCode setting.`);
+    this.disableIMSwitch();
+  }
+
+  private disableIMSwitch() {
+    const originConfig = configuration.autoSwitchInputMethod;
+    configuration.autoSwitchInputMethod = {
```
jpoon: Since these are separate configurations now, I think we can just go...

configuration.autoSwitchInputMethod.enable = false

jpoon: Can you add a test case for testing the scenarios with bad configs?

src/actions/plugins/imswitcher.ts
```diff
@@ -0,0 +1,138 @@
...
+    }
+  }
+
+  private isInsertLikeMode(mode: ModeName): boolean {
+    const insertLikeModes = new Set([
+      ModeName.Insert,
+      ModeName.Replace,
+      ModeName.SurroundInputMode,
+    ]);
+    if (insertLikeModes.has(mode)) {
```
jpoon: this can be simplified to...

return insertLikeModes.has(mode)

src/actions/plugins/imswitcher.ts
jpoon: Can you step through the code and see what operation is failing? We already have this pattern for a couple of other configurations like https://github.com/VSCodeVim/Vim/blob/master/src/configuration/configuration.ts#L247 so not sure why it wouldn't work.

daipeihust: ok, I'll try to figure it out tonight.

daipeihust: I found out the reason, the `vscode.workspace.getConfiguration()` function return a Proxy object instead of a object. The Proxy object's set method is replaced. So I got the `Cannot assign to read only property 'enable' of object.` error.

I've add commits to unwrap the Proxy object.

src/configuration/configuration.ts
```diff
@@ -170,6 +174,17 @@ class Configuration implements IConfiguration {
     );
   }
 
+  unproxify(obj: Object): Object {
```
jpoon: I don't quite understand why this is necessary.

jpoon: I removed that piece of code and made another change: https://github.com/VSCodeVim/Vim/commit/10409227b091c5f72c7326e6894127f014c37d70. All tests passed?

jpoon: With the changes here https://github.com/VSCodeVim/Vim/commit/10409227b091c5f72c7326e6894127f014c37d70, I was able to get your tests to pass.

jpoon: I don't quite follow what's going on. 

> The vscode.workspace.getConfiguration() function return a Proxy object instead of a object. The Proxy object's set method is replaced.

We never should be calling the Proxy object's set method.

```
    let vimConfigs: any = Globals.isTesting
      ? Globals.mockConfiguration
      : this.getConfiguration('vim');

    /* tslint:disable:forin */
    // Disable forin rule here as we make accessors enumerable.`
    for (const option in this) {
      const val = vimConfigs[option] as any;
      if (val) {
        this[option] = val;
      }
    }
```

`this[option] = val` is calling the proxy object's getter here. 

Additionally, we already have a pattern for setting configurations in this nested manner (see `vim.debug.logginglevel`) and I just submitted a PR to move some other configurations to the same pattern (https://github.com/VSCodeVim/Vim/pull/2915). Haven't come across the same error you are seeing.

daipeihust: I have search the whole project, the `vim.debug.logginglevel` only appear in package.json and README.md. I haven't found the logic like `vim.debug.logginglevel = xxx1` in code.
I added `vim.debug.logginglevel = 'warn'` in project and I got the same error:
```
TypeError: Cannot assign to read only property 'loggingLevel' of object。
Activating extension `vscodevim.vim` failed:  TypeError: Cannot assign to read only property 'loggingLevel' of object
```
You can try it yourself to see if you will get the same error!

jpoon: Ok, I see the issue now. Thanks for explaining.

.github/CONTRIBUTING.md
```diff
@@ -23,33 +23,33 @@ When submitting a PR, please fill out the template that is presented by GitHub w
...
-   # debug tab to build and run the extension.
-   # Or run tests by selecting the appropriate drop down option
-
-   # Alternatively, build and run tests through gulp and npm scripts
-   gulp build                  # build
-   npm test                    # test
-   gulp test                   # run tests inside Docker container
-   gulp test --grep testSuite  # run only tests/suites filtered by js regex inside container
-   ```
+1.  Install prerequisites:
```
jpoon: Do we need all these MD changes?

daipeihust: These changes is made by `gulp forceprettier` command.

DanSnow:
For somebody who is using fcitx under Linux

```json
{
  "vim.autoSwitchInputMethod.enable": true,
  "vim.autoSwitchInputMethod.defaultIM": "1",
  "vim.autoSwitchInputMethod.obtainIMCmd": "/usr/bin/fcitx-remote",
  "vim.autoSwitchInputMethod.switchIMCmd": "/usr/bin/fcitx-remote -t {im}"
}
```

transtone:
How can I set InputMethod to the defaultIm forever, never change back to last IM when insert?

cnshsliu:
How to NOT switch back to used IM method when enter back to Insert Mode?
say, I am using Chinese input in Insert Mode, then I press ESC, autoSwitch change IM automatically to ABC as expected, here I am able to navigate with hjkl etc, then, I press 'i' to enter Insert Mode, autoSwitch change IM to Chinese again.  These steps are all working as the autoSwitch design. 

However, what if I want autoSwitch does not switch back to Chinese IM when I enter Insert Mode again, but keep the ABC IM?

laishzh:
@cnshsliu I face the same problem too. My solution works fine, but seems not perfect.
```
{
    "vim.autoSwitchInputMethod.enable": true,
    "vim.autoSwitchInputMethod.obtainIMCmd": "/usr/local/bin/im-select",
    "vim.autoSwitchInputMethod.switchIMCmd": "/usr/local/bin/im-select com.apple.keylayout.ABC {im}",
    "vim.autoSwitchInputMethod.defaultIM": "com.apple.keylayout.ABC"
}
```

charlielin:
I got some issues while using these extension.
I use Chinese input method(Sogou pinyin) in insert mode to type Chinese. Then press ESC, autoSwitch to US to navigate with hjkl. It works very well most of time, but sometimes, when I enter the insert mode, even though I am using the Chinese input method(`im-select` returns `com.sogou.inputmethod.sogou.pinyin`), it still types abc, not Chinese expected.
here is my vscode settings:
```
    "vim.autoSwitchInputMethod.obtainIMCmd": "/usr/local/bin/im-select",
    "vim.autoSwitchInputMethod.enable": true,
    "vim.autoSwitchInputMethod.switchIMCmd": "/usr/local/bin/im-select {im}",
    "vim.autoSwitchInputMethod.defaultIM": "com.apple.keylayout.US"
```

Issue #2325, opened 2018-01-23
Ctrl+D interferes with Cmd+D multiple selection


timkendrick:
**Is this a BUG REPORT or FEATURE REQUEST?** (choose one): BUG REPORT

**What happened**: Since [#2314](https://github.com/VSCodeVim/Vim/pull/2314), pressing Cmd+D loses multiple selections unless `vim.handleKeys` is set to `{ "<C-d>": false }`

**What did you expect to happen**: Cmd+D should not be affected by Ctrl-D behavior

**How to reproduce it (as minimally and precisely as possible)**:

- In normal mode, position the cursor within a word that is repeated within the current buffer
- Press `viw` to select the current word in visual mode
- Press Cmd+D; another cursor is created at the next occurrence but the selection is lost and we leave visual mode (preventing the ability to use `c` to replace both selected occurrences).

N.B. Everything works correctly if VSCode is reloaded with `vim.handleKeys` set to `{ "<C-d>": false }`

**Environment**: macOS

- Extension (VsCodeVim) version: 0.10.3
- VSCode version: 1.19.2
- OS version: 10.12.6

timkendrick:
**UPDATE:** It looks like I was mistaken about this being related to Ctrl-D, as Cmd+D seems to work intermittently (I can't figure out what triggers the bug). I've updated the issue title/description to reflect this.

rhys-vdw:
I am experiencing this problem today as well - but not intermittently, it happens every time the second selection is made.

![bad-vim](https://user-images.githubusercontent.com/822105/35306348-3f3991ce-00f1-11e8-9ee9-a3299127ac48.gif)
_me pressing cmd+d repeatedly on `@extend`☝️_

jpoon:
We changed the default behaviour of `<C-d>` with #2269.

|                             | useCtrlKeys = true                     | useCtrlKeys  = false                |
|--------------------- | ------------------------------|-----------------------------|
| handleKeys = true | VSCodeVim move page half down | VSCodeVim move page half down
| handleKeys = false | remap to ```<D-d>``` | remap to ```<D-d>```
| undefined               | **VS Code default behavior** | VS Code default behavior

Try setting the `handleKeys` setting to the appropriate configuration you want.

timkendrick:
@jpoon as mentioned in my comment above, that doesn’t seem to fix the problem unfortunately - also it seems weird that the setting for Ctrl-D should have an effect on the Cmd-D behavior?

[...]

PR #2684, opened 2018-05-25
fixs  #2325 Cmd+D loses multiple selections

redhat008:

Fix the bug. Cmd+D loses multiple selections #2325

fixs  #2325 Cmd+D loses multiple selections

Special notes for your reviewer:
1. only test in my computer
2. I am a  programmer,  but not familiar with typescript/js and vscode extension development.  Just compare with 0.10 with 0.12 code . the  logic have changed ， so produce issue #2335

### DIFF
--- a/src/mode/modeHandler.ts
+++ b/src/mode/modeHandler.ts
@@ -236,11 +236,9 @@ export class ModeHandler implements vscode.Disposable {
           this.vimState.lastClickWasPastEol = false;
         }
 
-        if (
-          configuration.mouseSelectionGoesIntoVisualMode &&
-          !this.currentMode.isVisualMode &&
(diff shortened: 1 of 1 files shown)


xconverge:
I had recently added https://github.com/VSCodeVim/Vim/blob/master/test/multicursor.test.ts

Is there any chance you can add a test without your fix, and then make sure it passes with your fix?

These are tricky issues and I would love to catch them in tests if possible! Thanks

redhat008:
Sorry.  I'm not familiar with typescript/js and vscode extension development. I don't know how to add the test

xconverge:
The file I linked has 2 other tests, you can just mimic one of them and add it to that file even multicursor.test.ts

PR #2694, opened 2018-05-30
Fix vim-commentary description in README

Ran4:

Fixes bad description in README.


xconverge:
Thanks, I had to double check in our implementation vs vim-commentary, functionally all looks good though!

Issue #2696, opened 2018-05-30
Add support for new grid layout with splits


rodcloutier:
** Feature request **

Split command should support the upcoming grid layout feature from VsCode.

VsCode issue and PR:
* https://github.com/Microsoft/vscode/issues/14909
* https://github.com/Microsoft/vscode/pull/49599

jpoon:
Closed with https://github.com/VSCodeVim/Vim/pull/2697

elviskahoro:
@rodcloutier @jpoon I typically use <C-w>+s or <C-w>+v to create split panes. It stopped working for me so wanted to double check that this wasn't happening for others.

:vs still works!

jwtly10:
> @rodcloutier @jpoon I typically use +s or +v to create split panes. It stopped working for me so wanted to double check that this wasn't happening for others.
> 
> :vs still works!

This has also stopped working for me!

PR #2697, opened 2018-05-30
Add support for grid layout

rodcloutier:
Enables support for the new grid layout in VSCode 

VsCode issues and PR:
* https://github.com/Microsoft/vscode/issues/14909
* https://github.com/Microsoft/vscode/pull/49599

#2696 

**This is a work in progress**. It requires the latest insiders release.

Take note that I have sync the commands with the behaviour found in Vim.
* `ne` is not a split command thus removed
* Added `enew` command
* Split commands with file arguments or empty file remove the duplicate tab (this is up for dicussion and could be parametrized)
* I couldn't find exactly how to access the layout information to test that the command do indeed perform the proper split

**TODO**:
- [ ] Document that adding a space after the split command prompts a Open file dialog
- [ ] Parametrized the automatic tab close ?
- [ ] Find how to test that the command actually do what they claim


xconverge:
I have a feeling this is about to be requested heavily now that it is released in the latest stable vscode 1.25.0

Any major hurdles here?

rodcloutier:
No major hurdles on my side other than the test that are not very complete. I feel this would be a good first pass and we might want to add new features going forward (configurations, option for direction of splits, etc). 

The standard way VisualCode works with Tab history might be a little awkward and suprising for Vim users as it will close the current Tab in the current split but open the last open Tab which might be in another split. I don't this this is something that we should fix before releasing proper support.

rodcloutier:
As I am going on vacation next week, I will be available to work on moving this PR forward tomorrow (July 6) only. Otherwise we will need to have someone take over. Maybe @xconverge ?

dastergon:
Any updates on this PR?

rodcloutier:
Fixed some bugs introduced by a rebase. Should be working fine now.

src/cmd_line/commands/file.ts
```diff
@@ -8,16 +8,16 @@ import { Message } from './../../util/message';
...
+  NewWindowHorizontalSplit,
 }
 
 export interface IFileCommandArguments extends node.ICommandArgs {
   name: string | undefined;
   bang?: boolean;
   position?: FilePosition;
   lineNumber?: number;
-  createFileIfNotExists?: boolean;
+  new?: boolean; // Create a new untitled file
```
jpoon: I prefer the old variable name as it was descriptive and wouldn't require a preceding comment.

src/cmd_line/commands/file.ts
```diff
@@ -32,109 +32,94 @@ export class FileCommand extends node.CommandBase {
...
+      await vscode.commands.executeCommand('workbench.action.files.revert');
+      return;
     }
 
-    return active.viewColumn!;
-  }
-
-  getViewColumnToRight(): vscode.ViewColumn {
-    const active = vscode.window.activeTextEditor;
+    // Need to do this before the split since it looses the activeTextEditor
```
jpoon: | looses

typo.

src/cmd_line/commands/file.ts
```diff
@@ -32,109 +32,94 @@ export class FileCommand extends node.CommandBase {
...
-
-  async execute(): Promise<void> {
-    if (this._arguments.bang) {
-      await vscode.commands.executeCommand('workbench.action.files.revert');
-      return;
-    }
+    let hidePreviousEditor = async function() {
+      if (split === true) {
+        await vscode.commands.executeCommand('workbench.action.previousEditor');
+        await vscode.commands.executeCommand('workbench.action.closeActiveEditor');
```
jpoon: Why do we close the previous editor?

src/cmd_line/commands/file.ts
```diff
@@ -32,109 +32,94 @@ export class FileCommand extends node.CommandBase {
...
+      }
+    };
 
+    // No name was specified
     if (this._arguments.name === undefined) {
-      // Open an empty file
-      if (this._arguments.position === FilePosition.CurrentWindow) {
+      // Should we create an empty file?
```
jpoon: Not sure why you left a question as a comment?

src/cmd_line/commands/file.ts
```diff
@@ -32,109 +32,94 @@ export class FileCommand extends node.CommandBase {
...
-    );
-
-    if (this.arguments.lineNumber) {
-      vscode.window.activeTextEditor!.revealRange(
-        new vscode.Range(
-          new vscode.Position(this.arguments.lineNumber, 0),
-          new vscode.Position(this.arguments.lineNumber, 0)
-        )
-      );
+    if (filePath !== '') {
```
jpoon: I don't think there's a case where this would be false?

test/cmd_line/vsplit.test.ts
```diff
@@ -33,4 +33,25 @@ suite('Vertical split', () => {
...
+  test('Run :vnew', async () => {
+    await commandLine.Run('vnew', modeHandler.vimState);
+    await WaitForEditorsToClose(2);
+
+    assertEqual(vscode.window.visibleTextEditors.length, 2, 'Editor did not split in 1 sec');
+  });
+
+  test('Run :vne', async () => {
+    await commandLine.Run('vne', modeHandler.vimState);
+    await WaitForEditorsToClose(2);
```
jpoon: I'm looking at the implementation for `WaitForEditorsToClose`. Who is making it time out after 1s?

jpoon:
Sorry for the delay in the review. I left a couple of comments. Could you also add more complex test cases? ie. an existing window then `:vsplit`?

src/cmd_line/commands/file.ts
rodcloutier: This is to reflect the behaviour of Vim. 

In the case where we split specifying a file name (`vsplit bar.c`, `split bar.c`) or create an empty new file (`:new`, `:vnew`), we start by splitting the current editor which results in two splits with the same editor/file. We then create the empty file or load the specified file. 

So assuming the following command ran in the `code.go` editor

```
:vsplit bar.c
```

If we do not close the previous editor we end up with this

screen shot 2018-08-27 at 15 48 30

But we would rather have

screen shot 2018-08-27 at 15 50 02

src/cmd_line/commands/file.ts
rodcloutier: You are right. I have removed the check.

src/cmd_line/commands/file.ts
rodcloutier: It only reflected the underlying if condition. I have removed it.

src/cmd_line/commands/file.ts
rodcloutier: Faire enough.

src/cmd_line/commands/file.ts
rodcloutier: Fixed

src/cmd_line/commands/file.ts
jpoon: Ah, I see what you mean. Thanks for the explanation.

test/cmd_line/vsplit.test.ts
maxnordlund: Yes, it can be fixed with something like:

```typescript
export async function WaitForEditorsToClose(numExpectedEditors: number = 0, timeout: number = NaN): Promise<void> {
  let waitForTextEditorsToClose = new Promise((c, e) => {
    if (vscode.window.visibleTextEditors.length === numExpectedEditors) {
      return c();
    }

    vscode.window.onDidChangeVisibleTextEditors(() => {
      if (vscode.window.visibleTextEditors.length === numExpectedEditors) {
        c();
      }
    });

    if (!isNan(timeout)) {
      setTimeout(e, timeout);
    }
});
```

The `NaN` check should make it backwards compatible.

karthiksn:
Hi! Any more news on this?

jpoon:
Oops, I hadn't noticed you had address all the CR comments @rodcloutier. Is this ready to merge?

uloco:
Will this also add <kbd>ctrl+w s</kbd> functionality? Would be aweseome :)

nealot:
https://github.com/Microsoft/vscode/issues/47051 prevents navigation between viewlets.

xconverge:
Are we waiting on anything here? looks like everything was resolved @jpoon ?

Issue #2683, opened 2018-05-25
gf creates files when the given file does not exist


SuyogSoti:
**Is this a BUG REPORT or FEATURE REQUEST?** (choose one):
But Report

**What happened**:
I press gf in normal mode on any keywork, and the extension creates a file with the name as that keyword in the directory.

**What did you expect to happen**:
for the extension to not do anything. Please do not make it give me an error saying the file does not exist.

**How to reproduce it (as minimally and precisely as possible)**:
Go to any word in your code base and press gf for a file that does not already exist.

**Environment**:
Manjaro

- Extension (VsCodeVim) version: 0.12.0
- VSCode version: 1.23.1
- OS version: Manjaro

jpoon:
I don't quite understand the behaviour you expect. We do a best effort in trying to find the file (based off your relative path and/or workspace directory). However, if the file does not exist, you want the extension to do nothing? Don't create the file, and don't show an error? IMO that behaviour seems rather unfriendly to a user as there's no feedback to the user after command execution.

In any case, our original behaviour was to show an error dialog, but there was a relatively popular ask to automatically create the file -- https://github.com/VSCodeVim/Vim/issues/2274. For that reason, I'm inclined to close this bug as by-design.

SuyogSoti:
I do agree that #2274 is a great idea but I think that there should be a difference between how :e and gf operate. If possible I think that gf should only go to the file if it exists and :e should create the file if it does not exist. 

As for the notification, I think that it would be enough to the user to see that they did not go to a new file to see that that file does not exist. If there should be a notification, I think that it should be done in a more discrete way. The notification system in vscode is too distracting. To be perfectly honest, my request for the no notification more secondary to my request that gf should not open a new file.

SuyogSoti:
Hey,

I just wanted to check in with you about what you think about the gf creating new files if it does not exist. I did just a quick look in the code base and it seems like creation of the new file happens in src/cmd_line/commands/file.ts in the class FileCommand on line 105: `fs.closeSync(fs.openSync(filePath, 'w'))`.
A notification there instead of a new file would work great.
I checked and I see that this is the only place that FileCommand is being used. 

What I am trying to get at is I just want to check in about what we think about it not creating new files. If this is an issue of time, this seems like a simple change where I am more than happy to create a pull request. But if this is an issue of the design principle, then I would just loved to be looped in.

jpoon:
[...]

PR #2701, opened 2018-06-01
fix gf to be like issue #2683

SuyogSoti:
fix gf to be like issue #2683

#2683 

### DIFF
--- a/src/actions/commands/actions.ts
+++ b/src/actions/commands/actions.ts
@@ -2346,7 +2346,7 @@ class CommandOpenFile extends BaseCommand {
       const fileCommand = new FileCommand({
         name: filePath,
         lineNumber: lineNumber,
-        dontCreateFile: true,
+        createFileIfNotExists: false,
       });
       fileCommand.execute();
     }
--- a/src/actions/commands/actions.ts
+++ b/src/actions/commands/actions.ts
@@ -2343,7 +2343,11 @@ class CommandOpenFile extends BaseCommand {
     if (fileInfo) {
       const filePath = fileInfo[1];
       const lineNumber = parseInt(fileInfo[2], 10);
-      const fileCommand = new FileCommand({ name: filePath, lineNumber: lineNumber });
+      const fileCommand = new FileCommand({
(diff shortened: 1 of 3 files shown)


src/cmd_line/commands/file.ts
```diff
@@ -16,6 +16,7 @@ export interface IFileCommandArguments extends node.ICommandArgs {
   bang?: boolean;
   position?: FilePosition;
   lineNumber?: number;
+  dontCreateFile?: boolean;
```
jpoon: can you do the negation and rename something more descriptive like `createFileIfNotExists`

src/cmd_line/commands/file.ts
```diff
@@ -102,7 +103,12 @@ export class FileCommand extends node.CommandBase {
             filePath = pathWithExt;
           } else {
             // create file
-            fs.closeSync(fs.openSync(filePath, 'w'));
+            if (this.arguments.dontCreateFile === true) {
+              vscode.window.showErrorMessage('File path does not exits');
```
jpoon: there's a function in `util.ts` for this, can you use that? Also output the filepath.

src/cmd_line/commands/file.ts
jpoon: Thanks so much for this PR! <3

src/actions/commands/actions.ts
```diff
@@ -2346,7 +2346,7 @@ class CommandOpenFile extends BaseCommand {
       const fileCommand = new FileCommand({
         name: filePath,
         lineNumber: lineNumber,
-        dontCreateFile: true,
+        createFileIfNotExists: false,
```
jpoon: After changing the negation, the default value is 'false'. Which means you'll need to update the other spot (`:e`) to set this value to `true`.

xconverge:
@jpoon looks like you can finish your review here now if you get a chance!

SuyogSoti:
yes thanks!

jpoon:
Thanks for your contribution @SuyogSoti !

PR #2720, opened 2018-06-08
fix: revert our workaround cursor toggle as this has been fixed in vscode

jpoon:

* Going through the issue history, this was fixed by vscode
* Tested it for a bit and it does seem to be indeed the case which makes this piece of code redundant


xconverge:
This happened when switching tabs when 2 tabs were in different modes I believe, just for testing purposes

jpoon:
Just tried it out again. Seems to be fine. One tab in INSERT showed `|`, another tab in NORMAL showed the block cursor.

PR #2721, opened 2018-06-11
Fixed #2688. Revert "Removed viewcolumn from editoridentity information"

### LINKED ISSUE #2688: Wrong cursor position after using same file in two panels

It's a bug report

**What happened**:

5. I am taken to line 21

**Environment**:

- VSCode version: 1.23.1
- OS version: 10.13.2 (17C88)

**Update**

lukaszb:
I've pinpointed breaking change using git bisect. Just tried to revert this and see if it helps. Not sure if it breaks anything or relates to some other changes so feel free to treat this PR as a simple hint.

### DIFF
--- a/src/editorIdentity.ts
+++ b/src/editorIdentity.ts
@@ -2,13 +2,19 @@ import * as vscode from 'vscode';
 
 export class EditorIdentity {
   private _fileName: string;
+  private _viewColumn: vscode.ViewColumn;
 
   constructor(textEditor?: vscode.TextEditor) {
     this._fileName = (textEditor && textEditor.document && textEditor.document.fileName) || '';
+    this._viewColumn = (textEditor && textEditor.viewColumn) || vscode.ViewColumn.One;
   }
 
   get fileName() {
     return this._fileName;
+  }
+
+  get viewColumn() {
+    return this._viewColumn;
   }
 
   public hasSameBuffer(identity: EditorIdentity): boolean {
@@ -16,10 +22,10 @@ export class EditorIdentity {
   }
 
   public isEqual(other: EditorIdentity): boolean {
-    return this.fileName === other.fileName;
+    return this.fileName === other.fileName && this.viewColumn === other.viewColumn;
   }
 
   public toString() {
-    return this.fileName;
+    return this.fileName + this.viewColumn;
   }
 }


jpoon:
We [ @Chillee ] removed the `viewColumn` (https://github.com/VSCodeVim/Vim/pull/2559) as that was causing issues with the undo stack. 

Considering #2688, we need a better way to generate uniqueness for `editorIdentity`. I think we should add `viewColumn` back, but we also need to keep it in sync if a user moves the window from one column to another otherwise the states will be all messed up again.

lukaszb:
@jpoon thanks for clarification. However, imho opening same buffer side by side is more common than using undo action. That's only me, though.

jpoon:
> However, imho opening same buffer side by side is more common than using undo action. That's only me, though.

Are you advocating that we fix (1) same buffer side by side by re-introducing (2) https://github.com/VSCodeVim/Vim/issues/2007? Cause, we are not doing that....

lukaszb:
@jpoon well, for me broken undo is much less intrusive into my workflow than broken cursor position among multiple open buffers. I.e. I do lot of vue-js programming with single file components (markup, code and styles sit within same file) and have to work with same file opened at two panes pretty much all the time. I have not enough knowledge to be more helpful here - I don't even fully understand what is the impact of `editorIdentity` change on the undo or cursor position (apparently there are no tests for it so it's hard to tell what it breaks or what it supports).

Currently I simply replace `~/.vscode/extensions/vscodevim.vim-XXX` with repo of my PR's branch, run `npm i && npm run build` as a workaround.

jpoon:
`editorIdentity` is a unique identifier for the editor/file and is used to retrieve vim state. (https://github.com/VSCodeVim/Vim/blob/c13d2f7e2cb19d395a55a7f0bfca5016b4c6b94a/extension.ts#L38). As you change your active editor, we look up it's ID and then retrieve it's vim state.

Now that we've removed `viewColumn` (which IMO is a bug), any file with the same path will generate the same ID and we now see this issue of the same file opened in the editor having the same vim state (e.g. same cursor position across editors). 

As mentioned in an earlier comment,

> Considering #2688, we need a better way to generate uniqueness for editorIdentity. In addition to adding back the viewColumn to generate the editorIdentity, I think we'll also need to keep it in sync with it's actual position when a user moves the window from one column to another otherwise the states (e.g. undo stack) will be messed up again.

I think fixing this properly is the way to go. I am not cool with replacing one bug with another  especially as the undo issue was causing so much grief to so many users.

lukaszb:
@jpoon agree and thanks for explanation. Closing this PR then.

PR #2723, opened 2018-06-11
Fix #1348. ctrl+D/U correct position

### LINKED ISSUE #1348: Ctrl-D and Ctrl-U do not update cursor correctly

rebornix:

We added basic view information to the editor `editor.visibleRanges` in vscode, with which we can do some math to make Ctrl+D/U position correct.

#1348

It requires VSCode 1.22 at least so I updated the vscode dependency. Once we merge this RP and publish, users using VSCode prior to 1.22 will not get the update.

### DIFF
--- a/src/actions/commands/actions.ts
+++ b/src/actions/commands/actions.ts
@@ -653,6 +653,32 @@ class CommandMoveHalfPageDown extends CommandEditorScroll {
   keys = ['<C-d>'];
   to: EditorScrollDirection = 'down';
   by: EditorScrollByUnit = 'halfPage';
+
+  public async exec(position: Position, vimState: VimState): Promise<VimState> {
+    let lineOffset = 0;
+    let editor = vscode.window.activeTextEditor!;
+    let startColumn = vimState.cursorStartPosition.character;
+    let firstLine = editor.visibleRanges[0].start.line;
+    let currentSelectionLine = editor.selection.start.line;
+    lineOffset = currentSelectionLine - firstLine;
+
+    let timesToRepeat = vimState.recordedState.count || 1;
+    await vscode.commands.executeCommand('editorScroll', {
+      to: this.to,
+      by: this.by,
+      value: timesToRepeat,
+      revealCursor: false,
+      select:
+        [ModeName.Visual, ModeName.VisualBlock, ModeName.VisualLine].indexOf(
+          vimState.currentMode
+        ) >= 0,
+    });
+
+    let newFirstLine = editor.visibleRanges[0].start.line;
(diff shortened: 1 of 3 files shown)


rebornix:
This PR also demonstrate how we can get proper view information from VSCode API. We can probably improve commands which are not correct or perfect before cc @VSCodeVim/maintainers

src/actions/commands/actions.ts
```diff
@@ -653,13 +653,66 @@ class CommandMoveHalfPageDown extends CommandEditorScroll {
   keys = ['<C-d>'];
   to: EditorScrollDirection = 'down';
   by: EditorScrollByUnit = 'halfPage';
+
+  public async exec(position: Position, vimState: VimState): Promise<VimState> {
+    let lineOffset = 0;
```
jpoon: I think putting this in the base class `CommandEditorScroll` is a better place for this?

rebornix: I'll take a deeper look into how Ctrl+F and others work and then move it into them. I have a feeling that they don't work properly.

src/actions/commands/actions.ts
```diff
@@ -653,13 +653,66 @@ class CommandMoveHalfPageDown extends CommandEditorScroll {
...
+    let editor = vscode.window.activeTextEditor!;
+    let startColumn = vimState.cursorStartPosition.character;
```
xconverge: can we use const for a lot of these

rebornix: good suggestion ;)

jpoon:
I'm going to merge this and push a release. We can fix the `let` => `const` later.

chiuan:
still not working ?

jpoon:
Hm, you're right, `ctrl+d` works correctly, but the `ctrl+u` behaviour isn't just right yet. I see it hitting the top of the file.

PR #2726, opened 2018-06-13
Add visualModesKeyBindings

### LINKED ISSUE #2705: Add visualModeKeyBindings, in addition to otherModesKeyBindings

**Environment**:

- VSCode version: 0.12.0
- OS version: macOS 10.13.4

I'd like an option like `visualModeKeyBindings`/`visualModeKeyBindingsNonRecursive` in addition to `otherModesKeyBindings`/`otherModesKeyBindingsNonRecursive` so that we can add key bindings which only works in visual mode.

This is essentially `xnoremap` in vim. It's useful to do something like:
https://stackoverflow.com/questions/290465/how-to-paste-over-without-overwriting-register (this trick won't work if you set this keybinding to fire in normal mode).

chibicode:
Add `visualModeKeyBindings`/`visualModeKeyBindingsNonRecursive` in addition to `otherModesKeyBindings`/`otherModesKeyBindingsNonRecursive` so that we can add key bindings which only works in visual mode.

fixes #2705


chibicode:
Ok, tests are now passing. @jpoon

jpoon:
With the introduction of `visualModes` keybinding, the `otherModes` keybinding just means normal mode right? Kind of confusing to have BOTH `OtherModes` and `VisualModes` remappers handle the visual/visualline/visualblock modes.

I think we should rename OtherModes to `NormalMode` and have it just handle normal mode. However, this is a breaking change as people would need to migrate the configurations over....

chibicode:
@jpoon 

> otherModes keybinding just means normal mode right? 

I haven't changed what `otherModes` does on this PR, so it still handles both normal / visual modes. I agree it's confusing but I don't know if it's worth introducing a breaking change because most people seem to be happy w/ the current setup (insert vs other) as I didn't see anyone else wanting visual-only remapping.

icecreammatt:
@jpoon I came across this trying to search how to bind stuff in normal mode. Thanks for confirming for me that OtherModes is the same thing as normal mode. I was confused what that was for.

(trying to remap 0 and $ to shift+h and shift+l I'm hoping that is possible)

jpoon:
@chibicode i'm starting to think we should rename `otherModes` to be `normalMode`, but that'll be a breaking change for everybody's configurations.

Also posed the question to our Slack: https://vscodevim.slack.com/archives/C0ENQMTUM/p1529045308000348

README.md
```diff
@@ -312,13 +312,64 @@ There's several different mechanisms you can use to define custom remappings. Al
...
+    ],
+```
+
+* Bind `>` and `<` in visual mode to indent/outdent lines (repeatable)
+
+
+```json
+    "vim.visualModeKeyBindingsNonRecursive": [
+        {
+            "after": [],
```
jpoon: Can you remove all the empty arguments?

chibicode: @jpoon done https://github.com/VSCodeVim/Vim/pull/2726/commits/65c24f313f41d60d9a685bf4f205d967dd673dd6

adriaanbaelus:
I believe this just got updated? This broke essentially all my (vim) keybindings. It took me some time to figure out the changelog (well, VS Code did say vim.otherModesKeyBindingsNonRecursive is an unknown config setting now).

I don't have any mappings that apply only to normal mode or visual mode, they all apply to both. If I understand correctly, I now have to duplicate all my keybindings (once for normalModeKeyBindings and once for visualModeKeyBindings)?

Is there any way to configure keybindings for multiple modes at once?

jpoon:
No, you'll need to copy the keybindings over for both modes.

FelikZ:
I tried to use reference variables:
```json
{
   "vim.visualModeKeyBindingsNonRecursive": "${config:vim.normalModeKeyBindingsNonRecursive}"
}
```

However no luck. Vim bindings is a huge part of my settings and now I have to double them and maintain twice. Maybe there is a way to put a reference?

mdunicorn:
My key bindings stopped working without any notice. It took me some time to figure out what happened. It would be nice if I would get some warning when the extension automatically updated with a breaking change.

Why not have something for both normal and visual mode? Why not use the name `otherModesKeyBindings` for it, which does not break anything? I know that name is not good. "other" than what? At least it doesn't annoy existing users.

jpoon:
@mdunicorn, the best spot to look for this sort of information is our change log which we keep up-to-date with each release: https://github.com/VSCodeVim/Vim/blob/master/CHANGELOG.md#v0130-2018-06-18. This is viewable through GH and through VSCode's extension marketplace. We also announced the change on our Slack and on Twitter (https://twitter.com/jasonthepoon/status/1008784419354652672). 

We will eventually adopt semvar versioning (https://github.com/VSCodeVim/Vim/issues/2756), once we feel that we've reached a v1.0.0 milestone so that'll be another mechanism in which users can identify if there is a breaking change. 

If you have suggestions on how we can improve this process, we'd love to hear them.

> Why not use the name otherModesKeyBindings for it, which does not break anything? 

We debated this on the PR (see above) and on our slack. Ultimately, we decided to do the breaking change precisely for the point you brought up -- if we kept `otherModesKeyBindings` what would that actually mean as we would then have keybindings for normal, visual, insert so what would "other" mean? 
If you open your user settings, VSCode shows a red squiggly underneath any non-existent configurations. Considering that and the actual fix to migrate to the new syntax was fairly simple, and keeps our code base a bit cleaner, we went ahead with the change.

mdunicorn:
Hello @jpoon. Thanks for your kind response.

I really appreciate your effort on the extension and keeping the users informed. The thing is, not all the users of the extension, follow it on Twitter or subscribe to the Slack channel. And since VS Code automatically updates all the extensions (which is very good), the users will not notice the changes, unless something stops working.

Some extensions automatically show their Release Notes on every update (which is annoying, though!). I don't know how it works, but if the extension can use a hook and be informed by VS Code when it is updated, it would be very good that it warns the users of any breaking change.

Thanks for listening! :)

PR #2729, opened 2018-06-13
Fix autoindent on cc/S #2497

### LINKED ISSUE #2497: Commands cc and S do not respect indent level if executed before the first character
Bug report

**What happened**:
When you press cc or S to change a line, AND the cursor is anywhere before text, the cursor is not moved to the correct indentation space.

The cursor should move to the position right after the indentation.

Type this line:
`<tab>abcd`
Move cursor to position a, b, c or d. Press cc/S. Cursor is moved to the position of (now deleted) 'a' character. This is *correct*.

Using the same line, move cursor to positions 1, 2, 3 or 4 (assuming the tab is 4 spaces) and press cc/S. The cursor stays on that position. This is *incorrect*. Expected behaviour would be to place the cursor at the 5th position, right after the tab/space indent.

**Environment**:

- VSCode version: 1.21.1
- OS version: Windows 10 x64

dqsully:
Changes the `cc` and `S` commands to delete the whole line's contents and reindent it when `autoindent` is enabled. Originally these commands would only delete the text after the indent, which could leave the cursor inside the untouched indent text rather than moving it to the end of the indent.

I think that since commands like `o` already 'reindent' the new line that `cc` and `S` should too, and this implementation also moves the cursor to the end of the line automatically.

fixes #2497

### DIFF
--- a/src/actions/operator.ts
+++ b/src/actions/operator.ts
@@ -566,6 +566,10 @@ export class ChangeOperator extends BaseOperator {
   }
 
   public async runRepeat(vimState: VimState, position: Position, count: number): Promise<VimState> {
+    let thisLineIndent = vimState.editor.document.getText(
+      new vscode.Range(position.getLineBegin(), position.getLineBeginRespectingIndent())
+    );
+
     vimState.currentRegisterMode = RegisterMode.LineWise;
 
     vimState = await this.run(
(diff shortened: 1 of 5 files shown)


asilvadesigns:
👍

src/actions/operator.ts
```diff
@@ -566,21 +566,36 @@ export class ChangeOperator extends BaseOperator {
   }
 
   public async runRepeat(vimState: VimState, position: Position, count: number): Promise<VimState> {
-    const lineIsAllWhitespace = TextEditor.getLineAt(position).text.trim() === '';
+    let thisLineIndent = vimState.editor.document.getText(
```
xconverge: can you make this const

xconverge:
Awesome, LGTM! Thanks

Issue #2735, opened 2018-06-15
Long key chords does not trigger configured action.


regiontog:
**Is this a BUG REPORT or FEATURE REQUEST?** (choose one):
Bug
**What happened**:
With a long key chord like ```['<leader>', 'f', 'e', 's']``` configured, VSCodeVim  interpret `['<leader>', 'f', 'e']` and `['s']` as separate actions. The command bound to the key chord does not execute.
**What did you expect to happen**:
Expected the bound command to execute.

**How to reproduce it (as minimally and precisely as possible)**:
See unit test in 21659e7239d99f6d6f4fd454fd1a773964c62708

**Environment**:

- Extension (VsCodeVim) version: 0.12.0
- VSCode version: 1.24.0
- OS version: Linux 4.15.12

**Relevant issues**
#1405

jpoon:
I ran into this issue earlier. I think it's because `vim.leader` is being remapped using a key for an existing vim command. For instance, in the UT that you had created:

before: `,fes`
after: `v`

When you press `,` (which is the leader key), we will immediately execute: https://github.com/VSCodeVim/Vim/blob/9cff13d43d40e2b3e717f0234ff7520338030747/src/actions/motion.ts#L749 and not wait for further keys.

PR #2736, opened 2018-06-15
Add unit test for long user configured chords.

regiontog:
Unit test replicating #2735 

I don't understand the state here well enough to say if it's ok to return from ```handleKeyEventHelper``` like in f88f988f2e9b4bb6e697b6fd0395d0b5198ee7de. However I think the cause of the bug is when the ```NoPossibleMatch``` action is returned from ```Actions.getRelevantAction``` and there is a possible remap. In this case the ```NoPossibleMatch``` action is ran, causing ```recordedState.actionKeys``` to be set to ```[]```


jpoon:
Hm, I don't quite recall the reasoning behind that piece of code. Your logic does seem to make sense though.

regiontog:
@jpoon I guess the main assumption I'm making here is that a potential remap should alter the state in the same way a ```KeypressState.WaitingOnKeys``` action would. Is that correct or would a remap need to change the state in some additional way?

xconverge:
I don't recall the reasoning here either...

jpoon:
Sorry @regiontog this has taken so long. My hesitation is that I don't recall the reasoning why we had that in the first place, but your PR makes sense so I'm inclined to merge it. What's your take @xconverge?

regiontog:
@jpoon No problem at all, I only run into this problem when making weird keybinds for convenience functions. It seems odd to me that the extension works as well as it does if my reasoning is correct though, so I am cautious of this change as well. I don't know how many UT you guys have for user keybinds, but perhaps it would be good to have some more to increase confidence?

xconverge:
I am ok to merge it and move on :) I can't really recall either way

jpoon:
Yikes, that merge was a mess. Hope that resolves it.

Issue #2774, opened 2018-06-26
Remapping > to editor.fold


sulkhanp:
## BUG REPORT
**What happened**:
```
"vim.normalModeKeyBindings": [
        {
            "before": [">"],
            "commands": [{"command": "editor.unfold"}]
        },
        {
            "before": ["<"],
            "commands": [{"command": "editor.fold"}]
        }
]
```
These commands don't work.
**What did you expect to happen**:
Expected <, > to fold and unfold.
**How to reproduce it (as minimally and precisely as possible)**:
Copy these remaps into settings and go into foldable code, try to fold it with these keys.

**Environment**: Windows 10
- Extension (VsCodeVim) version: Latest
- VSCode version: 1.25 Insiders
- OS version: Win10

PR #2776, opened 2018-06-26
fix: fixes bug where null arguments to vscode executecommand would fail

jpoon:
Fixes an issue where undefined arguments in user remappings would cause vscode to error out when running `executeCommand` API.

#2774 

I couldn't figure out a more elegant one-liner for doing this :(


xconverge:
LGTM awesome

PR #2790, opened 2018-06-29
fixes #2606

### LINKED ISSUE #2606: Repeating a VISUAL LINE indentation is inconsistent with native vim behaviour

**What happened**: 

bug

1. Press `V` to enter VISUAL LINE mode
2. Press `k` e.g. 3 times to make a selection
3. Press `>` to indent these lines
4. Press `.` to repeat this indentation

The selected lines are indented once **and** the three previous lines are also indented.

expected

I expect **only** the selected lines to be indented (twice). This is what happens in native vim.

See above steps.

**Environment**:

- Extension (VsCodeVim) version: 0.11.5
- VSCode version: 1.23.0
- OS version: macOS 10.13.4

xconverge:
This is pretty wacky...but it is already special cased so 😢  

It is only a problem if your selection is from bottom up

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,23 @@
 # Change Log
 
-## [v0.13.1](https://github.com/vscodevim/vim/tree/HEAD)
-[Full Changelog](https://github.com/vscodevim/vim/compare/v0.13.0...HEAD)
+## [v0.14.0](https://github.com/vscodevim/vim/tree/HEAD)
+
+[Full Changelog](https://github.com/vscodevim/vim/compare/v0.13.1...HEAD)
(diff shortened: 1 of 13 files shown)


xconverge:
Ah shucks this falls apart near the top of the file since I can't get the full selection length, and prevSelection is already updated by this point. I need to take another approach at this

xconverge:
Interesting, in normal Vim, the dot operator always uses the previous visual selection, which we already have (gv for example). So the fix for this should NOT be in any action but in how we rerun a recorded state. If the recorded state contains any visual selection, we should just replace that with a gv instead, then replay the rest of the actions

Issue #2759, opened 2018-06-21
Bug: Remapping Numbers (0-9)


davidnorthetal:
**Is this a BUG REPORT or FEATURE REQUEST?** (choose one):
BUG

**What happened**:
I have the following user settings:
`"vim.normalModeKeyBindingsNonRecursive": [ {
	"before": ["0"],
	"after": ["^"]
} ],`

I type `0` in normal-mode and the same happens as it always does (it jumps to the very  beginning of the current line).

**What did you expect to happen**:
It should instead jump to the first charcter of the current line.

Changing the after from "^" to ":", which should focus the command-prompt instead also turns out to not do anything.

- Extension (VsCodeVim) version: 0.13.1
- VSCode version: 1.24.1
- OS version: Win10

jpoon:
To clarify, the bug is

xconverge:
For some reason this binding also makes my vscodevim experience INCREDIBLY slow, ~1 sec per keypress if holding j or k

PR #2793, opened 2018-06-29
fix: enable remapping of numbers

jpoon:

* This PR addresses issue #2759 to enable remapping of numbers (e.g. remapping `0` to `^`). Although this PR enables the remapping of ALL numbers, I don't see a real scenario where somebody would want to remap numbers 1-9 as that would override the default count behaviour for motions and operators. 
* Adds more checks for remapped configurations ignoring all that are invalid (e.g. duplicate or missing an `after` or a `command`)

#2759

### DIFF
--- a/src/actions/base.ts
+++ b/src/actions/base.ts
@@ -134,7 +134,7 @@ export class BaseAction {
     }
     if (
       this.mustBeFirstKey &&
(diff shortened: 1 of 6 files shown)


xconverge:
Did you experience the crazy slowdown when reproducing the issue before fixing it?

PR #2807, opened 2018-07-05
Added ability to register commands using simple strings (fixes #2806)

### LINKED ISSUE #2806: Allow registering keybindings commands using strings
# Feature request
Currently, custom commands can be added using the following syntax:

```json
{
        {
            "before":["<C-n>"],
            "commands": [
                {
                    "command": ":nohl"
                }
            ]
        }
    ]
}
```

This request proposes to add the following syntax, when arguments are unnecessary:

```json
{
        {
            "before":["<C-n>"],
            "commands": [ ":nohl" ]
        }
    ]
}
```

In the previous code block, `":nohl"` is interpreted as `{ "command": ":nohl" }`.

6A:
Enables the syntax specified in #2806 to work.

#2806.

**Notes:**
This simply makes the "command" object in `IKeyRemapping` be either the old object or a string. If it's a string, we then interpret the string as `{ command: str; args: [] }`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -257,6 +257,26 @@ Custom remappings are defined on a per-mode basis.
 ]
 ```
 
+* Bind `<leader>m` to add a bookmark and `<leader>b` to open the list of all bookmarks (using the [Bookmarks](https://github.com/alefragnani/vscode-bookmarks) extension):
+    * *Note:* `"commands": [ "cmd" ]` is equivalent to `"commands": [ { "command": "cmd", "args": [] } ]`.
+
+```json
+    "vim.normalModeKeyBindingsNonRecursive": [
+        {
+            "before": ["<leader>", "m"],
+            "commands": [
+                "bookmarks.toggle"
+            ]
+        },
+        {
+            "before": ["<leader>", "b"],
+            "commands": [
+                "bookmarks.list"
+            ]
+        }
+    ]
+```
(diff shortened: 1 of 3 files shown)


xconverge:
This looks good to me, I like the simplicity to the user

I think a note in the readme as an example would be nice

README.md
```diff
@@ -245,16 +245,36 @@ Custom remappings are defined on a per-mode basis.
...
-    {
-        "before": [":"],
-        "commands": [
-            {
-                "command": "workbench.action.showCommands",
-            }
-        ]
-    }
-]
+    "vim.normalModeKeyBindingsNonRecursive": [
```
jpoon: Awesome! Would you mind updating the other examples to use the more simplified version?

I'd also be sure to leave examples of how to specify multiple commands and such with both notations.

6A: Do you have an example of a command which also take arguments? I tried finding one to show a "complete" object but didn't succeed. 

In may case I'll change the README in a way that suggests the string is the default, and then when the command takes additional arguments you can use the fully fledged object.

6A:
Alright, I changed all the examples in the README to use that syntax, and also added a command that uses arguments to show the difference (`git.clone` with this repository).

xconverge:
Thanks!

PR #2830, opened 2018-07-11
Workaround surround bug

reujab:
When deleting surroundings using <kbd>ds</kbd>, this plugin doesn't recognize it as being handled, so the command list remains what it was before surround.ts handled it. This pull request fixes this by clearing the command list when a surround command finishes.

screencast

N/A

A better way to fix this bug would be to somehow have [this line of code](https://github.com/VSCodeVim/Vim/blob/e12c6f35edb35d777e5f2f8f1770057cb1e83ce1/src/mode/modeHandler.ts#L330) executed. However, I wasn't able to figure out how, so I came up with this workaround.

### DIFF
--- a/src/actions/plugins/surround.ts
+++ b/src/actions/plugins/surround.ts
@@ -347,6 +347,10 @@ export class CommandSurroundAddToReplacement extends BaseCommand {
       vimState.recordedState.surroundKeys.push(vimState.keyHistory[i]);
     }
 
+    // When deleting surroundings, handled == false. This workaround
+    // resets the command list anyway.
+    vimState.recordedState.resetCommandList();
+
     return false;
   }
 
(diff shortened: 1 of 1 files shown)


xconverge:
I think that there is a cleaner way to do it

If you have ```Finish()``` in surround.ts return true

We can then do something like this anytime we call TryToExecuteSurround():

```
    if (await CommandSurroundAddToReplacement.TryToExecuteSurround(vimState, position)) {
      this.isCompleteAction = true;
    }
```

xconverge:
there is also a ```vimState.recordedState.hasRunSurround``` that can be used in modeHandler perhaps

xconverge:
Do you think that you will get a chance to comment on my feedback in the near future?

reujab:
Your patch is certainly better than mine, and it works.

I'm going to try to read more into the code and get a better understanding of this plugin so I can hopefully fix a couple other things I've encountered.

xconverge:
I would highly recommend fixing things that will be convenient for YOU! Thanks for coming back to this, let me know if you have any questions or want any pointers on where to start with an issue that annoys you!

PR #2851, opened 2018-07-17
Interactive Substitute

### LINKED ISSUE #1010: Substitute interactive mode doesn't work
Please _thumbs-up_ 👍 this issue if it personally affects you! You can do this by clicking on the emoji-face on the top right of this post. Issues with more thumbs-up will be prioritized.

---
### What did you do?

since `ctrl+d` does not work proper trying to use the substitute for multi-place edits
:%s/someWord/anotherWord/gc

ask me to confirm before replacing the matches
### What happened instead?

c flag is ignored and everything is just replaced
### Technical details:
- VSCode Version:1.6x
- VsCodeVim Version: .4
- OS: ubuntu 16.4

kevintighe:
Adds support for the `c` flag to the substitute command.  Implements interactive substitute by using an `inputBox` to ask the user to confirm each replacement.  

Ex: `%s/foo/bar/gc`. 

fixes #1010 

I have not added any tests because I do not know how to interact with an `inputBox` in the tests.  Since this code opens an inputBox to ask for user input, the test would need to be able to send characters to the inputBox and I am not sure that is possible.


src/cmd_line/commands/substitute.ts
```diff
@@ -92,13 +94,87 @@ export class SubstituteCommand extends node.CommandBase {
...
+
+  async confirmReplacement(
+    replacement: string,
+    line: number,
+    vimState: VimState,
+    match: string,
+    matchIndex: number
+  ): Promise<boolean> {
+    const cancellationToken = new vscode.CancellationTokenSource();
+    let validSelections: string[] = ['y', 'n', 'a', 'q', 'l'];
```
xconverge: can this be const

kevintighe: Whoops, yep will fix.

xconverge:
overall this looks good to me from an implementation side, how well does it seem to work? Any issues?

kevintighe:
It works well as far as I can tell.  I've tested what I could think of to test and it behaves as I'd expect.

xconverge:
Just messed around with it, very cool!

xconverge:
it seems to have some issues actually :/

I made a blank file with 

```
test
test
test
```

then did :%s/test/1/gc

and pressed y y y, and it just replaces 1 character, not all of test

PR #2861, opened 2018-07-20
Cache mode so that calls to setContext is minimized

xconverge:
This call takes 2-5ms on my machine....that is a pretty huge for how easy this was to remove...


xconverge:
This actually might be really awesome for helping our performance...

Issue #1211, opened 2017-01-03
Configure automatic line wrapping?


jmreicha:
* Click *thumbs-up* 👍 on this issue if you want it!
* Click *confused* 😕 on this issue if not having it makes VSCodeVim unusable. 

The VSCodeVim team prioritizes issues based on reaction count. 

--------
### What did you do?

Go to the end of a line and press the right arrow key (or l) in normal mode.

### What did you expect to happen?

By default I don't expect the line wrapping to work but a config option to enable whichwrap style behavior would be great.

### What happened instead?

Automatically move up or down a line with arrow keys in normal mode, if configured.

### Technical details:

* VSCode Version: 1.8
* VsCodeVim Version: latest
* OS:  Win 10

xconverge:
are you referring to this?

http://vim.wikia.com/wiki/Automatically_wrap_left_and_right

jmreicha:
@xconverge Really sorry I missed your comment, yes that is exactly what I'm looking for.

Drakulix:
I would like to tackle this an provide a pull request, if somebody could point me into the right direction. The original patchset is not linked in #839.

dtinth:
The workaround that I am using is to disable Vim extension’s handling of Left and Right keys when inside Insert Mode:

```js
{ "key": "left",                  "command": "cursorLeft",
                                     "when": "textInputFocus" },
{ "key": "right",                 "command": "cursorRight",
                                     "when": "textInputFocus" },
{ "key": "left",                  "command": "extension.vim_left",
                                     "when": "editorTextFocus && vim.active && !inDebugRepl && vim.mode != 'Insert'" },
{ "key": "right",                 "command": "extension.vim_right",
									 "when": "editorTextFocus && vim.active && !inDebugRepl && vim.mode != 'Insert'" },
```

brtsos:
I added a whichwrap option to ideavim plugin. 

https://github.com/JetBrains/ideavim/pull/168

Ahsan40:
> 
> 
> The workaround that I am using is to disable Vim extension’s handling of Left and Right keys when inside Insert Mode:
> 
> ```js
> { "key": "left",                  "command": "cursorLeft",
>                                      "when": "textInputFocus" },
> { "key": "right",                 "command": "cursorRight",
>                                      "when": "textInputFocus" },
> { "key": "left",                  "command": "extension.vim_left",
[...]

PR #2864, opened 2018-07-21
Add whichwrap

davidmfoley:

Adds support for whichwrap
http://vim.wikia.com/wiki/Automatically_wrap_left_and_right

#1211

I am new to VS Code extension development, typescript, and this codebase. I tried to match style/etc. to the existing code. Let me know what changes are needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -436,6 +436,13 @@ Configuration settings that have been copied from vim. Vim settings are loaded i
 * What key should `<leader>` map to in key remappings?
 * Type: string (Default: `\`)
 
+#### `"vim.whichwrap"`
+
+* Controls wrapping at beginning and end of line.
+* Comma-separated set of keys that should wrap to next/previous line. Arrow keys are represented by `[` and `]` in insert mode, `<` and `>` in normal and visual mode.
+* Type: string (Default: ``)
+* To wrap "everything", set this to `h,l,<,>,[,]`
+
 ## 🖱️ Multi-Cursor Mode
 
 > :warning: Multi-Cursor mode is experimental. Please report issues in our [feedback thread.](https://github.com/VSCodeVim/Vim/issues/824)
--- a/src/actions/motion.ts
+++ b/src/actions/motion.ts
@@ -483,12 +483,31 @@ export class MarkMovement extends BaseMovement {
   }
 }
 
+const modes = {};
+
+modes[ModeName.Normal] = {
+  '<left>': '<',
+  '<right>': '>',
+};
+
+modes[ModeName.Visual] = modes[ModeName.Normal];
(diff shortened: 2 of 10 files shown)


src/common/motion/position.ts
```diff
@@ -410,13 +410,17 @@ export class Position extends vscode.Position {
     }
   }
 
-  public getRightThroughLineBreaks(): Position {
+  public getRightThroughLineBreaks(includeEol = false): Position {
```
davidmfoley: Note that this flag has the opposite default value from `getLeftThroughLineBreaks`, which is a bit confusing.
I erred on the side of changing fewer lines of code here, over making it more consistent.

src/actions/motion.ts
```diff
@@ -513,7 +516,11 @@ class MoveRight extends BaseMovement {
   keys = ['l'];
 
   public async execAction(position: Position, vimState: VimState): Promise<Position> {
-    return new Position(position.line, position.character + 1);
```
davidmfoley: I am not 100% sure that this change is correct. It works for the motion test cases below, and I think it is functionally equivalent in the non-whichwrap-enabled case, but someone with more knowledge should take a look.

README.md
```diff
@@ -436,6 +436,13 @@ Configuration settings that have been copied from vim. Vim settings are loaded i
 * What key should `<leader>` map to in key remappings?
 * Type: string (Default: `\`)
 
+#### `"vim.whichwrap"`
+
+* Controls wrapping at beginning and end of line.
+* Comma-separated set of keys that should wrap to next/previous line. Arrow keys are represented by `[` and `]` in insert mode, `<` and `>` in normal and visual mode.
```
davidmfoley: I chose to emulate whichwrap's (odd) syntax rather than lean on json to make a more understandable configuration option. Could go either way on that.

src/actions/motion.ts
xconverge: I saw this recently as well, your change makes sense to me either way, if it causes issues we can fix them

README.md
xconverge: I think this is a fine call to me, it is documented here, and is the vim way so why not

src/actions/motion.ts
```diff
@@ -513,7 +516,11 @@ class MoveRight extends BaseMovement {
...
+    if (shouldWrapKey(vimState, this.keysPressed)) {
+      const includeEol = vimState.currentMode === ModeName.Insert;
```
xconverge: is this even possible? l during insert will just insert an l

xconverge:
overall LGTM! Thanks!

PR #2866, opened 2018-07-22
Roadmap doc fix for visual mode case switching

pjlangley:

Adds an undocumented shortcut to the roadmap docs.

n/a

`~` works for case switching in visual mode, so I've assumed that the docs just aren't up to date.

It looks like the description text fell into the keyboard shortcut column... and the keyboard shortcut wasn't yet provided.


xconverge:
Thanks!

PR #2867, opened 2018-07-22
Improve foldfix performance and potentially fix some bugs(#1855 #2163)

### LINKED ISSUE #1855: Moving down at a fold that's at the end of the file causes an infinite loop
There's also the UX issue where moving down on a really huge wrapped lines is very annoying. I plan on implementing a hard limit on how far down it'll check for the wrapped line. This will theoretically introduce incorrect behavior on really long wrapped lines above folds, but that's a tradeoff I'm fine with.

VSCodeVim: 0.8.6
VSCode: 1.13
OS: Arch

### LINKED ISSUE #2163: Cursor jumps erratically before moving vertically

--------

**Environment**:

- **VSCode Version**: 1.18.1
- **VsCodeVim Version**: 0.10.4
- **OS**: Archlinux

**What happened**:
When using a count prefix before vertical moving (j,k), the cursor jumps erratically before moving to target position, as shown in the screencast (the actual rate is much faster than it's shown in gif due to fps limit):

record

The commands I used were, in order: `500j`, `200k`, `50j`, `10k`, `5j`, `gg`. It seems that the time the cursor spent jumping around is roughly proportional to the number of lines it should move pass.

Probably related: #1682, #1154. I tried both with and without neovim integration, but no change. Also since I'm not on mac, the mac setup mentioned in those two reports doesn't apply.

Cursor moves to target line directly, without any jumping, like what happens for `gg`

**How to reproduce it**:
Open any file with sufficient lines, and move cursor vertically by using number prefix `{count}j` or `{count}k`

xmbhasin:
Improved the foldfix setting behaviour for improved performance, reduced bugs, and hopefully easier extension and addition of more fold commands in the future.
Should fix #1855 and fix #2163, though it may require more testing. `gulp test` passed all tests with and without forcing `configuration.foldFix` to `true`, but more tests may need to be added.

One bug this may have introduced is flickering when prefixing a downward movement command with a large repeat. This is indicated in a comment inside `MoveDownFoldFix` but is still much better behaviour than before where an infinite loop would begin. To fix it we would need a more reliable way to detect the  last visual line of a file (must account for both wrapped lines and folds).

### DIFF
--- a/src/actions/motion.ts
+++ b/src/actions/motion.ts
@@ -260,25 +260,33 @@ class MoveDownFoldFix extends MoveByScreenLineMaintainDesiredColumn {
   value = 1;
 
   public async execAction(position: Position, vimState: VimState): Promise<Position | IMovement> {
-    if (position.line === TextEditor.getLineCount() - 1) {
+    if (position.line >= TextEditor.getLineCount() - 1) {
       return position;
     }
-    let t: Position;
-    let count = 0;
+    let t: Position | IMovement;
+    let prevLine: number = position.line;
+    let prevChar: number = position.character;
     const prevDesiredColumn = vimState.desiredColumn;
+    const moveDownByScreenLine = new MoveDownByScreenLine();
     do {
-      t = <Position>await new MoveDownByScreenLine().execAction(position, vimState);
-      count += 1;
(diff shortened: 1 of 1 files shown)


xconverge:
Tough one for me to review due to the nature of foldFix in general 😄 

Thanks for the contribution!

Issue #2879, opened 2018-07-24
Improve foldFix and add fold-awareness to more actions


xmbhasin:
**Is your feature request related to a problem? Please describe.**
`foldFix` currently does not adjust all actions for fold-awareness. `foldFix` also has no test configurations and does not have any specific tests for checking fold-awareness. 

**Describe the solution you'd like**
I have submitted some pull requests (#2867)  so far to address this and the following list keeps track of which actions I want to add fold-awareness to and which now have some support.
- [x] deleting
- [x] yanking

Testing seems like a difficult problem as there is no way currently to actually determine the visual or fold-aware state of the editor. It may be possible however to create a file with known folds and test if foldFix actions behave as expected.

**Describe alternatives you've considered**
The best alternative would be leveraging api features for fold-awareness, but since this is currently blocked (see #1004, Microsoft/vscode#22276) it would be nice to have better support in foldFix for those needing to work with heavily folded files.

**Additional context**

PR #2887, opened 2018-07-24
[WIP] Improve foldFix further by making more motion prefixed actions fold-aware

xmbhasin:
Improve foldFix further by making more motion prefixed actions fold-aware (including at least deleting and yanking)
See #2879 where I am tracking improvements to foldFix.

`gulp test` passes all tests with and without forcing `foldFix` to true. I'm not sure how to go about it, but I would like to add a foldFix configuration to `gulp test`. I don't want to make too many changes there as I'm worried it might break CI.


xconverge:
It would be very nice to have a few unit tests or at least a file that allow us to test behavior around folds

https://github.com/VSCodeVim/Vim/blob/master/test/multicursor.test.ts is an example what I think might be nice for starters, just something thin that allows us to have SOME metrics of how we are doing.

You should also be able to do something like: ```Configuration.foldFix = true``` or something in the test case/suite so that you don't impact other tests

xmbhasin:
I managed to get some test coverage for fold-aware movement and actions and fixed a few bugs along the way. It was rather tricky but I think there is a decent setup for adding more tests for foldfix in the future.

xmbhasin:
Some things I couldn't quite figure out and might need some help with:
 1. How can I set the `editor.wordWrap` and `editor.wordWrapColumn` settings? This is required for properly testing movement over wrapped lines.
2. How can I isolate a test file to be run in the docker container? This would help speed up testing and iterating.

xconverge:
O wow, I lost track of the fact you made changes here! Give me a little bit of time to go through the changes, I personally am not a heavy fold user so I am hoping that the changes make sense to you and provide a better experience for something that was a pain point for you in the current/past releases.

Regarding your questions

1. Would something like this help? 
```
await vscode.commands.executeCommand('editor.action.toggleWordWrap');
``` 

or something like:
```
let options = this.vimState.editor.options;
options.cursorStyle = cursorStyle;
this.vimState.editor.options = options;
```

2. Not sure, maybe @jpoon would have an idea

jpoon:
> How can I set the editor.wordWrap and editor.wordWrapColumn settings through the testConfiguration? This is required for properly testing movement over wrapped lines.

You can do something like this: https://github.com/VSCodeVim/Vim/blob/master/test/mode/modeNormal.test.ts#L14

You can take a peek inside `setupWorkspace` to see what it's doing with the configuration, but namely once you set it, you'll need to "reload the configuration" so you can either use the `setupWorkspace` helper function of manually reload the configuration. 

> How can I isolate a test file to be run in the docker container? This would help speed up testing and iterating.

If you find out, let us know :). Tests are executed using vscode's harness: `node ./node_modules/vscode/bin/test`. No idea what their test harness does, but we are using mocha for tests so there theoretically should be a way to just run a single test. 

If you are using the `newTest` syntax for the tests, you can swap it to `newTestOnly` and it'll only run that single test. 

The above applies to every which way you execute the tests: through vscode, `npm test`, or `gulp test` (ie. docker).

xmbhasin:
I checked out `./node_modules/vscode` and followed it back up to `./test/index.ts`. Turns out the mocha test configuration was there all along, but it was under the `testRunner` name! I've made a pull request adding --grep to `gulp test` here #2904.

jpoon:
Yikes, this part of the codebase I'm unfamiliar with. @Chillee can you CR?

Chillee:
@jpoon I'll take a look tomorrow.

xmbhasin:
I would like to refactor the changes here when I find the time as I did in #2921 by introducing a new subclass to reduce duplication of the `execActionWithCount` methods. Also, I stopped working on this for a while even though I had wanted to add fold support for opening lines and pasting lines. The problem however was that we rely right now on VSCode's built-in commands for inserting new lines, which from my testing is not itself fold-aware (at least not in the way vim is).

xmbhasin:
I also still need to adjust the test configuration to have predictable line wrapping so I can ensure this also fixes traversal of wrapped lines.

Chillee:
First of all, thanks a lot for the PR!

Sorry I didn't get to this earlier, but I'm not sure how I feel about extending upon `foldFix`. As I'm sure you're aware of if you took a look at the `foldfix` code, the implementation is incredibly hacky. We're literally "simulating" stepping up with the extension to obtain the necessary information that lets us step across folds. Thus, I'm inclined to not want to extend upon the `foldfix` code.

When I initially added the option, it was meant as a stopgap solution to make VSCodeVim usable for those who needed to use folds. Most of the time, the main frustration was caused by folds opening when trying to move over them with `j/k`.

On the other hand, there appear to be a fairly solid test suite, and the `foldfix` code is mostly segregated by an if statement (so it's not adding unmanageable complexity), so I wouldn't be too opposed to merging this PR.

What do you think @jpoon ?

xmbhasin:
As I noted, I don't think it will be possible to completely fix folds even if we continue to hack around vscode. But with the abstraction I added to the `BaseMovement` class in another PR, I think the foldfix code can be better isolated and made easier to modify or remove if in the future vscode provides a better api for folds. Give me some time to update this PR and I think it should be worthwhile to merge. I could also get rid of the support for deleting and yanking, and focus here only on cleaning up the core foldfix code.

Chillee:
@xmbhasin I don't mind keeping the deleting and yanking, but I would really like the abstraction of foldfix. Thanks!

jpoon:
Closing this PR due to the lack of activity. Please re-open once this is ready for review.

PR #2894, opened 2018-07-25
Updating README FAQ

augustnmonteiro:
To use native functions for Ctrl the vim.useCtrlKeys needs to be false

Update the readme FAQ

### DIFF
--- a/README.md
+++ b/README.md
@@ -626,7 +626,7 @@ Vim has a lot of nifty tricks and we try to preserve some of them:
 
 ### None of the vim `ctrl` (e.g. `ctrl+f`, `ctrl+v`) commands work
 
-Set the [`useCtrlKeys` setting](#vimusectrlkeys) to `true`.
+Set the [`useCtrlKeys` setting](#vimusectrlkeys) to `false`.
 
 ### Moving `j`/`k` over folds opens up the folds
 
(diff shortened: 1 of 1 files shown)


xconverge:
This is actually worded correctly for the intention. I believe at the time of writing the default value was false, and now it is true so the entire question/answer could be removed.

Likewise the entire question/answer could be rephrased to something like:
I want to use ctrl keys in native vscode but vscodevim keeps controlling them. With an answer like, set this to false, or use vim.handleKeys to disable particular control key combinations.

augustnmonteiro:
@xconverge thank you, I just updated the question : )

README.md
```diff
@@ -624,9 +624,9 @@ Vim has a lot of nifty tricks and we try to preserve some of them:
 
 ## 📚 F.A.Q.
 
-### None of the vim `ctrl` (e.g. `ctrl+f`, `ctrl+v`) commands work
+### None of the IDE `ctrl` (e.g. `ctrl+f`, `ctrl+v`) commands work
```
xconverge: Sorry for being pedantic, but can you change ```IDE``` to ```native Visual Studio Code```

augustnmonteiro: Sure : )

PR #2902, opened 2018-07-27
Remove unnecessary allocation every tick

xconverge:
I am not sure of the repercussions of this but it makes sense to me. We have a map of actions already allocated so lets use them! This function is called every tick for single key actions so it makes sense to me to minimize time spent allocating

### DIFF
--- a/src/actions/base.ts
+++ b/src/actions/base.ts
@@ -198,9 +198,8 @@ export class Actions {
       const { type, action } = possibleAction!;
 
       if (action.doesActionApply(vimState, keysPressed)) {
-        const result = new type();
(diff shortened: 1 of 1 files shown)


xconverge:
closing because the failing test scares me a bit. The performance gain is not worth the stability hit IMO

@jpoon you can consider this area since I know you were peeking around here with your refactor

jpoon:
Thanks! Had this in my local branch. Yep, made sense to me too.

jpoon:
weird. i had the exact change in my local branch, but all the tests were passing locally (but i'm on windows). after pushing to remote, see same issues you have

PR #2904, opened 2018-07-27
Add --grep flag to gulp test

xmbhasin:
Add --grep flag to `gulp test` for faster testing and iteration. The flag --grep takes js regex as a string and filters by test and test suite names. Also works with the default task since it includes `gulp test` (eg `gulp --grep foldFix`).


xmbhasin:
I think I'm done making changes to this PR for now, but I could also quickly add other [`mocha test` options](https://github.com/mochajs/mocha/wiki/Using-mocha-programmatically) as `gulp test` flags if you think any would be useful.

xconverge:
Really appreciate it, none of the other mocha test options scream out to me as super useful right now

PR #2915, opened 2018-07-30
fix: use full path for configs

jpoon:

* Enable auto-complete when trying to configure statusbar colors and custom cursors for mode. 
* Remove need for having to create a map of the custom cursors every tick. 


xconverge:
This might break some people's configs who are using more undocumented modes

jpoon:
True. Should add them all then.

Issue #2834, opened 2018-07-11
Support for Wildfire (visual selection  expansion using text objects)


expelledboy:
Hey guys, love the idea of using vim in vscode!

Would it be possible atm to implement something similar to [vim's wildfire](https://github.com/gcmt/wildfire.vim) plugin? It expands the visual selection according to next recognized text objects, searching in order of a configurable list.

![Wildfire Preview](https://github.com/gcmt/wildfire.vim/blob/master/_assets/preview.gif?raw=true)

I use it so much I actually cant use vim anywhere that doesn't support it, just breaks my concentration.

xconverge:
did you try our keybinding of ```af``` https://github.com/VSCodeVim/Vim#-vscodevim-tricks

You can rebind something else to af if that makes you more comfortable as well

expelledboy:
@xconverge Wow! That was easy, you should add it to the readme!

```json
    "vim.normalModeKeyBindingsNonRecursive": [
        {
            "before": ["enter"],
            "after": ["v", "a", "f"],
        }
    ],
    "vim.visualModeKeyBindingsNonRecursive": [
        {
            "before": ["enter"],
            "after": ["a", "f"],
        }
    ]
```

The only other powerful feature lacking here is then, having custom rules for text object expansions per language. For example in Erlang I expand inside `{` character according to what its encapulating, but these do no apply to any other lanugage. Even the order matters, if we cant define our own custom text objects.

But now I am invested, I think I am going to have a read through how this all works. This project has come along way!

Zyst:
@expelledboy I think it is, what do you think is missing there?

https://github.com/VSCodeVim/Vim#-vscodevim-tricks

If necessary it's possible to write more detailed docs for it.

expelledboy:
@Zyst My bindings with `af` posted above is what I was looking for, but the final step to get the full wildfire goodness is to have custom rules for text expansion per filetype.

If anyone could point me to where its implemented I would appreciate it!

cursorzz:
Is there any command that support select **Inside** block? for the example below

> af - visual mode command which selects increasingly large blocks of text. For example, if you had "blah (foo [bar 'ba|z'])" then it would select 'baz' first. If you pressed af again, it'd then select [bar 'baz'], and if you did it a third time it would select "(foo [bar 'baz'])".

I would like to select baz instead of 'baz'

Glench:
[...]

PR #2921, opened 2018-07-31
Fix multiple issues with expand selection commands and pair/block movement

### LINKED ISSUE #2506: va{a{ doesn't work

**What happened**: va{a{ doesn't work

**What did you expect to happen**: the `a{` part to select the parent braces

**How to reproduce it (as minimally and precisely as possible)**: find nested braces and do `va{a{`

**Environment**: OS X

- VSCode version: 1.2.1
- OS version: macOS Sierra 10.12.6

### LINKED ISSUE #2907: Expand selection with inner tag selection command
Hello everyone!

I come from Sublime Vintage mode context, and I was very used to type `vit` to select the inner content of a tag, which works in VSCode Vim, but when I typed `it` again and again in sublime it would keep expanding the selection up, which doesn't happen in VSCode Vim. 

Is there anyway to achieve this behavior? 

Thanks in advance!

xmbhasin:
Fixes multiple issues with bracket, tag, and other pair movement and selection commands. Adds backtick and angle brackets support to `af` command. Fixes `af` jumping to nonenclosing blocks. Adds `af` expanding over multiple types of blocks in succession. Fixes `afd` commands deleting and then returning cursor to the wrong position. Fixes expanding repeatedly over the same type of block ( eg using `v3a}` over `{{{|}}}` ). Adds a number of new tests around the behaviour of these changes.

Refactors `matcher.ts` and improves its performance.

Should get us closer to implementing more of the wildfire plugin as requested in #2834 in the form of `af`.
Should fix #2907 and fix #2506.

You can test out the difference these commits make in a file such as this:
```
<div>
  <ul>
    <li>
      <p>Lorem!</p>
    </li>
  </ul>
</div> 

do not delete

(
[
(
  {
    {
      (blah((foo))blah)
    }
  }
)
]
)

do not delete

test <div><p>[[{{((|))}}]]</p></div> test;
```


src/actions/motion.ts
```diff
@@ -1841,12 +1903,71 @@ abstract class MoveTagMatch extends BaseMovement {
...
 
+  public async execActionWithCount(
+    position: Position,
+    vimState: VimState,
+    count: number
+  ): Promise<Position | IMovement> {
+    let recordedState = vimState.recordedState;
+    let result: Position | IMovement = new Position(0, 0); // bogus init to satisfy typechecker
+
+    if (count < 1) {
```
jpoon: I know we've had this pattern but we should really move this sort of check somewhere outside this function (on a different PR).

src/actions/commands/actions.ts
```diff
@@ -3817,7 +3817,7 @@ class ActionChangeChar extends BaseCommand {
 
   public couldActionApply(vimState: VimState, keysPressed: string[]): boolean {
     return (
-      super.doesActionApply(vimState, keysPressed) &&
+      super.couldActionApply(vimState, keysPressed) &&
```
jpoon: Good catch.

src/actions/motion.ts
xmbhasin: I was thinking of abstracting more of the code in this function to private methods. This would probably allow us to add a couple of classes between `BaseMovement` and its subclasses and reduce duplication. That might also help in the future to refactor how count-prefixes are handled before reaching `BaseMovement`.

jpoon: Were you thinking of doing that in this PR or sometime later?

xmbhasin: I'll try to cleanup some of the changes in this PR and sprinkle in some more comments. If you have any other concerns or suggestions I'd like to organize what I need to do so I can plan some time to get to it.

xmbhasin:
I think I'm ready for a final review of this PR. I would have liked to fix count-prefixed actions with inner tag `it` movements. But it may have been a bit awkward with the way `it` alternates between selecting inner and outer tag content when repeated. Additionally, I'm not actually sure if vim even supports count-prefixed actions with pairs/blocks. I only realized that after having added it in this PR, but it is quite useful.

jpoon:
I'll merge when build finishes.

PR #2964, opened 2018-08-16
feat: implement 'changeWordIncludesWhitespace' option

darfink:
Vim uses a special implementation specifically for the `cw` (change word) action, it's implemented identical to `ce`. This makes the `cw` action inconsistent compared to other composed variants, such as `yw` and `dw` which include trailing whitespace. This reduces the utility of the composed commands, and introduces unexpected behavior.

This pull request adds a `changeWordIncludesWhitespace` option that enables the "consistent" behavior. Whilst this can be partially mimicked with mappings, it does not produce the same undo stack.

None has been created, if preferred I can create one.

Some related examples:
- evil-mode's  [`evil-want-change-word-to-end`](https://github.com/noctuid/evil-guide/blob/master/README.org#whats-the-equivalent-of-nnoremap-y-y) option.
- NeoVim's [`set cpoptions-=_`](https://github.com/neovim/neovim/issues/6234) option.
- Vim [vim-you-keep-using-that-word](https://github.com/ap/vim-you-keep-using-that-word) plugin.

I tried to choose a fitting name for the option, but you may prefer another.


jpoon:
Inline with `neovims` option, can we call this configuration: `cpoptions` as well? Or would that be too confusing?

darfink:
It may be a bit confusing, considering `cpoptions` is a flag register that contains multiple settings (e.g `E`, `C`, `M`, ...). It's only when the `_` flag is removed that this specific behavior is affected.

jpoon:
I'm cautious on adding extension-specific settings to describe vim behaviour as we should try to have 1:1 parity between the settings for vscodevim and vim.  

I don't use `cw` very much, but from your description, it seems like this behaviour should be the default? If so, I'd be OK with scrapping the need for having this configuration.

Chillee:
I'd err on the side on not making it default (simply because it doesn't preserve parity with regular vim).

i think as is, I'm fine with merging this PR. If we pick up more things like this (ie: small things that people often like changing from vim), we might want to create something similar to nvim's cpoptions.

xconverge:
LGTM can we add it to the readme so that someone besides you can use it? :)

darfink:
I've added an entry to the README that is hopefully sufficient.

PR #2993, opened 2018-08-25
I or A in visual/visual line mode creates multiple cursors #2167

### LINKED ISSUE #2167: [FEATURE REQUEST]visual line mode support A or I

--------

**This is a FEATURE REQUEST** :

**Environment**:

- **VSCode Version**: 1.18.1
- **VsCodeVim Version**: 0.10.5
- **OS**: windows 7

**Behavior**:

Press "shift + v" toggle visual line mode, then use "j" or "k" select serval lines, then press "A" or "I" to add cursors at the end or front of all the lines  selected and switch to insert mode.

This can be vary useful.

axs221:

This applies to Visual (`v`) and Visual Line (`shift+v`) modes, for keys `A` and `I`.

If you highlight text and press `I`, it will add cursors to the first non whitespace character of each line (or the visual cursor in Visual mode and on the first line).

Similarly, if you highlight text and press `A`, it will add cursors to the end of those lines.

Blank lines are skipped. This is what I prefer and assume others would prefer, feel free to state otherwise. Most use cases I would have would not involve the blank lines, and it would be more trouble than it is worth to get rid of the blank line cursors or modifications afterwards.

fixes #2167 

VisualBlock mode has similar behavior. I left it alone. It does have a bug in it I found where, if you visually block select a line but select no characters on that line (such as can happen easily if just selecting the end of ragged lines), it will delete the text of the line. I thought that would be best covered under a separate PR / issue. The code itself could possibly be shared, but it works slightly different in that it tries to handle blank lines (which is where the buggy code is), and it has different iteration logic.


jpoon:
LGTM. Thanks for the PR. Sorry for the delay in review, I'm in the midst of a job change.

PR #3025, opened 2018-09-04
fix: re-enable relativelinenumbers. closes #3020

### LINKED ISSUE #3020: Support smart relative line numbers
It would be nice if there was a configuration option such that line numbers are relative when jumping might be used (example:`10j`), and absolute when they could not.

My suggestion would be to have a configuration option `smartRelativeLineNumbers` that, when enabled, overrides user settings on line numbers and turns on relative line numbering in normal and visual modes, and absolute in insert mode.

Chintzy plugins.
aiotecism/amVim, which [enables this feature](https://github.com/aioutecism/amVim-for-VSCode/issues/197), but is otherwise not as good of a plugin for my needs.

**Additional context**
n/a

jpoon:

usage:

```
:set relativenumber
:set norelativenumber
```

#3020 

### DIFF
--- a/src/configuration/configuration.ts
+++ b/src/configuration/configuration.ts
@@ -18,10 +18,6 @@ const packagejson: {
 } = require('../../../package.json');
 
 type OptionValue = number | string | boolean;
-type ValueMapping = {
-  [key: number]: number | string | boolean;
-  [key: string]: number | string | boolean;
-};
 
 interface VSCodeKeybinding {
   key: string;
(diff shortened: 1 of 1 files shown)


d-gubert:
One interesting side-effect (or maybe it is intended) is that it is now possible to configure this in VSCode settings json file, even though the IntelliSense doesn't recognize the setting 🙈 

```
{
  // other configs
  "vim.relativenumber": true
}

PR #3028, opened 2018-09-05
Feature/improved jump list

### LINKED ISSUE #1933: Make small movement command not registered to Ctrl+o

--------

**Environment**:

- **VSCode Version**: 1.14.1
- **OS**:  macOS Sierra

**What happened**:

When using movement like j or k, it should not be undoable with Ctrl+o.

When going to definition for example, it can be reverted with Ctrl+o however if I press j or k a bit I need more Ctrl+o to finally revert the go to definition. In the original vim, Ctrl+o is specifically for big movements

### LINKED ISSUE #2461: Ctrl-o (go back) record every cursor move ?
FEATURE
**What happened**:
When I use 'gd' to go defined, it dump to another file, and press ctrl-o to go back.
But go back in vscode is based on step that record every cursor move and the history stack is limited.If I move many time, I could not go back to the last file.
wang2

Like the native vim: don't record every move or add new feature that support go-back-file-based.
wang

**Environment**:

- Extension (VsCodeVim) version:0.11.2
- VSCode version:1.12.1
- OS version:4.15.9-1-ARCH

shawnaxsom:

There have been several issues opened in the past that discuss the limitations of the current `<C-o>` and `<C-i>` jump list implementation, which relies on vscode internals.

The largest criticism is that small movements (such as j and k) are registered as jumps (vscode has an internal "jump" tracker, a history service, that is hard-coded with constant EDITOR_SELECTION_THRESHOLD to register a jump after moving 10 lines). The history was also limited. A common use case is to jump to definition in another file, navigate around that file to learn about how it works, and then to jump back. The user would have to jump back many times more than in vanilla Vim, and often would hit the start of history as old entries are truncated off.

This PR therefore implements our own jump list (or "Jump Tracker", in vein of the name of our "History Tracker" that is used for undo/redo). This gives us much more control of jump history, and I have also matched it with vanilla Vim behavior.

Motions and actions can now set an `isJump` flag to mark that they should be tracked as a jump. I have updated all of the existing motions and jumps to match vanilla Vim. Substitutes and some commands will also count jumps.

For testing, I enhanced `testSimplifier` to allow tracking jump position and jump entries. In the tests I wrote, I made sure to compare to vanilla Vim, matching the `:jumps` command in the expected jump results.

fixes #1933
fixes #2461

I ran into some issues trying to unit test, that I believe is unrelated to this PR. I tried to fix a few of the issues. I am still seeing this one:

```error: ModeHandler: error handling key=d. err=TypeError: Cannot read property 'document' of undefined.```

That seems to happen for the first test in the list if I run an individual test suite like this:

```MOCHA_GREP='Can record jumps for actions the same as vanilla Vim' npm test```

Anyway, here are some future enhancement ideas / limitations:

* Add `:jumps` and `:clearjumps` commands, matching Vim's built-in commands
* Neovim-based commands aren't tracked granularly, such as `:substitute`.
* Jumps where files were deleted may show a benign error `Unable to open...File not found...`. Behavior works fine, and you can continue to jump to previous entries. I wonder if there is a way to detect the error and remove the jump after the fact in case they jump to that point again.
* Split panes share the jump list. This seems similar to how vscode itself works with the built-in back functionality. I'm storing the JumpTracker in `GlobalState`. I don't know that we have per-pane state, so if we wanted to make it per-pane it would take a little more work.

### DIFF
--- a/src/jumps/jumpTracker.ts
+++ b/src/jumps/jumpTracker.ts
@@ -25,6 +25,67 @@ export class JumpTracker {
   }
 
   /**
+   * Current position in the list of jumps.
+   * This will be past last index if not traveling through history.
+   */
+  public get currentJumpIndex(): number {
+    return this._currentJumpIndex;
+  }
+
+  /**
+   * Current jump in the list of jumps
+   */
+  public get currentJump(): Jump {
+    return this._jumps[this._currentJumpIndex];
+  }
+
+  /**
+   * Current jump in the list of jumps
+   */
+  public get hasJumps(): boolean {
+    return this._jumps.length > 0;
+  }
+
+  /**
+   * Last jump in list of jumps
+   */
+  public get end(): Jump | null {
+    return this._jumps[this._jumps.length - 1];
+  }
+
+  /**
+   * First jump in list of jumps
+   */
+  public get start(): Jump | null {
+    return this._jumps[0];
+  }
+
+  /**
+   * Record that a jump occurred.
+   *
+   * If the current position is back in history,
+   * jumps after this position will be removed.
+   *
+   * @param from - File/position jumped from
+   * @param to - File/position jumped to
+   */
+  public recordJump(from: Jump, to?: Jump | null) {
+    const currentJump = this._jumps[this._currentJumpIndex];
+
+    if (to && from.onSameLine(to)) {
+      return;
+    }
+
+    this.clearJumpsOnSameLine(from);
+
+    if (from && !from.onSameLine(to)) {
+      this._jumps.push(from);
+    }
+
+    this._currentJumpIndex = this._jumps.length;
+  }
+
+  /**
    * Record that a jump occurred from one file to another.
    * This is likely only needed on a handler for
    * vscode.window.onDidChangeActiveTextEditor
@@ -57,54 +118,32 @@ export class JumpTracker {
   }
 
   /**
-   * Record that a jump occurred.
-   *
-   * If the current position is back in history,
-   * jumps after this position will be removed.
-   *
-   * @param from - File/position jumped from
-   * @param to - File/position jumped to
-   */
-  public recordJump(from: Jump, to?: Jump | null) {
-    const currentJump = this._jumps[this._currentJumpIndex];
-
-    if (to && from.onSameLine(to)) {
-      return;
-    }
-
-    if (currentJump && currentJump.onSameLine(from)) {
-      this.updateCurrentJumpColumn(from);
-      return;
-    }
-
-    this.clearJumpsOnSameLine(from);
-
-    if (from && !from.onSameLine(to)) {
-      this._jumps.push(from);
-    }
-
-    this._currentJumpIndex = this._jumps.length - 1;
-  }
-
-  /**
    * Get the previous jump in history.
    * Continues further back if the current line is on the same line.
    *
    * @param from - File/position jumped from
    */
   public back(from: Jump): Jump {
+    if (!this.hasJumps) {
+      return from;
+    }
+
     if (this._currentJumpIndex <= 0) {
       return this._jumps[0];
     }
 
-    this._currentJumpIndex = this._currentJumpIndex - 1;
-    const jump = this._jumps[this._currentJumpIndex];
+    let to: Jump;
 
-    if (jump && jump.onSameLine(from)) {
-      this.removeCurrentJump();
-      return this.back(from);
+    if (this._currentJumpIndex === this._jumps.length) {
+      to = this._jumps[this._currentJumpIndex - 1];
+      this.recordJump(from, to);
+      this._currentJumpIndex = this._currentJumpIndex - 2;
+    } else {
+      to = this._jumps[this._currentJumpIndex - 1];
+      this._currentJumpIndex = this._currentJumpIndex - 1;
     }
-    return jump;
+
+    return to;
   }
 
   /**
@@ -114,18 +153,22 @@ export class JumpTracker {
    * @param from - File/position jumped from
    */
   public forward(from: Jump): Jump {
-    if (this._currentJumpIndex >= this._jumps.length - 1) {
-      return this._jumps[this._jumps.length - 1];
+    if (!this.hasJumps) {
+      return from;
+    }
+
+    if (this._currentJumpIndex >= this._jumps.length) {
+      return from;
     }
 
     this._currentJumpIndex = Math.min(this._currentJumpIndex + 1, this._jumps.length - 1);
     const jump = this._jumps[this._currentJumpIndex];
 
-    if (jump && jump.onSameLine(from)) {
-      this.removeCurrentJump();
-      this._currentJumpIndex -= 1;
-      return this.forward(from);
-    }
+    // if (jump && jump.onSameLine(from)) {
+    //   this.removeCurrentJump();
+    //   this._currentJumpIndex -= 1;
+    //   return this.forward(from);
+    // }
     return jump;
   }
 
--- a/src/jumps/jumpTracker.ts
+++ b/src/jumps/jumpTracker.ts
@@ -1,4 +1,7 @@
+import * as vscode from 'vscode';
+
 import { Jump } from './jump';
+import { Position } from './../common/motion/position';
 
 /**
  * JumpTracker is a handrolled version of vscode's TextEditorState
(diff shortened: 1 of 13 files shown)

### CODE REVIEW

**Comment by Chillee** [comment]
Awesome!!! This is one of my personal biggest pet peeves with VSCodeVim, but I haven't had the time/energy to completely rewrite the jump tracking system.

I'll need to spend some time reviewing it (as well as probably some time trying it out), but I just wanted to say thanks first!

**Comment by Chillee** [comment]
Also, just a couple notes on some of the things that you noted:

Currently, state is not tracked per pane. It used to be, but I removed that due to it causing some really bad bugs with undo (where it would delete a lot of a user's history).

EDITOR_SELECTION_THRESHOLD is used in VSCode internally, but it actually doesn't help with our `<ctrl-o>` issues. I remember thinking that adding the EDITOR_SELECTION_THRESHOLD would help our issues, but they have a check that causes "changes made from an extension" to override the threshold. So every change we make is tracked by VSCode.

**Comment by axs221** [comment]
No problem! Thanks for the "thanks", and thank you for all of your hard work on this plugin. It's gotten me off of vanilla Vim for the longest time in recent years.

I should do a little more testing as well I'm sure. What is there seems to be working very well, but I'm not sure how well it play with other plugins. I am sure I need to handle some extra cases like `handleContentChangedFromDisk`, for example.

Yeah I did see some of the undo history discussion on #2007. That makes sense. The current behavior split pane doesn't bother me, though I don't use split panes often in vscode. Sometimes it might open a file in a different pane than expected, may be some work there.

File: src/jumps/jumpTracker.ts [hunk 1]
```diff
@@ -0,0 +1,265 @@
...
+    }
+
+    if (this._currentJumpNumber <= 0) {
+      return this._jumps[0];
+    }
+
+    const to: Jump = this._jumps[this._currentJumpNumber - 1];
+
+    if (this._currentJumpNumber === this._jumps.length) {
+      this.recordJump(from, to);
```
Chillee: Why is this necessary?

src/jumps/jumpTracker.ts
```diff
@@ -0,0 +1,265 @@
...
+  }
+
+  clearJumpsOnSameLine(jump: Jump): void {
+    this._jumps = this._jumps.filter(j => j === jump || !j.onSameLine(jump));
+  }
+
+  removeDuplicateJumps() {
+    for (let i = 0; i < this._jumps.length; i++) {
+      const jump = this._jumps[i];
+      this.clearJumpsOnSameLine(jump);
```
Chillee: Are we concerned about potential N^2 complexity here?

The max N=100, but 100^2 is still 10k, which is bordering on significant.

If any of the comparisons are expensive, then this might become an issue.

Chillee: Also, is it an issue which order we remove jumps?

src/jumps/jumpTracker.ts
```diff
@@ -0,0 +1,265 @@
...
+   *
+   * File jumps have extra checks in place, keeping in mind
+   * whether this plugin initiated the jump, whether the new file is
+   * a legitimate file.
+   *
+   * @param from - File/position jumped from
+   * @param to - File/position jumped to
+   */
+  public recordFileJump(from: Jump | null, to: Jump) {
+    if (this.isJumpingFiles) {
```
Chillee: Is there a potential race here, and does it matter?

Say we call `handleFileJump` twice before `recordFileJump` gets triggered once by the `onDidChangeActiveTextEditor`, we might record the second `handleFileJump` as a real jump.

src/jumps/jumpTracker.ts
axs221: That is how vanilla Vim appears to work, which lets you `<C-i>` after you've jumped back. Vim doesn't keep track of where you jumped to until you jump back. 

Here's video demonstrating that in Vim: https://cl.ly/b1d14cd87b45

Notice how the first time I show `:jump`, it doesn't show the current line of where I jumped to (`export class JumpTracker`), and that the `>` position arrow is outside of the jumps list, but then I press `<C-o>` and I can see a jump was added and the `>` position arrow is now on the `from` jump.

src/jumps/jumpTracker.ts
axs221: Good points. The latest jump should be preserved, to have the latest column number.

I can loop through the jumps in the list in reverse and keep track of the line numbers as I go, removing when I see a duplicate. That should get it down to O(n) at least and preserve the newer jump.

axs221: I changed it as described, and I added a new "Can handle text deleted from a file" test to `JumpTracker.tests.js` to validate behavior. Added tests, also added a test for Visual Mode instead of just Visual Line Mode.

src/jumps/jumpTracker.ts
axs221: It does matter if that race condition would occur. If you remove that if block, then it will record the `from` location as a jump again, which puts it back in the jump stack as the next location `<C-o>` will take you to. If you `<C-o>` again, same thing happens, so it circles between the two files so you can't get to older jumps.

But I haven't been able to reproduce the potential race condition. I tested by jumping to definition for two files in a row. I then manually `<C-o>` twice and `<C-i>` twice. Worked fine every time. I also tried `2<C-o>` and `2<C-i>` and that works fine as well.

I'd suggest it is fine as it is. I don't think there is a perfect solution for this. We could use a counter instead, but counters can be worse in their own ways if counter change events are missed for whatever reason. I also was going to originally just go by filename, seeing if it matches the previous or next in the  jump list, but that doesn't match vanilla Vim and might have unexpected behavior.

axs221:
I pushed up some changes based on the conversation so far. I also did a little refactoring, pulling in logic from `Actions.ts` into `JumpTracker.ts` itself to better encapsulate the logic, and to maybe be able to unit test it better.

src/jumps/jumpTracker.ts
Chillee: Alright, I tested it out, and I think the race condition occurring doesn't cause it to fail disastrously. I'm fine with it.

Chillee:
I think it looks good to me! Tell me if you think it's ready to merge.

axs221:
I think we are good! Should be ready to merge.

I tested file changes on disk this morning, and `vscode.workspace.onDidChangeTextDocument` fired and `textWasDeleted` returned true, so it handled that use case fine, which was the last one I was concerned about. Hopefully if it can handle that then it can handle whatever lines any plugin tries to add/delete from the files.

ohjames:
I was hoping the `''` binding would be fixed by this, but it's still as buggy in 0.16.5 as ctrl-o etc. was before.

shawnaxsom:
@ohjames Yes I didn't touch double single quote or double back tick keys. What is a specific use case that isn't working? Even though it is still using built-in vscode buggy commands, it often works fine when I am testing it given it just jumps back to previous jump, so reproduction steps would be helpful.

It wouldn't be too difficult to implement behavior for those keys. I could take a look at some point if you want to open a new GitHub issue, mention my name in it, and mention a specific use case that is failing.

devguydavid:
I know this is merged, but I must be missing something glaringly obvious. I'm having a hard time seeing the effect of this change. Example:

* ctrl-] to go to definition of a symbol
* j j j k k k to navigate line by line a bit
* ctrl-o to go back

Expected result: The one-line motions are skipped and I jump back to the original position.
Actual result: The ctrl-o follows the one-line motions.

I put this comment here only for posterity and hopefully to solicit a final comment describing if there's something that needs to be done to get this working as expected (or to set my expectations straight). If this is a newly introduced bug I'd be happy to create an Issue.

shawnaxsom:
@devguydavid Your expectations are correct. I can't reproduce an issue like you describe, though there are a few similar issues with jump to definition.

If I jump to a definition in a separate file, I have to press ctrl-o twice to return to the previous file.

If I jump to definition in the same file, it doesn't seem to be adding the "from" line as a jump.

To confirm, pressing `ctrl-] j j j k k k`, you are having to press `ctrl-o` 6 times? I'm not sure why it would be doing that. Each "action" (like pressing "j") has a flag to determine if it should create a jump or not, and those have that disabled.

devguydavid:
That is correct. I'd have to type `ctrl-o` 6 times just as before this pull request. I have tried multiple versions of this plugin including the one this was introduced in and it's the same result for me.

Good news, though! This morning I removed all my extensions and settings:

mv ~/.vscode ~/.vscode.orig
mv ~/Library/Application\ Support/Code ~/Library/Application\ Support/Code.orig

I then installed this plugin and the Go extension (since I'm testing on some go code) and it is now working as expected. There must have been a setting or an extension that was conflicting. VSCode itself showed that 0.17.0 was being loaded, so it's still a mystery to me. I'll dig in a little more, but at this point I at least have a properly functioning state to work from.

shawn-pe:
Okay great! Yes, there are some minor issues still as mentioned above, and also some non-Vim commands like `Go To Symbol In File` can't easily be tracked as jumps, but I hope it works better for you now otherwise.

PR #3036, opened 2018-09-10
Fix: Missing bindings to exit CommandMode. closes #3035

### LINKED ISSUE #3035: Exiting CommandMode should mimic Vim behavior
When in Command Mode (`:`) you can exit to Normal mode with `<Esc>` & `<C-[>` wich are identical. Also `<C-c>` (which is conceptually different) aborts Command Mode and takes you to Normal Mode.
This exit key bindings are out-of-the-box in vim.

**Proposed Solution:**
As discussed in the slack channel with @shawnaxsom we need to add `<C-c>` and `<C-[>` to actions.ts, as alternatives to exit Command Mode

mxlian:
This fixes #3035, adding the missing key bindings to exit CommandMode:
* <C-[>  which is the same as <Esc>
* <C-c>

### DIFF
--- a/src/actions/commands/actions.ts
+++ b/src/actions/commands/actions.ts
@@ -1856,7 +1856,7 @@ class CommandInsertInCommandline extends BaseCommand {
 @RegisterAction
 class CommandEscInCommandline extends BaseCommand {
   modes = [ModeName.CommandlineInProgress];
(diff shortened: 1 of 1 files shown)


jpoon:
Merging this guy in for now. We can add the missing bindings in a later PR.

PR #3041, opened 2018-09-12
Fixed Jump Tracker jumps when jumping from a file that auto closes

### LINKED ISSUE #3039: New Jump Tracker doesn't always handle that isn't left open in a tab
**Describe the bug**

I can take this one. I wrote the new Jump Tracker. Now that it is released, I noticed a use case that is failing.

Sometimes jumps between files isn't working in a manner than you can return to the previous file. It seems that jumps between files can fail if the previous file isn't left open (i.e. `View: Keep Editor`). It seems to work fine if you close editors and `<C-o>`.

**To Reproduce**

1.  Close all editors.
2.  Open an editor (without using `View: Keep Editor`) with an import statement (something you can `gd` to outside the file).
3.  Jump to definition found in another file.
4.  Press `<C-o>`. The jump back probably won't work.

**Expected behavior**

`<C-o>` should result in returning to the original file.

- Extension (VsCodeVim) version: v0.16.4
- VSCode version: 1.27.0
- OS: macOS OS X Sierra (10.12.6)

shawnaxsom:

I implemented the new Jump Tracker that handles `<C-o>` behavior. @Chillee and I noticed a few issues when trying to jump from or to files that are auto closed.

One use case, if you close all editors, then open temporary editor, then jump to definition to open another editor (auto-closing the previous editor), `<C-o>` often doesn't work. vscode ends up sending two separate `vscode.window.onDidChangeActiveTextEditor` events, once for closing one editor, once for opening the new editor, where as we were expecting just one event.

Another similar use case from @Chillee has the same cause:

```If you create a new tab that isn't saved, and you just switch back to the original tab
`ctrl-o` gets you stuck switching between the 2 tabs

So you're in file A
you create a new unsaved file B
so now you're in B
you switch back to file A through a mouse click
`ctrl o` then brings you back to file B
and repeated `ctrl o` cycles between file B -> A -> B ....```

fixes #3039

### DIFF
--- a/extension.ts
+++ b/extension.ts
@@ -27,6 +27,7 @@ import { CompositionState } from './src/state/compositionState';
 const globalState = new GlobalState();
 let extensionContext: vscode.ExtensionContext;
 let previousActiveEditorId: EditorIdentity | null = null;
+let lastClosedModeHandler: ModeHandler | null = null;
 
 interface ICodeKeybinding {
   after?: string[];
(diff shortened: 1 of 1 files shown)


Chillee:
LGTM!

PR #3043, opened 2018-09-12
Change status of :new and :sp Command

ShivanshuKantPrasad:
This commands have been implemented in newer versions of VSCodeVim as VSCode started supporting complex split screens.
I tested it and it seems to be working fine.


Chillee:
I believe the original checkmarks were in reference to horizontal splitting, which is blocked on this PR: https://github.com/VSCodeVim/Vim/pull/2697

ShivanshuKantPrasad:
My bad i tried using those commands and they split the screen (vertically instead of horizontally) but i being new to vim thought those commands were working currently.

Issue #2141, opened 2017-11-13
support line and file completion


jab:
* Click *thumbs-up* 👍 on this issue if you want it!
* Click *confused* 😕 on this issue if not having it makes VSCodeVim unusable. 

The VSCodeVim team prioritizes issues based on reaction count. 

--------

**Is this a BUG REPORT or FEATURE REQUEST?** (choose one): FEATURE REQUEST

In stock Vim you can press `<C-x><C-L>` to complete the rest of the current line from all matching lines in all open buffers. You can also press `<C-x><C-F>` to complete the current path from matching paths on the filesystem. Might there be any interest in adding support here for these powerful features? Thanks for your consideration and for the great work on VSCodeVim!

**Environment**:

- **VSCode Version**: 1.18.0
- **VsCodeVim Version**: 0.10.2
- **OS**: macOS 10.12.6

jab:
ref: http://vim.wikia.com/wiki/Line/word/file/whatever_completion

jaanauati:
this is called ctrl-x mode: 
http://vimdoc.sourceforge.net/htmldoc/insert.html#i_CTRL-X
and it seems that it is also not working in neovim mode: ('vim.enableNeovim': true`)
thanks.

xlucas:
This is something I miss a lot particularly in documentation to complete sentences with similar constructs.

axs221:
I can take this one. I looked at a line completion solution this morning. I have line completion working by just going through lines of visible editors and matching startsWith of the current line. If performance is fine with many files open then hopefully that is all we need.

shawnaxsom:
Quick update after I discussed with @Chillee. The completion UI will probably look like this:

https://cl.ly/5526d0bc36df

I am assuming this will be fine for users. You can still search through that quick pick list, it is just unexpected UI.

The issue is that vscode doesn't appear to allow you to open a standard completion list at the cursor with just your own entries. You can register a CompletionItemProvider, and trigger the completion menu to open, but your completions are mixed with other completions in the context. If you try to do a line completion where the previous character before the cursor is a space, then you get a huge list of completions and it is then difficult to find the actual line completions.

shawnaxsom:
PR is up with #3048 for line completion. I may look at file completion after that is merged. It will likely behave similarly, using Quick Pick.

tehmasterer:
Is autocomplete for file paths (`C-X><C-F>`) still planned?

adamency:
[...]

PR #3048, opened 2018-09-13
Support for line completion (<C-x><C-l>)

shawnaxsom:

This adds support for Vim's line completion mode:

* http://vimdoc.sourceforge.net/htmldoc/insert.html#i_CTRL-X_CTRL-L

The line completion hotkey `<C-x><C-l>` opens up a Quick Pick menu. Standard autocompletion wasn't used because extensions like ours can't limit the suggestions to just our own. I tried implementing it that way at first, and results were overwhelming. Quick Pick has the added benefit of being able to do additional searching within the list, so I actually prefer it for this use case.

Completions come from vscode's open editors list. The current file is prioritized, and search starts from the cursor position.

See a video here:

* https://cl.ly/f32a1aa9871e

#2141 - Just the line completion aspect.

I realized in testing it is slightly different than vanilla Vim, as results closest to the cursor are prioritized, but I'm leaving it as is and am happy to hear feedback. Usually I'd change it to mirror Vim, though in this case I think that is a worse experience (I'd be more likely to use a completion just below the cursor than at bottom of the file).

### DIFF
--- a/extension.ts
+++ b/extension.ts
@@ -10,7 +10,6 @@ import * as vscode from 'vscode';
 
 import { configuration } from './src/configuration/configuration';
 import { commandLine } from './src/cmd_line/commandLine';
-import { lineCompletionProvider } from './src/completion/lineCompletionProvider';
 import { Position } from './src/common/motion/position';
 import { EditorIdentity } from './src/editorIdentity';
 import { Globals } from './src/globals';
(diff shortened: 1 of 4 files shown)


Chillee:
Alright, I looked through it roughly, and it seems to LGTM!

PR #3059, opened 2018-09-19
Record external commands that change cursor position (like gotoSymbol) as jumps

### LINKED ISSUE #3052: <c-o> Please count "Go to Symbol" actions as jumps
**Problem Description**

`<c-o>`'s jump-list does not count VS Code **"Go to Symbol..."** action as jumps. ("Go to Symbol..." is the action `workbench.action.gotoSymbol` which is activated by `Ctrl+Shift+o` or `Ctrl+P, input @`.)

[#3028](https://github.com/VSCodeVim/Vim/pull/3028) includes a lot of jump cases, except this one. Please add it. Thanks ever so!

shawnaxsom:

I implemented the new internal Jump Tracker with #3028. Previously we let vscode manage jump capabilities, which was too noisy, not ignoring cases vanilla Vim wouldn't consider a jump.

But as far as external events, the new Jump Tracker was only handling file change events. It didn't yet record jumps for external commands that change cursor position but not file, like `workbench.action.gotoSymbol`.

This PR will record jumps for external commands. It also tracks `cursorPositionAfterLastEvent` to help determine what events were caused by our own actions and should be ignored.

fixes #3052

I am assuming most external commands that change cursor position should be recorded. There are cases where this could get noisy, like if another plugin was to implement its own hotkeys for `<Up>` and `<Down>`, having us record jumps line-by-line. I chose not to filter out single-line jumps, assuming that scenario is unlikely, and that it is preferred to capture any `workbench.action.gotoSymbol` event, even if the symbol ended up being just one line away.


shawnaxsom:
Something is going on with the integration tests in this PR since I last rebased. They were working before, but now they are failing in the jump tracker. But taking the same actions manually testing seems okay still, so I'm not sure what is going on.

shawnaxsom:
I'm going to close out this PR, this approach is too error prone. Using `vscode.window.onDidChangeTextEditorSelection`, there isn't a reliable way to discern between vscode commands and our own commands. We can track line number after we update, or track whether we are running an action or not, but these are unfortunately unreliable given the async nature; sometimes vscode sends the event too late and we no longer think we are executing a command, or it sends the event too early and we don't know what the new line number is to expect from the event.

PR #3071, opened 2018-09-23
feature: relative, plus/minus ranges. closes #2384

### LINKED ISSUE #2384: Search and replace doesn't work with current line (.) and relative lines

**What happened**: Using . to indicate current line and +<n>s in search and replace doesn't work

**What did you expect to happen**: It does the replacing from current line until the specified number of lines

**How to reproduce it (as minimally and precisely as possible)**: Set cursor to a line, and then try to replace any text in the line and the n next lines by typing :.,+ns/search/replace/g

**Environment**:

- VSCode version: 1.20.1
- OS version: Ubuntu 16.04.2 LTS

captaincaius:
This PR fixes the current line ranges functionality and also adds the ability to do relative line ranges like vim (e.g. :.,.+2s/this/that/).  A few bugs were fixed along the way too (e.g. specifying a single-line only, etc).

#2384

First of all, you'll notice I tried to make it behave like vim even when in undocumented cases.  If anything seems a little weird, there'll be a comment with "undocumented:" nearby so you can try it for yourself on vim ;)

Also, a few things worth noting that i came across while working on this:

- ~~I decided to defer the shorthand feature in vim, because it's unrelated - i.e. 4:s/this/that becomes :.,.+3s/this/that~~ done in this PR - it turned out to be a 4-liner ;)
- ~~I think TokenType.LineNumber probably needs a different name, like LineNumberOrOffset ;) - or we can do some surgery on the lexer, but I don't think you folks would think that's worthwhile.~~ I updated the lexer - I decided it's clearer to understand this way, even though it's a little more code to read.  If anyone disagrees, I'll be happy to revert.
- Another separate feature that needs a followup issue is related to cursor position after ranged commands - I noticed we're not mimicking vim on this (neither on relative nor absolute ranges).
- I haven't looked around too much, but is it intentional that interpreting the range is delegated to the commands themselves?  For example, the single-line bug I found was fixed only in substitute, but it might also exist in other commands.
- Speaking of which, I left a bit of residue for "Percent" in lineRefToPosition - again, since it seems that responsibility is left to the commands, I don't think it was being used.  But I didn't want to touch it without discussing it first.


captaincaius:
Since nobody's had time to review this yet, I decided to take care of the third point I mentioned so it's a little cleaner within this PR :).  I took the second approach (improving the lexer so LineNumber and Offset are semantically distinct).  I'll update it to master and then push it.

src/cmd_line/lexer.ts
```diff
@@ -60,7 +60,18 @@ namespace LexerFunctions {
         case '7':
         case '8':
         case '9':
-          return lexLineRef;
+          if (tokens.length < 1) {
+            // special case - first digitey token is always a line number
+            return lexDigits(TokenType.LineNumber);
+          } else {
+            // otherwise, use previous token to determine which flavor of digit lexer should be used
+            let previousTokenType = tokens[tokens.length - 1].type;
```
xconverge: const

src/cmd_line/node.ts
```diff
@@ -57,20 +71,30 @@ export class LineRange {
...
     vimState: VimState
   ): vscode.Position {
-    var first = toks[0];
-    switch (first.type) {
-      case token.TokenType.Dollar:
+    let currentLineNum: number;
+    let currentColumn = 0; // only mark does this differently
+    let currentOperation: LineRefOperation = undefined;
+
+    var firstToken = toks[0];
```
xconverge: const

src/cmd_line/node.ts
```diff
@@ -79,21 +103,94 @@ export class LineRange {
...
+      case token.TokenType.Minus:
+        currentLineNum -= 1;
+        currentColumn = 0;
+        break;
+    }
+
+    // finally, make sure current position is in bounds :)
+    currentLineNum = Math.max(0, currentLineNum);
+    currentLineNum = Math.min(doc.document.lineCount - 1, currentLineNum);
+    return new vscode.Position(currentLineNum, currentColumn);
```
xconverge: can you somehow use isValid in position.ts

captaincaius: I just looked and isValid won't tell why it's not valid (<0 vs >MAX).  I looked for something else in there that will push it in-bounds, but didn't find anything.

In reality, this was here before this PR - vim doesn't actually do this (it throws an "E16: Invalid range" instead).  I wasn't sure why this code was here, but I figured whoever put it there knew something I don't about vscode/this extension/etc, and changing this behavior would have been risky and out of the scope of this issue (and already hard-to-review PR), so I left it alone.

LMK if you want me to go the error route instead within this issue or defer this or let sleeping dogs lie.

src/cmd_line/lexer.ts
captaincaius: done

src/cmd_line/node.ts
captaincaius: done

PR #3078, opened 2018-09-28
Feature/insert mode optimizations

shawnaxsom:

There are many users reporting performance issues, especially typing in insert mode, though this is confounded by having seemingly different causes. For @cmcclellen and myself, we tested each piece of code in the override to the keys command. We found that:

Overall time per insert mode key press was 60ms

1. setContext for vim.mode took around 35ms of that, or around 60% of the total
2. runAction took around 13ms, or around 21% of the total, with 8 seconds handling insertTextVSCode
3. Some ms was also caused during async/await, though we weren't sure what else was on the event loop causing delay

This PR addresses (1) setContext above. We were sending setContext on every key press, but it only needs to be sent when the vim mode has changed.

#2021
#2216

This will help many users that are experiencing performance issues in #2021 and #2216, though there seem to be other causes as well. It had an impact for two users, @cmcclellen and @samueltlg, as discussed in #2021, making it fast enough for the extension to be usable for at least @cmcclellen. There will need to be more investigation and optimizations before marking those issues as closed.


jpoon:
Thanks @shawnaxsom! Awesome investigative work.

PR #3081, opened 2018-09-29
Feature/fix black hole operator mappings

### LINKED ISSUE #2672: Remapping `d` to always delete to black-hole
I used to have the following in my `.vimrc`:

```
" Redirect all delete-operations to black-hole
" Use 'm' ("move") for all cut-operations
nnoremap d "_d
nnoremap m d
```

What would be the VSCodeVim equivalent? I can't seem to make it work, I already tried the following:

```
    "vim.otherModesKeyBindings": [
        { "before": ["d"], "after": ["\"", "_", "d"] }
    ],
        { "before": ["m"], "after": ["d"] },
        { "before": ["m", "m"], "after": ["d", "d"] }
    ]
```

This makes the bindings for `m` work, but breaks `d`...

- Extension (VsCodeVim) version: v0.12.0
- VSCode version: 1.23.1
- OS version: Mac OS 10.13.4

shawnaxsom:

It is somewhat common in Vim to remap keys to copy to the black hole register, when the user doesn't want to modify the default register. In our current remapping system, this wasn't supported with operators when the operator was a repeated key press (e.g. `dd` with `"_d` mapped).

Vanilla Vim seems to disregard key mappings if an operator is in progress. `dd` with `"_d` mapped should result in `"_dd`, not `"_d"_d`. So this change seems in line with vanilla as far as that goes.

Therefore, with this PR, the following vscode setting should allow e.g. `dd` and `dw` to become `"_dd` and `"_dw`, thereby yanking to the black hole register and not overwriting the default register or the system clipboard.

```
{
        {
            "before": ["d"],
            "after": [ "\"", "_", "d" ]
        }
    ]
}
```

fixes #2672

Note that there are still issues with mapping, such as if you want to map both `d` and `dd`. But those don't appear to be regressions. Vanilla Vim handles the following scenarios:

```Mapping "d" should change the first "d" in "dw"
Mapping "w" does NOT affect the "w" in "dw"
Mapping "dd" should change "dd"
Mapping "d" and "dd", it uses "dd" remap if "dd" is pressed, otherwise it waits for a timeout to do "d" remap
Mapping just "d", it realizes there are no mappings that involve additional keystrokes, so it applies "d" remap immediately
```

It seems we would need to e.g. revamp how the remapper timeouts work in ModeHandler if we wanted to handle these. I could open a separate GitHub issue to tackle that.

### DIFF
--- a/src/actions/commands/actions.ts
+++ b/src/actions/commands/actions.ts
@@ -1872,7 +1872,7 @@ class CommandInsertInCommandline extends BaseCommand {
 @RegisterAction
 class CommandEscInCommandline extends BaseCommand {
   modes = [ModeName.CommandlineInProgress];
-  keys = [['<Esc>'], ['<C-c>'], ['<C-[>']];;
+  keys = [['<Esc>'], ['<C-c>'], ['<C-[>']];
   runsOnceForEveryCursor() {
     return this.keysPressed[0] === '\n';
   }
--- a/src/mode/modeHandler.ts
+++ b/src/mode/modeHandler.ts
@@ -321,6 +321,7 @@ export class ModeHandler implements vscode.Disposable {
        */
       if (
         !this.vimState.isCurrentlyPerformingRemapping &&
+        !this.vimState.recordedState.operator &&
         (withinTimeout || this.vimState.recordedState.commandList.length === 1)
       ) {
         handled = await this._remappers.sendKey(
--- a/test/configuration/remapper.test.ts
+++ b/test/configuration/remapper.test.ts
@@ -305,6 +305,24 @@ suite('Remapper', () => {
     actual = await Register.get(vimState);
     assert.equal(actual.text, expected);
   });
+
+  test('d -> black hole register delete in normal mode through modehandler', async () => {
+    assert.equal(modeHandler.currentMode.name, ModeName.Normal);
+
+    await modeHandler.handleMultipleKeyEvents(['<Esc>', 'g', 'g']);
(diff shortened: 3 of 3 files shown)


escamoteur:
I just tried this, but and it works if I press `dd`. But then I press `4d` it too is moved to the black hole. 
So I still can't map `dd`

berknam:
> 
> 
> I just tried this, but and it works if I press `dd`. But then I press `4d` it too is moved to the black hole.
> So I still can't map `dd`

This is an old PR. If you're using version 1.17+ then the remapper has been overhauled. Explain what you want to do. If you're trying to do something that works in normal VIM but doesn't with VSCodeVim then open an issue.

If you're trying to make all deletes using key `d` delete to black hole register then you can use the mapping showed above:
```jsonc
    {
        "before": ["d"],
        "after": [ "\"", "_", "d"]
    }
]
```

This will make all deletes in normal mode using 'd' delete to black hole register, like `dw`, `4dw` or `dd`. If you just want to change the `dd` for when deleting an entire line then you can do:
```jsonc
    {
        "before": ["d", "d"],
        "after": [ "\"", "_", "d", "d"]
    }
]
```
If instead you want to make the delete with 'd' go to black hole register except when deleting an entire line then you can do it like this:
```jsonc
    {
        "before": ["d"],
        "after": [ "\"", "_", "d"]
    },
    {
        "before": ["d", "d"],
        "after": [ "d", "d"]
    }
]
```

That last one is just to make sure that `dd` uses the normal `d` instead of the remapped `d`.

PR #3087, opened 2018-10-02
Multiline yank writes to 0 register; fixes #1214

### LINKED ISSUE #1214: numbered register 0 is not updated in Y
### What did you do?
yank a line with
```Y``` instead of ```yy```

The numbered register ```"0``` is updated.

### What happened instead?
The ```"0``` is not updated. 
Note: The ```"0``` works for ```yy``` though.

### Technical details:

* VSCode Version: 1.8.1
* VsCodeVim Version: 0.4.10
* OS: Win-7 x64

### PR DESCRIPTION
**Pull request by JKillian**
Makes sure that multiline yank commands (`5yy`, `5Y`, `V%y`) write to the `0` register.

#1214

See inline comment below for more explanation of this change

### DIFF
--- a/src/register/register.ts
+++ b/src/register/register.ts
@@ -302,9 +302,7 @@ export class Register {
    */
   private static processNumberedRegister(content: RegisterContent, vimState: VimState): void {
     // Find the BaseOperator of the current actions
-    const baseOperator = vimState.recordedState.actionsRun.find(value => {
-      return value instanceof BaseOperator || value instanceof BaseCommand;
-    });
+    const baseOperator = vimState.recordedState.operator || vimState.recordedState.command;
 
     if (baseOperator instanceof YankOperator || baseOperator instanceof CommandYankFullLine) {
       // 'yank' to 0 only if no register was specified
--- a/src/state/recordedState.ts
+++ b/src/state/recordedState.ts
@@ -126,7 +126,7 @@ export class RecordedState {
    * The command (e.g. i, ., R, /) the user wants to run, if there is one.
    */
   public get command(): BaseCommand {
-    const list = _.filter(this.actionsRun, a => a instanceof BaseCommand);
+    const list = _.filter(this.actionsRun, a => a instanceof BaseCommand).reverse();
 
     // TODO - disregard <Esc>, then assert this is of length 1.
 
--- a/test/register/register.test.ts
+++ b/test/register/register.test.ts
@@ -105,6 +105,49 @@ suite('register', () => {
     assertEqualLines(['test2', 'test2', 'test3']);
   });
 
+  test("Multiline yank (`[count]yy`) stores text in Register '0'", async () => {
+    modeHandler.vimState.editor = vscode.window.activeTextEditor!;
+
+    await modeHandler.handleMultipleKeyEvents('itest1\ntest2\ntest3'.split(''));
+
+    await modeHandler.handleMultipleKeyEvents([
+      '<Esc>',
+      'g',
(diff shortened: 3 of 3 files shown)

### CODE REVIEW

File: src/register/register.ts [hunk 1]
```diff
@@ -302,9 +302,7 @@ export class Register {
    */
   private static processNumberedRegister(content: RegisterContent, vimState: VimState): void {
     // Find the BaseOperator of the current actions
-    const baseOperator = vimState.recordedState.actionsRun.find(value => {
-      return value instanceof BaseOperator || value instanceof BaseCommand;
-    });
```
JKillian: Previously, this code was initializing `baseOperator` to be `CommandNumber` in some cases. For example, `5Y` is `[CommandNumber, CommandYankFullLine]`, and because [`CommandNumber` inherits from `BaseCommand`](https://github.com/VSCodeVim/Vim/blob/bd153ad9a8e7b3a445ebcbd596d7eeaf994e92a4/src/actions/commands/actions.ts#L284) and comes first, `baseOperator` would be set to that.

Then, the blocks below specifically to handle yanks wouldn't be hit.

https://github.com/VSCodeVim/Vim/blob/bd153ad9a8e7b3a445ebcbd596d7eeaf994e92a4/src/actions/commands/actions.ts#L284

src/state/recordedState.ts
```diff
@@ -126,7 +126,7 @@ export class RecordedState {
    * The command (e.g. i, ., R, /) the user wants to run, if there is one.
    */
   public get command(): BaseCommand {
-    const list = _.filter(this.actionsRun, a => a instanceof BaseCommand);
+    const list = _.filter(this.actionsRun, a => a instanceof BaseCommand).reverse();
```
JKillian: As far as I could tell, this `get command()` isn't actually being used anywhere yet, so this should be a safe change.

However, the comment below brought me pause: 

> // TODO - disregard <Esc>, then assert this is of length 1.

If there is only supposed to be one `BaseCommand`, how come `5Y` is two commands? Not familiar with the VSCode Vim codebase, but just wanted to mention this in case it matters.

xconverge: Hmm, what are your thoughts on handling this in the processNumberedRegister call instead? Changing this seems a little bit of a workaround since the get() is currently returning the stack the way it was executed (which seems OK to me)

Not too sure on the TODO comment :/

xconverge:
Thanks! This one actually bugs me a lot!

src/state/recordedState.ts
JKillian: Well, since nobody else seems to be calling `get command()`, I guess we can define it however we wish. To me, returning the last command seems to make more sense - for example if I ran `5@q` I'd expect `.command` to refer to the macro execution, `CommandExecuteMacro` as opposed to `5`, `CommandNumber`.  Or for something like `"aD`, I'd expect `.command` to refer to the deletion `D` command, `CommandDeleteToLineEnd`, instead of the register `"a` `CommandRegister` part.

That said, you know the code better than I do, so happy to do it either way in the end! Both ways will accomplish the same purpose and get this working better which is what matters 😃

xconverge: Interesting looks like the getter for operator is reversed, lets just go with this!

JKillian:
Thanks @xconverge! As a sidenote, I found this codebase really easy to jump into and start working on, so big 👍 for the simple code and organization! (And also for TypeScript which helps so much haha)

xconverge:
👍 ignore that travisbuddy warning, I merged in the middle of a build

PR #3095, opened 2018-10-04
fix - cursor position whenever changing editors - closes #2688

### LINKED ISSUE #2688: Wrong cursor position after using same file in two panels

captaincaius:
Updates vimState whenever the active editor changes so that changing splits doesn't have you making edits to the wrong line

#2688 

I hand-tested it, and it seems to solve the problem, and undo seems to work properly too.  But there's some questions:

1. I see some branches in the selection change handler for "easymotion" ... TBH I have no idea what that is, so I'm not sure this patch will do "the right thing" in situations involving it.
2. ~~Likewise with the jump tracker.~~ thanks to @shawnaxsom 's teaching me how this works, it looks like the jump tracker is A-Okay.
3. Is this testable at all?  I grepped for ":sp" in test/ and got no results :(.
4. Are there any potential race conditions introduced by changing vimState.editor right in the event handler (outside the task queue)?  I put it there because the selection handler checks it outside the queue, so putting it inside would very likely introduce a race condition.
5. Disclaimer - I have zero experience with the VSCode API and the timing of its events - so review this as though a child in a playground wrote it (a child that <3s vscodevim :P).  I ran tests twice and they don't seem to flake, but I'm sure you folks have a lot more experience with any gotchas in their APIs, especially with regard to timing, synchronization, promise resolution, etc.

### DIFF
--- a/src/mode/modeHandler.ts
+++ b/src/mode/modeHandler.ts
@@ -112,8 +112,42 @@ export class ModeHandler implements vscode.Disposable {
       }
     );
 
+    const onChangeActiveTextEditor = vscode.window.onDidChangeActiveTextEditor(e => {
+      if (configuration.disableExt) {
+        return;
+      }
+
+      if (Globals.isTesting) {
+        return;
+      }
+
+      if (e !== undefined) {
+        this.vimState.editor = e;
+        taskQueue.enqueueTask(
+          () => this.handleActiveEditorChange(e),
+          undefined,
+          /**
+           * We don't want these to become backlogged! If they do, we'll update
+           * the selection to an incorrect value and see a jittering cursor.
+           */
+          true
+        );
+      }
+    });
+
     this._disposables.push(onChangeTextEditorSelection);
+    this._disposables.push(onChangeActiveTextEditor);
     this._disposables.push(this.vimState);
+  }
+
+  private async handleSelectionChange(e: vscode.TextEditorSelectionChangeEvent): Promise<void> {
+    let onlyUpdateInternal = e.kind !== vscode.TextEditorSelectionChangeKind.Mouse;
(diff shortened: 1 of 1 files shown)


xconverge:
Did you look at onDidChangeActiveTextEditor in extension.ts? Could this have been handled there or in updateView? I am just hesitant because you may have noticed the big comment above the function you changed

Regarding easymotion, https://github.com/VSCodeVim/Vim#vim-easymotion

captaincaius:
Thanks for taking the time to review!

Dang, I did a search only in src/ for onDidChangeActiveTextEditor, so I didn't see it.  Thanks!  I agree that they should be combined.

I'm assuming it was in extension.ts for a good reason, so I'll augment the original handler instead and maybe make the handleActiveEditorChange call there.

Yup I noticed the big comment, and also how delicately the function seemed to be written, so I didn't want to change it from the point of view of the selection change event.

The selection change event isn't fired on a text editor change (even on a mouse click), so I broke down its input from the event into its constituents so that it could be used by both event handlers.

I'll give this some more careful thought and give it another stab later today when I have open-source-time :).

captaincaius:
@xconverge thanks for the point in the right direction!  You were right - we can avoid touching even the interface of that function, because most of what's being done this way was previously being done in the existing handler, besides a little bug introduced in the undo fixes :).

I'm gonna close this and open a new PR to keep the commit log clean.

Thanks again! :)

flowchartsman:
@captaincaius looks like a similar issue persisting with #3455, just to put it on your radar, since you were able to issue this fix

PR #3103, opened 2018-10-05
fix (simpler) - cursor whenever changing editors - closes #2688

### LINKED ISSUE #2688: Wrong cursor position after using same file in two panels

captaincaius:
ModeHandler.syncCursors() and updateView() stopped being called on active editor changes after the bugfix to EditorIdentity because a window change wasn't being detected as a window change when the filename-only identity is identical :D.

getAndUpdateModeHandler() is called in a ton of places, so this PR adds a force-sync option, and has the ActiveEditorChange handler use it.

#2688

This PR restores the behavior to how it was before.

I took a wrong turn on another PR, but because of that, I noticed that there's some differences between selection change handling and editor switches.  It's pretty clear from just the length of it ;), but for example, EOL-checking seems to behave slightly different when switching editors than when staying in the same editor.  It's probably outside the scope of this issue, but I thought I'd mention it.

### DIFF
--- a/extension.ts
+++ b/extension.ts
@@ -34,7 +34,7 @@ interface ICodeKeybinding {
   commands?: { command: string; args: any[] }[];
 }
 
-export async function getAndUpdateModeHandler(): Promise<ModeHandler> {
+export async function getAndUpdateModeHandler(forceSyncAndUpdate = false): Promise<ModeHandler> {
   const activeEditorId = new EditorIdentity(vscode.window.activeTextEditor);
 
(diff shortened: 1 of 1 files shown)


xconverge:
This looks pretty straight forward to me!

captaincaius:
it was your idea after all ;) thanks again!

PR #3108, opened 2018-10-06
Fix issues with keybindings when changing to an editor in different mode

### LINKED ISSUE #730: Pressing tab in insert-mode inserts the string `<tab>`
### What did you do?

Put the exact keys you pressed. `i[TAB]`

I expected the two spaces I have configured to be inserted.
### What happened instead?

The string `<tab>` was inserted.
### Technical details:

Running ionide-fsharp, latest version
- VSCode Version: 1.4.0
- VsCodeVim Version: latest, updated today
- OS: OS X 10.11.6

screen shot 2016-09-07 at 09 57 19

screen shot 2016-09-07 at 09 24 14

### LINKED ISSUE #3096: version 0.16.6 cause <tab> key insert string for unknown reason
**To Reproduce**

I don't know how to reproduce it.  It insert \<tab\> string when press tab key. BTW, Anyone knows how to install an older version?

screen shot 2018-10-04 at 6 58 04 pm

1.  Go to '...'
2.  Click on '....'
3.  Scroll down to '....'
4.  See error

**Expected behavior**

**Screenshots**

- VSCode version:
- OS:

**Additional context**

shawnaxsom:

Fix issues with keybindings when changing to an editor in different mode. This resulted in keybindings not working until changing modes again, including seeing `<Tab>` inserted as text in insert mode, if vscode thought we were in normal mode still.

fixes #3096 
fixes #730 
fixes https://github.com/Microsoft/vscode/issues/59925

I'm not exactly sure why we don't have tests between editors. It would be nice to add tests here. I know that in the onDidChangeActiveTextEditor handler we even return automatically if `Globals.isTesting` is true.

Anyway, there are two main changes here:

1. "setContext vim.mode" is called when switching editor windows (previously this wasn't getting called until pressing keys in the new editor window).
2. I also made _previousMode static, since setContext for vim.mode applies across editors (along with renaming to _lastVimModeSetForKeybindings to clarify purpose).

### DIFF
--- a/extension.ts
+++ b/extension.ts
@@ -187,6 +187,8 @@ export async function activate(context: vscode.ExtensionContext) {
       if (vscode.window.activeTextEditor !== undefined) {
         const mh: ModeHandler = await getAndUpdateModeHandler();
 
+        await mh.updateVimModeForKeybindings(mh.vimState.currentMode);
+
         await mh.updateView(mh.vimState, { drawSelection: false, revealRange: false });
 
         globalState.jumpTracker.handleFileJump(
--- a/src/mode/modeHandler.ts
+++ b/src/mode/modeHandler.ts
@@ -37,8 +37,13 @@ import { Jump } from '../jumps/jump';
 export class ModeHandler implements vscode.Disposable {
   private _disposables: vscode.Disposable[] = [];
   private _modes: Mode[];
-  private _previousMode: ModeName;
   private _remappers: Remappers;
+
+  /**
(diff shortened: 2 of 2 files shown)


shawnaxsom:
Yeah it is working in the test cases I've tried, changing tabs and opening new tabs in various mode states. This fixes the reported issue and I'm not aware of any related issues. There is an open issue with <Tab> inserted with replace mode, but that isn't the same root cause as far as I can tell.

PR #3116, opened 2018-10-08
Save search history to a file like commandline history

### LINKED ISSUE #3098: Previous searches are not saved across sessions
I press slash and want to edit or use previous searches. 
In vi or other emulators I press slash then using the UP / DOWN keys I can navigate to previous search terms.

The solution I expect is to save previous search terms and allow me to navigate to them. I should be able to edit it and execute it.
vi-search

xconverge:
fixes #3098


src/state/searchStateHistory.ts
```diff
@@ -0,0 +1,113 @@
...
+    });
+  }
+
+  private _loadFromFile() {
+    let data = '';
+
+    try {
+      data = fs.readFileSync(this._searchHistoryFilePath, 'utf-8');
+    } catch (err) {
+      if (err.code === 'ENOENT') {
```
jpoon: A recall writing code similar to this. Did you copy this code from somewhere else in this codebase?

jpoon: Ah, it was in https://github.com/VSCodeVim/Vim/blob/master/src/cmd_line/commandLineHistory.ts#L84, can we share the logic? Subclasses?

xconverge: Yeah I'll subclass

src/util/historyFile.ts
```diff
@@ -65,15 +67,15 @@ export class CommandLineHistory {
...
+          `Failed to create directory. path=${this._historyDir}. err=${err}.`
         );
         reject(err);
       }
 
       try {
         fs.writeFileSync(this._historyFilePath, JSON.stringify(this._history), 'utf-8');
       } catch (err) {
-        logger.error(`CommandLineHistory: Failed to save history. err=${err}.`);
+        logger.error(`Failed to Failed to save history. path=${this._historyDir}. err=${err}.`);
```
jpoon: Copy pasta error.

xconverge: Hmm impressive how I managed this

PR #3123, opened 2018-10-10
Convert synchronous funcs to async

kylecarbs:
Converts the extension to be entirely asynchronous. There were many unhandled asychronous functions that contained synchronous calls, which allowed tests to pass. Continuing with this could cause issues in the future.

None that I've found.

I'm not super familiar with this codebase so I'd be happy to hear feedback. I don't really enjoy the `load` function added to `historyFile.ts`.


jpoon:
Some of these places make sense, but I think we need to be a bit more tactical about where we `await` for things as it has perf implications.

kylecarbs:
All these changes await for funcs that were handled through executed some synchronous action (like fs.existsSync), so it should have similar perf as it was prior (its possible an async func slipped through).

I'd be happy to hear your suggestions.

kylecarbs:
Any updates on this?

src/actions/commands/actions.ts
```diff
@@ -140,7 +140,7 @@ export class DocumentContentChangeAction extends BaseAction {
         .add(last.positionDiff);
     }
 
-    vimState.currentMode = ModeName.Insert;
+    await vimState.setCurrentMode(ModeName.Insert);
```
kylecarbs: T

src/actions/commands/actions.ts
```diff
@@ -2367,7 +2367,7 @@ class CommandExitVisualMode extends BaseCommand {
   keys = ['v'];
 
   public async exec(position: Position, vimState: VimState): Promise<VimState> {
-    vimState.currentMode = ModeName.Normal;
+    await vimState.setCurrentMode(ModeName.Normal);
```
kylecarbs: LGTM, will this have any perf implications?

src/util/historyFile.ts
```diff
@@ -59,21 +59,23 @@ export class HistoryFile {
     }
   }
 
-  public async save(): Promise<void> {
-    return new Promise<void>((resolve, reject) => {
+  public save(): Promise<void> {
+    return new Promise<void>(async (resolve, reject) => {
```
jpoon: We can get rid of this `Promise` wrapper.

src/mode/modeHandlerMap.ts
```diff
@@ -8,7 +8,16 @@ class ModeHandlerMapImpl {
     let modeHandler = this.modeHandlerMap[key];
     if (!modeHandler) {
       isNew = true;
-      modeHandler = await new ModeHandler();
+      await new Promise((res, rej) => {
```
jpoon: Why was this necessary?

kylecarbs: `ModeHandler` had async calls in its constructor that were not handled. I added the callback as an _initializer_ of sorts. I'm not a big fan of this method. We could add a static function and make the constructor private?

Issue #1951, opened 2017-08-04
Selecting text backwards in insert mode only selects first character


rohitpaulk:
* Click *thumbs-up* 👍 on this issue if you want it!
* Click *confused* 😕 on this issue if not having it makes VSCodeVim unusable. 

The VSCodeVim team prioritizes issues based on reaction count. 

--------

**Environment**:

- **VSCode Version**: 1.14.2
- **VsCodeVim Version**: 0.9.0
- **OS**: macOS Sierra

**How to reproduce it**:

- Enter insert mode
- Type out a few characters
- Select those characters, by pressing `shift + left arrow` multiple times. 
- Copy using `cmd + c`
- Paste elsewhere using `cmd + v`.

![vimselect](https://user-images.githubusercontent.com/3893573/28972083-f866541c-794b-11e7-9039-297a228cd794.gif)

**What happened**:

- With the vim extension active, only the first character of the selection is copied. 
- Without the vim extension active, the entire selection is copied.

**What did you expect to happen**: 

I'd expect the entire selection to be copied, just like when the vim extension isn't active.

chunkeater:
Same issue on both Mac and Windows since months ago.. I need to retrain my workflow to not to backwards copy, which is pretty disruptive.  Any workarounds (without uninstalling VscodeVim) will be helpful, or information on the source code portions to be fixed and I could help. thank you! :)

tomasiser:
The issue was fixed in insert mode, but not in normal mode. Should a new issue be made about it?

flowkraD:
Is this still a bug or is it fixed now? If it's not fixed, I'll love to work on this!

PR #3124, opened 2018-10-10
[Fixed] Reverse selection in normal mode

flowkraD:
**What this PR does / why we need it**: This takes care of problems faced during reverse selection where only the first letter of the codeblock used to be selected and the rest was ignored.

**Which issue(s) this PR fixes** #1951 

**Special notes for your reviewer**: This seemed to be an issue with way getText() works on normal mode where it starts from the back and tries to move up finding where the start is and if it cannot find the start it returns only the first character as we began from the Right of stop. 
The fix implemented just emulates a forward select when backward selection is made by the user and effectively works like a forward select.


src/actions/commands/actions.ts
```diff
@@ -1010,8 +1010,10 @@ class CommandOverrideCopy extends BaseCommand {
         .map(range => {
           const start = Position.EarlierOf(range.start, range.stop);
           const stop = Position.LaterOf(range.start, range.stop);
-          return vimState.editor.document.getText(new vscode.Range(start, stop.getRight()));
-        })
+          return start < stop
```
jpoon: The docs for `EarlierOf` state:

```
 /**
   * Returns which of the 2 provided Positions comes earlier in the document.
   */
```

so start *should* always be >= stop. 

What's the case when `start < stop`?

flowkraD:
I think i read it wrong, and thought `EarlierOf` returned the position from where the selection started. Sorry for the inconvenience due to my misunderstanding.

PR #3129, opened 2018-10-11
Add yank highlighting

### LINKED ISSUE #2333: Flash on operate

Feature request

**What happened**:

yanked with no visual feedback

yank with visual feedback

https://github.com/t9md/atom-vim-mode-plus/wiki/GIFs#flash-on-operate

Context I'm coming from Atom but have configured neovim to do the same. Trivial I realize but I really like this feature.

**Environment**:

- VSCode version: 1.19.3
- OS version: Mac 10.13.2

### LINKED ISSUE #2916: Implement vim-highlightedyank
This is the extension I miss the most from Vim. If you accidentally copy the wrong text you'll only know it after you have put it to somewhere.

The idea is to temporarily change to background color of the selection which was yanked to make clear which text actually was yanked.

Here's the original plugin https://github.com/machakann/vim-highlightedyank

Also the SimpleVim vscode plugin implements it https://github.com/jpotterm/vscode-simple-vim#settings

### LINKED ISSUE #2991: word/line highlight when yy or shift + * is pressed
Currently when I try to yank a word or line, there is no visual cue on the editor showing that the word/line is being yanked. Same as when I am trying to search a word with `shift+*`. It would be nice to have some sort of visual cue when these actions are performed.

I understand that in vanilla vim there isn't any visual cue either when the above actions are performed, but as someone coming from sublime + vintageous and contemplating switching the VSCode, this nice feature was available in sublime + vintageous (they did it with one rectangle box flash around the word/line) and it will be a huge plus to have this in VSCode.

darfink:
This PR implements yank highlighting using decorations.

Partially fixes #2991
Fixes #2333
Fixes #2916

This is a quickly done but fully functional draft of yank highlighting, there are some things to note:

- ~~Enabled by default since I believe it's a discrete a QoL change.~~
- ~~There is no conditional around the logic, since it's a low-cost operation and full transparency can be configured to negate it (an alternativ may be to allow null for the option?).~~
- It creates a unique decoration instance for each yank operation, which allows highlighting when using multiple cursors.

This may be a sub-optimal implementation. Preferably `setDecorations` would only be called once when using multiple cursors, but I could not see an easy way to implement this. Therefore I couldn't use a static instance of `YankHighlight` so I instead opted to adding a method instead. Also the introduction of a decoration inside an operator may be considered a bit impure.


src/configuration/decoration.ts
```diff
@@ -24,4 +25,13 @@ export class Decoration {
   static readonly EasyMotion = vscode.window.createTextEditorDecorationType({
     backgroundColor: configuration.searchHighlightColor,
   });
+
+  static yankHighlight(vimState: VimState, range: vscode.Range[]) {
```
jpoon:
```suggestion
  static YankHighlight(vimState: VimState, range: vscode.Range[]) {
```

src/configuration/decoration.ts
```diff
@@ -24,4 +25,13 @@ export class Decoration {
...
+    const decoration = vscode.window.createTextEditorDecorationType({
+      backgroundColor: configuration.yankHighlightColor,
+    });
+
+    vimState.editor.setDecorations(decoration, range);
+    setTimeout(() => decoration.dispose(), 200);
```
jpoon: Interesting. Is this dispose necessary? As per the documentation:

```
		/**
		 * Remove this decoration type and all decorations on all text editors using it.
		 */
```

Doesn't sound like that's what we want.

jpoon: Can we do the actual `setDecorations` outside of this? As it stands, this class is primarily creating the decorationType and it's the responsibility of the caller to actually perform the set.

AckmanDESU:
This could be expanded upon to highlight other blocks of text which have been modified such as indentation changes.

nomad-software:
Can changes such as these be optional, please? These do not occur in Vim.

epeli:
Woo! Thank you! Just build this PR for testing and it seems to be working as expected. I will add this to my work machine on monday for more testing.

But I do agree @nomad-software that this should have an option to disable it.

iovis9:
I'd love to see this PR merged! And I agree it should be opt in behaviour

epeli:
I've been using this for two months and I can confirm it works well.

src/configuration/decoration.ts
darfink: Regarding the call to `dispose`; preferably no. The problem occurs when using multiple cursors.

When each cursor performs a yank operation there are multiple affected ranges. The best case scenario would be to aggregate all these ranges and call `setDecorations` only once (with a static `YankHighlight` instance), but I could not find an applicable place to implement this logic.

Therefore I could only see two alternatives:
- Use a static `YankHighlight` decorator instance and call `setDecorations` for each cursor. The problem with this approach is that each cursor replaces the previous cursor's decoration (since the instance is reused), and therefore only highlights the last yank operation.
- Create a unique `YankHighlight` decorator instance for each cursor, disposing it (i.e removing the highlight) after a specific timeout has passed (the current approach).

But if there is a reasonable way to aggregate all affected ranges and apply the `setDecorations` logic once all cursors' yank operation has finished executing I'd gladly implement it.

src/actions/operator.ts
```diff
@@ -278,6 +280,12 @@ export class YankOperator extends BaseOperator {
       text = text + '\n';
     }
 
+    if (configuration.yankHighlighting) {
+      const decoration = Decoration.YankHighlight;
+      vimState.editor.setDecorations(decoration, [range]);
+      setTimeout(() => decoration.dispose(), 200);
```
jpoon: make the 200 configurable?

README.md
```diff
@@ -122,6 +122,8 @@ These settings are specific to VSCodeVim.
...
 | vim.visualstar                   | In visual mode, start a search with `*` or `#` using the current selection                                                                                                                                                                                                                                                                                                                                                       | Boolean | false                    |
+| vim.yankHighlighting             | Enable highlighting when yanking                                                                                                                                                                                                                                                                                                                                                                                                 | Boolean | false                    |
```
jpoon: Can we add these under their own sections?

```
vim.highlightedyank.enable
vim.highlightedyank.color
vim.highlightedyank.duration
```

You can see `vim.debug` for reference.

src/configuration/decoration.ts
jpoon: Ah, sorry didn't realize that we wanted the `yank` to disappear. I think the way it is at present is fine.

asfaltboy:
Dear @jpoon any chance you could give this PR another look? It seems all the last requests were resolved in a1e309eaa5372f2b7c68b065586c6c938b6b86f9. Is there any more work / review / testing required before we can merge this?

jpoon:
Aside from the merge conflicts, LGTM.

ragavkumarv:
Hi @darfink ,
Kindly requesting to rebase and resolve conflicts. It would we awesome to see the feature in the coming release.
Thanks and Regards
Ragav

epeli:
I resolved the merge conflicts here https://github.com/VSCodeVim/Vim/pull/3593

jpoon:
Replaced with https://github.com/VSCodeVim/Vim/pull/3593

Issue #2752, opened 2018-06-19
Multi-Cursor + insertModeKeyBinding jk -> <Esc>


vivlim:
**Is this a BUG REPORT or FEATURE REQUEST?** (choose one): BUG REPORT

**What happened**: When using multi-cursor & jk to exit insert mode, the first cursor is out of alignment with the others.

**What did you expect to happen**: All of the cursors should be in the same position & behave the same.

**How to reproduce it (as minimally and precisely as possible)**:

1. Create multiple cursors vertically
2. Enter insert mode & type some text
3. type "jk" to exit insert mode
![misaligned](https://user-images.githubusercontent.com/1565930/41618146-2ee5a234-73b7-11e8-92cb-42abc0405673.gif)

User settings:
```
...
    "vim.insertModeKeyBindings": [{
        "before": ["j", "k"],
        "after": ["<Esc>"]
    }],
...
```
keybindings.json
```
    {
        "key":"ctrl+k",
        "command":"editor.action.insertCursorAbove",
        "when": "editorTextFocus"
    },
    {
        "key":"ctrl+j",
        "command":"editor.action.insertCursorBelow",
        "when": "editorTextFocus"
    }
```

**Environment**:

- Extension Name: vim
- Extension Version: 0.13.0
- OS Version: Windows_NT x64 10.0.17134
- VSCode version: 1.24.1

shawjak3:
I'm also experiencing this issue on Linux using the latest vscode.

nerdo:
Ditto. This is the only thing keeping me from using vscodevim. I heavily rely on multiple selections and vim. For now I will stick with amVim.

spacewaffle:
This issue is still happening if the cursors are on the same line and they are not on the end of the line
![Mar-14-2019 12-39-50](https://user-images.githubusercontent.com/1930218/54374837-56dd8280-4656-11e9-8558-47bc0d63518f.gif)

PR #3131, opened 2018-10-12
Fix for #2752

donald93:
This is a fix for issue #2752 

First pull request, let me know if I messed anything up. The issue was just that only the first cursor was being manipulated when using a remapped command in insert mode.


jpoon:
Cool. Thanks for the fix!

PR #3133, opened 2018-10-13
Force linewise when using put with indent

ropez:

It changes the behavior of `[p` and `]p`, so that they always put in line-wise mode. This is very useful in certain use cases, for instance when pulling some code out of a long line, and pasting it as a new line. With this patch, it's possible to do this by copying text character-wise, and then pasting line-wise.

This behavior is available in real Vim when using the plugin called "vim-unimpaired".

The behavior is not configurable (yet). I will add configuration if required, but frankly, I don't see any use-cases for the way the put-with-indent commands currently work when with character-wise register content.

None

I'm not sure what the policy is for deviating from real Vim behavior. This emulates a specific plugin, but it doesn't add all features from this plugin. I'm not sure whether this is a problem or not.

I don't expect this PR to be accepted right away, there are certainly some documentation that needs to be updated, and probably some tests, and maybe configuration is needed. I'm just putting this here, to get some initial feedback on what is needed.

### DIFF
--- a/src/actions/commands/actions.ts
+++ b/src/actions/commands/actions.ts
@@ -1333,6 +1333,9 @@ export class PutCommand extends BaseCommand {
     const register = await Register.get(vimState);
     const dest = after ? position : position.getRight();
 
+    // Force linewise when using put with indent (`[p` or `]p`)
+    const currentRegisterMode = adjustIndent ? RegisterMode.LineWise : register.registerMode;
+
     if (register.text instanceof RecordedState) {
       /**
        *  Paste content from recordedState. This one is actually complex as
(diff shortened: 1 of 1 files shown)


trkoch:
I like the feature, but I would prefer wrapping it as configuration/plugin to decrease the risk of surprising behavior and allow to keep vim as a reference where possible.

ropez:
Thanks for responding. It's been a while, but this is still a very much missed feature for me.

Now, there are a number of Vim plugins that are already emulated by this extension. Like surround.vim and commentary.vim. Those are configurable, but some are enabled by default. This feature is from unimpaired.vim, but as I wrote above, it doesn't include other features from that plugin.

Of course, for me having this feature would be great, whether it's on by default or not. Like I said, I don't think the default behavior of `[p` and `]p`, when the register contains character-wise selection, is very useful. It's basically the same as `P` and `p`.

trkoch:
How about wrapping as plugin, but enabling it by default? What features does unimpaired have apart from your enhancements? Specifically, does the amount of features implemented justify suggesting we have an emulation for this plugin?

J-Fields:
Sorry to keep this in limbo for so long. I'm going to pass on this because I think preserving Vim's behavior in this case (particularly for people who map `]p` to `p`) is important, and this hasn't garnered enough traction to warrant a feature flag. I appreciate the PR though!

Issue #2791, opened 2018-06-29
Make replacing space with newline indent next line


delgiudices:
**Is this a BUG REPORT or FEATURE REQUEST?** (choose one):
FEATURE REQUEST

**What happened**:
I pressed: r -> enter to remove space and break one line into multiple lines. the new line was not indented.

this is very common when I have a line like
```
fn({ param1, param2, param3 });
```
And I want to split it to look like:

```
fn({
  param1,
  param2,
  param3,
});
```
**What did you expect to happen**:

The replacing the space for new line would automatically indent the next line.

**How to reproduce it (as minimally and precisely as possible)**:o
Open vscode with vim plugin.

1.type in this line:
```
fn({ param1, param2, param3 });
```

2.move the cursor to the space after the first comma

3. r -> enter

**Environment**:

- Extension (VsCodeVim) version:0.14.0
- VSCode version: 1.24.1
- OS version: macOS 10.13.2

PR #3134, opened 2018-10-13
Correct intentation when replacing space with newline

ropez:

Changes the behavior of inserting a newline in the middle of a line, by removing whitespace to the right of the cursor on the original line. The result is that the new line is correctly indented.

The resulting behavior is consistent with real vim.

Fixes issue #2791

I first tried to implement this behavior by registering a new command for the return key. This resulted in exceptions, because there is some special handling of the `CommandInsertInInsertMode` inside `modeHandler.ts`. While it would be possible to handle my new command in the same way, it seemed simpler just to amend the existing command.

### DIFF
--- a/src/actions/commands/insert.ts
+++ b/src/actions/commands/insert.ts
@@ -282,6 +282,14 @@ export class CommandInsertInInsertMode extends BaseCommand {
           position: vimState.cursorPosition,
         });
       } else {
+        if (char === '\n') {
+          // Remove whitespace to the right of cursor before inserting newline
+          const whitespace = line.substring(position.character).search(/\S|$/);
+          if (whitespace > 0) {
+            await TextEditor.delete(new vscode.Range(position, position.getRight(whitespace)));
+          }
+        }
+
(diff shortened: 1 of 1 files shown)


ropez:
Found an issue: It does not repeat correctly with dot.

jpoon:
Thanks for the PR! Can you add tests?

ropez:
Thanks for the response, @jpoon 

I wanted to look into the testing, but the problem was I couldn't get the tests to run. It was something about libgtk.so (iirc) missing, when running the gulp test command, after it started the docker container and downloaded VSCode. Are the test supposed to work on windows?

jpoon:
The issue you describe should be resolved with https://github.com/VSCodeVim/Vim/pull/3153. Alternative, you can run the tests within vscode itself.

ropez:
OK, thanks. I will look into this later, after work.

jpoon:
As there hasn't been any recent activity on this PR, I am closing this out. Feel free to re-open after addressing comments.

PR #3170, opened 2018-10-27
Fix #2984: wrong tab selected after :quit

### LINKED ISSUE #2984: When closing a window with `:q` VS Code now selects the tab "before" the one you were previously on
1. Load 3 files, each in its own tab.
2. Move to the second tab, then the third tab.
3. Type `:q<enter>` to close the tab.

I would now expect the second tab to be selected, and this worked up until recently. Now the first tab will be selected instead. It always chooses the tab "one to the left of" the tab it should be selecting.

ohjames:

Fixes #2984, the wrong tab is being selected after `:q` which is very annoying.

#2984 

The previous code (that this PR deleted) seems to be there to work around a bug in VS Code, since the VS Code issue has been fixed the code now leads to this bug.


ohjames:
I debugged the code some more and I see that `this.activeTextEditor.viewColumn` never changes, no matter what tab I `:q` from. So the "select previous editor" logic that I'm deleting runs every single time, always leading you to the tab to the left of where you want to be.

ohjames:
I was concerned the problem might be caused by the interaction between VS Code Vim and another plugin, but after disabling all other plugins the issue still manifests.

ohjames:
@rebornix Is there any way I can help with this or you could help me get this merged? It's a pretty serious and visible bug to have been around for so long.

gohar94:
Any updates on this? It is EXTREMELY annoying to live with this!

rebornix:
@ohjames sorry for late reply, the code change looks good to me.

ohjames:
Thanks!

gohar94:
Thanks, guys!

PR #3183, opened 2018-11-01
fix: automerge renovate minor/patch

jpoon:
Dear @renovate-bot,

We don't want to be bothered. Just automerge patch/minor versions.

K. Thanks. Bye.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -2,5 +2,9 @@
   "extends": [
     "config:base",
     "default:pinDigestsDisabled"
-  ]
+  ],
+  "automerge": true,
+  "major": {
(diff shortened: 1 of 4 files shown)


rarkins:
Hi @jpoon, glad that it's working out well for you! Thanks for the feedback.

One problem here though - currently you have installed the "forking" version of Renovate, so it has no write access to the main repo. You would need to uninstall the [forking-renovate app](https://github.com/apps/forking-renovate) and install the regular [renovate app](https://github.com/apps/renovate) instead in order for the automerge to work, because otherwise Renovate is blocked from merging.

Issue #3155, opened 2018-10-21
Does not support nowrapscan


jnothstine:
**Describe the bug**
VIM has a nowrapscan option that stops searches when the end of the file is reached, as opposed to continuing back to the top of the file

**To Reproduce**

1. Search for an item with /
2. Press n, continue pressing n
3. When the bottom of the fie is reached, the search goes back to the top of the file

**Expected behavior**
When the bottom of the file is reached, the search should stop if nowrapscan can be set.

contrib15:
I really like the 'nowrapscan' feature of Vim because I often miss the 'search wrapping' message and start reading a second time a line that I have already read. It is also very handy when searching for a symbol from the current position to the end of file. It would be very nice to have this feature in Vscodevim!

XiaoXiaoLui:
I can't believe it that  so few people cares about this feature.  
When I search in vim, I don't need to care about the status bar and can concentrate on the text.  Also, if I want to do wrapscan at the last match, I can just press "ggn" So there's no reason to turn on wrapscan for me.
However, with wrapscan on, I need to pay  attension to the status bar to know if it have reach the end.  
This is really annoying.

juanlanus:
I care about being able to disable wrapscan, and I'm looking forward for it.  
It enhances vim usability.
It protects the user from some of his errors, like evaluating that there are no more instances of the searched text when there actually are. 
With nowrapscan one can search repeatedly until after the editor complains, and that's it. The software checks if there are no instances left, not the human or the developer.

jnothstine:
I think there is a pull request to fix this and I believe it has been there a long time.  I really wish this would get fixed.

J-Fields:
@jnothstine Where's the PR for this? I don't see one and am happy to implement this if there isn't an existing PR.

jnothstine:
My mistake, I thought I had seen one quite a while ago.  Don’t know why I thought that.  There isn’t one.

Sent from Mail for Windows 10

contrib15:
Partially inspired by aborted PR #3192, here is a patched version of searchState.js file that seems to do the trick : [searchState.zip](https://github.com/VSCodeVim/Vim/files/3559555/searchState.zip).

It replaces the file `extensions/vscodevim.vim-1.10.0/out/src/state/searchState.js`.

Wrapscan can be turned on by changing line 223 to `wrapscan = true`.

J-Fields:
@contrib15 any interest in submitting a PR?

[...]

PR #3192, opened 2018-11-04
added nowrap option

Rjongenburger:
This PR implements the wrapscan option.

Fixes: #3155

### DIFF
--- a/src/state/searchState.ts
+++ b/src/state/searchState.ts
@@ -227,10 +227,10 @@ export class SearchState {
 
     if (!configuration.wrapscan) {
       if (
-        (isForward && this._matchRanges[this._matchRanges.length - 1].contains(startPosition)) ||
-        (!isForward && this._matchRanges[0].contains(startPosition))
+        (!isForward && this._matchRanges[0].start.line === startPosition.line) ||
+        (isForward &&
+          this._matchRanges[this._matchRanges.length - 1].start.line === startPosition.line)
       ) {
-
         return { start: startPosition, end: startPosition, match: true };
       }
     }


src/state/searchState.ts
```diff
@@ -223,8 +223,19 @@ export class SearchState {
     }
 
     const effectiveDirection = direction * this._searchDirection;
+    const forward = effectiveDirection === SearchDirection.Forward;
```
jpoon: Can you rename this to `isForward`?

src/state/searchState.ts
```diff
@@ -223,8 +223,19 @@ export class SearchState {
...
+
+    if (!configuration.wrapscan) {
+      const last = this._matchRanges.length - 1;
+      if (
+        (forward && this._matchRanges[last].contains(startPosition)) ||
```
jpoon: No need to declare `last` var. I think you can combine it all to 231.

src/state/searchState.ts
```diff
@@ -223,8 +223,19 @@ export class SearchState {
...
+        (!forward && this._matchRanges[0].contains(startPosition))
+      ) {
+        return { start: startPosition, end: startPosition, match: false };
```
jpoon: `match: false`

I don't recall the code path, but what is this `match` used for? Shouldn't it be `true`?

package.json
```diff
@@ -661,6 +661,11 @@
           "type": "string",
           "description": "Comma-separated list of motion keys that should wrap to next/previous line.",
           "default": ""
+        },
+        "vim.wrapscan": {
+          "type": "boolean",
+          "description": "when \"search next\" reaches end of file, it wraps around to the beginning and vice versa",
```
jpoon: Can we use vim's help text here directly?

http://vimdoc.sourceforge.net/htmldoc/options.html#'wrapscan'.

```
Searches wrap around the end of the file.
```

PR #3218, opened 2018-11-25
display '?' or '/' in status bar in search mode

### LINKED ISSUE #3211: Finding with `?` renders `/` in the status bar instead of `?`
When you search forward with `/`, a `/` and your search term appears in the status bar on the bottom:

But if you search backwards with `?`, the same `/` is displayed (the actual backwards searching with `n` and `N` works as expected). What should display instead in the search bar is `?` followed by the search term.

**To Reproduce**

1. From normal mode, type `?`, then a few characters to search for.
2. That's it. A `/` with your search term appears in the status bar.

**Expected behavior**
What should display instead in the search bar is `?` followed by the search term.

Extension version: 0.16.12
VS Code version: Code 1.28.2 (7f3ce96ff4729c91352ae6def877e59c561f4850, 2018-10-17T00:23:51.859Z)
OS version: Windows_NT x64 10.0.1713

dsschnau:
, depending on search direction.

Fixes #3211.

This is a small thing but a behavior I rely on to tell if I've fat-fingered the shift-key when searching.

This is my first PR to this project. If there's anything wrong in this PR please let me know and I'll fix it.


src/mode/modes.ts
```diff
@@ -63,7 +64,12 @@ export class SearchInProgressMode extends Mode {
   }
 
   getStatusBarText(vimState: VimState): string {
-    return `/${vimState.globalState.searchState!.searchString}`;
+    const leadingChar =
```
jpoon: I think this would be an error if `globalState.searchState` is undefined. Can we add a check for that, if it is undefined, log an error and return an empty string and return null or something?

jpoon:
Awesome! Thanks for the fix. Left a small comment but other than that LGTM.

src/mode/modes.ts
dsschnau: sure thing. i added a check/logger call in 1c3a82f .

Issue #3252, opened 2018-12-09
Pressing `i` followed by another command yields the incorrect `this.vimState.recordedState.commandList`


lanza:
When you press `i` to enter insert mode and then another binding (e.g. here I'm using `<c-d>`), the `<c-d>` does not get remapped to whatever your `insertModeRemappings` dictates on the first press of the `<c-d>`.

Stepping through the debugger shows me that this fails to call the remapping because `modeHandler.ts:320` is inserting the new key into a list with a length of 1 and the only element is `i`.

So it seems that hitting `i` to enter insert mode ultimately does not properly reset the `vimState.recordedState.commandList` and thus the next key press is viewed as combination pending mode with `i` as the first key.

jpoon:
Fixed with https://github.com/VSCodeVim/Vim/pull/3424

PR #3253, opened 2018-12-09
Not quite working - Reset the command list in ModeHandler.runAction if in insert mode

lanza:
Line 382 was checking if there was an action ran AND if the current
state was not insert mode. This caused a bug which left the commandList
with a lingering `i` which broke remapped bindings for the following
key. Remove this check so that the command list is clear upon entering
insert mode.

* Define a `insertModeRemapping` e.g. `"<c-e>"` -> `["<c-o>","x"]`
* Press `i` to enter insert mode.
* Press `<c-e>`

Nothing happens. This is because the `vimState.recordedState.commandList` length is 2 and the `withinTimeout` variable is `false`. Thus no attempt to remap and the keymapping lookup fails to find anything. 

HOWEVER, this causes the `jj` -> `esc` map test to fail and does indeed invalidate such bindings in insert mode. 

I can't say I know what to look for at this point because how you guys handle this state is spread out throughout these classes. So if somebody could check this out and find the proper way to amend this diff then I'd be very pleased. 

https://github.com/VSCodeVim/Vim/issues/3252

### DIFF
--- a/src/mode/modeHandler.ts
+++ b/src/mode/modeHandler.ts
@@ -578,7 +578,7 @@ export class ModeHandler implements vscode.Disposable {
       }
     }
 
-    if (ranAction && vimState.currentMode !== ModeName.Insert) {
+    if (ranAction) {
       vimState.recordedState.resetCommandList();
     }
 


jpoon:
The root cause of the issue that this PR is trying to address has been fixed with https://github.com/VSCodeVim/Vim/pull/3424.

Issue #1812, opened 2017-06-08
Laggy paste


ryanlaws:
* Click *thumbs-up* 👍 on this issue if you want it!
* Click *confused* 😕 on this issue if not having it makes VSCodeVim unusable. 

The VSCodeVim team prioritizes issues based on reaction count. 

--------

- **VSCode Version**: 1.13.0
- **VsCodeVim Version**: 0.8.4
- **OS**: Windows 8.1 Pro

**What happened**: Pasting yanked text has a 2 second delay. I'm not sure why this is so slow but it adds up. It's better in more recent versions, whereas in old versions it was more like 5+ seconds and pretty much unusable. I can paste with the system clipboard in insert mode but that's pretty un-Vim-like.

**What did you expect to happen**: Instantaneous paste.

**How to reproduce it**: Yank some text and then paste.

Chillee:
@ryanlaws For any amount of text? What about yanking? Are you pasting from the system clipboard?

ryanlaws:
@Chillee Any amount of text, even with one character. The amount of lag seems to be the same regardless of length.

michaelHL:
Just like this...

![](http://i.imgur.com/8LrUXxf.gif)

Chillee:
@michaelHL The issue you're encountering seems a little bit different: https://github.com/VSCodeVim/Vim/issues/1647

It's possible that it's actually only a problem with paste though...

ryanlaws:
A couple more data points, hope these are helpful:
- I can reproduce this repeatedly by simply pasting. The yanking step is actually unimportant, as long as the clipboard is not empty.
- I do not have this issue whatsoever on OS X El Capitan.

Chillee:
@ryanlaws Would it be possible for you to post some kind of gif?

ryanlaws:
@Chillee Sure, hope this makes it a bit clearer.
![vscodevim-issue1812](https://user-images.githubusercontent.com/164690/27393134-444ffae8-565d-11e7-8a92-3eb037325ea3.gif)

Chillee:
@ryanlaws Are you pasting from the system clipboard?

ryanlaws:
@Chillee Yes. Yanked to it using VSCodeVim, then pasted from it.

Chillee:
What happens if you paste from a register.

ryanlaws:
@Chillee pasting from a register is instantaneous!

Chillee:
[...]

PR #3261, opened 2018-12-13
Change to use native vscode clipboard

xconverge:


xconverge:
fixes #1812

jpoon:
Sweet. Can we also get rid of `overrideCopy`?

package.json
```diff
@@ -674,7 +674,6 @@
   },
   "dependencies": {
     "appdirectory": "0.1.0",
-    "clipboardy": "1.2.3",
```
jpoon: Update https://github.com/VSCodeVim/Vim#linux-setup

src/util/clipboard.ts
```diff
@@ -1,16 +1,16 @@
...
-      clipboardy.writeSync(text);
+      await vscode.env.clipboard.writeText(text);
     } catch (e) {
       logger.error(e, `Clipboard: Error copying to clipboard. err=${e}`);
     }
   }
 
-  public static Paste(): string {
-    return clipboardy.readSync();
+  public static Paste(): Thenable<string> {
```
jpoon: return Promise instead?

xconverge: I actually am not really sure how to properly implement a thenable interface (which the clipboard read is) as a promise. Any help would be much appreciated...

jpoon: Should just be:

```suggestion
  public static async Paste(): Promise<string> {
    return vscode.env.clipboard.readText();
  }
```

xconverge:
> 
> 
> Sweet. Can we also get rid of `overrideCopy`?
> 
> Edit: maybe not, we have a bunch of special logic there --
> 
> [Vim/src/actions/commands/actions.ts](https://github.com/VSCodeVim/Vim/blob/0151438b4b802f6123eac1864886aeba8de5e92b/src/actions/commands/actions.ts#L992)
> 
> Line 992 in [0151438](/VSCodeVim/Vim/commit/0151438b4b802f6123eac1864886aeba8de5e92b)
>  class CommandOverrideCopy extends BaseCommand {

Yea it that functionality is really for a different purpose unfortunately

src/util/clipboard.ts
```diff
@@ -1,16 +1,16 @@
...
+      await vscode.env.clipboard.writeText(text);
     } catch (e) {
       logger.error(e, `Clipboard: Error copying to clipboard. err=${e}`);
     }
   }
 
-  public static Paste(): string {
-    return clipboardy.readSync();
+  public static async Paste(): Promise<string> {
+    return vscode.env.clipboard.readText();
```
xconverge: do I want to await here?

xconverge:
Tested on Win10 and Ubuntu

jpoon:
Cutting a new release

PR #3281, opened 2018-12-20
fix saving remote file  error

### LINKED ISSUE #2956: save (:w or :wq) with SSHFS and LiveShare guest don't work properly
I'm using VS Code with a remote workspace (via an extension that integrates with our dev-server) and saving files with ":w" always gives me error. It seems our private extension overloaded the "save" action in VS Code but the VIM extension is bypassing the editor's save API. 

We can have an additional flag in settings to bypass the local file write. A check of the setting can be added at here: https://github.com/VSCodeVim/Vim/blob/82abec31c996b381a4a544932a9cbaa89e80ea3b/src/cmd_line/commands/write.ts#L52

zhuzisheng:
This PR fixes the error when saving remote files (via ssh or others)

fixes #2956
 
scheme value in document's uri is "file" for disk files.
for remote files, the scheme value may be different.
To use the workbench's save function can work on saving remove files.

### DIFF
--- a/src/cmd_line/commands/write.ts
+++ b/src/cmd_line/commands/write.ts
@@ -49,7 +49,8 @@ export class WriteCommand extends node.CommandBase {
       Message.ShowError('Not implemented.');
       return;
     }
-
+    
+    // defer saving the file to vscode if file is new (to present file explorer) or if file is a remote file
     if (vimState.editor.document.isUntitled || vimState.editor.document.uri.scheme !== 'file') {
       await vscode.commands.executeCommand('workbench.action.files.save');
       return;
--- a/src/cmd_line/commands/write.ts
+++ b/src/cmd_line/commands/write.ts
@@ -50,7 +50,7 @@ export class WriteCommand extends node.CommandBase {
       return;
     }
 
-    if (vimState.editor.document.isUntitled) {
+    if (vimState.editor.document.isUntitled || vimState.editor.document.uri.scheme !== 'file') {
       await vscode.commands.executeCommand('workbench.action.files.save');
       return;
     }


src/cmd_line/commands/write.ts
```diff
@@ -50,7 +50,7 @@ export class WriteCommand extends node.CommandBase {
       return;
     }
 
-    if (vimState.editor.document.isUntitled) {
+    if (vimState.editor.document.isUntitled || vimState.editor.document.uri.scheme !== 'file') {
```
jpoon: Is this logic correct? If this is NOT a file, we want to save?

Is there an API we can instead leverage such that we can save for local and remote files?

zhuzisheng: I will check whether there is such api.

zhuzisheng: the issue lies in fs module which can only check the files with "file" schema.

jpoon: Would you mind adding a comment explaining this if clause? Maybe something along the lines of:

// defer saving the file to vscode if file is new (to present file explorer) or if file is a remote file

jpoon: What's the scheme on a remote file?

zhuzisheng: I'm using the package "Remote FS". It provides two ways to access remote files “ftp” and "sftp".
sftp is based on ssh. 
If I uses "sftp" to access remote file, the file scheme would be "sftp".

there is another extension which called "memfs" which is used to create a totally memory based file system. This extension is provided by VSCode team as tutorial on VSCode's "FileSystemProvider".
The files created by this "memfs" has the scheme "memfs". VSCodeVim also has the issue on saving this file scheme.

The real issue for saving files with non-"file" scheme is that VSCodeVim tries to check the file existence by "fs" module of nodejs before calling the "save" API.
VSCode's save API can work on different file schemes, but fs module can't. It only support “file:” protocol. 
https://nodejs.org/api/fs.html#fs_url_object_support
image

I haven't found an common api from VSCode which can check file's existence to replace fs access API.
Is it OK to remove the access checking on non-"file" scheme files?

jpoon: From my review:

> LGTM. If you could add that documentation, that would be great.

Issue #3013, opened 2018-08-31
All ex commands cause editor window to freeze


jplew:
**Describe the bug**

Vim works fine until I attempt to submit an ex command (eg: `:s/before/after/g`, `:t25 .`).

I see the keystrokes being recorded in the bottom status bar, but when I hit enter nothing happens and my cursor freezes. The command remains in the status bar, but I can't input any text into the editor window or issue any normal commands.

The menus still work, but the only way for me to enter text again is to close the window and and reopen it.

I've been experiencing this problem for over a month now.

**To Reproduce**

1.  Open any file
2.  Type `:t1 .`

**Expected behavior**
I expect line 1 to be copied.

**Environment (please complete the following information):**

- Extension Name: vim
- Extension Version: 0.16.2
- OS Version: Darwin x64 17.7.0
- VSCode version: 1.26.1

**Additional context**
I have replicated this problem in VSCode 1.27.0-insider as well.

The problem still occurs even after disabling all other extensions.

justinmk:
What if you remove your vimrc? 

See also https://github.com/VSCodeVim/Vim/issues/1735#issuecomment-429088199

jplew:
> What if you remove your vimrc?
> 
> See also [#1735 (comment)](https://github.com/VSCodeVim/Vim/issues/1735#issuecomment-429088199)

I don't have a ~/.vimrc

If I set ` "vim.enableNeovim": false` the problem goes away, and I can run substitutions (:s/foo/bar)

justinmk:
Sorry it's not clear to me if you mean the literal path `~/.vimrc`. Let me ask differently: do you see any errors or other prompts if you start `nvim` directly, in your terminal?

> If I set "vim.enableNeovim": false the problem goes away, and I can run substitutions (:s/foo/bar)

That doesn't help narrow down the cause of why `nvim` might be blocking for input ...

jplew:
I have no ~/.config/nvim directory. Any place else I should look?

Yes, I get the following error when I launch `nvim` from the Terminal: `E576: Error while reading ShaDa file: last entry specified that it occupies 101 bytes, but file ended earlier`.

jplew:
Update: I got it working. 

You were right about the neovim launch error. It seems my `main.shada` file was corrupted. I deleted `~/.local/share/nvim/shada/main.shada` and now neovim works as expected. Thanks @justinmk !

justinmk:
Two thoughts about that: 

[...]

PR #3288, opened 2018-12-23
fix: disable nvim shada

jpoon:
As recommended https://github.com/VSCodeVim/Vim/issues/3013#issuecomment-429335777, disable shada. @justinmk's second point has already been addressed: https://github.com/VSCodeVim/Vim/blob/master/src/neovim/neovim.ts#L32

https://github.com/VSCodeVim/Vim/issues/3013

### DIFF
--- a/src/neovim/neovim.ts
+++ b/src/neovim/neovim.ts
@@ -57,7 +57,7 @@ export class Neovim implements vscode.Disposable {
     if (!(await util.promisify(exists)(dir))) {
       dir = __dirname;
     }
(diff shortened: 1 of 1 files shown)


justinmk:
@jpoon Nvim 0.3.2 made a significant improvement to the behavior of `--embed` so that hacks like this aren't needed anymore. See:

https://github.com/neovim/neovim/wiki/Following-HEAD#20180922

Note that `--headless` flag should not be specified (and VSCodeVim has no reason to specifiy it, so it should not).

Issue #826, opened 2016-10-04
Add a setting to toggle whether <Esc> closes autocomplete in insert mode.


FSMaxB:
Please *thumbs-up* 👍 this issue if it personally affects you! You can do this by clicking on the emoji-face on the top right of this post. Issues with more thumbs-up will be prioritized.

-----

### What did you do?

I type something that opens the autocomplete dropdown (typing the beginning of the name of a known symbol for example). After I stop typing, I press Escape to go back to normal mode.

### What did you expect to happen?

The editor should change to normal mode.

### What happened instead?

The escape key closed the autocomplete dropdown, without changing the vim mode. If I want to change to normal mode, I have to press escape twice.

### Technical details:

* VSCode Version: 1.5.3
* VsCodeVim Version: 0.3.0
* OS: Windows 8.1

I am not completetely sure if this also happened with 0.2.0, but I don't think so. Because right now many times when I'm trying to change to normal mode and save the file, it types in :w<return> instead, which inserts a while loop in my code. This is really annoying because it completely destroys my workflow and it never saw that happening with 0.2.0.

johnfn:
This is actually on purpose: https://github.com/VSCodeVim/Vim/issues/804

FSMaxB:
Is there a setting to turn this off?

FSMaxB:
Or rather: Could you add one?

johnfn:
Yes, that could be done. PRs welcome... 😉

FSMaxB:
I see what I can do once I have time.

FSMaxB:
Regarding that: Is there a way to use settings from `settings.json` in conditions in the `when` part of keybindings?

FSMaxB:
The current way to set this setting is by putting the following in `keybindings.json`:
```JSON
 { "key": "escape",        "command": "extension.vim_escape",
                              "when": "editorTextFocus && !inDebugRepl" }
```

johnfn:
I've come around. You were right, this is incredibly annoying. If anyone wants the old behavior, they can add the setting. :P

leonzhu42:
Not working. I've checked the default `keybindings.json`. The solution above has been set default.

I also checked VSCode documentation. I didn't find `inDebugRepl` option. Instead, I saw `inDebugMode`. Tried. Still not working.

I also tried `suggestWidgetVisible`. Still.

zyujs:
you can try to add this to keybindings.json
`{
  "key": "escape",
[...]

Issue #804, opened 2016-09-26
<ESC> in auto complete mode exits insert mode


octref:
Please *thumbs-up* 👍 this issue if it personally affects you! You can do this by clicking on the emoji-face on the top right of this post. Issues with more thumbs-up will be prioritized.

### What did you do?

- Go to insert mode
- Get list of autocomplete
- Found nothing I want, exiting by `<ESC>`

### What did you expect to happen?

Still in insert mode, auto-complete cancelled

### What happened instead?

Back to normal mode

### Technical details:

* VSCode Version: 1.6 Insiders
* VsCodeVim Version: 0.2.0
* OS: OS X 10.11.6

johnfn:
Not too hard of a fix - just requires you to understand the vscode API...

rebornix:
@octref do we know whether the editor is in auto-complete or not?

octref:
@rebornix While editor is showing a list of suggestions, if I click `<ESC>`, it will close auto-complete but then also exit insert mode.

rebornix:
@octref actually I was asking whether Code has any idea indicating that's there is a suggestion popup but yes we do https://github.com/Microsoft/vscode/blob/846a1d8cccb3d6aa747af3cbc5b35c93bfbd393f/src/vs/editor/contrib/suggest/common/suggest.ts#L23 . Easy fix, it's my first commit above 10000 feet.

Eksapsy:
Pardon me if you can, but I don't understand, why is the issue closed? It's still happening after 2 years of this issue being closed.
How can I fix that? I have the exact same problem.

bluethon:
for anyone who google from here, I finally find the [solution](https://github.com/VSCodeVim/Vim/issues/826#issuecomment-346518581)

PR #3292, opened 2018-12-25
fix: esc when closing auto-complete widget will stay in insert mode

jpoon:

- When an auto-complete widget is visible when in Insert mode, `<Esc>` will close the auto-complete widget, but leave the editor in Insert mode.
- Wasn't able to find a VSCode API to detect if suggest widget was available. 
- Not sure if this will be a contentious change, but I like this new behaviour. Open to comments.

https://github.com/VSCodeVim/Vim/issues/826, https://github.com/VSCodeVim/Vim/issues/804

- It is assumed that while the suggestion widget is visible, the user's intention when pressing `<Esc>`, is to close the auto-suggest widget.

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,7 +55,7 @@
       {
         "key": "Escape",
         "command": "extension.vim_escape",
-        "when": "editorTextFocus && vim.active && !inDebugRepl"
+        "when": "editorTextFocus && !suggestWidgetVisible && vim.active && !inDebugRepl"
       },
       {
         "key": "Home",


xconverge:
This is contentious. It is a PITA to press esc twice last time we tried this if I remember correctly. I remember people being 50:50 everytime this was changed

jpoon:
Yah, gonna close this as a won't fix. Keeping with the status quo.

parkerault:
If anyone else lands here, the native `control-.` will close the suggest widget without leaving insert mode. I went deep down this rabbit hole before realizing that there's already a simple solution built in. :)

StanBright:
it's a rabbit hole for sure... however, `control-.` doesn't work for me :/

PR #3293, opened 2018-12-26
fix: slightly improve perf by caching vscode context

jpoon:

The following operation `vscode.commands.executeCommand('setContext', ...` takes roughly 1s to complete. Cache the value and only update vscode context when value changes.

We had done this for some of the context values. This PR does it for _all_ context values.

- sorted imports

### DIFF
--- a/src/util/vscode-context.ts
+++ b/src/util/vscode-context.ts
@@ -1,0 +1,19 @@
+import * as vscode from 'vscode';
+
+export class VsCodeContextImpl {
+  contextMap: { [key: string]: any } = {};
+
+  public async Set(key: string, value: any): Promise<void> {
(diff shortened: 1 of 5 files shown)


src/util/vscode-context.ts
```diff
@@ -0,0 +1,19 @@
+import * as vscode from 'vscode';
+
+export class VsCodeContextImpl {
+  contextMap: { [key: string]: any } = {};
+
+  public async Set(key: string, value: any): Promise<void> {
+    let prev = this.Get(key);
```
xconverge: const instead of let

xconverge:
This is a REALLY good idea, nice job

PR #3305, opened 2018-12-30
fix: sync editor.wordSeparators and vim.iskeyword. closes #3166

### LINKED ISSUE #3166: editor.wordSeparators setting is ignored
I use to consider ```this_example``` two words, because a word separator for me is the underscore.
Tried to change **editor.wordSeparators** but it was completely ignored, since I found later that there is another setting **vim.iskeyword** that does basically the same.

Is there a way to add a setting to use the default editor.wordSeparators instead a duplicate setting?

jpoon:

* Adds `iskeyword` as an overlap setting to `editor.wordSeparators`. As a result, the following will take precedence:

1. vim.iskeyword
2. if (1) is undefined, editor.wordSeparators
3. if (2) is undefined, then default value of `@,48-57,_,128-167,224-235` is used

* This change removes the default value for `iskeyword` so unless users have explicitly configured this value, `editor.wordSeparators` will be used. 

So in the standard use case where users were using the default values for `iskeyword` those folks will be moving from using this:
```
/\()\"':,.;<>~!@#$%^&*|+=[]{}`?-
``` 
to vscode's default for `editor.wordSeparators`

```
`~!@#$%^&*()-=+[{]}\|;:'",.<>/?
```

The two are the same set, but in different order so there should *not* be a change in behaviour. Note that if `vim.iskeyword` is set, it will override the value of `editor.wordSeparators`.

#3166 

### DIFF
--- a/package.json
+++ b/package.json
@@ -422,8 +422,7 @@
         },
         "vim.iskeyword": {
           "type": "string",
-          "description": "keywords contain alphanumeric characters and '_'",
-          "default": "/\\()\"':,.;<>~!@#$%^&*|+=[]{}`?-"
+          "description": "keywords contain alphanumeric characters and '_'. If not configured `editor.wordSeparators` is used."
         },
         "vim.ignorecase": {
           "type": "boolean",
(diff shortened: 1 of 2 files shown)


jpoon:
Not really `backwards-incompatible`, but more...breaking change.

jpoon:
Actually, looking at the two sets, they are the exact same, just in different orders so we are good here :)

PR #3315, opened 2019-01-01
feat: display vim errors as vscode user error windows

jpoon:
In an effort to make things easier to debug, this PR adds a new config `vim.debug.displayError` (default value: true) indicating that VSCodeVim extension errors will also be presented as VS Code information windows (https://code.visualstudio.com/api/references/vscode-api#window.showErrorMessage).

This configuration is separate from `vim.debug.loggingLevel` which denotes the logging level to console.


README.md
```diff
@@ -1,624 +1,625 @@
...
+| vim.textwidth    | Width to word-wrap when using `gq`                                                                                                                                                                                                                                    | Number  | 80            |
+| vim.timeout      | Timeout in milliseconds for remapped commands                                                                                                                                                                                                                         | Number  | 1000          |
+| vim.whichwrap    | Controls wrapping at beginning and end of line. Comma-separated set of keys that should wrap to next/previous line. Arrow keys are represented by `[` and `]` in insert mode, `<` and `>` in normal and visual mode. To wrap "everything", set this to `h,l
```
jpoon: Argh. Damn line-endings.

src/util/message.ts
```diff
@@ -2,10 +2,10 @@ import * as vscode from 'vscode';
 
 export class Message {
```
jpoon: I think I might get rid of this class. Seems kind of useless.

Issue #3323, opened 2019-01-04
Replacing with Neovim 0.3.2 enabled fails and require a VSCode restart


GuillaumeRochat:
**Describe the bug**
Replacing with Neovim enabled changes indentation (changed from 4 spaces to tabs in my case) and makes vim unresponsive, requiring to restart VSCode. Also, the ex command stays in the status bar as if it wasn't executed.

**To Reproduce**

1.  Enable Neovim
2.  Open any file with space indents and perform a full replace such as `:s/potato/carrot/g`
3.  Notice that the indent changed to tabs.
4.  Notice that the ex command is still appearing in the status bar
5.  Notice that navigating in a file using vim keybind no longer works.

**Expected behavior**
The text is successfully replaced with no other side effects.

**Environment (please complete the following information):**

- Extension (VsCodeVim) version: 0.17.3
- VSCode version: 1.30.1
- OS: Linux 4.20.0
- Neovim: 0.3.2

**Additional context**
It only started happening after I updated to the latest version of Neovim (0.3.2) so it may be related to #3290. Does not happen on Neovim 0.3.1

jpoon:
What does the debug console say? With the next version we'll be displaying errors as info boxes so it'll be more apparent I'd there is a setup error.

GuillaumeRochat:
@jpoon would that be the console in Help > Toggle Developer Tools? If so, nothing appears in that console. In the process explorer, I can see the `nvim -u NONE -i NONE -N --embed` command and it says it's using 0mb of memory.

jpoon:
You need to turn on debug logging. Its off by default. On my phone but it's something like vim.debug.logging

GuillaumeRochat:
I don't see anything not normal. I changed the setting `vim.debug.loggingLevel` to `debug` : 

```
console.ts:134 [Extension Host] debug: ModeHandler: handling key=<Esc>.
console.ts:134 [Extension Host] debug: Remapper: find matching remap. keys=<Esc>. mode=Normal.
console.ts:134 [Extension Host] debug: ModeHandler: handling key=:.
console.ts:134 [Extension Host] debug: Remapper: find matching remap. keys=:. mode=Normal.
console.ts:134 [Extension Host] debug: ModeHandler: handling key=%.
console.ts:134 [Extension Host] debug: ModeHandler: handling key=s.
console.ts:134 [Extension Host] debug: ModeHandler: handling key=/.
[...]

Issue #3290, opened 2018-12-23
bug: fix neovim start-up flags


jpoon:
Nvim 0.3.2 made a significant improvement to the behavior of `--embed` so that hacks like this aren't needed anymore. See:

https://github.com/neovim/neovim/wiki/Following-HEAD#20180922

Note that `--headless` flag should not be specified (and VSCodeVim has no reason to specifiy it, so it should not).

The NVim version is returned in the `version` field of the `nvim_get_api_info` API call.

_Originally posted by @justinmk in https://github.com/VSCodeVim/Vim/pull/3288#issuecomment-449632604_

PR #3334, opened 2019-01-05
Update to new neovim node library

xconverge:
Tested with both 0.3.1 and 0.3.3 and both work with the commands I tested (:sort mostly)

### DIFF
--- a/src/neovim/neovim.ts
+++ b/src/neovim/neovim.ts
@@ -18,6 +18,15 @@ export class NeovimWrapper implements vscode.Disposable {
   async run(vimState: VimState, command: string) {
     if (!this.nvim) {
       this.nvim = await this.startNeovim();
+
+      await this.nvim.uiAttach(999999999, 999999999, {
+        ext_cmdline: false,
+        ext_popupmenu: false,
+        ext_tabline: false,
+        ext_wildmenu: false,
+        rgb: false,
+      });
+
       const apiInfo = await this.nvim.apiInfo;
       const version = apiInfo[1].version;
       logger.debug(`Neovim Version: ${version.major}.${version.minor}.${version.patch}`);
@@ -63,7 +72,7 @@ export class NeovimWrapper implements vscode.Disposable {
     if (!(await util.promisify(exists)(dir))) {
       dir = __dirname;
     }
-    this.process = spawn(configuration.neovimPath, ['-u', 'NONE', '-n', '--embed', '--headless'], {
+    this.process = spawn(configuration.neovimPath, ['-u', 'NONE', '-n', '--embed'], {
       cwd: dir,
     });
 
--- a/src/neovim/neovim.ts
+++ b/src/neovim/neovim.ts
@@ -63,7 +63,7 @@ export class NeovimWrapper implements vscode.Disposable {
     if (!(await util.promisify(exists)(dir))) {
       dir = __dirname;
     }
-    this.process = spawn(configuration.neovimPath, ['-u', 'NONE', '-i', 'NONE', '-N', '--embed'], {
+    this.process = spawn(configuration.neovimPath, ['-u', 'NONE', '-n', '--embed', '--headless'], {
       cwd: dir,
     });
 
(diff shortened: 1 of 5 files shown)


src/neovim/neovim.ts
```diff
@@ -30,7 +30,8 @@ export class Neovim implements vscode.Disposable {
     // Execute the command
     logger.debug(`Neovim: Running ${command}.`);
     await this.nvim.input(command);
-    if ((await this.nvim.getMode()).blocking) {
+    const mode = await this.nvim.mode;
```
xconverge: @justinmk this is not returning, any ideas? Also any advice in general for this PR would be great

xconverge:
Relates to #3323 and #3290

src/neovim/neovim.ts
```diff
@@ -59,25 +63,31 @@ export class Neovim implements vscode.Disposable {
     if (!(await util.promisify(exists)(dir))) {
       dir = __dirname;
     }
-    this.process = spawn(configuration.neovimPath, ['-u', 'NONE', '-i', 'NONE', '-N', '--embed'], {
+    this.process = spawn(configuration.neovimPath, ['-u', 'NONE', '-n', '--embed', '--headless'], {
```
xconverge: so now that I am using --headless do I still need the -i for SHADA stuff? A bit confused here

justinmk: `-i NONE` is not needed for 0.3.2.

In 0.3.1,  `--headless` is useless; in 0.3.2+ it is worse than useless because it changes the behavior of `--embed`.

src/neovim/neovim.ts
```diff
@@ -4,20 +4,23 @@ import { Position } from './../common/motion/position';
...
-import { attach, Nvim } from 'promised-neovim-client';
 import { configuration } from '../configuration/configuration';
 import { dirname } from 'path';
 import { exists } from 'fs';
 import { logger } from '../util/logger';
 import { spawn, ChildProcess } from 'child_process';
+import { attach, Neovim } from 'neovim';
 
-export class Neovim implements vscode.Disposable {
+export class NeovimWrapper implements vscode.Disposable {
```
justinmk: why change all of these names?

src/neovim/neovim.ts
justinmk: call `nvim_ui_attach` after connecting. Nvim waits until then before continuing, so that UI messages can be received.

src/neovim/neovim.ts
xconverge: To not mix it up with the exposed interface from the neovim npm package, and it is a wrapper

I understand that there are other ways of doing imports in neovim.ts but this seemed to be the simplest and clearest to me

src/neovim/neovim.ts
xconverge: Got it, thanks! Since I am only partially using the ui features, I am specifying a bogus width/height... seems bad but works for now..

src/neovim/neovim.ts
xconverge: Got it, there is not a great way to spawn the process, then check the version to know whether or not -i NONE is needed, I am ok with just leaving it omitted for now (since we also got away without too many bug reports for many many months) if it only will impact 0.3.1 users

justinmk:
Something I forgot to mention in https://github.com/VSCodeVim/Vim/issues/3290 , because I forgot that VSCodeVim does not attach as a UI: to handle dialogs like the "shada" dialog, the UI (VSCodeVim) would need to somehow detect the dialog and then prompt the user.

With Nvim 0.3.2 the UI can enable `ext_linegrid` in the `nvim_ui_attach` call. See `:help ui-linegrid`. This will send redraw events to the UI as lines instead of characters. That makes it somewhat easy to "scrape" the UI for dialogs when `nvim_get_mode` returns `blocked=true`. 

Of course, you could instead simply dismiss the dialog with `nvim_input('<c-c>')`. But that's basically equivalent to `-i NONE`. Actually, `-i NONE` may not be a bad choice since it's probably not relevant for the way VSCodeVim uses Nvim currently. And it might also be a good idea to disables swapfiles by specifying `-n` .

xconverge:
I appreciate the tips, I already had swapfiles disabled

Thanks for the help!

PR #3380, opened 2019-01-17
fixes #3374

### LINKED ISSUE #3374: status bar color unexpectedly changes to red
Hi there.

**Describe the bug**
I'm using `statusBarColorControl`.  
When I use `: key action` (like `:w`) or `easymotion`, status bar changes to red.  
I never set this red for status bar.

**To Reproduce**
I set my `settings.json` like this.

>
    "vim.statusBarColorControl": true,
    "vim.statusBarColors.normal": "#0e4569",
    "vim.statusBarColors.insert": "#424d00",
    "vim.statusBarColors.visual": "#691b41",
    "vim.statusBarColors.visualblock": "#691b41",
    "vim.statusBarColors.visualline": "#691b41",
    "vim.statusBarColors.replace": "#770000",        //    <-- this red is not the problem red.

After I set above, this was automatically added.

>
    "workbench.colorCustomizations": {
        "statusBar.background": "#0e4569",
        "statusBar.noFolderBackground": "#0e4569",
        "statusBar.debuggingBackground": "#0e4569"
    },

**Expected behavior**
Don't change to red.

**Screenshots**

- Extension (VsCodeVim) version: v1.0.1
- VSCode version: 1.30.2
- OS: windows10, ubuntu 16.04 (I found this problem in my Windows laptop, and the same happens in Linux Desktop PC.)

**Additional context**
It worked correctly few weeks ago.

This very annoys me.  
Does anyone have an idea?

Thanks.

xconverge:


src/statusBar.ts
```diff
@@ -63,7 +63,13 @@ class StatusBarImpl implements vscode.Disposable {
     let foreground;
     let background;
 
-    const colorToSet = configuration.statusBarColors[ModeName[mode].toLowerCase()];
+    let colorToSet = configuration.statusBarColors[ModeName[mode].toLowerCase()];
+
+    // If no color defined, fallback to normal mode color
+    if (colorToSet === undefined) {
+      colorToSet = configuration.statusBarColors[ModeName[ModeName.Normal].toLowerCase()];
```
jpoon: What is there's no normal mode color defined? Should we just log a debug message and return instead?

xconverge: The problem I have with that is the existing color will be left as active. If no normal is defined apparently it will be annoying bright red

jpoon: Where does the red come from? It's just strange that we are using the normal mode color as default as it may be null so maybe we should either:

1) validate the config and if they don't provide all the mode colors, we disable the status bar color.
2) if it's undefined, delete the `colorCustomizations` so we return back to vscode defaults.

Issue #3267, opened 2018-12-14
Support digraphs


FelipeLema:
Insert digraphs using `Ctrl-K` and two keys in insert mode.

This is fully documented in [`:help digraph`](https://github.com/neovim/neovim/blob/master/runtime/doc/digraph.txt).

It's been implemented from scratch in [Evil (Emacs) package](https://github.com/emacs-evil/evil/blob/82c65dcfe23aff3d764cafc78124d92940c5bd59/evil-digraphs.el).

Don't really care about the implementation or adding custom digraphs.

FelipeLema:
I would implement it myself, but I couldn't find a starting place (I tried looking for tests for "ctrl-r in insert mode" and (consequently) "registers", but no go).

jpoon:
Closed by #3407

Issue #1980, opened 2017-08-25
Bug: Can't navigate in autocompletion with "Ctrl+j" and "Ctrl+k".


Nexmean:
* Click *thumbs-up* 👍 on this issue if you want it!
* Click *confused* 😕 on this issue if not having it makes VSCodeVim unusable. 

The VSCodeVim team prioritizes issues based on reaction count. 

--------

**Is this a BUG REPORT or FEATURE REQUEST?** (choose one): BUG REPORT

**Environment**:

- **VSCode Version**: 1.15.1 - 41abd21afdf7424c89319ee7cb0445cc6f376959
- **VsCodeVim Version**: 0.9.0
- **OS**: Manjaro 17.0.2 Gellivara

**What happened**: Can't navigate in autocompletions with `<C-j>` and `<C-k>`. Instead of navigation opens terminal panel, but after removing of this shortcut navigation still don't works. Before, in Code 1.14 or below and VsCodeVim 0.6.* this worked fine.

**What did you expect to happen**: 

**How to reproduce it**: 

My VsCode config:
```json
{
    "window.menuBarVisibility": "toggle",
    "workbench.colorTheme": "Nord",
    "workbench.iconTheme": "material-icon-theme",

    // vim
    "vim.disableAnnoyingNeovimMessage": true,
    "vim.leader": " ",
    "vim.easymotion": true,

    // editor
    "editor.fontFamily": "'Fira Code', 'Droid Sans Mono', 'Courier New', monospace, 'Droid Sans Fallback'",
    "editor.fontLigatures": true,
    "editor.renderWhitespace": "boundary"
}
```
No other configurations or keymap plugins.

**See related gifs below:**
When I'm pushing `<C-j>` :
![ctrl-j](https://user-images.githubusercontent.com/3821520/29709743-e7a54e0e-89d0-11e7-8812-66a2364fa73c.gif)

When I'm pushing `<C-k>` :
![ctrl-k](https://user-images.githubusercontent.com/3821520/29709744-e7aee342-89d0-11e7-84ed-a4e830bb7510.gif)

Thank you for very good extension. I really love it.

Chillee:
Somebody brought up this issue on the slack, and the fundamental issue is this line: 
https://github.com/VSCodeVim/Vim/blob/master/package.json#L145

I'm not sure, why it's there. 
@xwvvvvwx Do you happen to remember?

xwvvvvwx:
[...]

PR #3407, opened 2019-01-23
Digraph support

jbaiter:
This PR implements support for digraphs.
All of the digraphs available in the current neovim version can be entered with `<C-k> <any> <any>`.
Additionally, users can configure custom digraphs with the `vim.customDigraphs` configuration key. It maps two-character shorthands to a pair of a descriptive string and one or more UTF16 code points:

```json
{ "vim.customDigraphs": {
    "o:": ["ö", 62976],
    "R!": ["🚀", [55357, 56960]]
  }
}
```
A filterable list of all available digraphs can be obtained with the ex command `:dig[raphs]`, same as in Vim.

**Which issue(s) this PR fixes**: #3267 

**Special notes for your reviewer**: Unfortunately I had to remove the `<C-k>` shortcut for `CommandNavigateAutocompleteUp`, since it prevented the `<C-k> <any> <any>` shortcut from being called.


package.json
```diff
@@ -675,6 +675,13 @@
           "type": "number",
           "description": "Threshold for reporting number of lines changed.",
           "default": 2
+        },
+        "vim.customDigraphs": {
+          "type": "object",
+          "description": "Custom digraph shortcuts for inserting special characters, expressed as UTF16 code points",
+          "default": {
+            "R!": ["🚀", [55357, 56960]]
```
jpoon: Did you mean to leave this default digraph here? What shall happen if a user puts a custom digraph here which is already in the default? Walking through the code, looks like we'll add it twice?

src/cmd_line/commands/digraph.ts
```diff
@@ -0,0 +1,54 @@
...
+      digraphKeyAndContent.push({
+        label: digraphKey,
+        description: charDesc,
+      });
+    }
+
+    vscode.window.showQuickPick(digraphKeyAndContent).then(async val => {
+      if (val) {
+        // TODO: Can we insert the selected digraph at cursor position?
+        vscode.window.showInformationMessage(`${val.label} ${val.description}`);
```
jpoon: Yes, we should probably insert it at the current position rather than outputting it to an info window.

src/cmd_line/commands/digraph.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    for (let digraphKey of Object.keys(configuration.customDigraphs)) {
+      let charDesc = configuration.customDigraphs[digraphKey][0];
+      digraphKeyAndContent.push({
+        label: digraphKey,
+        description: `${charDesc} (user)`,
+      });
+    }
+
+    for (let digraphKey of Object.keys(DefaultDigraphs)) {
```
jpoon: Dupe code from above, can we combine these two maps?

src/cmd_line/subparser.ts
```diff
@@ -99,6 +100,11 @@ export const commandParsers = {
 
   reg: parseRegisterCommandArgs,
 
+  dig: parseDigraphCommandArgs,
+  digr: parseDigraphCommandArgs,
+  digraph: parseDigraphCommandArgs,
+  digraphs: parseDigraphCommandArgs,
+
```
jpoon: I'm not familiar with the vim digraph command. Are all these commands synonymous?

package.json
jpoon: If you meant to put this here for tests, there's a `testConfiguration` you can add it to.

Edit: nevermind, i see you already did that.

jpoon: In vim, is there a way to configure this in `.vimrc`? If so, we should match similar naming.

src/actions/commands/insert.ts
```diff
@@ -491,7 +540,7 @@ class CommandNavigateAutocompleteDown extends BaseCommand {
 @RegisterAction
 class CommandNavigateAutocompleteUp extends BaseCommand {
   modes = [ModeName.Insert];
-  keys = [['<C-p>'], ['<C-k>']];
+  keys = ['<C-p>'];
```
jpoon: This would be a breaking change. Not certain if this is ok...

src/cmd_line/subparser.ts
jbaiter: Yeah, they all map to the same two commands: https://github.com/vim/vim/blob/master/runtime/doc/digraph.txt#L27-L39

src/actions/commands/insert.ts
jbaiter: I couldn't get it to work otherwise, since the `<C-k>` shortcut will always prevent the `<C-k> <any> <any>` shortcut from firing.
I'm not even sure if `<C-k>` is actually referenced anywhere in the original vim for navigating up in the omnicomplete popup. All references to `<C-k>` in the neovim source code refer to digraphs.

test/configuration/configuration.test.ts
```diff
@@ -84,4 +85,10 @@ suite('Configuration', () => {
     end: ['|'],
     endMode: ModeName.Visual,
   });
+  
+  test('Can handle custom digraph insert', async () => {
```
jbaiter: With the latest release, this tests fails since nothing gets inserted in the buffer. Do you know what might cause this?

package.json
jpoon: > Should I rename it to just digraphs instead of customDigraphs
Yes, I think that would be best.

src/actions/commands/insert.ts
jpoon: I'm trying to figure out what the appropriate keybinding would be for `CommandNavigateAutocompleteUp` and `CommandNavigateAutocompleteDown`.

These were initially added as a part of https://github.com/VSCodeVim/Vim/issues/1980. Likewise, I don't see `<c-j>` documented as autocompletedown in vim docs.

test/configuration/configuration.test.ts
jpoon: You can take a look at the changelog, but nothing apparent comes to mind.

PR #3408, opened 2019-01-23
fix #3157: register single onDidChangeTextDocument handler and delegate to appropriate mode handler

### LINKED ISSUE #3157: Cursor in different spot than where cursor appears and text gets deleted in different location
**Describe the bug**
When switching from one app to VSCode when VIM was in insert mode and I delete some text sporadically text is deleted from a different location than where the cursor appears.  Also, the select next occurrence doesn't function as it should when the VIM extension is installed.

Disabling this extension everything work fine, so it's definitely something when this extension is installed.

**To Reproduce**
1.  Go to '...'
2.  Click on '....'
3.  Scroll down to '....'
4.  See error

- Extension (VsCodeVim) version: v0.16.11
- VSCode version: 1.28.2
- OS: MacOS Mojave, and it happened on High Sierra as well

jpoon:
This seemed to do the trick to fix #3157. Instead of having an event handler for each modehandler (ie. text editor), declared a single event handler and delegated the event to the right mode handler. This should have the added benefit of improving perf.

To be perfectly honest however, I'm not entirely sure how this fixed the issue as the same callback code to update the position was still being called, albeit unnecessarily for all the other open text editors.

This PR also includes a minor refactor to add a wrapper around registering event handlers to ensure that we are disposing the events properly.

#3157 

### DIFF
--- a/extension.ts
+++ b/extension.ts
@@ -83,20 +83,23 @@ export async function activate(context: vscode.ExtensionContext) {
     configuration.reload();
   });
 
-  const textWasDeleted = event =>
-    event.contentChanges.length === 1 &&
-    event.contentChanges[0].text === '' &&
-    event.contentChanges[0].range.start.line !== event.contentChanges[0].range.end.line;
-
-  const textWasAdded = event =>
-    event.contentChanges.length === 1 &&
-    (event.contentChanges[0].text === '\n' || event.contentChanges[0].text === '\r\n') &&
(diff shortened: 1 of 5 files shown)


jpoon:
> To be perfectly honest however, I'm not entirely sure how this fixed the issue as the same callback code to update the position was still being called, albeit unnecessarily for all the other open text editors.

Yep, this assessment was correct. This PR doesn't actually fix the issue. Turns out I was using the wrong steps to repro. Once you exit and re-enter insert mode, the issue goes away.

PR #3416, opened 2019-01-28
Run `closeMarkersNavigation` on ESC. Fix #3367

### LINKED ISSUE #3367: Close error window <F8>
When I press F8, the hover error window shows up while the focus is still on the editor. In that case Esc cannot close the hover window. I have to click the window to focus on it then press Esc, which is really annoying.

Is it possible to pass the Esc to VS code so that it can close the error window without change focus? or altenatly is there a way to focus on that hover window without use mouse?

octref:

After running F8 to jump to next diagnostic location, ESC cannot dismiss the inline problems view (unless you click to focus into it).

See https://github.com/Microsoft/vscode/issues/66684#event-2101313157 for the command added.

- https://github.com/VSCodeVim/Vim/issues/3367
- https://github.com/Microsoft/vscode/issues/67086

Same addition as https://github.com/VSCodeVim/Vim/pull/1707.

### DIFF
--- a/src/actions/commands/actions.ts
+++ b/src/actions/commands/actions.ts
@@ -519,6 +519,8 @@ class CommandEsc extends BaseCommand {
       // of vscode's default "close notification" actions. I think we should
       // just add to this list as needed.
       await vscode.commands.executeCommand('closeReferenceSearchEditor');
+      await vscode.commands.executeCommand('closeMarkersNavigation');
+
       return vimState;
     }
 


xconverge:
LGTM thanks

octref:
@xconverge Can we make this a setting? A list of commands to run when pressing ESC. Try Option + F3 — ESC do not close the inline diff view either.

Through https://github.com/Microsoft/vscode/issues/36566#issuecomment-459231768 I get that `closeDirtyDiff` is another one. As you start to search for ESC you see a huge list.

image

PR #3418, opened 2019-01-29
fix: validate configurations once, instead of every key press

jpoon:

* load and validate remappings on configuration file changes instead of every single keypress. 
* fixed this circular dependency between `configuration` and `logger`. logger depended on configuration to configure its levels, configuration required logger to log config errors. 


jpoon:
I ended up deleting this unit test that rebinded leader,f,e,s. I’m not sure how that rebounding worked initially as f,<character> is an movement.

PR #3426, opened 2019-01-31
fix ds" with nested quotes and add some tests - fixes #3415

### LINKED ISSUE #3415: "Delete surrounding quotes" doesn't work in certain cases
**Describe the bug**

Delete surrounding quotes (`ds"`) doesn't work in certain cases.

**To Reproduce**

Enter the following to a buffer, place your cursor on `f` and press `ds"`.

```
""foo""
```

**Expected behavior**

It should remove the inner quotes. The end result should be `"foo"`. Instead, nothing happens. Similarly, `""foo"` should become `"foo`. Interestingly it works with other surrounding characters e.g. `((foo))` correctly becomes `(foo)`. 

- Extension (VsCodeVim) version: 1.0.3
- VSCode version: 1.30.2 
- OS: macOS

esetnik:

Fixes `ds"` with nested quotes e.g. `""te|st""` should become `"te|st"`.
Fixes `ds"` with cursor on close quote e.g. `"test|"` should become `test|`

#3415 

### DIFF
--- a/src/actions/motion.ts
+++ b/src/actions/motion.ts
@@ -1685,8 +1685,15 @@ export abstract class MoveQuoteMatch extends BaseMovement {
   public async execAction(position: Position, vimState: VimState): Promise<IMovement> {
     const text = TextEditor.getLineAt(position).text;
     const quoteMatcher = new QuoteMatcher(this.charToMatch, text);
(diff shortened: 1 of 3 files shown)


jpoon:
This is dope! Thanks especially for adding tests.

PR #3432, opened 2019-01-31
Adopt latest list navigation support

joaomoreno:

This PR introduces a new keybinding when in the list context (<kbd>/</kbd>) which invokes the `list.toggleKeyboardNavigation` in order to use the new upcoming keyboard navigation feature in VS Code 1.31.

The introduction is backwards compatible since the keybinding is only enabled if `listSupportsKeyboardNavigation` is true which is the case only from 1.31 onwards.

Related VS Code issue: https://github.com/Microsoft/vscode/issues/66995

This should be merged before VS Code 1.31 releases.


extension.ts
```diff
@@ -360,6 +360,8 @@ export async function activate(context: vscode.ExtensionContext) {
   }
 
   await Promise.all([
+    // This is in order to disable automatic keyboard navigation in lists
+    vscode.commands.executeCommand('setContext', 'listAutomaticKeyboardNavigation', false),
```
jpoon: We ended up wrapping this as we've found that `setContext` sometimes takes awhile to complete. It doesn't make much difference in this context, but to follow with the pattern, can you call:

```
  await VsCodeContext.Set('listAutomaticKeyboardNavigation', false);
```

package.json
```diff
@@ -718,5 +723,10 @@
     "tslint": "5.12.1",
     "typescript": "3.2.4",
     "vscode": "1.1.28"
+  },
+  "__metadata": {
```
jpoon: Ohh, secret fields -- what is this?

joaomoreno:
@jpoon Alright, merged and cleaned up! :+1:

joaomoreno:
Ahah I even created a commit which left those bad whitespaces in simply because I didn't want more changes in the PR. But sure, bot, I can leave them in!

test/configuration/configuration.test.ts
```diff
@@ -78,7 +78,7 @@ suite('Configuration', () => {
   });
 
   test('neovim disabled on missing path', async () => {
-    assert.equal(false, srcConfiguration.configuration.enableNeovim)
+    assert.equal(false, srcConfiguration.configuration.enableNeovim);
```
jpoon: Yikes, thanks for fixing this.

joaomoreno:
Thanks @jpoon!

bcpugh:
thanks guys! slightly off-topic:
any advice on setting **listSupportsKeyboardNavigation = false** without implementing VSCodeContext.set() inside of an extension (i.e. a settings.json entry?)

joaomoreno:
@bcpugh Do you have single-letter keybindings for lists regardless of using VIM?

FelikZ:
So when I am focused on (for example) file explorer, vim mode enabled and hit `/` then it should enter a search mode?

I did not get it how it works.

Because I have updated VSCode and vim plugin and whenever I hit `/` I got the old behaviour (which is good).

joaomoreno:
What was the old behaviour?

FelikZ:
@joaomoreno 
**Old:**
- Focus on editor, hit `/` - toggles vim search
- Focus on file explorer, hit `/` - nothing expected to happen
**New:**
- Focus on editor, same behaviour
- Focus on file explorer, hit `/` - filter/highlight `/` char (new feature)

So my question is, what this PR actually did, since in file editor it is the same behaviour as before and in file explorer (other lists) vim is not active there (was it?).

joaomoreno:
VIM sets `hjkl` as keybindings to navigate in file explorer.

FelikZ:
Thank you, it is clear now! Awesome job!

bcpugh:
@joaomoreno

We lost functionality for key binding of “a” to create new file when the file tree is focused (and other similar bindings). Only some of our team uses VIM, but I wanted to ask here since the latest VScode  release notes pointed us here. Can we somehow revert back to the old behavior without installing an extension like VIM?

joaomoreno:
@bcpugh My recommendation would be to simply switch to a keybinding with modifier key, eg <kbd>Ctrl N</kbd>.

asilvadesigns:
I'm very excited about this feature, awesome work, but it does not work for me.
With VIM enabled, and `workbench.list.keyboardNavigation = "simple"`, focusing on the tree and typing "/" does not enter highlight or filter mode. What am I missing, I'm not sure to enable this?

nomad-software:
How can I disable this feature added by this pull request?

macksta:
This is an epic feature!!  Good work!!

Molunerfinn:
@asilvadesigns Hi~ I have found how to fix your problem!
You just need to set the 
`"workbench.list.keyboardNavigation": "filter"` or 
`"workbench.list.keyboardNavigation": "highlight"`
Since `list.toggleKeyboardNavigation` is to toggle navigation mode to `filter` or `highlight`. While you just set to `simple`, so the navigation mode is just `simple` after toggling(whether it is on or off).

controversial:
Is there a way to disable the tree keyboard navigation without installing a plugin?

nomad-software:
This is a really annoying change, can we please have a way of disabling this?

thomazmoura:
> Is there a way to disable the tree keyboard navigation without installing a plugin?

> This is a really annoying change, can we please have a way of disabling this?

There is no need to install any plugin to disable this feature. If you just want to turn off the tree widget completely, use this setting on VS Code:

```json
"workbench.list.keyboardNavigation": "simple"
```

joaomoreno:
> Is there a way to disable the tree keyboard navigation without installing a plugin?

> This is a really annoying change, can we please have a way of disabling this?

I don't understand guys. This PR exists in order not to make keyboard filtering annoying for VIM users, who use `hjkl` keys to navigate in lists/trees.

What is annoying about this?

nomad-software:
@joaomoreno because you're are forcing your way of working on other people. This has nothing to do with Vim, this is an _editor_ change that alters the tree view and not an optional one at that so we _all_ have to have it now! When I install a Vim plugin for any given editor, I expect it to give me Vim keybindings for the editor pane, not to start changing things outside of that unless it's strictly opt-in.

I wish the maintainer of this repo would limit pull requests to implement actual features of Vim and stop straying into altering the actual editor itself.

joaomoreno:
@nomad-software Can you explain me in steps what currently doesn't work for you?

thomazmoura:
> @joaomoreno because you're are forcing your way of working on other people. This has nothing to do with Vim, this is an _editor_ change that alters the tree view and not an optional one at that so we _all_ have to have it now! When I install a Vim plugin for any given editor, I expect it to give me Vim keybindings for the editor pane, not to start changing things outside of that unless it's strictly opt-in.
> 
> I wish the maintainer of this repo would limit pull requests to implement actual features of Vim and stop straying into altering the actual editor itself.

@nomad-software I believe there is misunderstanding going on here about what this PR is about. It fixes the feature **the VS Code team added to do VS Code** and that was conflicting with VIM users. It doesn't add any feature to VS Code itself - **the feature that adds the tree widget came directly from the VS Code team on their repo and it's not directly influenced by this repo or related to this PR**.

And although I agree it might have been a better choice to leave it as opt-in (or at least easier to turn off  after the upgrade) it's still opt-out (as most VS Code features are). Just check my former answer:

> There is no need to install any plugin to disable this feature. If you just want to turn off the tree widget completely, use this setting on VS Code:
> ``` json
> "workbench.list.keyboardNavigation": "simple"
> ```

So just to try to make things clearer: **if you don't use or want to use VIM this PR does not affect you any way and there's no need or point in installing this extension**. And if you want to disable the feature added to VS Code on 1.31 just use the code above.

dmazin:
Has anyone figured out how to actually open a file in this mode? `o` toggles directory expansion, but doesn't open files for editing.

Issue #3433, opened 2019-01-31
Extension crash or hangs when failing to call nvim


rodcloutier:
**Describe the bug**

vscode extension crash/hangs when failing to call `nvim`. Receive a `ENOENT` error.
Extension cannot be used anymore.

Console output:
```
/usr/share/code/resources/app/out/vs/workbench/workbench.main.js:303 [Extension Host] Neovim: error: Error spawning neovim. Error=spawn /usr/local/bin/nvim ENOENT.
t.log @ /usr/share/code/resources/app/out/vs/workbench/workbench.main.js:303
/usr/share/code/resources/app/out/vs/workbench/workbench.main.js:2508 Neovim: Error spawning neovim. Error=spawn /usr/local/bin/nvim ENOENT.
t.onDidNotificationChange @ /usr/share/code/resources/app/out/vs/workbench/workbench.main.js:2508
2/usr/share/code/resources/app/out/vs/workbench/workbench.main.js:303 [Extension Host] Error: Unexpected SIGPIPE
	at process.on (/usr/share/code/resources/app/out/bootstrap.js:5:80)
	at emitNone (events.js:106:13)
	at process.emit (events.js:208:7)
	at Signal.process.on.wrap.onsignal (internal/process.js:208:44)
t.log @ /usr/share/code/resources/app/out/vs/workbench/workbench.main.js:303
```
**To Reproduce**
1. Create new file
2. Type `:W`

**Expected behavior**
Vscode not freezing or crashing

**Environment (please complete the following information):**

- Extension (VsCodeVim) version: 1.0.5
- No other extension enabled
- Nvim:  0.3.3
- VSCode
Version: 1.30.2
Commit: 61122f88f0bf01e2ac16bdb9e1bc4571755f5bd8
Date: 2019-01-07T22:49:48.319Z
Electron: 2.0.12
Chrome: 61.0.3163.100
Node.js: 8.9.3
V8: 6.1.534.41
OS: Linux x64 4.19.15-300.fc29.x86_64

bradamelton:
I am also experiencing this. It works for a few seconds after startup, then crashes and no typing is possible. Just started today.

buwilliams:
+1

jpoon:
1. Was neovim working before w/ vscodevim < 1.0.5?
2. Is this happening for all ex-commands?

jpoon:
With nvim 0.3.3 and vscodevim (on master), I can't repro.

[...]

PR #3437, opened 2019-02-01
fix: check neovim configurations and timeout on nvim attach

jpoon:
Check neovim configurations for validity and disables neovim if anything is amiss. Timeout when trying to attach neovim process.

#3433


xconverge:
This broke my setup where I just had 'nvim' as the path

This then lets me use the same config (synced with settings sync) and environment variables with both Linux and Windows

jpoon:
Does that mean you have nvim running in the same working directory as vscodevim?

PR #3467, opened 2019-02-07
fix: remember position when jumping to definition

ercolanelli-leo:

It records a jump whenever following a symbol to its definition, allowing one to easily go back after the definition has been checked using `C-o`.

### DIFF
--- a/src/actions/commands/actions.ts
+++ b/src/actions/commands/actions.ts
@@ -2620,6 +2620,8 @@ class CommandGoToDefinition extends BaseCommand {
   public async exec(position: Position, vimState: VimState): Promise<VimState> {
     const oldActiveEditor = vimState.editor;
 
+    const jumpFrom = Jump.fromStateNow(vimState);
+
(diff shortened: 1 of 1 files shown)


src/actions/commands/actions.ts
```diff
@@ -2629,6 +2631,8 @@ class CommandGoToDefinition extends BaseCommand {
       vimState.cursorStopPosition = Position.FromVSCodePosition(vimState.editor.selection.start);
     }
 
+    vimState.globalState.jumpTracker.recordJump(jumpFrom, Jump.fromStateNow(vimState));
```
jpoon: We should already be doing this here: https://github.com/VSCodeVim/Vim/blob/master/src/mode/modeHandler.ts#L400. 

Is there a bug somewhere else then?

jpoon:
As there hasn't been any recent activity on this PR, I am closing this out. Feel free to re-open after addressing comments.

PR #3472, opened 2019-02-07
Set < and > marks when yanking in visual mode.

### LINKED ISSUE #3452: Error when using the `> motion
**Describe the bug**
`TaskQueue: Error running task. Failed to handle key=>. Cannot read property 'position' of undefined.` when doing `` `> `` after a yank from visual mode.

**To Reproduce**

1.  Select a few rows in visual mode.
2. Yank
3. Press `` `> ``

**Expected behavior**
Cursor should jump to end of the selection made in step 1 above. See http://vimdoc.sourceforge.net/htmldoc/motion.html#%60%3E .

- Extension Name: `vim`
- Extension Version: `1.0.7`
- OS Version: `Darwin x64 17.7.0`
- VSCode version: `1.30.2`

rickythefox:
Marks `<` and `>` should be set when yanking in visual mode. 
See http://vimdoc.sourceforge.net/htmldoc/motion.html#%60%3E .

fixes #3452


src/actions/operator.ts
```diff
@@ -280,6 +280,12 @@ export class YankOperator extends BaseOperator {
 
     Register.put(text, vimState, this.multicursorIndex);
 
+    if (vimState.currentMode === ModeName.Visual || vimState.currentMode === ModeName.VisualLine) {
+      end = new Position(end.line, end.character - 1);
```
jpoon: What is `end.character` = 0?

rickythefox: Higher up in the same method it's `end = new Position(end.line, end.character + 1);`. I just reset it.

jpoon: As this is specific to the `if` block, maybe create a temp variable? Nobody is using `end` after this, but this would be prone to bugs if we modified it's value and left it like this.

rickythefox: Yep, good idea.

src/actions/operator.ts
```diff
@@ -261,25 +261,30 @@ export class YankOperator extends BaseOperator {
     if (end.isEarlierThan(start)) {
       [start, end] = [end, start];
     }
-    end = new Position(end.line, end.character + 1);
+    let extendedEnd = new Position(end.line, end.character + 1);
```
rickythefox: Instead of modifying the `end` parameter we create a local variable.

src/actions/operator.ts
rickythefox: b854385 fixes this.

src/actions/operator.ts
jpoon: Err, sorry I meant just within the `if` block.

rickythefox: I understand. 

The original code mutad the `end` parameter by adding 1 to its length, which is rather ugly.

I introduced a variable to hold the mutated parameter and used the original parameter in my code.

What's your opinion?

Issue #3522, opened 2019-02-25
Activating extension 'vscodevim.vim' failed: Invalid regular expression


tetchel:
**Describe the bug**
When I launch VS Code in Extension Development Host mode, I get an error popup (newlines added by me) saying the following:

```
Activating extension 'vscodevim.vim' failed: Invalid regular expression: 
/([^\s`~!@#\$%\^&\*\(\)\-=\+\[\{\]\}\\\|;:'",\.<>/\?](?:(?:(?<=[A-Z_])[A-Z](?=[\sA-Z0-9`~!@#\$%\^&\*\(\)\-=\+\[\{\]\}\\\|;:'",\.<>/\?_]))+|(?:(?<=[0-9_])[0-9](?=[\sA-Z0-9`~!@#\$%\^&\*\(\)\-=\+\[\{\]\}\\\|;:'",\.<>/\?_]))+|(?:(?<=[_])[_](?=[\s`~!@#\$%\^&\*\(\)\-=\+\[\{\]\}\\\|;:'",\.<>/\?_]))+|[^\sA-Z0-9`~!@#\$%\^&\*\(\)\-=\+\[\{\]\}\\\|;:'",\.<>/\?_]*))|[`~!@#\$%\^&\*\(\)\-=\+\[\{\]\}\\\|;:'",\.<>/\?]+|$^/: 
Invalid group.
```

My condolences to whoever has to work on anything to do with that regex 😨!

**To Reproduce**
1.  Launch VS Code in Extension Development Host mode.
2. That's it, the error appears every time.

I can't get the error to pop up when **not** using Extension Development Host mode. In addition, I tried launching the EDH with two different extensions I work on, and both had the same activation failure message appear, so I don't think it's related to either of my extensions.

**Expected behavior**
I expect the extension to activate without failing :P

**Screenshots**
![image](https://user-images.githubusercontent.com/8888438/53371323-ec220c80-391d-11e9-808b-ce5f78120d72.png)

**Environment (please complete the following information):**

- Extension (VsCodeVim) version: 1.1.0
- VSCode version: 1.30.2
- OS: MacOS Mojave

**Additional context**
I just started noticing this error today. I have been using your extension to develop my own extension for ~6 months so this seems to be a new issue.

Here are my settings, they haven't changed in a long time though so I don't think they are the cause.
```
"vim.insertModeKeyBindings": [
        {
            "before": ["j", "j"],
            "after": ["<Esc>"]
        },
        {
[...]

PR #3525, opened 2019-02-26
Only use regex lookbehind where supported

JKillian:

Users on older versions of VSCode were hitting bugs because of lookbehinds in the camel case motion regex. This should fix the issue for those users.

Fixes https://github.com/VSCodeVim/Vim/issues/3522

I didn't get to test this myself on an older version of VSCode. But I think it should work?


jshap70:
This is very similar to what I was testing. Looks fine. Special note that if a user is building vscodevim on an older branch that some test cases will fail, but seems fine to just say that is disallowed.

JKillian:
Closing this since nobody has complained that they can't upgrade VSCode yet. Will keep the branch around for now though in case it becomes necessary

CompuIves:
Hey! We're using VSCodeVIM inside CodeSandbox (https://codesandbox.io/s/new), which runs VSCode and can run VSCode extensions. For our VIM Mode we use VSCodeVIM (thanks for building this!).

This worked great so far, but we found out that lookbehind regexpressions are not supported in Firefox. I can imagine that this is a very unusual request but would you be open to a PR that adds the lookbehind check and disables it if it detects Firefox/Safari?

JKillian:
reopening for consideration since it seems like there are some legit use cases

JoeSchr:
is this working? I'm trying to access code-server via firefox reality and it gives me this error.

would be really cool if I could code in VR and so far this is the only roadblock :(

anyways thanks for building VSCodeVim, it's really fine!

shritesh:
This was removed again in https://github.com/VSCodeVim/Vim/commit/1d273ebb334b3f7fbf381c32a609b911e5402858#diff-4eb584e50b8970dfd9f200f6d934f418

DAddYE:
> This was removed again in [1d273eb#diff-4eb584e50b8970dfd9f200f6d934f418](https://github.com/VSCodeVim/Vim/commit/1d273ebb334b3f7fbf381c32a609b911e5402858#diff-4eb584e50b8970dfd9f200f6d934f418)

I'm using vscode 1.46.1 through code-server on ipad and I'm getting:

```
Activating extension 'vscodevim.vim' failed: Invalid regular expression: invalid group specifier name.
_logMessageInConsole — abstractExtensionService.ts:396
_logOrShowMessage — abstractExtensionService.ts:410
(anonymous function) — mainThreadExtensionService.ts:70
asyncFunctionResume
_invokeHandler — rpcProtocol.ts:387
_receiveRequest — rpcProtocol.ts:303
_receiveOneMessage — rpcProtocol.ts:230
fire — event.ts:587
(anonymous function) — webWorkerExtensionHostStarter.ts:70
console.ts:137
```

J-Fields:
Ah, I hadn't considered us running in the browser when I removed that. I'll fix it momentarily.

DAddYE:
> Ah, I hadn't considered us running in the browser when I removed that. I'll fix it momentarily.

Thanks so much guys!

DAddYE:
> Ah, I hadn't considered us running in the browser when I removed that. I'll fix it momentarily.

Seems is still not working, I tested from master and still has errors with some grouped regex.
I'll try in the next few days to see if I can profile/debug the extension, because currently the inspector isn't super helpful.

J-Fields:
@DAddYE Just created #5072, I'll try to resolve when I have a chance

Immortalin:
@J-Fields any luck?

PR #3531, opened 2019-02-27
Naive fix for #2032

### LINKED ISSUE #2032: g$ moves to next visual line

--------

**This is  a BUG REPORT**:

**Environment**:

- **VSCode Version**: 1.16.1
- **OS**: macOS 10.12.6

**What happened**:
Pressing `g$` moved to the *first* character in the *next* visual line.

moved to the *last* character in the *current* visual line.

**How to reproduce it**:
Press `g$` in the start of a soft-wrapped line.

randomPoison:

Adds a naive fix for #2032 that simply moves the cursor back by one character when moving to the end of a wrapped line.

Fixes #2032

This provides a naive fix for the issue in #2032 based on a [previous attempt by another contributor](https://github.com/Strafos/Vim/commit/262f22784b75eb3046e9ee663280bd71e46884b3) that appears to never have been submitted in a PR. While this fix seems to work in my testing (and may well be sufficient to fix the bug), it strikes me as being a bit of a hack. Would it make more sense to dig into `Position.FromVSCodePosition()` and see why it's returning the wrong position in the first place? I don't have any experience with this extension's codebase, so I wanted to at least get some feedback on the naive version before digging into a more involved solution.

If it seems like a more involved fix is called for, I'd appreciate any pointers you can offer for putting together a better fix! :grin:


jpoon:
Hi, sorry for the delayed review. I'm not certain if this will address the issue. Particularly, we need to a way to figure out the line wrapping with VSCode. This fix doesn't take that into account so if I'm reading it correctly it'll move the cursor left regardless of if it's soft wrapped or not.

randomPoison:
Ah yeah, since I put this PR up I noticed that it doesn't handle wrapped lines. That said, I've also noticed that on the whole VSCodeVim doesn't seem to handle wrapped lines well, so I wonder if deeper changes are necessary in order to make movement handling more robust 🤔 Unfortunately I haven't had the time to dig into the issues I've seen, so I can't provide more details at the moment 😞 

Do you think this change is worth merging if it only works for non-wrapped lines?

jpoon:
> Do you think this change is worth merging if it only works for non-wrapped lines?

No, as this will result in the incorrect behaviour on non-wrapped lines.

randomPoison:
I'm going to close this, then, since I don't have time to follow up with a more robust fix right now 😞

PeterDavidson:
I added a comment to #2032 that has another (partial) fix for this problem. It would be nice to find the root cause, but I don't understand the code enough to attempt a fix (yet).

Issue #3455, opened 2019-02-05
Cursor jumps to position in vertical split after delete


flowchartsman:
**Describe the bug**
If I have the same file open in a vertical split, and the cursor is at another position in the other split, if I hit "j" or "k" to move after deleting, the cursor will jump to the same line as it is in the other split

**To Reproduce**

1.  Open a file
2.  `:vs`
3.  move to another line in the split with j or k, preferably one off screen to heighten the problem
4.  return to the first buffer and navigate with j or k to a higher line
5. `dd`
6. `j` or `k`

**Expected behavior**
Cursor moves up or down one line normally

**Actual behavior** cursor jumps

**Screenshots**
In the following gif, I delete a line with `dd`, and hit `j` in an attempt to scroll down.
![vsjumpbug](https://user-images.githubusercontent.com/926884/52297225-40a12000-294e-11e9-948f-675b6d4ee9f2.gif)

**Environment (please complete the following information):**

- Extension (VsCodeVim) version: 1.0.7
- VSCode version: 1.30.2
- OS: macOS 10.13.6

**Additional context**

flowchartsman:
Seems related to #2688, but that should be fixed in this version, so I'm submitting again.

captaincaius:
I can say I've noticed this myself recently too, but with horizontal splits.

Hopefully I'll be able to look at it by this weekend or next, unless someone beats me to the punch.

c10b10:
It's happening to me too. Any quickfix to this?

flowchartsman:
> It's happening to me too. Any quickfix to this?

For future reference, and because I was guilty of this so many times, it's generally encouraged to just pile on the thumbs up on the first comment to say you are also affected by it instead of making a comment.

jpoon:
Upstream issue: https://github.com/Microsoft/vscode/issues/15178.

Full background: https://github.com/Microsoft/vscode/issues/51001#issuecomment-397177893

flowchartsman:
<img width="476" alt="api access to open editors issue 15178 microsoft vscode 2019-02-24 20-52-37" src="https://user-images.githubusercontent.com/926884/53309515-4071b180-3876-11e9-818e-65da65fcc639.png">

Oof. Looks like this one could be with us for awhile.

coddingtonbear:
I'm not completely convinced I've fixed the issue via #3539, but I believe I have.  If somebody could give it a shot and let me know if that works for them, too, I'd appreciate it.

[...]

PR #3539, opened 2019-02-28
Possible fix for #3455: Compare EditorIdentity when handling window.onDidChangeTextEditorSelection.

coddingtonbear:

This PR changes two things:
* Alters EditorIdentity such that it uses the `ViewColumn` of the editor for differentiating between multiple editors open to the same file using changes long ago proposed by @lukaszb (thanks for posting a patch!).
* Updates our handler for `vscode.window.onDidChangeTextEditorSelection` such that it uses that EditorIdentity when deciding whether to handle an incoming event instead of simply comparing documents.

#3455 

### DIFF
--- a/extension.ts
+++ b/extension.ts
@@ -254,9 +254,11 @@ export async function activate(context: vscode.ExtensionContext) {
     context,
     vscode.window.onDidChangeTextEditorSelection,
     async (e: vscode.TextEditorSelectionChangeEvent) => {
+      const eventEditorIdentity = new EditorIdentity(e.textEditor);
+      const activeEditorId = new EditorIdentity(vscode.window.activeTextEditor);
       if (
         vscode.window.activeTextEditor === undefined ||
-        e.textEditor.document !== vscode.window.activeTextEditor.document
+        !eventEditorIdentity.isEqual(activeEditorId)
       ) {
         // we don't care if there is no active editor
         // or user selection changed in a paneled window (e.g debug console/terminal)
--- a/src/editorIdentity.ts
+++ b/src/editorIdentity.ts
@@ -2,20 +2,30 @@ import * as vscode from 'vscode';
 
 export class EditorIdentity {
   private _fileName: string;
+  private _viewColumn: number;
 
   constructor(textEditor?: vscode.TextEditor) {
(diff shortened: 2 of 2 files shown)


jpoon:
Iirc, there is a case where this breaks undo. https://github.com/VSCodeVim/Vim/issues/2688

coddingtonbear:
Yes!  I have seen a situation in which it seems like the undo history is shared between panes now, and when undoing something in one pane, it might undo an action that took place in the other.  I wasn't immediately sure that that was a regression this change created, but am ?glad to hear that you can confirm that it is.

Would it be your opinion that this approach toward solving this problem is unworkable (meaning, I shouldn't bother spending time figuring out how vscode-vim handles undo queues), or is it worthwhile to keep digging into things to see if I can fix undo handling with the changes in this PR?

jpoon:
This was an old issue so not sure if I recall the details but the current behaviour was the lesser of two evils given the limitation with the vscode API there's an upstream issue somewhere (On my phone so can't link).

At the moment, we only have one notion of an editor id but I feel like we might need to split this up as there Is state specific to:

* Unique to the text editor -- cursor position, movements, action history
* Unique to the file -- undo/redo

coddingtonbear:
Just to be clear, if we're expecting undo/redo to be shared among all open editors to the same file -- and in hindsight, I can't imagine it any other way -- I *think* this is operating as expected for me so far in the ways you've described above.

The weird thing I _do_ think isn't working is that search/replace within a visual selection isn't working. I haven't dug into it very deeply, but given the things this PR changed, I can think of several ideas about what might've become broken and will try to find a little time to dig deeper this week.

J-Fields:
Hey @coddingtonbear, thanks for this PR!
I'm seeing some misbehavior with undo/redo. Here's an example:
1. Open new document
2. `:vsp`
3. `o1<esc>o2<esc>o3<esc>`
4. Undo/redo to see they operate on single line additions
5. Go to other side and see undo undoes all 3 lines
6. Try to redo on first side - doesn't work

coddingtonbear:
Oh, sure, this totally doesn't work in the latest version of vscodevim, @J-Fields, and I don't have a lot of time to update it, so I think I'm just going to close this PR :man_shrugging: .

PR #3544, opened 2019-03-01
Add small deletions to small delete register

### LINKED ISSUE #3492: The small delete register "- doesn't work
**Describe the bug**
Charactars deleted by "nx" in one line shoud be stored in the small delete register "-.
But they are stored in the unnamed register "".

- Extension (VsCodeVim) version:1.0.8
- VSCode version:1.31.1
- OS:win7 64

rickythefox:
Adds small deletions to small delete register. 
Fixes #3492.

> Small delete register "-
> This register contains text from commands that delete less than one line,
> except when the command specifies a register with ["x].
> {not in Vi}


rickythefox:
Will fix the failing test and re-open.

src/register/register.ts
```diff
@@ -315,19 +320,27 @@ export class Register {
         Register.registers['0'].registerMode = vimState.effectiveRegisterMode;
       }
     } else if (
-      baseOperator instanceof DeleteOperator &&
+      (baseOperator instanceof DeleteOperator || baseOperator instanceof ActionDeleteChar) &&
```
JKillian: should this code also check for `ActionDeleteCharWithDeleteKey` and `ActionDeleteLastChar`? I think they also would trigger writes to the `-` register.

You may also fix some slightly orthogonal bugs with your code here. As far as I can tell, VsCode Vim isn't correctly writing multiline deletes with `x`, `X`, or `<del>` to the numbered registers. By checking for those actions, you'd actually fix that bug along with the bug you intended to fix. (Not sure if there's a ticket or not for this one.)

I wonder if there's an easier way to just check if the underlying operator is `DeleteOperator` (since `x`, `X`, and `<del>` all use it under the hood) instead of needing to check for each incantation of it separately.

rickythefox: Thanks for the suggestion, will fix!

test/register/register.test.ts
```diff
@@ -256,13 +256,54 @@ suite('register', () => {
     }
   });
 
-  test('Small deleteion is stored in small delete register', async () => {
+  test('Small deleteion using x is stored in small delete register', async () => {
```
JKillian: typo, "deleteion" (also in tests below)

test/register/register.test.ts
```diff
@@ -256,13 +256,54 @@ suite('register', () => {
...
     await modeHandler.handleMultipleKeyEvents('itest1\ntest2\ntest3'.split(''));
 
     await modeHandler.handleMultipleKeyEvents(['<Esc>', 'g', 'g', '2', 'x', 'j', '"', '-', 'p']);
 
     assertEqualLines(['st1', 'tteest2', 'test3']);
   });
-});
\ No newline at end of file
+
+  test('Small deleteion using x is stored in small delete register', async () => {
```
JKillian: test name is incorrect "x" -> "<Del>"

JKillian:
Looks like a build flake to me though, you might be able to repush to trigger another build:

```
Current working directory: /home/travis/build/VSCodeVim/Vim
Downloading VS Code into "/home/travis/build/VSCodeVim/Vim/.vscode-test/stable" from: https://vscode-update.azurewebsites.net/1.31.1/linux-x64/stable
internal/streams/legacy.js:59
      throw er; // Unhandled stream error in pipe.
      ^
Error: read ECONNRESET
    at TLSWrap.onread (net.js:622:25)
The command "npm test --silent" exited with 1.
Done. Your build exited with 1.
```

PR #3580, opened 2019-03-13
 fix can't activation with Java Pack

rectcircle:
to see https://github.com/Microsoft/vscode-java-pack/issues/113


jpoon:
I don't quite understand. What's the bug? Looking over the code, I don't see anything apparent.

rectcircle:
Sorry, this problem should be solved by the Microsoft/vscode-java-pack project. The purpose of this PR is to circumvent the problem of Microsoft/vscode-java-pack, so that the VIM extension can be used in the case of Microsoft/vscode-java-pack. https://github.com/Microsoft/vscode-java-pack/issues/113#issuecomment-478891322

PR #3617, opened 2019-03-23
When a search is cancelled, revert to previous search state

### LINKED ISSUE #3616: Search state should not change until a new search command is completed
**Describe the bug**
The search state is currently changed when you **start** entering a search (when you first hit `/`). It should only change when you **finish** (when you hit enter after `/`, when you use `*`, etc.)

**To Reproduce**

1.  Do a search with `/abc<enter>`
2.  Start another search with `/xyz` but **do not** press enter
3.  Cancel the search with `<esc>`
4.  Observe: if you have `hlsearch` enabled, you'll see instances of `xyz` highlighted. In addition, hitting `n` will go to the next instance of `xyz`

**Expected behavior**
After the search for `xyz` is cancelled, instances of `abc` should still be highlighted and hitting `n` should cycle through them.

- Extension (VsCodeVim) version: 1.2.0
- VSCode version: 1.32.3
- OS: Windows 10

J-Fields:
When a search is cancelled (`<Esc>` after `/`), the search state should revert to the previous state, rather than being overwritten. This means the previous search should still be highlighted if `hlsearch` is enabled and `n` should go to the next instance of that search.

Fixes #3616

### DIFF
--- a/src/actions/commands/actions.ts
+++ b/src/actions/commands/actions.ts
@@ -502,7 +502,6 @@ class CommandEsc extends BaseCommand {
     ModeName.VisualLine,
     ModeName.VisualBlock,
     ModeName.Normal,
-    ModeName.SearchInProgressMode,
     ModeName.SurroundInputMode,
     ModeName.EasyMotionMode,
     ModeName.EasyMotionInputMode,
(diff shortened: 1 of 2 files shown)


src/actions/commands/actions.ts
```diff
@@ -958,8 +951,18 @@ class CommandEscInSearchMode extends BaseCommand {
...
 
   public async exec(position: Position, vimState: VimState): Promise<VimState> {
-    await vimState.setCurrentMode(ModeName.Normal);
-    vimState.globalState.searchState = undefined;
+    const searchState = vimState.globalState.searchState!;
+
+    vimState.cursorStopPosition = searchState.searchCursorStartPosition;
+
+    const prevSearchList = vimState.globalState.searchStatePrevious;
+    if (prevSearchList) {
```
jpoon: nit: can be 

```
vimState.globalState.searchState = prevSearchList ? ...

J-Fields: Done.

Issue #3507, opened 2019-02-21
Preview file from explorer is not tracked as jump


yaegaki:
**Describe the bug**
Preview file from explorer is not tracked as jump
so i can not back to previous file.

**To Reproduce**

1.  set `workbench.editor.enablePreview` to `true`.
2.  preview file from explorer.
3.  preview other file from explorer.
4.  type `<C-o>`
5. can not back to first preview file.

**Expected behavior**
Preview file is tracked as jump.

**Screenshots**
![vim](https://user-images.githubusercontent.com/4941773/53181680-5c3d4380-363b-11e9-84b2-7dabc7ff4157.gif)

* preview `lexter.text.ts` -> `parser.test.ts` -> `scanner.test.ts`.
* type `<C-o>`
* expect `parser.test.ts` was opend, but not

**Environment (please complete the following information):**

- Extension Version: `1.0.8`
- OS Version: `Windows_NT x64 10.0.17134`
- VSCode version: `1.31.1`

**Additional context**

`vscode.workspace.onDidCloseTextDocument` event is fired before `vscode.window.onDidChangeActiveTextEditor` event.
so old ModeHandler has already been deleted from ModeHandlerMap when fire `vscode.window.onDidChangeActiveTextEditor`.

Before this commit(61cba7a628f5e9a1f72d9df84fa8c96d2b654718),  Deletion was delayed because await was used in `vscode.workspace.onDidCloseTextDocument` handler.

PR #3630, opened 2019-03-27
Assign lastClosedModeHandler when onDidCloseTextDocument.

yaegaki:
Can track file open from explorer/quick open when `workbench.editor.enablePreview` is enabled.

#3507


jpoon:
Sorry for the delay in review, we had a 👶 recently so my spare time is pretty much filled with 🍼 and 😴 .

Good catch!

PR #3658, opened 2019-04-05
Fixes #2754. Ctrl+d/u pull cursor along when screen moves past cursor

### LINKED ISSUE #2754: editor.smoothScrolling breaks C-d / C-u

**What happened**: Pressed Ctrl-D, cursor does not stay in its position on screen. When it is at the top, the scroll happens but then the window bounces back.

**What did you expect to happen**: Scrolling to not bounce back.

**How to reproduce it (as minimally and precisely as possible)**: Enable `editor.smoothScrolling` and try out the new behavior of C-d / C-u.

Basically, I'm about 99% sure it comes down to [this line](https://github.com/VSCodeVim/Vim/pull/2723/files#diff-ab811cd6f8cfa81bd0cb8223c86e72bfR677) and the equivalent in the duplicate function for C-u. When smooth scrolling is active, the await resolves before the visible ranges have been updated, so the "new first line" is always equal to the old first line, which locks the cursor within the visible screen area. Cc @rebornix 

- Extension (VsCodeVim) version: 0.13.1
- VSCode version: 1.24.1
- OS version: Windows 10 b17692

mayhewluke:

Current implementation of ctrl+d/ctrl+u does not allow scrolling past the cursor's original position when smooth scrolling is enabled (see #2754). With this PR, when ctrl+d/ctrl+u scroll the screen such that the cursor would be off-screen, the cursor is placed at the first/last line respectively rather than the screen snapping back to keep the cursor in view.

#2754 

As I understand it, when VSCodeVim executes ctrl+d or ctrl+u, it checks the range of lines on the screen and then places the cursor at an offset (relative to the new first line) equal to the offset prior to scrolling. This is the ideal behavior normally, but when smooth scrolling is enabled the new range of lines is not available to the VSCodeVim command (due to scrolling happening asynchronously). It ends up accessing the original range of lines instead, which means the new position is calculated relative to (and identical to) the original position. This causes the cursor to not be moved and prevents the screen from scrolling any further than the cursor.

This change "fixes" the issue by using a built-in feature of VSCode's scrolling command. By setting `revealCursor` to true in the scroll command VSCode will ensure the cursor gets "pulled along" when scrolling it off-screen, and telling VSCodeVim to use that new position ensures VSCodeVim doesn't reset it back.

**Caveat:** This doesn't keep the proper vertical offset when scrolling, so isn't quite a "true" fix, especially as the functionality differs slightly between having smooth scrolling on and off (it as at least usable now, though).


mayhewluke:
@jpoon Pretty sure this PR is ready for merge, assuming the tradeoffs are acceptable. Got time for a :+1:/:-1:? :smile:

src/actions/commands/actions.ts
```diff
@@ -671,23 +672,27 @@ class CommandMoveHalfPageDown extends CommandEditorScroll {
       to: this.to,
       by: this.by,
       value: timesToRepeat,
-      revealCursor: false,
+      revealCursor: smoothScrolling ? true : false,
```
jpoon: Can be simplified to just `smoothScrolling`? No need for ternary operator.

src/actions/commands/actions.ts
```diff
@@ -671,23 +672,27 @@ class CommandMoveHalfPageDown extends CommandEditorScroll {
...
         [ModeName.Visual, ModeName.VisualBlock, ModeName.VisualLine].indexOf(
           vimState.currentMode
         ) >= 0,
     });
 
-    const newFirstLine = editor.visibleRanges[0].start.line;
-    const newPositionLine = newFirstLine + lineOffset;
-    let newPosition;
+    let newPosition: Position;
+    if (!smoothScrolling) {
```
jpoon: nit: mind reversing this logic such that it's `if (smoothScrolling)`. IMO, affirmation always easier to read than negation.

src/actions/commands/actions.ts
```diff
@@ -671,23 +672,27 @@ class CommandMoveHalfPageDown extends CommandEditorScroll {
...
+      const newFirstLine = editor.visibleRanges[0].start.line;
```
jpoon: same as `firstLine` var?

src/actions/commands/actions.ts
```diff
@@ -718,15 +724,20 @@ class CommandMoveHalfPageUp extends CommandEditorScroll {
       to: this.to,
       by: this.by,
       value: timesToRepeat,
-      revealCursor: false,
+      revealCursor: smoothScrolling ? true : false,
```
jpoon: see comment above.

src/actions/commands/actions.ts
```diff
@@ -718,15 +724,20 @@ class CommandMoveHalfPageUp extends CommandEditorScroll {
...
       select:
         [ModeName.Visual, ModeName.VisualBlock, ModeName.VisualLine].indexOf(
           vimState.currentMode
         ) >= 0,
     });
 
-    let newFirstLine = editor.visibleRanges[0].start.line;
-    let newPosition = new Position(newFirstLine + lineOffset, startColumn);
+    let newPosition: Position;
+    if (!smoothScrolling) {
```
jpoon: see similar comment above.

src/actions/commands/actions.ts
mayhewluke: Same as `newFirstLine` from before this commit ([line `681`](https://github.com/VSCodeVim/Vim/pull/3658/files#diff-ab811cd6f8cfa81bd0cb8223c86e72bfL681) above). `firstLine` is the first line before the scroll, `newFirstLine` is the first line after the scroll.

src/actions/commands/actions.ts
mayhewluke: :man_facepalming: Good catch, my bad.

mayhewluke:
Thanks for the review @jpoon - I believe everything should be fixed/addressed now.

Issue #3665, opened 2019-04-08
Word seperate doesn't works well


yifeidai:
**Describe the bug**
The word-seperate doesn't work well. For example, "a.b.c" is regarded as a word but not three words. In function "hellowWorld(param1, param2)", "hellowWorld(param1" is regarded as a word. It works well a few days ago, but it works wrong now. I don't change any configuration these days. So I can't use keys like "w", "e", "b", "*" properly.
It is now proved that **in old vim version: "v1.2.0"** it works well.

**To Reproduce**
Here is my vim config in vscode.
```
    "vim.disableExtension": false,
    "vim.easymotion": true,
    "vim.sneak": true,
    "vim.incsearch": true,
    "vim.useSystemClipboard": true,
    "vim.useCtrlKeys": true,
    "vim.hlsearch": true,
    "vim.leader": "<space>",
    "vim.handleKeys": {
      "<C-a>": false,
      "<C-f>": false,
      "<C-c>": false,
      "<C-v>": false,
      "<C-h>": false,
      "<C-shift-l>": false,
      "<C-alt-up>": false,
      "<C-alt-down>": false,
      "<C-[>": false,
      "<C-]>": false,
      "<C-w>": false,
      "<C-x>": false,
    },
```
And here is word seperate config.
```
    "editor.wordSeparators": "`~!@%^&*()=+[{]}\\|;:'\",.<>-/?",
```

The vscode's word seperate works well. It means that if I use Ctrl+arrow, "a.b.c" isn't regarded as a whole word.

**Environment (please complete the following information):**

- Extension (VsCodeVim) version: 1.3.0
- VSCode version: 1.32.3
- OS: Windows_NT x64 10.0.17134

ajalab:
The recent change in terms of word separation is responsible for me.
I pasted your configuration script to my `settings.json` and tried it, but actually the problem was not reproduced on my Windows machine.
Would you please show the whole text which produces the problem?

Environment:

- Extension Version: `1.4.0`
- OS Version: `Windows_NT x64 10.0.17134`
- VSCode version: `1.33.0`

I also tried it on VsCodeVim version `1.3.0`.

hugoj-goncalves:
may this be related with this? Working with diacritics is really annoying atm...
![issue-2019-04-08 20-04](https://user-images.githubusercontent.com/10927715/55762584-d7b35280-5a39-11e9-93c1-baec8a0f66b0.gif)

ajalab:
[...]

PR #3667, opened 2019-04-09
Fix word separation

ajalab:

This PR fixes the degradation issue that words which contain diacritics (e.g., `Städte`) are separated in word motions.

Two issues are reported in #3665:

1.  A string which contains both Latin alphabets and `vim.iskeyword` characters (e.g., `helloWorld(param1`) is wrongly regarded as a single word https://github.com/VSCodeVim/Vim/issues/3665#issue-430275917i.
1.  A string which contains both Latin ASCII alphabets and diacritics (e.g., `adbdémm`, `Städte`) is wrongly regarded as separate words https://github.com/VSCodeVim/Vim/issues/3665#issuecomment-481040030.

This PR fixes the second one. I've not been able to reproduce the original first problem...

- Auxiliary methods (`makeAsciiWordSegments` and `makeUnicodeWordSegments`) are removed.
- The behavior of the regex can be checked in https://regex101.com/r/X1agK6/1.


ajalab:
Thanks to the issue report #3680, it was found that [Latin-1 symbols and punctuations](https://en.wikipedia.org/wiki/List_of_Unicode_characters#Latin-1_Supplement) were wrongly classified as word characters. I overlooked that the original Vim treats characters whose code point is less than 0x100 (both ASCII chars *and Latin-1 chars*) as a special case in [mbyte.c#L2879](https://github.com/vim/vim/blob/55d3bdbbe2bfc7a78b4aa17763788dbddf87cab0/src/mbyte.c#L2879). So I added Latin-1 symbols and punctuations (code point range: `0x00a1`-`0x00bf`) to the symbol table in https://github.com/VSCodeVim/Vim/pull/3667/commits/e831849a3d068a157ef4679de39564fc50180fbb.

This PR fixes #3680.

src/common/motion/position.ts
```diff
@@ -946,10 +908,11 @@ export class Position extends vscode.Position {
       Hangul,
     }
 
+    // List of printable characters (code point intervals) and their character kinds.
+    // Latin alphabets (e.g., ASCII alphabets and numbers,  Latin-1 Supplement, European Latin) are excluded.
     // Imported from utf_class_buf in src/mbyte.c of Vim.
-    // Spelling alphabets are not listed here since they are covered as non-white letters.
-    // TODO(ajalab): add Emoji
-    const codePointRanges: [[number, number], CharKind][] = [
+    const table: [[number, number], CharKind][] = [
```
jpoon: nit: rename this a more descriptive name.

ajalab: how about `symbolTable`?

jpoon: Works for me.

Issue #3370, opened 2019-01-14
Slack invite link doesn't resolve


areinmeyer:
**Describe the bug**
Clicking on the link in the README that says `If you need to ask us any questions, join us on Slack` generates an error

**To Reproduce**

1.  Go to https://vscodevim-slackin.azurewebsites.net
2.  See page error out.

I've tried this link on and off for about a week and it's never resolved, so I don't think this is a random networking issue.

**Expected behavior**
I expected to see a form that would allow me to request a login or resolve to a login.  

Note that https://vscodevim.slack.com works, but there's no way to request a new account from that page.

jpoon:
Looks like my free azure account is no longer free after leaving Microsoft. We need another place to host the slackin server.

YamadaKY:
I want this too!
  '<,'> normal.

PR #3679, opened 2019-04-15
docs: update slackin link

khoitd1997:
Change to https for existing links and convert to heroku for azure ones

**What this PR does / why we need it**: Slack has moved to heroku apps but some of them still has the old azure link, also some heroku one has http instead of https

**Which issue(s) this PR fixes**: This doesn't start a fix for a particular issue but rather finish the fix for #3370


jpoon:
Ah. Oops, Didn't update all the places. Thanks!

Issue #2457, opened 2018-03-20
Default selector via shift+arrow does not become visual mode


WangRongda:
**Is this a BUG REPORT or FEATURE REQUEST?** (choose one):
Bug
**What happened**:
Shift+arrow to select text not become visual mode, but shift+<mouse-left> become visual mode
**What did you expect to happen**:
Shift+arrow to select should become visual mode 
**How to reproduce it (as minimally and precisely as possible)**:

**Environment**:

- Extension (VsCodeVim) version: 0.11.2
- VSCode version: 1.21.1
- OS version: Arch 4.15.9-1

cybertim:
Shift-End/Home does also not enter visual mode.
Partial mouse selection does however enter visual mode.

**What did you expect to happen:**
All selection methods should enter visual-mode - or not - but not half the selection modes.

lilixiong2018:
I modify src\mode\modeHandler.ts, change 
```
if ( e.kind !== vscode.TextEditorSelectionChangeKind.Mouse ) {
...
} 

```
to
```
   if (
      e.kind !== vscode.TextEditorSelectionChangeKind.Mouse &&
      e.kind !== vscode.TextEditorSelectionChangeKind.Keyboard
    ) {
...
}
```

Now it support shift+arrow/shift+home/shift+end to goto visual mode.

PR #3727, opened 2019-05-05
switch to visual mode when shift selection

lilixiong2018:
For windows developer, it is quite often to make selection with shift + arrow, shift + Home, shift + End., switch to visual mode will be convenient for replace within selection, etc.

#2457


lilixiong2018:
Withdraw because there are some bugs need to fix.

PR #3728, opened 2019-05-05
Various improvements to registers

### LINKED ISSUE #1775: Is there a way to repeat a CLI command?
For example, can we repeat a :s/foo/bar/g without retyping it?
I've looked around but haven't been able to find anything about this (maybe I've missed it)

### LINKED ISSUE #3542: Support the / register

In standard Vim, the most recently searched-for string is stored in the `/` register. This makes it easy to paste it into a regex or reverse the direction of your search (e.g., if searching forwards and you really want to search backwards, you can just use `?<C-r>/`).

The only current solution I've found is to retype the search expression, as you cannot even copy out of the status bar entry in VSCodeVim.

**Additional context**

### LINKED ISSUE #3604: The . (last inserted text) register should be read-only
**Describe the bug**
You can currently copy arbitrary text into the `.` register. This is not how vim behaves.

**To Reproduce**

1.  Insert some text
2.  Use `:reg` to verify the inserted text is in the `.` register
3.  Copy some other text into the register with `".yw`, for instance
4.  Use `:reg` to verify the copied text is in the `.` register

**Expected behavior**
Step 3. above should be a no-op, and in step 4., the originally inserted text should show in `:reg`

- Extension (VsCodeVim) version: 1.2.0
- VSCode version: 1.32.3
- OS: Windows 10

### LINKED ISSUE #3605: Implement the % (file name) and : (last executed command) registers
The `%` register should contain the current filename, and the `:` register should contain the last executed command. They should be read-only.

### LINKED ISSUE #3606: :reg should not show the _ (black hole) register
**Describe the bug**
`:reg` should not show the `_` (black hole) register. This is not consistent with how vim behaves, and it's not useful because it can never contain anything.

**To Reproduce**

1.  Run `:reg`
2.  Observe that the `_` register is listed

**Expected behavior**
It should not be listed.

- Extension (VsCodeVim) version: 1.2.0
- VSCode version: 1.32.3
- OS: Windows 10

### LINKED ISSUE #3610: :reg should show multiple registers if given multiple arguments
**Describe the bug**
Currently, `:reg` ignores all but its first argument

**To Reproduce**

1.  Call `:reg 0 1 2`
2.  Observe that an info box with the contents of register `0` pops up

**Expected behavior**
Same as `:reg` with no arguments, but list only the registers given to the command

- Extension (VsCodeVim) version: 1.2.0
- VSCode version: 1.32.3
- OS: Windows 10

J-Fields:

- Implement the `/`, `:`, and `%` registers
- Stops read-only registers from being written to directly
- Allows the `:` register to be used as a macro to repeat the last command
- Improves the behavior of `:reg` (hiding `_` register, allowing multiple arguments)

Fixes #3542
Fixes #3604
Fixes #3605
Fixes #1775
Fixes #3606
Fixes #3610

This PR does a lot of different things, but they're all small changes and mostly related. I'm happy to break this up if it's easier for you; let me know!


jpoon:
Wow 😮 . Thanks for all the recent PRs!

jpoon:
LGTM. Will merge after this build passes.

Issue #3607, opened 2019-03-21
vim.statusBarColors.normal reports type error


EntilZha:
**Describe the bug**
The editor reports that the type of this setting is incorrect, expecting a string instead of array after setting things according to https://github.com/VSCodeVim/Vim#vim-airline

It looks like the type specification here is responsible https://github.com/VSCodeVim/Vim/blob/master/package.json#L579 but easily fixed.

**To Reproduce**

1.  Follow the instructions here https://github.com/VSCodeVim/Vim#vim-airline to enable color bars
2.  Let linters run
3.  Observe in the errors/warnings menu that VS code reports a type error in `settings.json` that the expected type is string

**Expected behavior**
There should not be a warning

**Suggested Fix**
My quickfix to see if my suspicion was correct was to edit `~/.vscode/extensions/vscodevim.vim-1.2.0/package.json` to specify the type of that setting from `"type": "string"` to `"type": ["string", "array"]`. After reloading the editor this seems to have made the warning go away. A simple modification to the `package.json` I think fixes this.

**Environment (please complete the following information):**

better-toml | bun | 0.3.2
-- | -- | --
latex-workshop | Jam | 6.2.2
vscode-rust | kal | 0.4.2
python | ms- | 2019.2.5558
vscode-docker | Pet | 0.6.1
vim | vsc | 1.2.0

- Extension (VsCodeVim) version: 1.2.0
- VSCode version: 1.32.2
- OS: Arch Linux

EntilZha:
If that fix seems reasonable I can open a PR, but for a one line fix, might be easier to have a committer make the minor change.

PR #3767, opened 2019-05-13
Fix statusBarColors linting in vscode user settings.

faldah:
User settings linting for `vim.statusBarColors` shows errors when configuring a background and foreground colors. User settings expects a string, but `vim.statusBarColors` supports `string` and `string[]`.

https://github.com/VSCodeVim/Vim/issues/3607


jpoon:
Thanks!

PR #3822, opened 2019-06-07
Make indent-object LineWise:

burke:
Given the block:

      1234567
    1.def foo
    2.  a
    3.  b
    4.end

putting the cursor at r2c1 and running `yii` should copy the content at the current indentation level -- lines 2 and 3. Currently, the region is copied in CharacterWise mode as "  a\n  b", meaning that when it's pasted, it's inserted into the current line; that is, given:

      1234567
    1.//word
    4.//etc.

Running `p` at r1c1 gives:

      12345678
    1./  a
    2.  b/word
    3.//etc.

While what we want is:

      12345678
    1.//word
    2.  a
    3.  b
    4.//etc.

I can't think of any situation where indent-object wouldn't make sense as LineWise, but I don't have any context on this codebase so maybe I'm way off here.

### DIFF
--- a/src/actions/textobject.ts
+++ b/src/actions/textobject.ts
@@ -523,6 +523,8 @@ abstract class IndentObjectMatch extends TextObjectMovement {
     const firstValidLine = TextEditor.getLineAt(new Position(firstValidLineNumber, 0));
     const cursorIndent = firstValidLine.firstNonWhitespaceCharacterIndex;
 
+    vimState.currentRegisterMode = RegisterMode.LineWise;
+
     // let startLineNumber = findRangeStart(firstValidLineNumber, cursorIndent);
     let startLineNumber = IndentObjectMatch.findRangeStartOrEnd(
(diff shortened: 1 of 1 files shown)


xiaoq1024:
Add a new line at the indent-object beginning
```
// #1 init 
"vim.easymotion": true,
"vim.handleKeys": {
    "<C-d>": true,
    "<C-n>": false,// cursor here typing: yaI
},
"workbench.tree.indent": 8,

// #2 paste
"vim.easymotion": true,
"vim.handleKeys": {
    "<C-d>": true,
    "<C-n>": false,
},// cursor here typing: p
"workbench.tree.indent": 8, 

// #3 result
"vim.easymotion": true,
"vim.handleKeys": {
    "<C-d>": true,
    "<C-n>": false,
}"vim.handleKeys": {
    "<C-d>": true,
    "<C-n>": false,
},,
"workbench.tree.indent": 8,
```

Issue #3756, opened 2019-05-10
Allow to rebind camelCase motion


pomdtr:
**Is your feature request related to a problem? Please describe.**
I use VSCodeVim in conjonction with this set of settings to replicate a spacemacs setup. 
However, the sequence <leader>+<w>, <leader>+b and <leader>+<e> are already dedicated to start some commands.

**Describe the solution you'd like**
A set of settings to rebind the access to these motion.

jpoon:
If you don't use `camelCaseMotion` you can disable it? https://github.com/VSCodeVim/Vim#camelcasemotion

genesy:
The issue is about allowing to use `camelCaseMotion `with other keys. How do I use `camelCaseMotion `if I disable it? @jpoon

dakom:
I was going to file a new issue but I think it's a dup of this... for example, instead of `<operator>i<leader>w` to delete the camelCase section, I'd like it to be just `<operator>w`. big-W could default to non-camelCase

genesy:
Yeah that's a dupe of this. They don't care about this specific issue tho.

dakom:
Without knowing the code base, isn't it as simple as making this array configurable?

https://github.com/VSCodeVim/Vim/blob/d8564f90a66b414c8698af1514becc5549973263/src/actions/plugins/camelCaseMotion.ts#L81

jbriales:
I was also looking how to customize bindings for these operators. Concretely I wanted to use `w` instead of `<leader>w`, etc.
It seems remapping via settings works fine:
```js
"vim.normalModeKeyBindingsNonRecursive": [
  {"before": ["w"], "after": ["\\", "w"]},
  {"before": ["b"], "after": ["\\", "b"]},
  {"before": ["e"], "after": ["\\", "e"]},
],
"vim.operatorPendingModeKeyBindingsNonRecursive": [
  {"before": ["i","w"], "after": ["i", "\\", "w"]},
]
```

coindegen:
> I was also looking how to customize bindings for these operators. Concretely I wanted to use `w` instead of `<leader>w`, etc. It seems remapping via settings works fine:
> 
> ```js
> "vim.normalModeKeyBindingsNonRecursive": [
>   {"before": ["w"], "after": ["\\", "w"]},
>   {"before": ["b"], "after": ["\\", "b"]},
>   {"before": ["e"], "after": ["\\", "e"]},
> ],
[...]

PR #3837, opened 2019-06-16
Adds option to remap camelCase keys

genesy:
Fixes issue #3756 
Allows user to rebind camelCaseMotion keys


genesy:
Should I force prettier here? it's prettying files I didn't touch

PR #3889, opened 2019-07-12
WebPack builds for improved loading times

### LINKED ISSUE #3127: Add Webpack Bundling
The startups times are quite heavy now and can be cut by adding webpack bundling.

ianjfrosst:
Add WebPack support via a `gulp` plugin

Partially fixes #3127

In my tests, loading times (`Running Extensions` -> `Startup Activation`) decrease from 2-4s to <1s.

The `webpack-stream` package is a little outdated, and raises some deprecation and vulnerability warnings. The best alternative would be to migrate fully from `gulp` to `webpack`, but that requires some discussion.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,8 @@ script:
     echo "Prettier Failed. Run `gulp forceprettier` and commit changes to resolve.";
     exit 1;
   fi
-- npm run build
+- gulp build
+- gulp prepare-test
 - npm test
 
 before_deploy:
--- a/.vscodeignore
+++ b/.vscodeignore
@@ -14,4 +14,4 @@ typings/**
 node_modules/
 src/
 tsconfig.json
-webpack.config.json
+webpack.config.js
--- a/.vscodeignore
+++ b/.vscodeignore
@@ -10,3 +10,8 @@ typings/**
 .github_changelog_generator
 .gitignore
 *.yml
+# Not needed, due to webpack
+node_modules/
+src/
+tsconfig.json
+webpack.config.json
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -131,6 +131,12 @@ function updateVersion(done) {
     });
 }
 
+function copyPackageJson() {
+  return gulp
+    .src('./package.json')
+    .pipe(gulp.dest('out'))
+}
+
 gulp.task('tsc', function() {
   var isError = false;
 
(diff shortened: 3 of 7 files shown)


xconverge:
@jpoon Any thoughts here? I am all for webpack and think it will be a great improvement to start up times, however I like our gulp infrastructure

jpoon:
I haven't touched webpack in awhile but IIRC, they aren't exclusive. We can still continue using our gulp as a task runner for running tests, prettier, etc. and let webpack do it's thing with transpiling the TS and bundling? Or has the JS community deemed gulp un-cool like grunt and we should webpack everything?

ianjfrosst:
You can certainly run both. That's the way things are set up right now (on this branch). `webpack` is just another gulp task, like an alternative to `tsc`. The problem I'm having is that the webpacked output isn't suitable for running the tests. The recommended approach is to use `tsc` for the tests only, bu that causes a problem with the relative imports (`src/` vs. `out/src/`).

jpoon:
> The problem I'm having is that the webpacked output isn't suitable for running the tests. 

Webpack newb here. How come?

>  The recommended approach is to use tsc for the tests only, bu that causes a problem with the relative imports (src/ vs. out/src/).

Yikes, doesn't that technically mean the transpiled JS code that we test against isn't the same as that we launch with?

ianjfrosst:
I'm also a webpack newb, so I might not be the best resource. The [vscode developer docs](https://code.visualstudio.com/api/working-with-extensions/bundling-extension) have most of the info I'm going off of. 

As I understand, webpack bases itself off the entrypoints of an application (in this case, `extension.ts`) and culls anything that isn't used from there. There are options for multiple entrypoints (with more confusing documentation), but even then, it may optimize them to into 2 separate files (in production mode). I don't believe there's a way to run the tests on the same transpiled JS code as the main output using webpack. 

However, that is the way that all _compiled_ projects run unit tests. It just depends on how much you trust the transpile+optimize+minify steps.

ianjfrosst:
@jpoon @xconverge what do you feel is left to do for this?

xconverge:
I feel really bad, basically I want this and support this, however we are a little light on time to maintain right now, and I am not up to speed with how webpack works really at all. I really apologize, we have been in a bit of a tough situation for a few months.

J-Fields:
@ianjfrosst Can you merge/rebase with latest master? This looks good to me.

J-Fields:
@ianjfrosst Could you update the branch? It'd be awesome to get this in the next release.

J-Fields:
Can confirm that this does reduce measured startup time significantly (it does also feel a bit faster to start up). Also reduces `.vsix` size to ~33% of what it was. I'll do another sanity check later and then merge this in.

jpoon:
Wow, that is amazing performance improvements! 🎉

J-Fields:
Thanks @ianjfrosst!

PR #3890, opened 2019-07-15
Fix gj/gk so it maintains cursor position

hetmankp:
This fixes the gj & gk motion commands so that they maintain column position when jumping across short lines. This fixes issue #1323 .


jpoon:
Can you add tests?

hetmankp:
Done.

test/mode/modeVisual.test.ts
```diff
@@ -226,6 +226,22 @@ suite('Mode Visual', () => {
     });
   });
 
+  suite('Various motions in Visual Mode', () => {
```
jpoon: Can we use a more descriptive name? Like `Display line motions` or something?

hetmankp:
Ok, try that.

test/mode/modeVisual.test.ts
hetmankp: Ok, try that.

flip1995:
Is something blocking this PR? I'd really like to see to get this merged :)

test/mode/normalModeTests/motions.test.ts
```diff
@@ -725,4 +725,18 @@ suite('Motions in Normal Mode', () => {
...
+  newTest({
+    title: "Can handle 'gk'",
+    start: ['blah', 'duh', '|dur', 'hur'],
+    keysPressed: 'gk',
+    end: ['blah', '|duh', 'dur', 'hur'],
+  });
+
+  newTest({
+    title: "Can handle 'gj'",
+    start: ['blah', 'duh', '|dur', 'hur'],
```
J-Fields: These tests don't seem like they exhibit the bug in #1323 (since the cursor is at the front of the line each time). Am I wrong? If not, we should have some regression tests.

jpoon:
I’m not sure if there is a better mechanism but as a maintainer it’s hard to figure out when a PR is ready for another round of reviews. After you’ve addressed all the comments, let us know so we can take another look.

hetmankp:
Yes, I had wanted to add those tests a while ago but then realised visual mode behaviour was completely broken with the existing code and had to abandon it.

I've reworked the original fix with better dependency hygiene, fixed visual mode behaviour so it behaves like VIM and added the requested tests.

Issue #2895, opened 2018-07-25
Pasting from a register in visual mode overwrites the wrong register


justinr636:
**Is this a BUG REPORT or FEATURE REQUEST? (choose one):**

BUG REPORT

**What happened:**

Pasting from a register in visual mode puts the highlighted/deleted text in the register you pasted from.

**What did you expect to happen:**

Pasting from a register in visual mode should not change the contents of the register I am pasting from.

**How to reproduce it (as minimally and precisely as possible):**

> <kbd>"</kbd><kbd>a</kbd><kbd>y</kbd><kbd>y</kbd>
> Yank a line (e.g. line no. 1) into register a
> 
> <kbd>2</kbd><kbd>j</kbd><kbd>V</kbd>
> Highlight a different line (e.g. line no. 3) in visual mode
> 
> <kbd>"</kbd><kbd>a</kbd><kbd>p</kbd>
> Paste from register a
> 
> <kbd>"</kbd><kbd>a</kbd><kbd>p</kbd>
> Pasting from register a again will paste the data you overwrote (line no. 3) instead of the original data you copied into the register (line no. 1)
> 
> Register a should be unchanged. The highlighted/deleted content should be placed in the " register.

**Environment:**

Extension (VsCodeVim) version: 0.15.7
VSCode version: 1.25.1
OS version: Ubuntu 18.04 LTS

hetmankp:
I can confirm this is also causing me problems.

MrSurly:
I'm also experiencing this.

Issue #2939, opened 2018-08-07
Visual line causes named register to be overwritten with selection on paste


samkhal:
**Describe the bug**
When pasting from a named register, the selection is sometimes copied into that register. This seems to happen when the original selection used visual line mode.

**To Reproduce**

Start with the file
```
aa
bb
cc
```
With the cursor on the first character in normal mode.

Type `Vj"ayjjve"ap"ap`.

The result is
```
aa
bb

acca
bb
```
Register "a" now contains "cc".

**Expected behavior**
Output in vim:
```
aa
bb

aa
aa
bb
bb
```
Register "a" still contains "aa\nbb"

**Environment (please complete the following information):**

- Extension Name: vim
- Extension Version: 0.15.7
- OS Version: Linux x64 4.15.0-29-generic
- VSCode version: 1.25.1

J-Fields:
Closing in favor of #2895

PR #3891, opened 2019-07-15
Fix behaviour when pasting lines over selection

hetmankp:
Currently when pasting lines of text (i.e. text in register is linewise) over an existing selection, the code incorrectly yanks the selected lines of text into the register specified by the command. This is incorrect behaviour as the selected text should always be yanked into the unnamed register (or clipboard register if appropriate).

Fixes issues #2895 and #2939 .


xconverge:
Can you please add test cases

hetmankp:
This was already partially covered by the `Vp updates register content` test but I have added an extra test `"xVp only updates default register content` for the situation that was previously incorrect.

hetmankp:
Hi @xconverge , not sure how to request a review for this but it would be great to have someone look at merging it.

PR #3908, opened 2019-07-21
Read key remappings from .vimrc

rdnlsmith:

Reads key remapping commands from $HOME/.vimrc, $HOME/_vimrc, or a user-specified Vim configuration file. For each, builds an IKeyRemapping object and appends it to the appropriate collection, _if_ doing so will not override a remapping specified in the VS Code settings.

Partially addresses #463.

This implementation borrows heavily from [Sheepolution/vimrc-to-json](https://github.com/Sheepolution/vimrc-to-json).

I haven't done much with node or TypeScript before this, so apologies in advance if I've made any "obvious" mistakes.


JohnnyUrosevic:
I would love to see this feature go live!

rdnlsmith:
At this point, I figure I should rebase the branch and replace the `fs` dependency with the [the new VS Code filesystem API](https://github.com/microsoft/vscode/issues/48034). It also occurs to me that it probably wouldn’t behave as expected if someone has remappings specified as part of an autocommand. But, it would be nice to know whether the maintainers are interested in this PR before I put any more work into it.

J-Fields:
@rdnlsmith Sorry the team is so backed up on PRs. I'm definitely interested in this - will skim now and revisit later.

package.json
```diff
@@ -743,6 +752,7 @@
     "appdirectory": "0.1.0",
     "diff-match-patch": "1.0.4",
     "lodash": "4.17.15",
+    "fs": "0.0.1-security",
```
J-Fields: As you mentioned, `vscode.workspace.fs` seems to be preferred

package.json
```diff
@@ -665,6 +665,15 @@
           "description": "Path to run neovim executable. For example, /usr/bin/nvim, or  C:\\Program Files\\Neovim\\bin\\nvim.exe",
           "default": "nvim"
         },
+        "vim.enableVimrc": {
+          "type": "boolean",
+          "description": "Use key mappings from a .vimrc file",
```
J-Fields: Descriptions should always end with a `.`

package.json
```diff
@@ -665,6 +665,15 @@
...
+          "default": "true"
+        },
+        "vim.vimrcPath": {
+          "type": "string",
+          "description": "Path to a Vim configuration file. If unset, it will check for $HOME/.vimrc or $HOME/_vimrc"
```
J-Fields: Remind me - is `_vimrc` a gVim thing?

J-Fields: What about `$XDG_CONFIG_HOME/nvim/init.vim`? (https://neovim.io/doc/user/starting.html#vimrc)

Feel free to disagree, but my I think my preferred approach for this is have `vim.vimrcPath` default to an empty string. Then note here that if it's unset, we fall back on default locations (like `$HOME/.vimrc`, `$HOME/_vimrc`, and `$XDG_CONFIG_HOME/nvim/init.vim`) and then link to the part of the `README` which lists these locations.

A nice-to-have would be a pop up the first time it defaults to a location (which is probably immediately after you enable this feature) that says something like "We couldn't find your .vimrc. Click **here** to set the location" or "We found a .vimrc here: /path/to/.vimrc. Click **here** to use it"

src/configuration/vimrc.ts
```diff
@@ -0,0 +1,96 @@
...
+        collection = configuration.visualModeKeyBindingsNonRecursive;
+        break;
+      case 'inoremap':
+        collection = configuration.insertModeKeyBindingsNonRecursive;
+        break;
+      default:
+        return false;
+    }
+
+    // Don't override a mapping present in settings.json; those are more specific to VSCodeVim.
```
J-Fields: Does this mean `set scrolloff=5` in my `.vimrc` won't override an unset `editor.cursorSurroundingLines`? Or only a *set* `editor.cursorSurroundingLines` will have higher priority?

src/configuration/vimrcKeyRemappingBuilder.ts
```diff
@@ -0,0 +1,60 @@
...
+  private static readonly KEY_LIST_REG_EX = /(<[^>]+>|.)/g;
+  private static readonly COMMAND_REG_EX = /(:\w+)/;
+
+  public build(line: string): IVimrcKeyRemapping | null {
+    let matches = VimrcKeyRemappingBuilderImpl.KEY_REMAPPING_REG_EX.exec(line);
+    if (!matches || matches.length < 4) {
+      return null;
+    }
+
+    let type = matches[1];
```
J-Fields: Nit: we prefer `const` to `let` when possible

test/configuration/vimrcKeyRemappingBuilder.test.ts
```diff
@@ -0,0 +1,64 @@
...
+        keyRemapping: {
+          before: ['<C-s>'],
+          commands: [':w'],
+        },
+        keyRemappingType: 'nnoremap',
+        expectNull: false
+      },
+      {
+        // Ignore non-mapping lines
+        vimrcLine: 'set scrolloff=8',
```
J-Fields: Since you opened this PR, `scrolloff` has been introduced as `editor.cursorSurroundingLines`

package.json
rdnlsmith: It might be a gVim thing; I know of it as more generally a Windows thing. Windows hasn't always supported file names that start with `.`, and even now some programs / dialogs don't allow it, so Windows versions of Vim and Vim derivatives tend to support a `_vimrc` file.

rdnlsmith: > What about `$XDG_CONFIG_HOME/nvim/init.vim`? (https://neovim.io/doc/user/starting.html#vimrc)

I skipped this initially because `$XDG_CONFIG_HOME` wasn't set by default on my Ubuntu machine. But, I can certainly add it, plus the default paths from your link.

> Feel free to disagree, but my I think my preferred approach for this is have `vim.vimrcPath` default to an empty string. Then note here that if it's unset, we fall back on default locations (like `$HOME/.vimrc`, `$HOME/_vimrc`, and `$XDG_CONFIG_HOME/nvim/init.vim`) and then link to the part of the `README` which lists these locations.

I'm not sure I understand how this is different from what I did?

> A nice-to-have would be a pop up the first time it defaults to a location (which is probably immediately after you enable this feature) that says something like "We couldn't find your .vimrc. Click **here** to set the location" or "We found a .vimrc here: /path/to/.vimrc. Click **here** to use it"

I like this idea! I'll see what I can do.

src/configuration/vimrc.ts
rdnlsmith: To be clear&mdash;currently, this PR only supports key remappings, and not other settings. 

The idea was that a user might want to bind the same key(s) differently within VS Code, _e.g._ so they could take advantage of VS Code's built-in commands. So, I called my new code after the remappings are read in from VS Code's settings, and checked to make sure a given key combination from the `.vimrc` was not present before adding it.

I wanted to support other settings as well, but for those I wasn't sure if it was possible to tell the difference between a default setting vs. a user setting. I was also uncertain about determining (automatically) which Vim settings are supported by VS Code Vim, or which might reasonably translate to built-in VS Code settings (like the new `editor.cursorSurroundingLines` you mentioned in another comment). It seemed like most of the people commenting in the issue were primarily concerned with key remappings, so I figured I would submit what I had, and perhaps and revisit the other settings later.

rdnlsmith:
@J-Fields Do you prefer `merge` or `rebase` for bringing my branch back up-to-date with `master`?

J-Fields:
> 
> 
> @J-Fields Do you prefer `merge` or `rebase` for bringing my branch back up-to-date with `master`?

Up to you

src/configuration/vimrc.ts
J-Fields: Ah, gotcha. As long as we design with an eye forward (because if we support `.vimrc` at all, we'll be expected to support at least all the basics), that's fine, we can tackle it later.

I think we should probably maintain an explicit whitelist of supported settings, but these design decisions probably warrant a broader discussion.

package.json
J-Fields: Gotcha, makes sense

test/configuration/vimrcKeyRemappingBuilder.test.ts
J-Fields: Resolving this because it's out of scope for the PR

package.json
J-Fields: > I'm not sure I understand how this is different from what I did?

Sorry about that - I think we should support more default locations and list them externally (with a link in the description).

J-Fields:
I think an `Edit .vimrc` command would be helpful. Maybe a `Source .vimrc` command as well?

package.json
```diff
@@ -665,6 +665,15 @@
           "description": "Path to run neovim executable. For example, /usr/bin/nvim, or  C:\\Program Files\\Neovim\\bin\\nvim.exe",
           "default": "nvim"
         },
+        "vim.enableVimrc": {
```
jpoon: Can we move the configs to under something like this:

`vim.vimrc.enable`
`vim.vimrc.path`

src/configuration/vimrc.ts
```diff
@@ -0,0 +1,96 @@
...
+import * as fs from 'fs';
+import * as path from 'path';
+import { IConfiguration, IKeyRemapping, IVimrcKeyRemapping } from './iconfiguration';
+import { vimrcKeyRemappingBuilder } from './vimrcKeyRemappingBuilder';
+
+class VimrcImpl {
+  public load(configuration: IConfiguration) {
+    if (configuration.vimrcPath) {
+      configuration.vimrcPath = VimrcImpl.expandHome(configuration.vimrcPath);
+      if (!fs.existsSync(configuration.vimrcPath)) {
```
jpoon: Add a new config validator? 

See: src/configuration/validators/remappingValidator.ts

We should error and present a message to the user if this path doesn't exist. The validator will handle that.

src/configuration/vimrc.ts
```diff
@@ -0,0 +1,96 @@
...
+      default:
+        return false;
+    }
+
+    // Don't override a mapping present in settings.json; those are more specific to VSCodeVim.
+    if (!collection.some(r => _.isEqual(r.before, mapping!.keyRemapping.before))) {
+      collection.push(mapping.keyRemapping);
+    }
+
+    return true;
```
jpoon: You don't do anything with this `true`. Is it necessary to return it?

src/configuration/vimrc.ts
```diff
@@ -0,0 +1,96 @@
...
+      case 'inoremap':
+        collection = configuration.insertModeKeyBindingsNonRecursive;
+        break;
+      default:
+        return false;
+    }
+
+    // Don't override a mapping present in settings.json; those are more specific to VSCodeVim.
+    if (!collection.some(r => _.isEqual(r.before, mapping!.keyRemapping.before))) {
+      collection.push(mapping.keyRemapping);
```
jpoon: What if duplicates are added? What if there is a malformed remapping -- can we have these settings undergo `src/configuration/validators/remappingValidator.ts`?

src/configuration/vimrc.ts
```diff
@@ -0,0 +1,96 @@
...
+      }
+    }
+
+    let vimrcContent = fs.readFileSync(configuration.vimrcPath, { encoding: 'utf8' });
+    let lines = vimrcContent.split(/\r?\n/);
+
+    for (const line of lines) {
+      VimrcImpl.tryKeyRemapping(configuration, line);
+    }
+  }
```
jpoon: Should we save the updated vscodevim configuration? Otherwise, we'll need to do this re-syncing every time vscode is launched. 

If i update my vimrc, I'll need to re-launch vscode? Is that what most users will expect? Changes to the vscodevim settings on the otherhand don't require a restart.

src/configuration/vimrc.ts
jpoon: If things go wrong, how do users debug this?

jpoon:
https://github.com/VSCodeVim/Vim/pull/3908#issuecomment-528627428

@rdnlsmith it doesn't matter. we'll squash when we merge the PR :)

src/configuration/vimrc.ts
J-Fields: We could re-evaluate the vimrc if the user saves it from VSCode. We could also have a "source" command to manually re-evaluate

jpoon:
Let us know when the PR is ready for review again.

src/configuration/vimrc.ts
J-Fields: Actually, I don't think it's possible that a bad remapping could be generated, based on what's currently being validated in `remappingValidator`. That said, remappings from .vimrc are added to the config before the config is validated.

src/configuration/vimrc.ts
J-Fields: Resolving this chain because we'll handle other .vimrc features down the line - we're just doing remaps here.

J-Fields:
@jpoon I've fixed most of the issues that were brought up. Another round of review would be great when you get the chance.

src/configuration/configuration.ts
```diff
@@ -303,6 +308,11 @@ class Configuration implements IConfiguration {
   enableNeovim = false;
   neovimPath = '';
 
+  vimrc = {
+    enable: true,
```
jpoon: false?

src/configuration/validators/vimrcValidator.ts
```diff
@@ -0,0 +1,22 @@
+import * as fs from 'fs';
+import { IConfiguration } from "../iconfiguration";
+import { IConfigurationValidator, ValidatorResults } from "../iconfigurationValidator";
+
+export class VimrcValidator implements IConfigurationValidator {
+  async validate(config: IConfiguration): Promise<ValidatorResults> {
+    const result = new ValidatorResults();
+
+    if (config.vimrc.enable && !fs.existsSync(config.vimrc.path) ) {
```
jpoon: The code in `vimrc.ts` has a lot more logic in checking the vimrc path. `VimrcImpl.expandHome`

test/configuration/vimrcKeyRemappingBuilder.test.ts
```diff
@@ -0,0 +1,65 @@
+import * as assert from 'assert';
+import { IKeyRemapping, IVimrcKeyRemapping } from '../../src/configuration/iconfiguration';
+import { vimrcKeyRemappingBuilder } from '../../src/configuration/vimrcKeyRemappingBuilder';
+
+suite('VimrcKeyRemappingBuilder', () => {
+  test('Build IKeyRemapping objects from .vimrc lines', () => {
```
jpoon: maybe a test case with a multi-line vimrc? and with other options that are ignored?

rdnlsmith:
@J-Fields Hey, I'm sorry I never got back to this&mdash;it's been a busy month+ and time got away from me. I'm glad to see that you're fixing it up!

J-Fields:
No worries @rdnlsmith, thanks for getting the ball rolling on this feature!

Sheepolution:
README.md currently says:
> :warning: Vimscript is not supported; therefore, we are not able to load your .vimrc or use .vim plugins. You have to replicate these using our Settings and Emulated plugins.

With this change we can remove this right? And perhaps add a small section on how to let it read your .vimrc file.

J-Fields:
@Sheepolution Yeah, there are still a few small changes I need to make, thanks for pointing that out.

PR #3940, opened 2019-08-02
Implement `scrolloff`

### LINKED ISSUE #1427: Please add support for scrolloff
Please add support for the vim setting `scrolloff`

Thanks! :)

J-Fields:
Implements `scrolloff`, which keeps a buffer between the cursor and the top/bottom of the viewport.

Fixes #1427

As noted on the ticket, this feature is best addressed in VSCode itself, but this at least works more consistently with VSCodeVim's feature set than separate existing extensions.
Known issues:
* It counts lines, not "screen lines", so it does not play nicely with word wrap.
* It does not play nice with, for instance, <C-d>, when smooth scrolling is enabled. This is already buggy due to race conditions relating to scrolling, though, so I'm not worrying about it here.

I'm not sure if this is auto-testable in the existing framework, so I'd encourage the reviewer and anyone just passing by to check this branch out and play around with it yourself.

### DIFF
--- a/extension.ts
+++ b/extension.ts
@@ -52,7 +52,11 @@ export async function getAndUpdateModeHandler(forceSyncAndUpdate = false): Promi
     !previousActiveEditorId.isEqual(activeEditorId)
   ) {
     curHandler.syncCursors();
-    await curHandler.updateView(curHandler.vimState, { drawSelection: false, revealRange: false });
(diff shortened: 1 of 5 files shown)


J-Fields:
Closing as this was implemented upstream very recently. It's available in VSCode insiders as `editor.cursorSurroundingLines`.

PR #3943, opened 2019-08-04
Allow <C-p> and <C-n> to be used as prev/next when entering a command or search

### LINKED ISSUE #3942: Ctrl-P and Ctrl-N should be equivalent to Up / Down when entering a command or search
See title. I don't think this interferes too much with the VSCode's default Ctrl+P, and if it does for someone, they can disable it (via `handleKeys` or by changing the `when` clause of `extension.vim_ctrl+p`).

J-Fields:
Contributes keybindings and corresponding support in code for `<C-p>` and `<C-n>` to act as prev/next when entering a command or search term.

Fixes #3942

### DIFF
--- a/package.json
+++ b/package.json
@@ -214,6 +214,16 @@
         "when": "editorTextFocus && vim.active && vim.use<C-n> && !inDebugRepl"
       },
       {
+        "key": "ctrl+n",
+        "command": "extension.vim_ctrl+n",
+        "when": "vim.mode == 'SearchInProgressMode' && vim.active && vim.use<C-n> && !inDebugRepl"
+      },
+      {
+        "key": "ctrl+n",
+        "command": "extension.vim_ctrl+n",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -223,6 +233,16 @@
         "command": "extension.vim_ctrl+p",
         "when": "suggestWidgetVisible && vim.active && vim.use<C-p>"
       },
+      {
+        "key": "ctrl+p",
+        "command": "extension.vim_ctrl+p",
+        "when": "vim.mode == 'SearchInProgressMode' && vim.active && vim.use<C-p>"
```
xconverge: I think it was a mistake to not have !inDebugRepl in the previous one

J-Fields: Should I add it to the existing ctrl+p binding?

J-Fields: I'm assuming the answer to that is "yes". Resolved.

xconverge: thanks! Sorry to make you assume :)

Issue #3949, opened 2019-08-06
Support :his[tory]


J-Fields:
**Is your feature request related to a problem? Please describe.**
I've done a command a while ago, but don't want to type it out again and going back through history one at a time is cumbersome.

**Describe the solution you'd like**
`:his[tory]` should open a quickpick that will show me my command history, from which I can select a command to run.

**Describe alternatives you've considered**
None really. Is there another existing way of doing this?

J-Fields:
Apparently this also works for searches. That'd be nice too.

J-Fields:
So in vim `:history` is different from `q:`, but they mostly do the same thing as far as I'm concerned, and we support `q:` already! Cool! Now let's support `q/`/`q?` and activate them via `Ctrl-F` when doing a command/search.

PR #3954, opened 2019-08-06
`<C-f>` shows command history when pressed on command line

J-Fields:
This lets you search through your command history so you don't have to retype everything. Equivalent to `q:`.
Refs #3949


xconverge:
Please resolve the conflict with your other change :)

PR #3955, opened 2019-08-06
When the `c` (confirm) flag is used in a `:s` command, don't use neovim

### LINKED ISSUE #3371: Search-replace with confirm not working with neovim
**Describe the bug**
This is an extension of #1010, with a focus on the Neovim integration. Search-replace with confirm does not appear to work with the Neovim integration enabled.

**To Reproduce**

Run a search replace command like so:

```
:%s/foo/bar/gc
```

When you press enter, nothing happens. If you took out the `c` flag, then the replace will work correctly.

**Expected behavior**
A prompt window appears in VS Code to receive input for the confirmation (ie, `y/n/...`).

I am using the latest of all versions as of writing:

- Extension (VsCodeVim) version: 1.0.1
- VSCode version: 1.30.2
- OS: MacOS 10.14.2

J-Fields:
We need to use VSCode's quickpick capabilities to do confirmation.
Fixes #3371

### DIFF
--- a/src/cmd_line/commandLine.ts
+++ b/src/cmd_line/commandLine.ts
@@ -84,7 +84,7 @@ class CommandLine {
 
     try {
       const cmd = parser.parse(command);
-      const useNeovim = configuration.enableNeovim && cmd.command && cmd.command.neovimCapable;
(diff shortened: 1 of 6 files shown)


xconverge:
Another good one! Thanks!

Issue #3922, opened 2019-07-24
:wq does not close window if there are unsaved changes


zombiezen:
**Describe the bug**
When I try to use `:wq` to write and close a _modified_ file, it will write the file, but not close the window, failing with the error `E37: No write since last change (add ! to override).` in the status bar.

**To Reproduce**

1.  `code --wait foo.txt`
2.  Make a change (e.g. type in "foo")
3.  `:wq`

**Expected behavior**
The file is written and the window is closed.

**Environment (please complete the following information):**

- Extension (VsCodeVim) version: `1.8.2`
- VSCode version: `1.36.1`
- OS: `Windows_NT x64 10.0.18362`

**Additional context**

Not sure if relevant, but I am using this in conjunction with the Remote Development - WSL extension.

**EDIT 2019-07-26:** Edited to clarify that this behavior only occurs for modified files.

J-Fields:
Seems to me this only happens when the file does not exist. @zombiezen can you confirm?

zombiezen:
On my machine, the existence or non-existence of the file does not affect the behavior. I just verified that running `touch foo.txt` before going through the reproduction steps produces the same results.

J-Fields:
Interesting. I think I'm seeing a manifestation of the same bug because if I create a file via `code --wait test.txt`, I see the behavior you describe. But once it exists (if, for instance, I do `:w` before step 2 above), `:wq` works as expected.
My environment:
* Extension version: `1.8.2`
* VS Code version: `1.36.1`
* OS version: `Windows_NT x64 10.0.16299`

stevenguh:
This has been a issue for a while since the remote ssh. I never thought much of it.
https://github.com/VSCodeVim/Vim/blob/master/src/cmd_line/commands/write.ts#L55
Once I added an await for the execute call, I no longer get the error after executing `:wq` with remote ssh.

stevenguh:
Sure. But I have been get some unrelated test fails. I am wonder if this is a Windows problem.

stevenguh:
Exactly my thought. But okay. I will get a PR going and see if it pass on CI. Thanks.

Issue #3777, opened 2019-05-17
Vim extension UI "blocks" on remote development save


suo:
**Describe the bug**
I am using the remote development extension. When the connection is slow, using `:w` to save the file causes the VSCodeVim to "block", i.e. no input is accepted until the save is finished.

I can tell that it's VS Code itself that's blocked, because non-VSCodeVim commands are still responsive.

**To Reproduce**

1. Get VS Code insiders and install the remote development extension pack. I am using it over ssh.
2. Your connection has to be slow (mine has to tunnel through a jumphost on my employer's corporate network and back out).
3. Make some changes, save, then try to use the movement keys immediately after.
4. Observe that the Vim extension is locked up for a second or two.

**Expected behavior**
I would expect saving to occur asynchronously and for the vim extension to continue processing input, like the rest of VS Code does.

**Screenshots**
Here's a gif of what I mean. Notice I input a bunch of movement commands after the `:w` but nothing happens.
![example](https://user-images.githubusercontent.com/1617424/57903674-f7514e00-7823-11e9-9a8f-b4e8b5f35fcf.gif)

**Environment (please complete the following information):**

- Extension (VsCodeVim) version: 1.8.0
- VSCode version: Version 1.34.20-insider
- OS: Mac

**Additional context**
I'm happy to pitch in and fix this issue if given pointers. I think this is really important, as remote development will become more popular over time.

PR #3998, opened 2019-08-27
fix :wq in remote

stevenguh:
Fix the issue where `:wq` can't close a file in vscode remote

https://github.com/VSCodeVim/Vim/issues/3922

### DIFF
--- a/src/cmd_line/commands/write.ts
+++ b/src/cmd_line/commands/write.ts
@@ -52,7 +52,7 @@ export class WriteCommand extends node.CommandBase {
 
     // defer saving the file to vscode if file is new (to present file explorer) or if file is a remote file
     if (vimState.editor.document.isUntitled || vimState.editor.document.uri.scheme !== 'file') {
-      vscode.commands.executeCommand('workbench.action.files.save');
+      await vscode.commands.executeCommand('workbench.action.files.save');
       return;
     }
 


suo:
Hi @J-Fields, this change is a regression for me (I actually removed the await in a previous PR to fix the following issue: https://github.com/VSCodeVim/Vim/issues/3777).

This change causes all saves to block the Vim extension until the save is completed. With the remote development extension, this save can actually take multiple seconds, depending on the connection quality and file size (see the above issue for an example).

I would argue that completely blocking the Vim extension (no shortcuts, no ex commands, etc.) is worse than failing the save in this instance. Especially since `:w` is likely more common than `:wq`. Is it possible to revert this PR?

I suppose the ideal solution is to have sequential commands await each other. Or for whatever state machine handles the vim state not block on an ex command finishing. Happy to implement something along those lines if there's consensus on the right path.

cc @jpoon @Chillee

J-Fields:
@suo Thanks for pointing this out. I think the solution might be to `await` the save only for `:wq`, to avoid the race condition while not blocking for `:w`.

stevenguh:
Ok. I see where you are coming from. I assumed that's the cost working on remote, because all other write commands like `:wa`, and `:wqa are blocking as well.

@J-Fields That's what I am thinking as well. I will looking into the code. It seems like we can do that.

PR #4013, opened 2019-08-29
:w write in background

stevenguh:
This changes the :w behavior to write in background to unfreeze the VS Code UI when the user is using VS Code Remote while allowing operation like wq and wa to work in VS Code

This fix the regression caused by https://github.com/VSCodeVim/Vim/pull/3998

https://github.com/VSCodeVim/Vim/pull/3998
https://github.com/VSCodeVim/Vim/issues/3922
https://github.com/VSCodeVim/Vim/issues/3777

### DIFF
--- a/src/cmd_line/commands/write.ts
+++ b/src/cmd_line/commands/write.ts
@@ -1,8 +1,8 @@
 import * as fs from 'fs';
 import * as node from '../node';
 import * as path from 'path';
-import * as util from 'util';
 import * as vscode from 'vscode';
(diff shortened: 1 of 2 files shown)


J-Fields:
@suo If you get a chance, could you verify that this fixes the issue you mentioned in https://github.com/VSCodeVim/Vim/pull/3998#issuecomment-525950869

J-Fields:
`:wa` probably doesn't need to block either, but I see that's implemented separately

PR #4031, opened 2019-09-02
Remove word in command line with <C-w>

### LINKED ISSUE #4027: Command line does not support Ctrl-W
In the command/search bar, `<C-W>` should delete the previous word.
http://vimdoc.sourceforge.net/htmldoc/cmdline.html#c_CTRL-W

See also: #3243

stevenguh:
Remove word in command line with <C-w>

Fixes #4027


stevenguh:
Btw, I am still looking into why some of the tests failed

src/common/motion/position.ts
```diff
@@ -1068,21 +1084,29 @@ export class Position extends vscode.Position {
     return positions;
   }
 
+  private static getWordLeftWtihRegex(
```
J-Fields: `s/Wtih/With`

J-Fields:
Can you add support for `SearchInProgressMode` as well?

src/actions/commands/actions.ts
```diff
@@ -2165,6 +2165,32 @@ class CommandEscInCommandline extends BaseCommand {
   }
 }
 
+@RegisterAction
+class CommandRemoveWordCommandline extends BaseCommand {
+  modes = [ModeName.CommandlineInProgress];
+  keys = ['<C-w>'];
+  runsOnceForEveryCursor() {
+    return this.keysPressed[0] === '\n';
```
J-Fields: Is this here for a reason?

stevenguh: runsOnceForEveryCursor()? I think it should be false since this action doesn't need to be execute twice when there are multiple cursors.

stevenguh:
> Can you add support for `SearchInProgressMode` as well?

Okay I will add that. I forgot about that mode.

src/actions/commands/actions.ts
J-Fields: Yeah I agree - let's just return `false`

stevenguh: I think there are some other methods that need to change to return false as well, but I only corrected the methods I added

J-Fields:
Don't we need to modify the `when` condition on the `<C-w>` keybinding so we capture the key stroke in command line mode?

stevenguh:
https://github.com/VSCodeVim/Vim/blob/f70c8aa9f782881ebfe9f9de33952faa5e9a7f2c/package.json#L259
@J-Fields  It should be captured?

PR #4032, opened 2019-09-02
Refactor the existing file opening and auto completion

### LINKED ISSUE #3815: :o fails in remote SSH
**Describe the bug**
When I connect to a remote SSH host using Remote-SSH, using :o and then hitting tab for autocomplete results in:

```
TaskQueue: Error running task. Failed to handle key=<tab>. ENOENT: no such file or directory, scandir '\path\goes\here'.
```

**To Reproduce**

1.  Connect to remote host with Remote-SSH
2.  Open a file
3.  :o ./<TAB>
4.  See error

**Expected behavior**

- Extension (VsCodeVim) version: 1.8.1
- VSCode version: 1.35.0-insider
- Host OS: Windows 10
- Remote OS: Ubuntu 18.04

**Additional context**
Note that the SSH remote is Ubuntu, could be a contributing issue.

### LINKED ISSUE #3824: :split Can't Open Files With Names That Include Spaces
I feel like I'm missing something blindingly obvious here but I can't quite work it out.

**Describe the bug**
`:sp` command can't open a file if its name includes a space.

**To Reproduce**
Change directory to a folder that includes files with spaces in their names. Type `:sp` plus the first few characters of the file name, hit tab to complete (which happens correctly, including the spaces). Press enter and a notification pops claiming that `File: <path_to_working_directory>\<first_word_of_filename> does not exist`.

For example - `C:\Temp\My Test File.txt`
Resulting notification - `File: C:\Temp\My does not exist.`

I've tried enclosing the file name in single quotes, double quotes and just about anything else I think might work without success.

Files without spaces open just fine.

**Expected behavior**
A horizontal split will open showing the file specified.

**Environment**
- Extension (VsCodeVim) version: 1.8.1
- VSCode version: 1.35.0
- OS: Windows 10

All advice very gratefully received!

stevenguh:
The current implementation of file completion and file opening has way too many cases that don't work, and behavior that are not consistent with Vim. To name a few:
- Not working with remote
- Not working when a file contains space
- Minimal Windows support
- <tab> can't cycle like you you would expect
- There is no trailing / to indicate the completion option is a directory
- many more....

This pull request contains commits that refactor the existing auto complete mechanism to be more consistent with Vim while adding support for VS Code Remote.

This is the list of the expected behaviors:
- Tab complete base on the content before the cursor and the content on the right of the cursor will remind after completion
- File completion will use / as a suffix to indicate a directory like ":e " => <tab> => ":e testDir/"
- File completion will add ../ and ./ as completion options only if the . is at the end of File Completion command. e.g. ":e ." or ":e ~/."
- Repeating <tab> will cycle through all completion options, and any key press other than <tab> like <right> will finalize the completion
- Completion will be finalized on <tab> if the the completion only have one completion option. The subsequent <tab> will start a new completion base on current text in cmd
- File completion on untitled document of an non-remote instance is limited to either full path or ~. e.g. you can need to at least specify ":e /", ":e ~/" or ":e C:\" for the completion to work.
- No file completion on untitled document of a remote instance
- Use untitled document for nonexisting file path in cmd
Concrete examples: (The ":e" command can be any file commands like ":vsp")
- A <tab> after ":e " or ":e ./" will provide completion to the current directory of the active tab
- A <tab> after ":e ../" will provide completion to the directory one up of the current directory of the active tab
- A <tab> after ":e ~/" will only complete when the currently active tab is on local files system.
- A <tab> after ":e ./abc" will provide completion to any files or directories on the directory of the current active non-untitled tab where its name begin with abc

Windows Only behaviors:
- File Completion will convert any / to \ when the active tab is a local file or an untitled document. e.g "C:/Users/" will convert to ":e C:\Users\"
- Can complete and open path with drive letter like C:\ on Windows
- Can complete and open network drive path like //name/dir/

Fixes #3815
Fixes #3824

In the future, we can resolve the current directory of an untitled tab with workspace. However, this is complicated by the fact that multi-folder workspace is possible. But I chose to not do that, because it would complicate this pull requests.

Since this pull request combined a few features and refactor. It may be easier to view the changes commit by commit.


src/actions/commands/actions.ts
```diff
@@ -2902,9 +2935,14 @@ class CommandGoToDefinition extends BaseCommand {
...
   public async exec(position: Position, vimState: VimState): Promise<VimState> {
-    await vscode.commands.executeCommand('editor.action.goToDeclaration');
+    const oldActiveEditor = vimState.editor;
 
-    if (vimState.editor === vscode.window.activeTextEditor) {
+    await vscode.commands.executeCommand('editor.action.goToDeclaration');
+    // `executeCommand` returns immediately before cursor is updated
+    // wait for the editor to update before updating the vim state
+    // https://github.com/VSCodeVim/Vim/issues/3277
+    await waitForCursorSync(1000);
```
jpoon: We can remove this hack. VSCode fixed the underlying issue.

src/actions/commands/actions.ts
```diff
@@ -1968,83 +1968,109 @@ class CommandNavigateInCommandlineOrSearchMode extends BaseCommand {
...
-    }
-    return vimState;
+    vimState.currentCommandlineText = evalCmd + autoCompleteItems[commandLine.autoCompleteIndex];
+    vimState.statusBarCursorCharacterPos = vimState.currentCommandlineText.length;
+    vimState.currentCommandlineText += restCmd;
   }
 
   public async exec(position: Position, vimState: VimState): Promise<VimState> {
     const key = this.keysPressed[0];
+    const tabForward = key === '<tab>';
```
jpoon: nit: can you rename to `isTabForward`

src/actions/commands/actions.ts
```diff
@@ -1968,83 +1968,109 @@ class CommandNavigateInCommandlineOrSearchMode extends BaseCommand {
...
+
+    if (
+      commandLine.autoCompleteItems.length !== 0 &&
+      (commandLine.lastKeyPressed === '<tab>' || commandLine.lastKeyPressed === '<shift+tab>')
```
jpoon: `  (commandLine.lastKeyPressed === '<tab>' || commandLine.lastKeyPressed === '<shift+tab>')` this check shouldn't be necessary. it is handled by the base class.

src/actions/commands/actions.ts
```diff
@@ -1968,83 +1968,109 @@ class CommandNavigateInCommandlineOrSearchMode extends BaseCommand {
...
-        const search = <RegExpExecArray>(
-          /(?:.* .*\/|.* )(.*)/g.exec(vimState.currentCommandlineText)
-        );
-        commandLine.autoCompleteText = search[1];
-        commandLine.autoCompleteIndex = 0;
-      } else {
-        commandLine.autoCompleteText = vimState.currentCommandlineText;
-        commandLine.autoCompleteIndex = 0;
-      }
+  private cycleCompletion(vimState: VimState, tabForward: boolean) {
```
jpoon: nit: rename to `isTabForward`

src/actions/commands/actions.ts
```diff
@@ -1968,83 +1968,109 @@ class CommandNavigateInCommandlineOrSearchMode extends BaseCommand {
...
-      vimState.statusBarCursorCharacterPos = result.length;
-    } else if (result !== undefined) {
-      const searchArray = <RegExpExecArray>/(.* .*\/|.* )/g.exec(vimState.currentCommandlineText);
-      vimState.currentCommandlineText = searchArray[0] + result;
-      vimState.statusBarCursorCharacterPos = vimState.currentCommandlineText.length;
-    }
-    return vimState;
+    vimState.currentCommandlineText = evalCmd + autoCompleteItems[commandLine.autoCompleteIndex];
+    vimState.statusBarCursorCharacterPos = vimState.currentCommandlineText.length;
+    vimState.currentCommandlineText += restCmd;
```
jpoon: Is there a reason why you separated line 2001 and 2003? Why is it not:

```
 vimState.currentCommandlineText = evalCmd + autoCompleteItems[commandLine.autoCompleteIndex] + restCmd
```

src/actions/commands/actions.ts
```diff
@@ -1968,83 +1968,109 @@ class CommandNavigateInCommandlineOrSearchMode extends BaseCommand {
...
+      evalCmd = evalCmd.slice(0, fileRegex.lastIndex) + partialPath;
+
+      // test if the baseName is . or ..
+      const shouldAddDotItems = /^\.\.?$/g.test(baseName);
+      newCompletionItems = await readDirectory(
+        currentUri,
+        fullDirPath,
+        p.sep,
+        shouldAddDotItems
+      ).then(dirItems =>
```
jpoon: Let's not mix promises and async/awaits.

src/actions/commands/actions.ts
```diff
@@ -1968,83 +1968,109 @@ class CommandNavigateInCommandlineOrSearchMode extends BaseCommand {
...
+        dirItems
+          .filter(d => d.path.startsWith(baseName))
+          .map(r => r.path.slice(r.path.search(baseName) + baseName.length))
+          .sort()
+      );
     }
 
+    const newIndex = tabForward ? 0 : newCompletionItems.length - 1;
```
jpoon: nit: as noted above, i am a fan of having `is` prepended to boolean variables.

src/actions/commands/actions.ts
```diff
@@ -1968,83 +1968,109 @@ class CommandNavigateInCommandlineOrSearchMode extends BaseCommand {
...
+    commandLine.autoCompleteIndex = newIndex;
+    // If here only one items we fill cmd direct, so the next tab will not cycle the one item array
+    commandLine.autoCompleteItems = newCompletionItems.length <= 1 ? [] : newCompletionItems;
+    commandLine.preCompleteCharacterPos = cursorPos;
+    commandLine.preCompleteCommand = evalCmd + restCmd;
+
+    const completion = newCompletionItems.length === 0 ? '' : newCompletionItems[newIndex];
+    vimState.currentCommandlineText = evalCmd + completion;
+    vimState.statusBarCursorCharacterPos = vimState.currentCommandlineText.length;
+    vimState.currentCommandlineText += restCmd;
```
jpoon: Why is the assignment separated (lines 2070, 2072)?

src/actions/commands/actions.ts
stevenguh: Auto complete works on the content before the cursor, so when the `<tab>` is pressed the content on the right of the cursor (restCmd) should be appended after the new cursor pos.

That part of code can be confusing, so I changed it to the following
```javascript
vimState.currentCommandlineText =
  evalCmd + autoCompleteItems[commandLine.autoCompleteIndex] + restCmd;
vimState.statusBarCursorCharacterPos = vimState.currentCommandlineText.length - restCmd.length;
```

src/actions/commands/actions.ts
stevenguh: That probably came from a poor merge. Deleted.

src/actions/commands/actions.ts
stevenguh: Ok. Changed it. It actually reads better.

src/actions/commands/actions.ts
stevenguh: Same as the reason above. Change it to
```javascript
vimState.currentCommandlineText = evalCmd + completion + restCmd;
vimState.statusBarCursorCharacterPos = vimState.currentCommandlineText.length - restCmd.length;
```

src/actions/commands/actions.ts
stevenguh: Renamed

src/actions/commands/actions.ts
stevenguh: Changed :)

src/actions/commands/actions.ts
stevenguh: Renamed

src/actions/commands/actions.ts
```diff
@@ -1958,83 +1958,102 @@ class CommandNavigateInCommandlineOrSearchMode extends BaseCommand {
...
+      commandLine.lastKeyPressed = key;
+      return vimState;
+    }
+
+    let newCompletionItems: string[] = [];
+    const currentCmd = vimState.currentCommandlineText;
+    const cursorPos = vimState.statusBarCursorCharacterPos;
 
-    if (!/ /g.test(vimState.currentCommandlineText)) {
+    // Sub string since vim do completion before the cursor
```
jpoon: `vim do completion before the cursor`

typo: vim does completion?

src/cmd_line/commands/file.ts
```diff
@@ -67,59 +78,76 @@ export class FileCommand extends node.CommandBase {
...
-        ? path.normalize(this._arguments.name)
-        : path.join(path.dirname(editorFilePath), this._arguments.name);
+      // Using a filename, open or create the file
+      const { fullPath, path: p } = getFullPath(this.arguments.name);
+      // Only if the expanded path of the full path is different than
+      // the currently opened window path
+      if (fullPath !== editorFilePath) {
+        const uriPath = getUriPath(editorFileUri, fullPath, p.sep);
+        if (uriPath === null) {
+          // return the if path is invalid
```
jpoon: extraneous the?

src/cmd_line/commands/file.ts
```diff
@@ -67,59 +78,76 @@ export class FileCommand extends node.CommandBase {
...
-          const pathWithExt = filePath + path.extname(editorFilePath);
-          fileExists = await doesFileExist(pathWithExt);
-          if (fileExists) {
-            filePath = pathWithExt;
+          const pathWithExt = fullPath + p.extname(editorFilePath);
+          const uriPathWithExt = getUriPath(editorFileUri, pathWithExt, p.sep);
+          if (uriPathWithExt !== null) {
+            fileExists = await doesFileExist(uriPathWithExt);
+            if (fileExists) {
+              // if the file with the added ext exists
```
jpoon: change comment to `file with the added ext exists`

src/cmd_line/commands/file.ts
```diff
@@ -67,59 +78,76 @@ export class FileCommand extends node.CommandBase {
...
+              fileUri = uriPathWithExt;
+            }
           }
+        } else {
+          // If the file without the added ext exists
```
jpoon: change comment to `file without the added ext exists`

src/cmd_line/commands/file.ts
```diff
@@ -67,59 +78,76 @@ export class FileCommand extends node.CommandBase {
...
+          const pathWithExt = fullPath + p.extname(editorFilePath);
+          const uriPathWithExt = getUriPath(editorFileUri, pathWithExt, p.sep);
+          if (uriPathWithExt !== null) {
+            fileExists = await doesFileExist(uriPathWithExt);
+            if (fileExists) {
+              // if the file with the added ext exists
+              fileUri = uriPathWithExt;
+            }
           }
+        } else {
```
jpoon: can you reverse the logic?

if (fileExists) {
} else {
}

always easier to read affirmation over negation.

src/cmd_line/commands/file.ts
```diff
@@ -67,59 +78,76 @@ export class FileCommand extends node.CommandBase {
...
         if (!fileExists) {
-          if (this._arguments.createFileIfNotExists) {
-            await util.promisify(fs.close)(await util.promisify(fs.open)(filePath, 'w'));
-          } else {
-            this._logger.error(`${filePath} does not exist.`);
-            return;
+          if (this.arguments.createFileIfNotExists) {
+            // Change the scheme to untitled to open an
+            // untitled tab
+            fileUri = uriPath.with({ scheme: 'untitled' });
```
jpoon: what if `createFileIfNotExists` = false?

src/util/path.ts
```diff
@@ -2,26 +2,133 @@ import * as vscode from 'vscode';
...
-    basePath = '/';
-  } else if (partialPath.startsWith('~/')) {
-    basePath = <string>untildify(partialPath);
-    partialPath = '';
-  } else if (partialPath.startsWith('./')) {
-    basePath = path.dirname(editorFilePath);
-    partialPath = partialPath.replace('./', '');
-  } else if (partialPath.startsWith('../')) {
-    basePath = path.dirname(editorFilePath) + '/';
+export function separatePath(searchPath: string, separator: string) {
```
jpoon: can you add a function header/comment explaining this function? especially as you are returning multiple things, we should document what's being returned and the order.

src/util/path.ts
```diff
@@ -2,26 +2,133 @@ import * as vscode from 'vscode';
...
+        // If there isn't a complete UNC path,
+        // return the incomplete UNC as baseName
+        // e.g. \\test-server is an incomplete path
+        // and \\test-server\ is a complete path
+        return [searchPath, ''];
+      }
+    }
+  }
+
+  let baseNameIndex = searchPath.lastIndexOf(separator) + 1;
```
jpoon: const?

src/util/path.ts
```diff
@@ -2,26 +2,133 @@ import * as vscode from 'vscode';
...
+  const baseName = searchPath.slice(baseNameIndex);
+  const dirName = searchPath.slice(0, baseNameIndex);
+  return [dirName, baseName];
+}
+
+export function getFullPath(partialPath: string) {
```
jpoon: same as above. add functoin header explaining function + return types

src/util/path.ts
```diff
@@ -2,26 +2,133 @@ import * as vscode from 'vscode';
...
+  addCurrentAndUp: boolean
+) {
+  try {
+    const directoryUri = getUriPath(currentUri, absolutePath, sep);
+    if (directoryUri === null) {
+      return [];
+    }
+    const directoryResult = await vscode.workspace.fs.readDirectory(directoryUri);
+    return directoryResult
+      .map(d => ({ path: d[0] + (d[1] === vscode.FileType.Directory ? sep : ''), type: d[1] }))
```
jpoon: can't figure out what's going on here...

test/cmd_line/tabCompletion.test.ts
```diff
@@ -80,4 +78,162 @@ suite('cmd_line tabComplete', () => {
     await modeHandler.handleKeyEvent('<Esc>');
     assert.notEqual(statusBarBeforeTab, statusBarAfterTab, 'Status Bar did not change');
   });
+
+  test('command line file tab completion directory with / at the end', async () => {
+    const dirPath = await t.createRandomDir();
```
jpoon: wrap in try/finally

jpoon: and same for all the remaining tests.

test/cmd_line/tabCompletion.test.ts
```diff
@@ -80,4 +78,162 @@ suite('cmd_line tabComplete', () => {
...
+
+    await modeHandler.handleMultipleKeyEvents(['<left>', '<left>']);
+    statusBarAfterTab = StatusBar.Get().trim();
+    assert.equal(statusBarAfterTab, ':ed|it', 'Failed to move the cursor to the left');
+
+    await modeHandler.handleKeyEvent('<tab>');
+    statusBarAfterTab = StatusBar.Get().trim();
+    assert.equal(statusBarAfterTab, ':edit|it', 'Failed to complete content left of the cursor');
+
+    await modeHandler.handleKeyEvent('<Esc>');
```
jpoon: Validate that we've exited tab completion?

src/util/path.ts
```diff
@@ -2,26 +2,133 @@ import * as vscode from 'vscode';
 import * as path from 'path';
 import untildify = require('untildify');
```
jpoon: Add tests for these utils?

src/actions/commands/actions.ts
jpoon: Ah gotcha. Sorry missed that.

src/actions/commands/actions.ts
stevenguh: Yes. Fixed.

src/cmd_line/commands/file.ts
stevenguh: Fix the comment

src/cmd_line/commands/file.ts
stevenguh: Agree changed

src/cmd_line/commands/file.ts
stevenguh: If you search for when `createFileIfNotExists = false`, there are two instances -- `gf` and performFileJump.

I added the original else handling back
```
this._logger.error(`${this.arguments.name} does not exist.`);
return;
```

src/util/path.ts
stevenguh: Done

src/util/path.ts
stevenguh: Refactored the path util to be more unit testable, but one method is still hard to test since it depends on the platform.

test/cmd_line/tabCompletion.test.ts
stevenguh: Not sure if we need to, and how.

test/cmd_line/tabCompletion.test.ts
stevenguh: Done

src/util/path.ts
stevenguh: f[0] is the item name in a directory, and d[1] is the type of the item.
Basically, I am appending separator at the end as the indicator of directory item.
I refactored this method a bit so it only returns an array of strings.

src/util/path.ts
stevenguh: Added

src/util/path.ts
stevenguh: Added

test/cmd_line/tabCompletion.test.ts
jpoon: let's remove line 162 then ?

stevenguh: If we don't have <ESC> at the end, it will messed up other tests.

jpoon:
Let us know when the PR is ready for review again.

stevenguh:
Oh..it's ready.

test/cmd_line/tabCompletion.test.ts
jpoon: Tests shouldn't be dependent on each other. If this test fails causing other test failures, it will be difficult to deduce the cause.

stevenguh: Sure. How about I call <Esc> always after a test case?

Issue #2655, opened 2018-05-13
[Feature Request] Configure Sneak keymap


liamdiprose:
The sneak plugin offers similar functionality to find command (`f`). Some vim users believe sneak's usefulness eclipses the find command, and remap `f` to sneak [1]. 

For vim users that map this, it means a consistent key can be used [2], and returns the original map of the `s` key (the substitute command).

This is a request to add an option to the plugin's configuration to define a custom key to activate the sneak plugin. Users that rebind the key will have the original map of `s` returned to them. 

I have looked though the [source](https://github.com/VSCodeVim/Vim/blob/master/src/actions/plugins/sneak.ts#L12) and feel I could implement this myself, if I wasn't so busy. I will make a pull request in the future; if someone doesn't beat me to it.

---
[1] Vim-Sneak Readme: https://github.com/justinmk/vim-sneak#how-can-i-replace-f-with-sneak
[2] By default, Sneak uses `s` for moving, and `z` for text operations

longdeqidao:
Look forward to this feature. I use `s` to delete and jump into insert mode a lot.

Jayphen:
+1 I'd love to replace `f`

notiv-nt:
+1

mmqmzk:
Put this in your config:

```json
"vim.normalModeKeyBindingsNonRecursive": [
    {
      "before": [
        "f"
      ],
      "after": [
        "s"
      ]
    },
    {
      "before": [
        "F"
      ],
      "after": [
        "S"
      ]
    },
    {
      "before": [
        "s"
      ],
      "after": [
        "c",
        "l"
      ]
    },
  ],
```

This works for me, replace `f` and `F` key as you like.

J-Fields:
We have the `vim.sneakReplacesF` option, which does as you say

liamdiprose:
Sweet, thanks for implementing the feature @J-Fields :tada:

josephan:
Agree with above comment. Would be great if we can sneak 2 chars using f.

PR #4048, opened 2019-09-06
Add flag to replace `f` with a single-character sneak

### LINKED ISSUE #4012: Allow for placeholders in rebindings
This would augment the rebinding capabilities in the extension to allow us to do some more complex rebinds.

Example:
```
        {
            "before": ["f", "<placeholder>"],
            "after": ["/", "<placeholder", "<enter>"]
        },
```
This would remap f <letter> to / <letter> enter. I came up with this idea as I was trying to rebind f to  single char sneak (s <letter> enter) to f, but couldn't find any current way to do this.

J-Fields:
It's called `sneakReplacesF` and is nice because it basically makes `f` work across newlines.

I had to pull a few interfaces and a base class out from motion.ts to avoid a circular import.

Closes #4012

### DIFF
--- a/package.json
+++ b/package.json
@@ -566,6 +566,11 @@
           "markdownDescription": "Case sensitivity is determined by `ignorecase` and `smartcase`.",
           "default": false
         },
+        "vim.sneakReplacesF": {
+          "type": "boolean",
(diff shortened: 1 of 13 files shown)


liamdiprose:
Also closes #2655

Issue #3929, opened 2019-07-30
Text reflow doesn't respect tabs


aykevl:
When reflowing lines that are indented using tabs (using `gq` for example), tabs are replaced by spaces.

1.  Select a few commented lines with `Shift` `V`, for example the two commented lines in the following code (note the tabs at the beginning):
    ```go
    	// Gather the list of parameters for every call we're going to
    	// make.
    	callParams := make([]llvm.Value, call.OperandsCount()-1)
    	for i := range callParams {
    		callParams[i] = call.Operand(i)
    	}
    ```
2.  Press `g` `q`
3.  Notice how the line has become the following, with a single space for indentation completely breaking indentation:
    ```go
     // Gather the list of parameters for every call we're going to make.
    	callParams := make([]llvm.Value, call.OperandsCount()-1)
    	for i := range callParams {
    		callParams[i] = call.Operand(i)
    	}

    ```

I would expect this to become like the following:

```go
	// Gather the list of parameters for every call we're going to make.
	callParams := make([]llvm.Value, call.OperandsCount()-1)
	for i := range callParams {
		callParams[i] = call.Operand(i)
	}
```

Screenshot after reflowing:

![Screenshot at 2019-07-30 17-58-27](https://user-images.githubusercontent.com/729697/62145441-c6959680-b2f3-11e9-8c44-712e10bcb0f0.png)

**Environment:**

- Extension (VsCodeVim) version: 1.9.0
- VSCode version: 1.36.1
- OS: Linux (Debian Stretch)

Some settings that might be relevant:

```javascript
    "window.zoomLevel": 1,
    "editor.fontSize": 13,
    "editor.renderWhitespace": "boundary",
    "vim.searchHighlightColor": "rgba(255, 255, 0, 0.5)",
    "vim.hlsearch": true,
    "editor.wordWrap": "on",
    "editor.wrappingIndent": "deepIndent",
    "editor.minimap.enabled": false,
    "editor.hover.enabled": false,
    "[html]": {
        "editor.tabSize": 2,
    },
    "files.insertFinalNewline": true
```

I'm testing this with Go, which basically always uses tabs for indentation so this is a real problem for me.

PR #4050, opened 2019-09-07
Fix gq to handle tab indentation

orn688:
**What this PR does / why we need it**: The `gq` command to "reflow" selected text in visual line mode didn't properly handle tabs, replacing them with spaces.

**Which issue(s) this PR fixes**: (none, I can create one if needed)

**Special notes for your reviewer**: I added another commented-out `gq` test case under the assumption that it would also "run poorly on Travis". It passes locally though (along with the two old commented-out `gq` tests).

### DIFF
--- a/src/actions/operator.ts
+++ b/src/actions/operator.ts
@@ -841,22 +841,27 @@ class ActionVisualReflowParagraph extends BaseOperator {
     { singleLine: true, start: '' },
   ];
 
-  public getIndentationLevel(s: string): number {
+  public getIndentation(s: string): string {
+    // Use the indentation of the first non-whitespace line, if any such line is
+    // selected.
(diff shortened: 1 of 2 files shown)


J-Fields:
Fixes #3929

orn688:
Ah thanks! I checked a couple months ago to see if there was an issue for this and (incorrectly) assumed none had been created in the interim 🙂

J-Fields:
LGTM. Why do you assume it won't pass on CI though?

orn688:
> LGTM. Why do you assume it won't pass on CI though?

There's a comment above the old commented-out tests that says they don't work on Travis – I can try uncommenting them all to see if they succeed.

orn688:
The gq tests are pretty slow, but they're passing!

PR #4059, opened 2019-09-09
build(webpack): enable tree shaking, code splitting for webpack bundle

kwonoj:

This PR applies optimization to webpack configuration enable tree shaking & code splitting. When transpile target is set to cjs, webpack can't analyze import tree correctly for tree shaking, ends up bundles all codes. This PR applies webpac-specific ts transpilation config to emit esm import syntax instead, so webpack can tree shake dead codes in bundle. This change ends up shrinking down bundle size as below, about ~50kb of size gain for free.

- before
image

- after
image

Further there might be some other optimizations can be considered like targeting higher ecma (current vscode accepts es2017, in further es2018 can be used) / optimize some of bundle (`winston` is heavyweight) / `await import` for code splitting, but that's not a scope of this PR aims for now.


src/cmd_line/commands/file.ts
```diff
@@ -3,7 +3,7 @@ import * as node from '../node';
 import * as path from 'path';
 import * as util from 'util';
 import * as vscode from 'vscode';
-import untildify = require('untildify');
+import * as untildify from 'untildify';
```
kwonoj: in esm output `import = require` is being treated as non valid (only allowed to import types) and stripped out.

image

src/history/historyTracker.ts
```diff
@@ -9,7 +9,7 @@
  *
  * Undo/Redo will advance forward or backwards through Steps.
  */
-import DiffMatchPatch = require('diff-match-patch');
+import * as DiffMatchPatch from 'diff-match-patch';
```
kwonoj: in esm output import = require is being treated as non valid (only allowed to import types) and stripped out.

image

src/cmd_line/commands/file.ts
```diff
@@ -3,9 +3,10 @@ import * as node from '../node';
 import * as path from 'path';
 import * as util from 'util';
 import * as vscode from 'vscode';
-import untildify = require('untildify');
 import { Logger } from '../../util/logger';
 
+const untildify: typeof import('untildify') = require('untildify');
```
kwonoj: unlike esm import https://github.com/VSCodeVim/Vim/pull/4059/files#r322551630 , this still uses cjs require as replacement of `import = require` syntax: this 3rd party dep uses export default, either requires enable tsconfig `esmoduleInterop` or use cjs. instead of changing global config, import specific modules only via cjs.

xconverge:
Heres context for anyone interested:

https://webpack.js.org/guides/tree-shaking/

src/cmd_line/commands/file.ts
J-Fields: I'm not super well-versed in this stuff. My understanding is that `import` is needed rather than `require` because `import` can't be done dynamically anywhere, so static analysis like tree shaking is more possible? If that's the case, do we get the benefit of tree shaking, etc. for modules imported like this?

kwonoj: This `import` syntax is typescript specific: assign `typeof` imported module to assigned var (`const untildify`) from cjs `require`. Ts doesn't import types from cjs require, but allow import types via this syntax as needed - and runtime code will not contain this as same as other type information does.

kwonoj: This not cause of tree shaking: as described in comment, it's way to import `default` export modules without enabling tsconfig's esmoduleinterop.

kwonoj: For these modules I didn't try hard to enable esm import: it's being used lot / small and no huge gain via tree shaking as far as I can tell.

jpoon:
I am not familiar at all with webpack so I'll defer to the other contributors :)

kwonoj:
fyi my extension (https://github.com/kwonoj/vscode-activedocumentindicator) applied same practices for webpack config already.

webpack.config.js
```diff
@@ -31,7 +34,17 @@ const config = {
       {
         test: /\.ts$/,
         exclude: /node_modules/,
-        use: ['ts-loader'],
+        use: [
+          {
+            loader: 'ts-loader',
+            options: {
+              transpileOnly: true,
```
J-Fields: It says [here](https://github.com/TypeStrong/ts-loader#faster-builds) that `transpileOnly` will disable type checking... is this true?

Maybe we should make use of https://github.com/TypeStrong/fork-ts-checker-webpack-plugin

webpack.config.js
```diff
@@ -18,6 +18,9 @@ const config = {
     libraryTarget: 'commonjs2',
     devtoolModuleFilenameTemplate: '../[resource-path]',
   },
+
+  node: false, // prevent to include webpack's node polyfill, 📖 -> https://webpack.js.org/configuration/node/
```
J-Fields: This is saying we don't need node polyfills because we're running in node?

kwonoj: Yes.

webpack.config.js
kwonoj: Yes, and current gulp setup already have separate tsc runner I don't think webpack need duplicated type checking.

J-Fields: Yeah, but it doesn't seem like any type checking happens on, say, `vsce package`. I'm a bit skeptical about that, even though it would fail the tests, etc. @jpoon any thoughts?

kwonoj: There are lot of debates of webpack need to take care of type checking, in my opinion 1. Gulp build command already have seperate typecheck 2. And it should catch build failure instead of webpack 3. Webpack does typecheck only will increase build time by doing redundant type check.

jpoon: We definitely want type-checking. Right now we do it in parallel on build: 

```
gulp.task('build', gulp.series('prettier', gulp.parallel('webpack', 'tslint'), 'commit-hash'));
```

It's not clear to me if we have any value-add for moving to doing this with webpack.

J-Fields: Ah, you're right. If we're explicitly doing `tslint` alongside webpack, I think the proposed change makes sense.

jpoon:
If there's no other concerns, I'll merge this after build goes green.

J-Fields:
Actually @jpoon I tested this out the other day and `gulp build` doesn't catch type errors. Maybe the `tslint` that's run in parallel with webpack isn't working? I haven't looked closely into it.

jpoon:
We had to revert the original move to webpack as it was causing issues with neovim. As such, this PR is not in a merge-able state until that is resolved.

jpoon:
Due to the lack of activity, I'm closing this PR. Please re-open once the comments have been addressed.

Issue #4062, opened 2019-09-10
mapping control-something to escape in insert doesn't work


kyazdani42:
**Describe the bug**
Hi,
in insert mode, mapping C-something to Esc doesn't work properly.
when typing, you need to wait for 1/2 seconds before pressing the keys, else it produces an output like that:
```
Remapper: debug: trying to find matching remap. keys=a,<C-j>,<C-j>,<C-j>,<C-j>,<C-j>,<C-j>,<C-j>,<C-j>,<C-j>. mode=Insert. keybindings=insertModeKeyBindingsMap.
```
and cannot match the appropriate behavior. Tested with `<C-j>` and `<C-e>` \
`<C-c>` and `<C-[>` work fine though. It also works perfectly in visual mode

**To Reproduce**
in settings.json:
```
{
    "keyboard.dispatch": "keyCode",
    "vim.debug.loggingLevelForConsole": "debug",
    "vim.insertModeKeyBindings": [
        {
            "before": ["<c-j>"],
            "after": ["<Esc>"]
        }
    ],
    "vim.visualModeKeyBindings": [
        {
            "before": ["<c-j>"],
            "after": ["<Esc>"]
        }
    ]
}
```

**Expected behavior**
mapping c-something to escape in insert mode should properly be detected

**Environment (please complete the following information):**

- Extension (VsCodeVim) version: 1.10.2
- VSCode version: 1.38.0
- OS: Linux 5.2.11-1-MANJARO x86_64
- I also use Xmodmap to map caps lock to control but that shouldn't affect this as it worked fine with mapping caps lock to escape before

kyazdani42:
duplicate of #3530

Issue #3530, opened 2019-02-27
When entering visual mode, vim.visualModeKeyBindingsNonRecursive don't take effect for about a second


ohjames:
**Describe the bug**
Add to `vim.visualModeKeyBindingsNonRecursive`:

```
  { "before": ["n"], "after": ["j"] }
```

In an editor window, enter visual mode and quickly push `n` several times, the first few `n` key-presses will do a "search next" then the next few will be remapped as desired.

It takes about a second before the remaps take effect after entering visual mode.

- Extension (VsCodeVim) version: 1.1.0
- VSCode version: 1.31.1
- OS: archlinux

This is not an issue in vim or neovim.

sobjornstad:
I'm suspicious this may be related to #3221 (insert mode remapping doesn't work within a second of typing other characters in insert mode).

ohjames:
@sobjornstad I have that issue too. I'm having a lot of issues with remapping... VsCodeVim isn't very useable unless you are using the default mappings unfortunately and it's been that way for two years.

peitalin:
+1 same issue with the delay when switching to insert mode.

Issue #3221, opened 2018-11-26
Cannot remap <C-e> in insert mode


sobjornstad:
**Describe the bug**
The following setting has no effect:

    "vim.insertModeKeyBindingsNonRecursive": [
        {
            "before": ["<C-e>"],
            "after": ["<esc>"]
        }
    ],

(I don't actually want it to remap to just `<esc>`, but I'm starting simple.)

**To Reproduce**
Add the above text to `settings.json` and save. Enter insert mode and try pressing `C-e`. The character on the line below the current one is copied in -- rather than performing the mapped action of `<esc>`.

**Expected behavior**
VSCodeVim acts as if Escape was pressed.

**Environment (please complete the following information):**

- Extension (VsCodeVim) version: 0.16.12
- VSCode version: 1.29.1
- OS: Windows 10.0.15063

**Additional context**
This does not appear to be a duplicate of [Unable to remap `<C-s`> in insert mode](https://github.com/VSCodeVim/Vim/issues/2411). In that issue, the keys that people were trying to remap were not handled by VSCodeVim. In this case, `<C-e>` is [listed in packages.json](https://github.com/VSCodeVim/Vim/blob/1a5f358a1a57c62d5079093ad0dd12c2bf018bba/package.json#L164), and VSCodeVim is doing something with the key (copying the character below the cursor).

janhesters:
Same behavior. I tried to remap:

```json
"vim.insertModeKeyBindingsNonRecursive": [{ "before": ["<C-o>", "$"], "after": ["<C-e>"] }]
```

Apparently there is [trouble with remapping `ctrl` commands](https://stackoverflow.com/a/53897785/8331756) in general.

sobjornstad:
The workaround is good, but it doesn't help in my case, as I use Ctrl-E to scroll in normal mode without moving the cursor, and not having that is much more annoying than not being able to bind it in insert mode. The `keybindings.json` is not aware of VSCodeVim's modes, so I can't override only in insert mode.

I am not sure that the parent issue @janhesters listed applies to the behavior for `C-e` in particular.. Maybe I'm misunderstanding, but the `package.json` file appears to be binding to Ctrl-E, and VSCodeVim is able to do things specific to the extension when I press Ctrl-E, so this seems like it may be a different issue.

jpoon:
Dupe of https://github.com/VSCodeVim/Vim/issues/3221

sobjornstad:
[...]

PR #4063, opened 2019-09-10
fix insertion mappings

kyazdani42:
There is an issue in insert modes with key mappings not working properly. \
When typing, keymaps are just not recognized.
 \
This PR tries to fix this by resetting the keys buffer on insert mode if the mapping didn't seem to be a potential remap.

However, it seems there are new issues that i encountered. For instance, this mapping:
```
{
  before: ["<C-e>", "<C-e>"],
  after: ["<Esc>"]
}
```
works, but it erases the previous character from the screen, which is a strange behavior i must say.

#3221 #3530 #4062

first time working on an open source project


dimxdim:
Tried this fix with current vscode.
I confirm this fixes the issue of needing to wait for timeout to expire before <C-x> insert mode mapping would take place.
Minimal settings to reproduce:
```
{
    "vim.insertModeKeyBindingsNonRecursive": [
        {
            "before": [ "<c-x>" ],
            "after": [ "<ESC>"]
        }
    ],
    "vim.debug.loggingLevelForConsole": "debug"
}
```
Type quickly `a` `C-x`, we now exit insert mode correctly, as expected.

kyazdani42:
Hello,
thanks for your feedback.

Indeed, this fix is just enough for simple sequence mappings, this is why it says 'short' in the PR name. I needed more time to think about this before coming up with a proper solution.

And so i did, i might have found an other way that would fix all mapping issues in insert mode, but i'd have to dig in the code more deeply. I will start working on this again when i have some free time.

dimxdim:
Hi. I really look forward to your next patch. It's a pity there are still bugs in basic mappings, which make it harder to customize and prevent people from switching over from vim/nvim. I am too in-between those. I will be using your patch for the time being and report if there are any related issues. Thx again!

dimxdim:
Today I have tested your new fix. I confirm it works and there is no delay required to trigger either single or double <c-something>.

Regarding the issue you describe with `<c-e><c-e>`, I have come across it too when testing. It is not specific to <c-e> and it is not caused by your fix. It is reproducible with the latest vscodevim in the same manner, just by waiting 2 sec before pressing <c-e><c-e> (or another combination `<c-smth><c-smth>`).

Here is what I found. Using this settings.json file (to disable auto-update overwriting the extension, also using this as the to-be-edited file)"
```json
{
    "vim.insertModeKeyBindingsNonRecursive": [
        {
            "before": [ "<c-x>", "<c-x>" ],
            "after": [ "<esc>" ]
        }
    ],
    "vim.debug.loggingLevelForConsole": "debug",
    "update.mode": "manual",
    "update.enableWindowsBackgroundUpdates": false,
    "extensions.autoCheckUpdates": false,
    "extensions.autoUpdate": false
}
```
Go to line before last, move to the end of the line: `Gk$`.
Enter insert by changing inside the last word: `ciw`.
The word `false` has disappeared and we are in insert.
Then (after waiting 2 sec if using the latest non-patched vscodevim): `<c-x><c-x>`
The change will be undone! (bug)
The same behavior is observed if there was a change to the document before entering insert mode.
If the change was bigger, like appending `"test": "test"`, then multiple `I<c-x><c-x>` will result in multiple "undos", in a step-by-step manner, not like vim.

Try re-opening vscode and without modifying the document: `a<c-x><c-x>`. Will successfully exit insert mode without other modifications (probably because there is no history at this point?). But this time the debug logs: 
```[Extension Host] DocumentChange: warn: Something bad happened in removeChange```

Finally, there is another clue the above "successful" exiting is not the same as just `<esc>`: the cursor is moved 1 space to the left from the position is it supposed to be (like end of line).

It seems to me something is messing up current file's vim-specific history or the file's state.

kyazdani42:
thanks again for the feedback, i feel like this should be tested by more people, although mapping keys in insert mode is not that common.

This bug might need to be issued  properly, it's kind of outside the scope of this PR. It might be good to address this soon, and i might just make another PR if i can. 

in the meantime, i'll wait for the maintainers to take a look at this.

PR #4074, opened 2019-09-13
Fix incorrect position when editing the same file in 2 splits

### LINKED ISSUE #3713: 'dd' followed by any character jumps cursor to end of file.
**Describe the bug**
After deleting a line with `dd` any character afterwards jumps me to the end of the file.

This issue has been marked closed in an [old Closed Issue](https://github.com/VSCodeVim/Vim/issues/2136)

**To Reproduce**

- Open any file
- Remove a line in normal mode ('dd')
- Enter any character

**Expected behavior**
Not jumping to end of file.

**Environment:**
- Extension (VsCodeVim) Version: `1.7.0`
- VSCode version: `1.34.0-insider` (user setup)
- OS: Windows_NT x64 `10.0.17134`
- Date: `2019-04-30T08:17.612Z`

**Additional context**
<details>
<summary>VsCodeVim settings</summary>

```json
{
    "vim.easymotion": true,
    "vim.textwidth": 140,
    "vim.useCtrlKeys": false,
    "vim.history": 100,
        {
            "before": ["j", "j"],
            "after": ["<Esc>"]
        }
    ],
        {
            "before": ["<C-n>"],
            "commands": [":nohl"]
        }
    ],
    "vim.leader": ",",
    "vim.sneak": false,
    "vim.incsearch": true,
    "vim.useSystemClipboard": true,
    "vim.hlsearch": true,
    "vim.searchHighlightColor": "rgba(150, 150, 255, 0.3)",
    "vim.surround": true
}
```

uHOOCCOOHu:
Fix setting wrong (position) states when editing the same file in two panels.

Fixes #3713 , hopefully.
I've checked [the reproduction steps](https://github.com/VSCodeVim/Vim/issues/3713#issuecomment-508525383) from @elprans , the test mentioned in [comments](https://github.com/VSCodeVim/Vim/blob/4a5fff2b50a5f1ebece157327dd3aaadd39903ea/src/mode/modeHandler.ts#L109),  and some other random tests.
It is no longer triggered anymore.


J-Fields:
I've got to dig into this some more, but a few things spring to mind:
This handles the same misbehavior as https://github.com/VSCodeVim/Vim/pull/3539, right? Why is this solution better?
Why don't we avoid sending the event in the first place (by modifying [this conditional](https://github.com/VSCodeVim/Vim/blob/master/extension.ts#L255), say)?

uHOOCCOOHu:
> This handles the same misbehavior as #3539, right? Why is this solution better?

Oops, I just have not seen that. But yes, it does also fix #3455 , and **not** [break undo](https://github.com/VSCodeVim/Vim/pull/3539#issuecomment-526885116).
 I just keep `EditorIdentity` untouched. Since `viewColumn` can be changed, I don't know if it would cause some problems.
Well, I think `EditorIdentity` should be another issue.

> Why don't we avoid sending the event in the first place (by modifying this conditional, say)?

You are right.

uHOOCCOOHu:
Rebased to master.

J-Fields:
Thanks for tracking this down, @uHOOCCOOHu! I'll play around with it a bit more, but this seems to work.

kdnk:
When will this feature be released?

PR #4089, opened 2019-09-22
Add shift+tab support for cmd line

stevenguh:
Add shift+tab to go back on autocompletion on cmd line

This is related to
#4032

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,6 +89,11 @@
         "when": "editorFocus && vim.active && vim.mode != 'Insert' && !inDebugRepl"
       },
       {
+        "key": "shift+tab",
+        "command": "extension.vim_shift+tab",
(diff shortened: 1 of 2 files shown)


J-Fields:
Cool, thanks a bunch @stevenguh!

It occurs to me that we should at some point make a distinction between commands that work and commands that will autocomplete. For instance, `:wri` will work, but vim will never autocomplete to `:wri` - it'll go all the way to `:write`.

stevenguh:
The current implementation does not autocomplete to `:wri`, but I wouldn't be surprise if there are instances that are like that in the current implementation.

J-Fields:
Yeah, that was just an example. Currently we need to choose between an "intermediate" command (one that is not fully specified nor fully abbreviated) working or not clogging up our autocomplete options. I'll write up a ticket - I think I've got a nice way of accomplishing this.

PR #4105, opened 2019-09-27
Do not override foreground color

wSedlacek:
The foreground color is not in the config settings so by overriding it on every mode change it just prevents me from setting it to fix it the default from the Material Theme.


wSedlacek:
Looks like you can set as an array like. This was my bad.
```
"vim.statusBarColors.insert": ["#4CAF50", "#FFF"]
```

PR #4122, opened 2019-10-02
WIP: Add support for <C-a> and g<C-a> in visual modes

### LINKED ISSUE #3226: g Ctrl+A column increment
In Vim 8, the first number in a selection can be incremented by pressing Ctrl-A. If the selection covers several lines, the first number in the selection on each line is incremented. Alternatively, numbers in a selection covering several lines can be converted to a sequence by typing g Ctrl-A. For example, start with the following line:

my_array[0] = 0;
Then copy it using Y6p (copy the line and paste it six times). The result is:

my_array[0] = 0;
my_array[0] = 0;
my_array[0] = 0;
my_array[0] = 0;
my_array[0] = 0;
my_array[0] = 0;
my_array[0] = 0;
With the cursor on the first 0 in the first line, start a blockwise select by pressing Ctrl-V (or Ctrl-Q if you use Ctrl-V for pasting). Move the cursor down to select the first column of zeros, then press g Ctrl-A. Using Vim 8, that will give:

my_array[1] = 0;
my_array[2] = 0;
my_array[3] = 0;
my_array[4] = 0;
my_array[5] = 0;
my_array[6] = 0;
my_array[7] = 0;
Vim 7 and earlier need a script or a macro to create a sequence.

### LINKED ISSUE #4112: <C-a> does not work in visual mode
`<C-a>` and `<C-x>` (increment and decrement) do not work in visual modes.
It should, on each line, increment the number by `count`. After this works, I think `g<C-a>` will be pretty straightforward.

J-Fields:
This is WIP. Still to-do:

- When a 9 rolls over and it goes from i digits to i+1 digits, we need to put the cursor 1 space to the left
- Visual-block mode 1 column wide should only affect that column
- Sometimes text afterward is deleted

Fixes #4112, fixes #3226

### DIFF
--- a/src/actions/commands/actions.ts
+++ b/src/actions/commands/actions.ts
@@ -13,7 +13,7 @@ import { Position, PositionDiff } from './../../common/motion/position';
 import { Range } from './../../common/motion/range';
 import { NumericString } from './../../common/number/numericString';
 import { configuration } from './../../configuration/configuration';
-import { ModeName, Mode } from './../../mode/mode';
+import { ModeName } from './../../mode/mode';
 import { VisualBlockMode } from './../../mode/modes';
 import { Register, RegisterMode } from './../../register/register';
 import { SearchDirection, SearchState } from './../../state/searchState';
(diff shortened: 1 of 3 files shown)


J-Fields:
Blocked by #4469. Even when that refactor goes through, I'll likely want to re-write this mostly from scratch.

J-Fields:
#4469 is now merged, so this isn't blocked. Hoping to get back to this before `1.15.0`, though like I said earlier, it'll likely entail a re-write

PR #4125, opened 2019-10-02
Adding backspace motion in visual mode

### LINKED ISSUE #3220: Backspace doesn't move cursor back in visual mode (like vim)
Backspace doesn't move the cursor backwards in 'visual mode' like vim does, it should perform the same action as left cursor key, useful for altering your visual selection.

Note it works as expected in 'normal' mode

Thanks

pedroapfilho:
Fixes #3220

### DIFF
--- a/src/actions/motion.ts
+++ b/src/actions/motion.ts
@@ -440,7 +440,9 @@ class BackSpaceInVisualMode extends BaseMovement {
   keys = ['<BS>'];
 
   public async execAction(position: Position, vimState: VimState): Promise<Position> {
-    return position.getLeftThroughLineBreaks();
+    return configuration.whichwrap.includes('b')
+      ? position.getLeftThroughLineBreaks()
+      : position.getLeft();
   }
 }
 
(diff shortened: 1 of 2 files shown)


J-Fields:
Thanks! Could you add support for `VisualBlock` mode as well?
Also, a few basic test cases would be awesome.

pedroapfilho:
Of course I can! Just to know, on VisualBlock, it should select the last block, right?

J-Fields:
Not quite sure what you mean tbh. Backspace should work the same as the left arrow in visual block mode

src/actions/motion.ts
```diff
@@ -434,6 +434,16 @@ class BackSpaceInNormalMode extends BaseMovement {
   }
 }
 
+@RegisterAction
+class BackSpaceInVisualMode extends BaseMovement {
+  modes = [ModeName.Visual, ModeName.VisualBlock];
+  keys = ['<BS>'];
+
+  public async execAction(position: Position, vimState: VimState): Promise<Position> {
+    return position.getLeftThroughLineBreaks();
```
J-Fields: It should only go through line breaks if `whichwrap` contains `b`
http://vimdoc.sourceforge.net/htmldoc/options.html#'whichwrap'

pedroapfilho: Done!

PR #4127, opened 2019-10-03
Fixes gj/gk behaviour on wrapped lines; closes #4120

### LINKED ISSUE #4120: Cursor movement in wrapped lines changed from 1.10.2 to 1.11.0
**Describe the bug**
Using the `gj` and `gk` binding to [move by wrapped line instead of the file line](https://vim.fandom.com/wiki/Move_cursor_by_display_lines_when_wrapping) does not work correctly in version 1.11.0, but works fine when downgrading to 1.10.2.

Specifically, here are the issues I've experienced:
1) sometimes it will move up (`gj`) by file line, but down (`gk`) by wrapped line. Though this behavior is inconsistent and reverses itself sometimes.
2) when it does successfully move by wrapped line, it'll change the relative character position, sometimes moving to the last character in the wrapped line, sometimes the first, sometimes by a (seemingly) random number of characters in either direction. (Note: this is moving in a block of text where there are several wrapped lines for every file line)

**To Reproduce**
Go to file with lots of wrapped lines (or take a normal file and make the window really narrow) and move up and down with `gj` and `gk`. 

**Expected behavior**
It to move by one wrapped line and maintain the same lateral character position (assuming there is a character present in that lateral character position).

- Extension (VsCodeVim) version: 1.11.0
- VSCode version: 1.37.1
- OS: Manjaro 18.1 

Note: Manjaro currently doesn't have vscode 1.38.x on it's stable branch. This is why I'm a version behind the latest.

hetmankp:
Closes #4120. This more or less reverts the functionality introduced by commit
4b2e07933 and raised in issue #3890.

Unfortunately it is not currently possible to fully implement display
line movements over wrapped lines that will always preserve cursor
position. One of two things is necessary to change this, either the
VSCode API needs to be extended to return information about wrapped
lines (but this was rejected in issue microsoft/vscode#23045).
Alternatively VSCodeVim would need to reïimplement all motions using
vscode.commands.executeCommand() commands rather than manipulating
vscode.window.activeTextEditor.selections directly, as the former
preserves column position while the latter does not. However it's not
clear to me if all VIM beahviour could be reïmplemented this way without
first trying this out and sounds like a bit of a ground up rewrite.

It seems for now we're stuck with gj/gk not preserving column position
when jumping across short lines.

### DIFF
--- a/src/actions/motion.ts
+++ b/src/actions/motion.ts
@@ -163,6 +163,19 @@ abstract class MoveByScreenLineMaintainDesiredColumn extends MoveByScreenLine {
   }
 }
 
+class MoveDownByScreenLineMaintainDesiredColumn extends MoveByScreenLineMaintainDesiredColumn {
+  movementType: CursorMovePosition = 'down';
+  by: CursorMoveByUnit = 'wrappedLine';
+  value = 1;
+}
+
+
+class MoveUpByScreenLineMaintainDesiredColumn extends MoveByScreenLineMaintainDesiredColumn {
+  movementType: CursorMovePosition = 'up';
+  by: CursorMoveByUnit = 'wrappedLine';
+  value = 1;
+}
+
 class MoveDownFoldFix extends MoveByScreenLineMaintainDesiredColumn {
   movementType: CursorMovePosition = 'down';
   by: CursorMoveByUnit = 'line';
@@ -785,7 +798,7 @@ class MoveScreenLineCenter extends MoveByScreenLine {
 }
 
 @RegisterAction
-export class MoveUpByScreenLineMaintainDesiredColumn extends MoveByScreenLineMaintainDesiredColumn {
+export class MoveUpByDisplayLine extends MoveByScreenLine {
   modes = [ModeName.Insert, ModeName.Normal, ModeName.Visual];
   keys = [['g', 'k'], ['g', '<up>']];
   movementType: CursorMovePosition = 'up';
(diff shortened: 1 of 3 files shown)


J-Fields:
> Alternatively VSCodeVim would need to reïimplement ***all motions*** using
vscode.commands.executeCommand() commands rather than manipulating
vscode.window.activeTextEditor.selections directly

Why's this? Aren't basically all motions other than `gj` agnostic to whether lines are wrapped?

hetmankp:
> > Alternatively VSCodeVim would need to reïimplement _**all motions**_ using
> > vscode.commands.executeCommand() commands rather than manipulating
> > vscode.window.activeTextEditor.selections directly
> 
> Why's this? Aren't basically all motions other than `gj` agnostic to whether lines are wrapped?

@J-Fields, this is specific to maintaining the column position. What I gathered from microsoft/vscode#23045 is that VSCode already maintains column position for us. This works OK as long as we manipulate the cursor via `vscode.commands.executeCommand()` but the desired column is reset every time you modify the `vscode.window.activeTextEditor.selection` & `.selections` attributes directly.

Unfortunately, the only code with an awareness of the display column position when wrapping is turned on, is the internals of VSCode and it seems the developers are uninterested in exposing this. So the only option would be to implement all motions using `executeCommand()`, to prevent the desired column being reset.

Without trying this I'm not sure if there would be any corner cases where VIM behaviour could not be expressed using existing VSCode commands, however the developers did indicate that they would be open to adding additional commands as needed on a case by case basis.

J-Fields:
Got it, thanks! Shame there isn't an API for this.

PR #4128, opened 2019-10-03
Update CONTRIBUTING.md after webpack bundling addition

hetmankp:
With the addition of Webpack bundling (issue #3127, commit 1f80b2d)
the instructions for how to run tests were no longer working.


hetmankp:
The comments in this change may need to be modified as they're basically my best guess at the intent.

Issue #3975, opened 2019-08-09
Exit from Visual-Line mode moves cursor


pattersongp:
**Describe the bug**
When you enter visual-line mode, then immediately exit, your cursor is taken to the end of the line.

**To Reproduce**

1. Enter visual line mode, eg shift+v
2.  Escape
3. Observe

**Expected behavior**
The cursor should not move at all

**Environment**
- Extension (VsCodeVim) version: v1.9.0
- VSCode version: 1.36.1
- OS: macos x 10.14.5

caleywoods:
I'd like to take a stab at this one @J-Fields  - any details to offer on where a good starting point in the code would be? What handles entering visual line selection mode when the user enters `shift + v`?

Is [this](https://github.com/VSCodeVim/Vim/blob/38c268a26dadbd0fed1ea297923e7640f358097c/src/mode/modeHandler.ts#L112) involved? Sounds like a fun time :D

J-Fields:
@caleywoods It seems like [this](https://github.com/VSCodeVim/Vim/blob/master/src%2Fmode%2FmodeHandler.ts#L1219) is where we create the selection and [this](
https://github.com/VSCodeVim/Vim/blob/master/src%2Fmode%2FmodeHandler.ts#L610) is where we set the cursor position after going back to normal mode

caleywoods:
I think I've got a solution for this but I'm having a hard time running the existing test suite. Removing `.vscode-test` and then executing `gulp test` does all of the docker magic but then it just seems to exit (with code 0) successfully without any of the test output I remember from before.

I edited the dockerfile to remove the entrypoint and ran it again, then went into a bash prompt on the container and manually tried `npm test` and they do the exact same thing. 

[Screenshot](https://i.imgur.com/Z03sv5G.png)

caleywoods:
Confirming that it would be expected for some of the visual line selection tests written prior to this fix would now fail such as `gq handles spaces after single comments correctly`. The old test expects the cursors end position to be on line 1 (the newly created line from the wrap). Post-fix, however, the cursor will reside on line 0 (the original line) since the cursor now maintains the position that it started at instead of moving elsewhere. 

[Current test](https://github.com/VSCodeVim/Vim/blob/master/test/mode/modeNormal.test.ts#L1334)

I'm thinking the new test would be:
```
  newTest({
    title: 'gq handles spaces after single line comments correctly',
    start: [
[...]

PR #4141, opened 2019-10-06
Return cursor to start position when immediately exiting visual line mode

caleywoods:
This PR allows the cursor to return to its original position when the user enters visual line selection mode and then immediately exits it with `<Esc>`. This replicates VIM's behavior when pressing `V<Esc>`.

[This issue](https://github.com/VSCodeVim/Vim/issues/3975)

I noticed when working on this issue that in VIM when you enter visual line mode with `V` and then you select more lines with `j` for example, the cursor moves to that same column/position on the line below but it looks like VSCodeVIM always jumps to the end of the next line. 

Not sure if that's intended, perhaps it should be filed as another issue.

### DIFF
--- a/test/mode/modeVisualLine.test.ts
+++ b/test/mode/modeVisualLine.test.ts
@@ -471,4 +471,26 @@ suite('Mode Visual Line', () => {
       end: ['111', '222_|', ' ', '444_', '555'],
     });
   });
+
+  suite('Cursor behavior when immediately exiting', () => {
+    newTest({
+      title: 'Returns cursor to original position when immediately exiting visual line',
+      start: ['rocinante', 'nauvoo', '|anubis', 'canterbury'],
+      keysPressed: 'V<Esc>',
+      end: ['rocinante', 'nauvoo', '|anubis', 'canterbury'],
+    });
+
+    newTest({
+      title: 'Does not return cursor if a movement is made before exiting visual line',
+      start: [
+        'Don\'t cry because it\'s over|,',
+        ':smile because it happened.',
+      ],
+      keysPressed: 'Vj<Esc>',
+      end: [
+        'Don\'t cry because it\'s over,',
+        ':smile because it happened|.',
+      ],
+    });
+  });
 });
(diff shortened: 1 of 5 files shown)


J-Fields:
In Vim, you can move your cursor anywhere within the selected region when in visual line mode. VSCode has no concept of a selection that does not have a cursor at the start or end, so we can't faithfully recreate this in a way that wouldn't be confusing, I think. Maybe we should handle more complicated cases than this, but I think `V<Esc>` is a good start, at least.

J-Fields:
Actually, I've rethought this - I think we should put the cursor on the original column, even if they move down or up lines. This is a bit more limited than Vim, but it's more consistent & intuitive than only preserving the column when you immediately exit.

caleywoods:
Cool, will be just a bit more work to fix the tests as I had started doing originally. I can address it in the evenings this week.

test/mode/modeVisualLine.test.ts
```diff
@@ -471,4 +471,26 @@ suite('Mode Visual Line', () => {
       end: ['111', '222_|', ' ', '444_', '555'],
     });
   });
+
+  suite('Cursor behavior when immediately exiting', () => {
+    newTest({
+      title: 'Returns cursor to original position when immediately exiting visual line',
+      start: ['rocinante', 'nauvoo', '|anubis', 'canterbury'],
```
xconverge: I like the cut of your jib

caleywoods: Remember the cant

J-Fields:
Hey @caleywoods any update on this?

caleywoods:
@J-Fields I'll see if I can sink some time into it this week. IIRC my original change was what we're now after, it had ~20 some odd failing tests. Of those 20, some, if not all, need re-written to respect the new cursor behavior.

Sorry for falling off the map on this one, I've been spending my evenings working through some interview practice books/problems - Thanks for the reminder, have a great week.

J-Fields:
Any update @caleywoods? If you're too busy with other things, I'm happy to take a crack at polishing this up!

caleywoods:
@J-Fields I would appreciate you doing that. The functionality was pretty simple and it seemed like there were roughly ~20 tests that needed updated which had their final cursor position altered by this chance.

J-Fields:
Will do - thanks for getting this off the ground! Hope you have a great new year 🙂

PR #4148, opened 2019-10-08
Added support for :sort u (sort + unique)

jdxcode:

Vim supports this functionality so it brings it up to parity.

None (that I found)

💙 this extension!


src/cmd_line/subparsers/sort.ts
```diff
@@ -3,15 +3,17 @@ import { Scanner } from '../scanner';
...
-  let scannedArgs: node.ISortCommandArguments = { reverse: false, ignoreCase: false };
+  let scannedArgs: node.ISortCommandArguments = { reverse: false, ignoreCase: false, unique: false };
   let scanner = new Scanner(args);
   const c = scanner.next();
   scannedArgs.reverse = c === '!';
 
-  scannedArgs.ignoreCase = scanner.nextWord() === 'i';
+  const nextWord = scanner.nextWord();
+  scannedArgs.ignoreCase = nextWord === 'i';
+  scannedArgs.unique = nextWord === 'u';
```
J-Fields: Is it possible to do both unique and ignorecase?

J-Fields: You're probably right that it's not a common usage. I'll leave it up to you, but at least write a comment mentioning that this case isn't covered. Thanks!

Issue #4055, opened 2019-09-08
Use `ExtensionContext.globalStoragePath` instead of `appdirectory` library


J-Fields:
VSCode tells us where to put our files, so we should be good citizens and store history files there. It'd be ideal to move existing stuff there the first time.
We can also kill the `mkdirp` dependency while we're at it, since the parent directory of `ExtensionContext.globalStoragePath` is guaranteed to exist.

PR #4151, opened 2019-10-09
Simplify storage directory to use supplied extension directory

xconverge:
This is pretty ugly, any suggestions would be appreciated...

We do not get the extension context until the extension is activated (and we need this to get the storage folder), yet we construct some stuff at startup (prior to activation), so now there is a weird dependency in order of operations.


J-Fields:
Fixes #4055

xconverge:
I couldnt kill mkdirp because we need to create the subfolder unfortunately

J-Fields:
https://code.visualstudio.com/api/references/vscode-api#ExtensionContext
> An absolute file path in which the extension can store global state. The directory might not exist on disk and creation is up to the extension. However, the parent directory is guaranteed to be existent.

Isn't it the case we could just use `fs.mktdir` instead, since we only need to create one directory?

xconverge:
Yes I will do that, duh! Thanks

J-Fields:
Users will lose their history files after updating, right? Probably not a big enough deal to bother copying things over the first time, but we should be aware of that

xconverge:
We have done this in the past without any complaints. Not ideal but I imagine it will be ok, we had far more complaints about WHERE we were storing stuff

PR #4152, opened 2019-10-09
Update neovim path to be machine-overridable for remote use

xconverge:
This is really minor but at least is a slight improvement with no downsides


package.json
```diff
@@ -354,7 +354,7 @@
       }
     ],
     "configuration": {
-      "title": "Vim Configuration",
```
xconverge: I noticed in my settings extension list, we were the odd-man out here with this verbosity

J-Fields:
We've got `"extensionKind": "ui"` in package.json. I thought this meant that the extension runs on the user's machine, not remotely, and so this wouldn't benefit us? Only vaguely familiar with this stuff, though, so I could be wrong.

xconverge:
Yea the documentation for machine-overridable is not great, I am fine if we skip this for now, I just figured in the offchance it helps someone I couldn't see any harm.

jpoon:
LGTM. Ship it.

xconverge:
Sure, not fully sure what it does, but pretty sure it doesn't hurt. So if it helps someone then 🤷‍♂

Issue #4047, opened 2019-09-06
A failed `f`/`F` cannot be retried with `;` or `,`


J-Fields:
**To Reproduce**

1.  With text `abc | xyz`
2.  `fb` ... no match, does nothing as expected
3.  `,` ... will not work

**Expected behavior**
Cursor jumps back to the `b`. It actually uses the *previous* character to search backward.

PR #4157, opened 2019-10-10
Fix back search when no matching (#4047)

seachicken:
For find character action (e.g `f`/`F`, `t`/`T`) when no matching, then after back search is not working. This PR solved the problems.
`lastSemicolonRepeatableMovement` and `lastCommaRepeatableMovement` setting are necessary on find character action failed too. Because back search uses those variables.

#4047


J-Fields:
Thank you!

Issue #1453, opened 2017-03-30
`za` for toggling folds does not work


hk0i:
Please *thumbs-up* 👍 this issue if it personally affects you! You can do this by clicking on the emoji-face on the top right of this post. Issues with more thumbs-up will be prioritized.

-----

### What did you do?

Place cursor over a section I want to fold then press <kbd>Z, a</kbd>

From vim manual:

```
							*za*
za		When on a closed fold: open it.  When folds are nested, you
		may have to use "za" several times.  When a count is given,
		that many closed folds are opened.
		When on an open fold: close it and set 'foldenable'.  This
		will only close one level, since using "za" again will open
		the fold.  When a count is given that many folds will be
		closed (that's not the same as repeating "za" that many
		times).
```

### What did you expect to happen?

Section is folded on first press, and unfolded when pressing again.

### What happened instead?

Nothing

### Side Note:

<kbd>Z, c</kbd> doesn't seem to work well either, it closes the section but then expands again after a second or so.

### Technical details:

* VSCode Version: 1.10.1
* VsCodeVim Version: 0.6.11
* OS: OS X

xconverge:
https://github.com/VSCodeVim/Vim/issues/1004 this is part of what you are seeing

hk0i:
Is there any work around to make code folds work? What I really wanted to do was map the space key to toggle folds like I do in vim, but VSCode seems to have nothing built in for toggling folds.

Chillee:
I think this is currently impossible using the fold API. Even ignoring our issues with auto-expanding folds (which I am hopeful that https://github.com/VSCodeVim/Vim/issues/1004#issuecomment-301677076 can fix), all we have for folds is an API call for opening them and an API call for closing them. As there is no API call for telling us whether we are currently within a fold or not, we don't know which one to call.


Actually..., I think I just thought of a hack for solving this problem.

sudowork:
I was attempting to map `za` to `editor.unfold` and `zA` to `editor.unfoldRecursively`; however, the commands don't seem to work properly. Is this an expected limitation? I know there was some discussion previously regarding limitations of the folding API, but directly invoking these commands from the command palette seem to work properly. Here's an excerpt from my settings:

```
    "vim.otherModesKeyBindingsNonRecursive": [
        {
            "before": ["z", "a"],
            "after": [],
            "commands": [
                {
                    "command": "editor.unfold",
                    "args": []
                }
            ]
        },
        {
[...]

PR #4160, opened 2019-10-11
Add support for za Vim binding

aquova:
Adds support for the Vim 'za' keybinding, which toggles a fold.

#1453 

This simply binds the new VSCode editor.toggleFold option to the 'za' combination. I was unable to find a way to implement specifying a number modifier, as you need to know whether or not to travel up or down.

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -567,7 +567,7 @@ Pretty much everything fold-related is blocked by [this issue](https://github.co
 | :white_check_mark: | zO                       | Open all folds under the cursor recursively.                                                                 |
 | :white_check_mark: | zc                       | Close one fold under the cursor. When a count is given, that many folds deep are closed.                     |
 | :white_check_mark: | zC                       | Close all folds under the cursor recursively.                                                                |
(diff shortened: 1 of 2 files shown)


xconverge:
@RegisterAction
class CommandToggleFold extends CommandFold {
  keys = ['z', 'a'];
  commandName = 'editor.toggleFold';
    await vscode.commands.executeCommand(this.commandName);
    vimState.cursors = await getCursorsAfterSync();
    await vimState.setCurrentMode(ModeName.Normal);
    return vimState;
  }
}

This seems to work better for me? are you using the foldFix configuration option?

xconverge:
Without the getCursors sync and foldFix when I toggle it closed it reopens right away unless you are perfectly on the start of the fold (ie. not within the folded region"

aquova:
Yes, it seems you're right. When foldFix is disabled, I'm seeing it reopen unless you're directly on the start of the fold.

xconverge:
Looks good and when I tested it it seems to work fine-ish, except for the whole foldFix concept but thats another ticket!

PR #4161, opened 2019-10-11
Try to parse neovim from PATH if neovimPath is empty

### LINKED ISSUE #3584: Search PATH for nvim when enableNeovim is true and neovimPath is empty
I believe this is a different request than https://github.com/VSCodeVim/Vim/issues/3458 but it is related. It could make sense to do both, so I opened a new issue.

I share my VSCode settings between machines and the neovim path is not always the same on each machine (OS differences as well as just the directory in which things are installed). However, I do always make sure `nvim`/`nvim.exe` is on my `PATH`.

I suggest VSCode Vim automatically search the `PATH` directories for `nvim`/`nvim.exe` when `vim.neovimPath` is empty (but `vim.enableNeovim` is enabled). That way as long as I ensure my `PATH` is up-to-date, everything just works between machines with no need to mess with config (or have OS-specific paths as suggested in #3458)

xconverge:
Fixes  #3584 

Tested on windows so far, will also test on linux


README.md
```diff
@@ -140,7 +140,7 @@ To leverage neovim for Ex-commands,
 | Setting          | Description                    | Type    | Default Value |
 | ---------------- | ------------------------------ | ------- | ------------- |
 | vim.enableNeovim | Enable Neovim                  | Boolean | false         |
-| vim.neovimPath   | Full path to neovim executable | String  |               |
+| vim.neovimPath   | Full path to neovim executable. If left empty, PATH environment variable will be automatically checked for neovim path automatically. | String  |               |
```
jpoon: One too many automatically's :)

src/configuration/validators/neovimValidator.ts
```diff
@@ -8,12 +8,31 @@ export class NeovimValidator implements IConfigurationValidator {
...
+                return;
+              }
+            });
+          }
+        }
         await promisify(execFile)(config.neovimPath, ['--version']);
       } catch (e) {
+        let errorMessage = `Invalid neovimPath. ${e.message}.`;
+        if (triedToParsePath) {
+          errorMessage += 'Tried to parse PATH.';
```
jpoon: nit: should we output the path we tried?  `config.neovimPath

src/configuration/validators/neovimValidator.ts
```diff
@@ -8,12 +8,31 @@ export class NeovimValidator implements IConfigurationValidator {
...
+      let triedToParsePath = false;
       try {
+        // Try to find nvim in path if it is not defined
+        if (config.neovimPath === '') {
+          const pathVar = process.env.PATH;
+          if (pathVar) {
+            pathVar.split(';').forEach(element => {
+              if (element.toLocaleLowerCase().includes('neovim')) {
+                const path = require('path');
+                config.neovimPath = element + path.sep + 'nvim';
```
jpoon: path.join?

src/configuration/validators/neovimValidator.ts
xconverge: Sure, would be easy

src/configuration/validators/neovimValidator.ts
xconverge: done

README.md
xconverge: oops oops

src/configuration/validators/neovimValidator.ts
```diff
@@ -2,18 +2,37 @@ import { IConfiguration } from '../iconfiguration';
...
     const result = new ValidatorResults();
 
     if (config.enableNeovim) {
+      let triedToParsePath = false;
       try {
+        // Try to find nvim in path if it is not defined
+        if (config.neovimPath === '') {
+          const pathVar = process.env.PATH;
+          if (pathVar) {
+            pathVar.split(';').forEach(element => {
```
xconverge: This is not going to work, on ubuntu the path would be /usr/bin/, I need to come up with another way

PR #4165, opened 2019-10-11
Add configuration for highlight/yank highlight text colors

### LINKED ISSUE #4164: Feature request: Highlight text color
There's no way to define a highlight text color that goes with the background color defined by `vim.searchHighlightColor`, making it hard to find a good `vim.searchHighlightColor` that works in contrast with every text color in your theme.

Screenshot 2019-10-11 at 13 48 58

Making it possible to define a `vim.searchHighlightTextColor` would eliminate any contrast issues. 

**Additional context**
see #1052

J-Fields:
Fixes #4164

### DIFF
--- a/package.json
+++ b/package.json
@@ -408,6 +408,10 @@
           "type": "string",
           "description": "Color of the search highlight."
         },
+        "vim.searchHighlightTextColor": {
+          "type": "string",
+          "description": "Color of the search highlight text."
+        },
         "vim.highlightedyank.enable": {
           "type": "boolean",
           "description": "Enable highlighting when yanking.",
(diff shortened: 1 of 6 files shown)


J-Fields:
@jpoon @xconverge Any chance you guys think we should merge `searchHighlightColor` and `searchHighlightTextColor` like `highlightedyank`? I feel like package.json could be cleaned up quite a bit but I don't want to break backward compatibility needlessly. Maybe we plan for a breaking 2.0 update, or do a slower deprecation?

jpoon:
> Maybe we plan for a breaking 2.0 update, or do a slower deprecation?

We could write some code to migrate folks to the new config?

J-Fields:
Yeah that's true. I only made additive changes in this MR - we can deal with naming inconsistencies and such later.

Issue #4126, opened 2019-10-02
Copy command broke in 1.11.0 release


tjb-wiley:
**Describe the bug**
On both my Mac and Ubuntu laptops, if I try to copy/paste code using the `co` command, like this:
```
:1,20co20
```
I get an error:
```
TaskQueue: Error running task. Failed to handle key= . nvim_buf_set_lines: Wrong type for argument 1, expecting Buffer.
Source: Vim (Extension)
```

**To Reproduce**

1.  Open any source file
2.  With "enable neovim" check in setting, try to copy/paste any lines

**Expected behavior**
Lines 1,20 should be copied after line 20 

**Screenshots**
If remapping-related, please attach log output: https://github.com/VSCodeVim/Vim#debugging-remappings.

**Environment (please complete the following information):**

- Extension (VsCodeVim) version: 1.11.0
- VSCode version: 1.38.1
- OS: MacOS 10.14.6

**Additional context**

Since this extension is using `neovim`, i upgraded it via brew to ensure that that didn't fix the issue. Even after upgrading from 0.3.4 -> 0.4.2, however, still get the same error.

![Screen Shot 2019-10-02 at 5 57 22 PM](https://user-images.githubusercontent.com/47123536/66085307-da9d8480-e53e-11e9-9376-b847cb6115a5.png)

tjb-wiley:
I just confirmed that downgrading to `1.10.2` works, so that is a short-term workaround

haifengkao:
same here. actually I get get this error by any command, e.g. `:nothiscommand`

dreadwail:
Global file search/replace via `:s` is also busted with the same error message. Can confirm downgrading to `1.10.2` solves it as well.

jpoon:
Going through the diff, we didn't change anything explicitly with neovim. The neovim npm package we rely on is at the same version and likewise no code change in the interface. May have been our migration to webpack?

jpoon:
It was indeed the webpack change.

```
git revert 1f80b2d4cc4d4d108acb18d9fca8c49b8891b696
```

Seems to have fixed neovim.

jpoon:
Ok, that didn't work. Reverting this commit and pushing a new version as that commit completely broke the extension if neovim was enabled.

Hultner:
I see a mention of this working if we downgrade to `1.10.2`, how would one go about doing this?


[...]

PR #4169, opened 2019-10-14
fix: exclude neovim from webpack module. #4126

jpoon:
Excludes neovim module from whatever magic webpack is doing.

#4126 

We need better tests, but I'm not even sure how we can even test this on Travis...


jpoon:
Oh dammit. That did not work. Need to revert this commit and the release.

jpoon:
Argh, I should have read the docs a bit better.

https://webpack.js.org/configuration/externals/

```
Instead, the created bundle relies on that dependency to be present in the consumer's environment. 
```

It was true in my case, but obviously on distribution this won't work.

PR #4171, opened 2019-10-14
upgrade to node v12

jpoon:

upgrade to node v12 used for `gulp test` and corresponding docs.


J-Fields:
node 13 is out at this point - is it worth upgrading straight to that?

PR #4185, opened 2019-10-16
Add vim.statusBarColors for 5 remaining modes

thisisrandy:
**What this PR does / why we need it**: Adds vim.statusBarColors for 5 remaining modes, i.e. CommandInProgress, etc. Note that this already works without the change, but the five modes added in this commit aren't officially listed as options. The default coloring for each appears to be the same as the standard VSCode coloring, which happens to be #007ACC

(none)

**Special notes for your reviewer**: Builds fine, but testing fails for me in `master` with
```
Error: Error: Cannot find module '/home/randy/git/VSCodeVim/test'
	at t.ExtHostExtensionService._doHandleExtensionTests (/home/randy/git/VSCodeVim/.vscode-test/vscode-1.39.2/VSCode-linux-64/resources/app/out/vs/workbench/services/extensions/node/extensionHostProcess.js:670:756)

Error: Cannot find module '/home/randy/git/VSCodeVim/test': Error: Error: Cannot find module '/home/randy/git/VSCodeVim/test'
	at t.ExtHostExtensionService._doHandleExtensionTests (/home/randy/git/VSCodeVim/.vscode-test/vscode-1.39.2/VSCode-linux-64/resources/app/out/vs/workbench/services/extensions/node/extensionHostProcess.js:670:756)
```
This is probably a configuration issue on my end (I'm unfamiliar with vscode test), so if you'd like to point me in the right direction, I can confirm that the test passes as well. To be clear, running
```
> yarn install
> yarn test
```
on a clean clone of `master` produces the above output.


ifuseandprint:
øl'

Issue #2499, opened 2018-04-01
Visual mode replace with tab adds <tab> text


kroppt:
**Is this a BUG REPORT or FEATURE REQUEST?**:
Bug Report

**What happened**:
1. Visual highlight
2. Press `r` then press `tab`
3. Every character is replaced with the text `<tab>`

**What did you expect to happen**:
Every character is replaced with a tab

**How to reproduce it (as minimally and precisely as possible)**:
Every time when highlighting more than 1 character and replacing with tab.

**Environment**:

Tested on
- Extension (VsCodeVim) version: 0.11.3
- VSCode version: 1.21.0
- OS version: Linux kernel 4.14.31-gentoo 

and

- Extension (VsCodeVim) version: 0.11.3
- VSCode version: 1.21.1
- OS version: Linux kernel 4.15.14-gentoo 

and

- Extension (VsCodeVim) version: 0.11.3
- VSCode version: 1.21.1
- OS version: Windows 10

peawyoyoyin:
hello. May I try this one?

J-Fields:
Fixed by https://github.com/VSCodeVim/Vim/pull/4195

PR #4195, opened 2019-10-19
fix replace commands from visual and visual block mode not inserting tab

peawyoyoyin:
Currently, when replacing characters using replace command from Visual and Visual block mode. Pressing tab replaces characters with `<tab>` instead of the tab character. This PR fixes the issue so that every character in the selection is replaced by the tab character (affected by `tabstop` and `expandtab` configuration).

#2499 

This is my first time contributing to an open source project. Please be kind 😄 .

### DIFF
--- a/src/textEditor.ts
+++ b/src/textEditor.ts
@@ -160,6 +160,14 @@ export class TextEditor {
     return word;
   }
 
+  static getTabCharacter(): string {
+    const { tabstop, expandtab } = configuration;
+    if (expandtab) {
+      return ' '.repeat(tabstop);
+    }
+    return '\t';
+  }
+
   static isFirstLine(position: vscode.Position): boolean {
     return position.line === 0;
   }
--- a/src/textEditor.ts
+++ b/src/textEditor.ts
@@ -2,6 +2,7 @@ import * as vscode from 'vscode';
 
 import { Position } from './common/motion/position';
 import { configuration } from './configuration/configuration';
+import { VimState } from './state/vimState';
 
 /**
  * Collection of helper functions around vscode.window.activeTextEditor
@@ -160,10 +161,11 @@ export class TextEditor {
     return word;
   }
 
-  static getTabCharacter(): string {
-    const { tabstop, expandtab } = configuration;
-    if (expandtab) {
-      return ' '.repeat(tabstop);
+  static getTabCharacter(vimState: VimState): string {
+    if (vimState.editor.options.insertSpaces) {
+      // This will always be a number when we're getting it from the options
+      const tabSize = vimState.editor.options.tabSize as number;
+      return ' '.repeat(tabSize);
     }
     return '\t';
   }
--- a/src/textEditor.ts
+++ b/src/textEditor.ts
@@ -161,10 +161,10 @@ export class TextEditor {
     return word;
   }
 
-  static getTabCharacter(vimState: VimState): string {
-    if (vimState.editor.options.insertSpaces) {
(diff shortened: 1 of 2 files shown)


src/textEditor.ts
```diff
@@ -160,6 +160,14 @@ export class TextEditor {
     return word;
   }
 
+  static getTabCharacter(): string {
```
J-Fields: I know we handle `expandtab` in other places. Are you sure you aren't duplicating this logic? Or maybe we've done it ad-hoc in a few places and this should replace those.

J-Fields:
Thanks for the PR, @peawyoyoyin!

src/textEditor.ts
peawyoyoyin: Thanks for the review. I'll check and get back shortly.

peawyoyoyin: After quick global search for `expandtab`, I found one method that I think could be refactored to use this `getTabCharacter()` method. But I don't know how, it's in `TextEditor.setIndentationLevel`.

```javascript
  static setIndentationLevel(line: string, screenCharacters: number): string {
    let tabSize = configuration.tabstop;
    let insertTabAsSpaces = configuration.expandtab;

    if (screenCharacters < 0) {
      screenCharacters = 0;
    }

    let indentString = '';

    if (insertTabAsSpaces) {
      indentString += new Array(screenCharacters + 1).join(' ');
    } else {
      if (screenCharacters / tabSize > 0) {
        indentString += new Array(Math.floor(screenCharacters / tabSize) + 1).join('\t');
      }

      indentString += new Array((screenCharacters % tabSize) + 1).join(' ');
    }

    let firstNonWhiteSpace = 0;
    let lineCheck = line.match(/^\s*/);
    if (lineCheck) {
      firstNonWhiteSpace = lineCheck[0].length;
    }

    return indentString + line.substring(firstNonWhiteSpace, line.length);
  }
```

J-Fields: Okay, no worries - thanks for checking. We can always come back to this for a refactor.

src/textEditor.ts
```diff
@@ -160,6 +160,14 @@ export class TextEditor {
...
+    const { tabstop, expandtab } = configuration;
```
J-Fields: It seems `tabstop` does not get updated on a per-document basis. So if VSCode auto-detects a file is tab-indented, or indented with 2 spaces instead of 4, this won't account for that. I wonder if there's a way to get that info.

J-Fields: Maybe `TextEditorOptions.tabSize`? ([see here](https://code.visualstudio.com/api/references/vscode-api#TextEditorOptions))

J-Fields: I changed it to use `TextEditorOptions` which seems to work. We should probably change other instances, but this is fine for now. Could you take a look and make sure this is reasonable? Thanks!

peawyoyoyin: should the information in the `configuration` object be stored per-document as well?

peawyoyoyin: ah I see. configuration is instantiated only once.

I'm not sure about making `getTabCharacter()` requires a `VimState` object, perhaps we can make the `TextEditor` class aware of the vim state by itself?

J-Fields: I'm not sure I love that, since `TextEditor` is more of a module than a class (it probably *should* be a module, actually) and there are multiple `VimState`s floating around. Maybe it's better to just pass the editor itself into the function, since that's all we need?

peawyoyoyin: Then that makes sense to me

PR #4198, opened 2019-10-20
Update CONTRIBUTING.md to recommend npx gulp

nateroling:
The CONTRIBUTING.md guide recommends installing gulp globally, but you can instead run gulp with `npx gulp`, which avoids versioning issues inherent in global Gulp installs.

No issue, just noticed while looking into contributing ☺️


J-Fields:
Seems reasonable to me. Thanks!

PR #4217, opened 2019-10-25
Revert "WebPack builds for improved loading times (#3889)"

rickpr:
This reverts commit 1f80b2d4cc4d4d108acb18d9fca8c49b8891b696.

**What this PR does / why we need it**: Neovim integration is broken due to Webpack. This reverts the Webpack commit to fix Neovim.

**Which issue(s) this PR fixes** https://github.com/VSCodeVim/Vim/issues/4126

**Special notes for your reviewer**: See comment below, I tested this locally to check that it fixes Neovim.

PR #4222, opened 2019-10-27
Remove prepare-test step from CONTRIBUTING.md

JKillian:
With the revert of the webpack build PR (#4217), we also need to revert the documentation change in CONTRIBUTING.md. 

(Just noticed this when making a PR earlier today, figured it would be good so that new contributors aren't confused!)

Reverts #4128 manually


J-Fields:
Good catch - thanks!

Issue #3775, opened 2019-05-16
Cursor is one column off after leaving multi-cursor mode


trkoch:
**Describe the bug**
When in multi-cursor mode, pressing escape twice to return to normal mode leaves cursor one column to the left.

**To Reproduce**

1. Given the text 
```
after_action
after_action
```
2. With cursor at `n`
```
after_actio>n<
after_action
``` 
3. Enable multi-cursor mode by pressing `cmd+d`
4. Leave multi-cursor mode py pressing `esc`, `esc`
5. Cursor is at
```
after_acti>o<n
after_action
``` 

**Expected behavior**
Ideally cursor should be were it was before entering multi-cursor mode.

With regards to the example given above, I expect the cursor to be at.
```
after_actio>n<
after_action
``` 

**Screenshots**
Cursor before leaving multi-cursor mode

![image](https://user-images.githubusercontent.com/73974/57849692-3c519200-77dc-11e9-855a-0c89c2303974.png)

Cursor after leaving multi-cursor mode

![image](https://user-images.githubusercontent.com/73974/57849756-6f942100-77dc-11e9-8976-116f949215cd.png)

**Environment (please complete the following information):**

- Extension (VsCodeVim) version: 1.8.0
- VSCode version: 1.33.1
- OS: macOS

trkoch:
A tracked down this line

https://github.com/VSCodeVim/Vim/blob/59261b7d606f2d2b57ae2d0a68796e50979ea141/src/actions/commands/actions.ts#L546

```diff
```

trkoch:
> However, I'm pretty sure it is there for a reason?

When exiting insert mode, both VSCode and Vim shift the cursor one column to the left (e.g. `i, <esc>`) so I'm guessing that's why it's done here as well. However, I believe this to be an error and inconsistent with visual mode. When exiting visual mode, the cursor stays at the position. Since multi-cursor mode seems to be a variation of the visual mode (as indicated in the status bar), I suggest to match its reset behavior.

stevenguh:
This fixed the bug for me, but I am not familiar with this part of code enough to if there is no regression.
```diff
diff --git a/src/actions/commands/actions.ts b/src/actions/commands/actions.ts
index e2480b2c..5a0dbc99 100644
--- a/src/actions/commands/actions.ts
+++ b/src/actions/commands/actions.ts
@@ -573,12 +573,12 @@ class CommandEsc extends BaseCommand {
     if (
[...]

PR #4237, opened 2019-11-03
Fix cursor off after leaving multi-cursor mode

trkoch:
Previously when leaving multi-cursor mode with `<Esc>` the cursor would move one column to the left. For instance, when triggering multi-cursor mode on a single word, the cursor would end at the second to last character. While this is expected behavior when leaving insert mode, it is not for visual mode (which multi-cursor mode is based on). To fix, remove special case handling altogether.

#3775 

After giving it some thoughts, I decided to remove the special case completely. I could not reproduce any other case where conditions would match other than move-left after multi-cursor mode. To verify, I added a `console.log()` in the branch in question, and ran the complete test suite. Only one test wrote a message to the console (the one I changed to expected behavior).

### DIFF
--- a/src/actions/commands/actions.ts
+++ b/src/actions/commands/actions.ts
@@ -590,18 +590,6 @@ class CommandEsc extends BaseCommand {
       return vimState;
     }
 
-    if (
-      vimState.currentMode !== Mode.Visual &&
-      vimState.currentMode !== Mode.VisualLine &&
-      vimState.currentMode !== Mode.EasyMotionMode
-    ) {
-      // Normally, you don't have to iterate over all cursors,
-      // as that is handled for you by the state machine. ESC is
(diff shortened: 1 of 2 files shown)


trkoch:
Deleted code was introduced with 2cdc5579.

trkoch:
Any objections?

stevenguh:
Maybe @jpoon knows more about this code. Also, should we have a specific unit test for this case you are trying to fix?

trkoch:
It seems this case was not covered by a test, so in order to guard against regressions, adding a test may be good idea. However I did not add new functionality but rather removed (apparently) misbehaving code.

J-Fields:
I agree - there doesn't seem to be any reason why exiting visual mode with multiple cursors would act differently than with only one.

Issue #4246, opened 2019-11-05
`gp` incorrectly places cursor when pasting more than three lines


trkoch:
**Describe the bug**
`gp` always places the cursor at the third line of pasted text when pasted text has more than three lines.

**To Reproduce**

Given

```
|1
2
3
4
```

`4yyGgp`

```
1
2
3
4
1
2
|3
4
```

**Expected behavior**

As in Vim, I expect the cursor to be placed at the end of the pasted text.

```
1
2
3
4
1
2
3
|4
```

**Environment (please complete the following information):**

- Extension version: 1.11.3
- VS Code version: Code 1.39.2 (6ab598523be7a800d7f3eb4d92d7ab9a66069390, 2019-10-15T15:33:00.827Z)
- OS version: Darwin x64 19.0.0

PR #4247, opened 2019-11-05
Fix `gp` when pasting more than three lines

trkoch:
Make `gp` (and `gP`) correctly place cursor after pasting 4+ lines.

#4246

This looks to me like a minor slip of the original author. Tests did not reveal the bug and it is not super obvious in day to day usage.

### DIFF
--- a/src/actions/commands/actions.ts
+++ b/src/actions/commands/actions.ts
@@ -1868,12 +1868,9 @@ export class GPutBeforeCommand extends BaseCommand {
     }
 
     if (vimState.effectiveRegisterMode === RegisterMode.LineWise) {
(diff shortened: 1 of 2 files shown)


trkoch:
Turns out this only works as expected in simple cases, but still has issues when there is existing text after paste (i.e. paste in the middle of a file). Mind you it is broken in both `master` and this PR. What's happening? 🤷‍♂

trkoch:
I simplified and removed the obscure (to me) `addAnotherLine` switch. Also expanded the fix to `gP`. Judging from some quick testing this seems to work as expected now. I'll continue testing in my local installation.

Issue #4266, opened 2019-11-15
Searching backwards with offset is unaware of current word under cursor


spoike:
**Describe the bug**
Apparently searching backwards using an offset doesn't seem to take in account for the current search result under the cursor.

**To Reproduce**

1.  Search for a recurring word and use offset `e`. E.g. `/word/e`
2.  The cursor should now be at the end of the next result
3.  Repeat search backwards with `N`
4.  The cursor is still on the current result

**Expected behavior**
The cursor should go to the previous result with the cursor at the end of it

**Screenshots**
N/A

**Environment (please complete the following information):**

- Extension (VsCodeVim) version: 1.11.3
- VSCode version: 1.40.0
- OS: MacOS Mojave 10.14.6

**Additional context**
Doing the same steps in vim8 will move the cursor to the end of the previous result.

J-Fields:
I'm assigning this to myself since I implemented the original feature and so I remember to actually fix this. If someone wants to give it a go before I get around to it, just give me a heads up.

bdomanski:
Hi, me and @micklett are working on an open source project for one of our classes and would like to take a look at this! We would hope to have a pull request in the next week.

J-Fields:
@bdomanski Amazing! I'll assign this issue to you - feel free to bug me if you've got questions about the code base.

J-Fields:
Fixed by #4294

PR #4294, opened 2019-11-22
Fixed backward search from end of word (VSCodeVim #4266)

bdomanski:
Fixed backward search from end of word (VSCodeVim #4266)
- Searching with 'N' after /e at the end of search fixed
- Changed comparison for next search result
- Change made in getNextSearchMatchRange in searchState.ts

Fix the issue reported,
1. Search for a recurring word and use offset e. E.g.` /word/e`
2. The cursor should now be at the end of the next result
3. Repeat search backwards with `N`
4. The cursor is still on the current result

Also fixed the larger issue, where backward search was broken in general. For example, the previous steps could be repeated with `?word?e`, where backward search with `n` also did move the cursor. 

#4266 

We tested this manually to verify it works, please let us know if we should add actual tests to the suite.

### DIFF
--- a/src/state/searchState.ts
+++ b/src/state/searchState.ts
@@ -325,7 +325,7 @@ export class SearchState {
         .slice(0)
         .reverse()
         .entries()) {
-        if (matchRange.start.compareTo(startPosition) < 0) {
+        if (matchRange.end.compareTo(startPosition) <= 0) {
           return {
             start: Position.FromVSCodePosition(matchRange.start),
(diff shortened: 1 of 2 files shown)


J-Fields:
Thanks for the PR! A regression test would be great.

bdomanski:
Alright the regression test is up, let me know if everything looks good!

J-Fields:
LGTM, thanks for the PR! I noticed while testing that if `wrapscan` is off, there's still some strangeness, but that can be fixed separately.

Issue #4296, opened 2019-11-23
zz doesn't maintain horizontal cursor position


schu34:
**Describe the bug**
in actual vim, the `zz` command maintains the cursor's horizontal position, even if it's past te end of the current line. but in vscode vim it gets reset when the cursor position is past the end of the current line

**To Reproduce**

1. initial state
```
this is a very long line of text
                               ^-cursor is here
this one is shorter
```
2. I press `j`, now
```
this is a very long line of text
                               
this one is shorter
                  ^-cursor is here
```
3. I press `zz`, nothing changes visibly in this example
4. I press `k`, now 

```
this is a very long line of text
                  ^-cursor is here                             
this one is shorter
```

**Expected behavior**
```
this is a very long line of text
                               ^-cursor should be here
this one is shorter
```

**Environment (please complete the following information):**

- Extension Name: `vim`
- Extension Version: `1.11.3`
- OS Version: `Darwin x64 18.7.0`
- VSCode version: `1.39.1`

**Additional context**
happy to work on this as long as everyone can agree its a bug :)

J-Fields:
Can confirm this is a bug. We store the column that the cursor "wants" to go to (the last one, in this case) as `vimState.desiredColumn`. Not sure where that's getting changed.

PR #4298, opened 2019-11-25
actions wont update desired column by default #4296

schu34:
This PR adds the `doesntChangeDesiredColumn` property to `BaseAction` and defaults it's value to `true`. The current behaviour was that after any non-movement action we were updating `vimState.desiredColumn` to be whatever the current position of the cursor is, which usually isn't what we want in most cases. There are a few non-movement actions that _should_ update `desiredColumn` (yanks, and puts) so I've added `doesntChangeDesiredColumn = false` to those.
 
This also fixes the bug below, which was just a symptom of the incorrect behaviour of updating `desiredColumn` on non-movement actions.

#4296
a bunch of other bugs that haven't been noticed.


J-Fields:
I hate that we have a boolean that starts with "doesnt", but it's better to keep things consistent for now - we can change that later.
This looks good to me, but I'd like to do some manual testing before merging.

schu34:
@J-Fields Any ETA on when this could be merged? Happy to make any tweaks if needed.

J-Fields:
Sorry, I've been busy recently with other things. I'll try to do some testing tonight

schu34:
No problem, totally understand. Just want to make sure it doesn't forgotten completely :)

J-Fields:
Sorry for the delay, @schu34. How confident are you you handled all the commands that should have `doesntChangeDesiredColumn = true`? Also, would you mind writing a test case for the case you described in #4296?

schu34:
@J-Fields I'm fairly confident confident that the case where `doesntChangeDesiredColumn = true` is handled correctly, I'm less confident that I've found every place where it should be `false`. I'm not certain of the functionality of real vim so I just added `doesntChangeDesiredColumn = false` everywhere that was causing tests to fail. 

I'll try and get a test written for this this week

schu34:
@J-Fields added tests, but I've been playing around with this a bit more today and I've found a few bugs related to updating `desiredColumn` when switching back to normal mode from insert mode. going to try and look into it and add a few tests before I go back to work next week.

J-Fields:
Thanks @schu34, I appreciate that. Let me know if there's anything I can do to help you out.

schu34:
@J-Fields the more I played around with this the more weird bugs I found. I opened #4424 which takes mostly the same approach, but defaults all actions to updating `desiredColumn` and just explicitly sets `zz` not to.

I kept this one open in case you or anyone else wants to take a look but feel free to close it if you prefer the other approach.

Issue #4072, opened 2019-09-13
Visual Mode + 'S' Not Working


DanDPanda:
**Describe the bug**
When I grab multiple lines of text in visual mode and click `S`, it is expected that it deletes all lines that are selected and puts me into insert mode. But what is happening instead is that it puts me into insert mode and allows me to insert one character at the start and beginning of the selection.

**To Reproduce**

1.  Open the file and focus to some text.
2. Click `v` to enter visual mode.
3. Click `jj` to select two lines down.
4. Click `S` to attempt to delete the selected lines.
5. Click `e` to see that an 'e' has been inserted into the location where you started and ended the selection.

**Expected behavior**
When I grab multiple lines of text in visual mode and click `S`, it is expected that it deletes all lines that are selected and puts me into insert mode.

**Screenshots**
![Select](https://user-images.githubusercontent.com/15399401/64870675-59229880-d609-11e9-880b-1ce99f62b35c.gif)

**Environment (please complete the following information):**

Extension (VsCodeVim) version: 1.10.2
VSCode version: 1.38.1
OS: macOS Mojave 10.14.6

**Additional context**
I'm no VIM expert, but it seems like `S` isn't the only key to have issues in visual mode. Selecting in visual mode and clicking `I` or `A` also seem to reproduce unintentional behavior.

DanDPanda:
So just a quick question, was there ever a version in which this command worked as intended? I swear I had no problem with this plugin back then, but ever since I tried it out again recently, it has been the only thing that has been holding me back from using the plugin.

J-Fields:
No, as far as I can tell, it's never been supported in visual mode. I'll get to this when I have the chance; in the meantime, you can use `c` in visual-line mode to get the same effect, though I know changing muscle is stubborn.

1ntEgr8:
@J-Fields I want to give this a try

J-Fields:
@1ntEgr8 Awesome, thanks! I'll assign the issue to you.

micklett:
Hi, myself and @bdomanski would like to take a look at this one as well as we are close to finishing up our first PR.

J-Fields:
Sweet, thanks @micklett!

anuragsati:
Hello @J-Fields i would love to work on this issue.

J-Fields:
Duplicate of #4812

This was fixed by https://github.com/VSCodeVim/Vim/commit/d43a6e1ef49da86a4e6978760299f5196d137431

[...]

PR #4299, opened 2019-11-25
Fixed Visual Mode + 'S' Not Working (VSCodeVim #4072)

micklett:
Fixed Visual Mode + 'S' Not Working (VSCodeVim #4072)

- Combined C and R commands in Visual Mode since they had the same code
- Added a command for S to delete highlighted lines in visual mode
- Added check to disable S when using sneak or surround
- Pressing C, R, or S now changes to insert mode

Fix the issue reported, 

1. Open the file and focus to some text.
2. Click `v` to enter visual mode.
3. Click `jj` to select two lines down.
4. Click `S` to attempt to delete the selected lines.
5. Click e to see that an 'e' has been inserted into the location where you started and ended the selection.

Also changed it so that `C`, `R`, and `S` all put the user in Insert Mode

#4072 

Tested locally and tried to write a test; unfortunately, since vim.surround is on by default, `S` has a different behavior (as it should when vim.surround is on), and so testing will not pass.

Also, VSCode automatically changes the arrays to multiple lines. If this is a problem, I can try to amend my last commit with the arrays on the same line.


J-Fields:
Thanks for the PR!

> Also, VSCode automatically changes the arrays to multiple lines. If this is a problem, I can try to amend my last commit with the arrays on the same line.

Don't worry about it, we use prettier to keep everything consistent and it likes to do that.

> Tested locally and tried to write a test; unfortunately, since vim.surround is on by default, S has a different behavior (as it should when vim.surround is on), and so testing will not pass.

You can modify the config for a test [like this](https://github.com/VSCodeVim/Vim/blob/master/test/mode/modeNormal.test.ts#L13).

micklett:
OK, I've added a test and updated a few others. Sorry about the strange commits; trying to get my local branch up to date with the recent changes.

J-Fields:
Awesome, thanks! Sorry for the mayhem caused by the refactor hahaha

test/mode/modeVisual.test.ts
```diff
@@ -1222,7 +1237,7 @@ suite('Mode Visual', () => {
         'ggvep'.split('') // replace 'replace'
       );
       await modeHandler.handleMultipleKeyEvents(['g', 'v']);
-
+
```
J-Fields: Run `gulp forceprettier`, please

J-Fields:
Any reason this doesn't also extend support to visual line and visual block modes?

J-Fields:
Hey @micklett, do you plan on coming back to this PR?

PR #4304, opened 2019-11-28
Close sidebar and bottom panel with :only command

kizza:

- Currently `:only` in this plugin simply reduces the number of editors (splits) to one
- In vim, `:only` closes all open views except the one with focus (eg. it closes NERDTree, fzf and all other native/plugin windows)
- The sidebar and bottom panels already have builtin support for toggling their visibility (just like vim). 
- I believe better alignment to the native vim experience is to close everything except the single pane with focus

**Current behaviour**
Two splits, navigation panel and bottom panel => become identical but with a single split

| Before | After :only |
| ------ | ----- |
| before | after |

**Proposed behaviour**
Two splits, navigation panel and bottom panel => become a single editor view

| Before | After :only |
| ------ | ----- |
| before | after |

I couldn't find any via a search on "only"

- The functionality is achieved easily via builtin vscode functionality
- I could not find a "native" method to assert against the visibility of the sidebar panel
- I didn't want to _presume_ the state left from potential previous tests changing this
- I ended up calculating its visibility by diffing the number of visible lines due to the presence/absence of the panel


J-Fields:
Thanks!

PR #4312, opened 2019-11-30
Fix laggy scrolling with <C-e> and <C-y> (WIP)

### LINKED ISSUE #4309: Scrolling with <C-e> and <C-y> is very laggy
**Describe the bug**
When holding down (or quickly tapping) &lt;C-e&gt; or &lt;C-y&gt;, scrolling soon becomes very laggy and the editor only scrolls at a rate of about two lines per second. The keystrokes get queued up and after letting go of the keys the editor continues to scroll slowly and try to catch up.

&lt;C-u&gt; and &lt;C-d&gt; don't suffer from this problem. I noticed in the source that they don't use `vimState.postponedCodeViewChanges.push()` and instead call `vscode.commands.executeCommand()` directly. Maybe this has something to do with it?

Similar to #1943 and #1985

**To Reproduce**

1. Using the latest stable (or insider) version of VS Code on Mac, run `/Applications/Visual Studio Code.app/Contents/Resources/app/bin/code --user-data-dir /tmp/something`.
2. Ensure only the latest VS Code Vim extension is installed.
3. Open a file and try &lt;C-e&gt; or &lt;C-y&gt;

**Expected behavior**
Performant scrolling

**Screenshots**
Screen Shot 2019-11-30 at 3 35 04 PM copy

- Extension (VsCodeVim) version: 1.11.3
- VSCode version: 1.40.2 (or 1.41.0-insider)
- OS: macOS Catalina 10.15.1 (19B88)

raiskila:
To fix #4309

This change fixes the performance issue, but causes an issue where &lt;C-y&gt; stops working when the cursor is at the bottom of the screen. This may not be the right fix.

Contributions to this PR are appreciated from people more familiar with the codebase.

### DIFF
--- a/src/actions/commands/actions.ts
+++ b/src/actions/commands/actions.ts
@@ -690,15 +690,12 @@ abstract class CommandEditorScroll extends BaseCommand {
       vimState.cursors[0] = vimState.cursors[0].getDown(scrolloff - linesBelowCursor);
     }
 
-    vimState.postponedCodeViewChanges.push({
-      command: 'editorScroll',
-      args: {
-        to: this.to,
-        by: this.by,
-        value: timesToRepeat,
-        revealCursor: true,
-        select: [Mode.Visual, Mode.VisualBlock, Mode.VisualLine].includes(vimState.currentMode),
-      },
+    await vscode.commands.executeCommand('editorScroll', {
+      to: this.to,
+      by: this.by,
+      value: timesToRepeat,
+      revealCursor: true,
+      select: [Mode.Visual, Mode.VisualBlock, Mode.VisualLine].includes(vimState.currentMode),
     });
     return vimState;
   }


J-Fields:
So we *generally* want to keep actions as pure as possible - ideally they'd just change `vimState` and `ModeHandler` would handle actually modifying the document. If there's a good reason to make an exception (like performance), then that's fine.

Removing `getCursorsAfterSync()` (see #4432) fixes this lag without needing to make an exception. I just need to make sure that this doesn't in fact screw anything up...

J-Fields:
@raiskila I'm closing this PR because I merged #4433. Would you mind smoke testing the latest `master`? I'd like confirmation that that change didn't break the extension and also that it fixed the lag of `<C-e>`/`<C-y>`

raiskila:
Thanks! I'm no longer experiencing the problem on the master branch.

PR #4337, opened 2019-12-08
Fix fs.readFileSync() receiving a possibly undefined path

### LINKED ISSUE #4336: Extension fails to activate when vimrc path is undefined
**Describe the bug**

Extension fails to activate when vimrc is enabled and vimrc.path is not defined in the config file, which are the default settings.

The error given is:
```
Activating extension `vscodevim.vim` failed:  ENOENT: no such file or directory, open
```

**To Reproduce**
Set vim.vimrc.enable = true and leave vim.vimrc.path unset.

**Expected behavior**

$HOME/.vimrc or $HOME/_vimrc should have been checked and extension should activate.

- Issue Type: `Bug`
- Extension Name: `vim`
- Extension Version: `1.12.0`
- OS Version: `Linux x64 4.15.0-72-generic`
- VSCode version: `1.40.2`

**Additional context**

The issue comes from fs.readFileSync() receiving vimrc.path when it is undefined. 
$HOME/.vimrc or $HOME/_vimrc. are correctly being set as the default file path when vimrc.path is undefined, however the variable that stores the correct file path is not the one being passed to readFileSync.

CVVPK:

When config.vimrc.path is undefined, fs.readFileSync() will throw an error which results in the extension failing to activate. This PR fixes that issue by changing the variable passed to fs.readFileSync() to one that is guaranteed to not be undefined.

Fixes #4336

The correct default path ($HOME/.vimrc or $HOME/_vimrc) is already being assigned to the local _path variable which is properly being checked before using it to set the vimrcPath property. Thus, it is safe to assume vimrcPath will never be an invalid path and is safe to be passed to fs.readFileSync().

### DIFF
--- a/src/cmd_line/commands/substitute.ts
+++ b/src/cmd_line/commands/substitute.ts
@@ -124,12 +124,12 @@ export class SubstituteCommand extends node.CommandBase {
     if (args.pattern === undefined) {
       // If no pattern is entered, use previous SUBSTITUTION state and don't update search state
       // i.e. :s
-      const prevSubstiteState = globalState.substituteState;
-      if (prevSubstiteState === undefined || prevSubstiteState.searchPattern === '') {
+      const prevSubstituteState = globalState.substituteState;
+      if (prevSubstituteState === undefined || prevSubstituteState.searchPattern === '') {
         throw VimError.fromCode(ErrorCode.E35);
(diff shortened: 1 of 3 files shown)


J-Fields:
Cool, thanks @CVVPK!

PR #4360, opened 2019-12-10
Make sure DocumentChange.timestamp is always set

### LINKED ISSUE #4351: Undo After Replace Causes Error: "Cannot read property 'getTime' of undefined."
**Describe the bug**
When successfully performing a replace command via :s/blah/bleh, immediately undoing this replacement will cause the error 'TaskQueue: Error running task. Failed to handle key=u. Cannot read property 'getTime' of undefined.'

**To Reproduce**

1.  Type 'blah' in a document
2.  Type :s/blah/bleh to replace 'blah' with 'bleh'
3.  Hit 'u' (default key binding) to undo this replacement

The undo will work as expected, but the stated error will appear. Furthermore, ':' or '/' will have to then be hit twice to bring the command prompt back up. This seems to happen with or without neovim integration. Error does not occur if the replacement is undone after entering insert mode and then going back to normal mode.

**Expected behavior**
The error doesn't occur.

* VSCode version: `1.40.2`
* VSCodeVim version: `1.12.2`
* OS: `Windows_NT x64 10.0.18362.476 (Win 10 Pro)`

J-Fields:
This seems like it could be incorrect sometimes, and we should make sure all changes are either declared finished at their creation or finished before they can be undone.
Also remove an instance of suite.only() that slipped in...whoops
Fixes #4351

### DIFF
--- a/src/cmd_line/commandLine.ts
+++ b/src/cmd_line/commandLine.ts
@@ -98,6 +98,9 @@ class CommandLine {
     }
   }
 
+  /**
+   * Prompts the user for a command using an InputBox, and runs the provided command
+   */
   public async PromptAndRun(initialText: string, vimState: VimState): Promise<void> {
     if (!vscode.window.activeTextEditor) {
       this._logger.debug('No active document');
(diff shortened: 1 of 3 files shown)


raudabaugh:
When is the next release expected? Would love to start using this fix 👍

J-Fields:
@raudabaugh Sorry, was meaning to get a few other small things into the patch release but life got busy. I'll do my best to release tonight.

changhz:
this annoying error has been torturing me 😭😭

PR #4380, opened 2019-12-14
Improved performance of addChange()

### LINKED ISSUE #3920: Investigate optimizing HistoryTracker.addChange()
I stumbled across a comment @johnfn made in this function which I think deserves some testing and some thought; this ticket exists to remind me to do both. On large files, I suspect this is _very_ inefficient and may contribute to the slowdown we experience. First step is to do some profiling.

CVVPK:
Currently, HistoryTracker.addChange() has a major impact on performance, specially when working on large files. The issue stems largely due to processing the entire document on every call, and that the function gets called on every key press. My proposed changes are as follows:

**Check if offset at EOF is equals to oldText.length**

Currently the first thing addChange() does is grab the text of the entire document, which is a pretty expensive operation. We then check the newText against the oldText to see if there are any changes that need to be processed. There really is no need to get the entire document to know this.

The zero based offset at the document's end represents the location of the character that would follow the last character, thus it is equal to the length of the document's text. We can easily get this offset by using vscode.TextDocument.offsetAt() and passing it Position.getDocumentEnd(). If there are no changes this offset will be equals to historyTracker.oldText.length. 

This would increase the performance when navigating through large files.

**Improvements while on insert mode**

The major issue is that addChange() gets called on every key press, thus the entire document is getting processed every few milliseconds if we type too fast. Which is why we experience more lag the faster we make changes on large files. Currently the first change after going into insert mode triggers _addnewHistoryStep(), consequent changes are added to the historyStep and then when we get out of insert mode we finalise that historyStep.

We shouldn't really need to log each substep while we are on Insert Mode since we only go back one historyStep at a time. We can grab a snapshot of when we first start making changes and one at the end. I had initially thought we could just grab the last change, but we do need the first step to be logged so that we don't brake historyTracker.goBackHistoryStepsOnLine().

We can add a condition that checks if the currentMode === Mode.Insert to get out of addChange() and prevent processing the document while changes are still in progress. The condition shouldn't trigger if we just created a new history step, so that we may log the first step. When we get out of Insert Mode, the call from to addChange() within ModeHandler.runAction() will process the final step.

This change, while it has a major positive impact on performance, does change the workflow of the code a little bit, since calls to addChange() while we are on insert mode will not do anything. With the exception of the first change that creates the historyStep, the call made to addChange() from within ModeHandler.handleKeyEventHelper() will be go largely ignored. I don't believe this has any negative side effects, but is something we might want to be aware of.

Possibly fixes #3920, #2021, #2216

### DIFF
--- a/src/history/historyTracker.ts
+++ b/src/history/historyTracker.ts
@@ -18,6 +18,7 @@ import { Logger } from './../util/logger';
 import { VimState } from './../state/vimState';
 import { TextEditor } from './../textEditor';
 import { StatusBar } from '../statusBar';
+import { Mode } from '../mode/mode';
 
 const diffEngine = new DiffMatchPatch.diff_match_patch();
 diffEngine.Diff_Timeout = 1; // 1 second
@@ -29,7 +30,7 @@ class DocumentChange {
    * true => addition
    * false => deletion
    */
-   // TODO: support replacement, which would cut the number of changes for :s/foo/bar in half
+  // TODO: support replacement, which would cut the number of changes for :s/foo/bar in half
   public isAdd: boolean;
 
   private _end: Position;
@@ -427,14 +428,16 @@ export class HistoryTracker {
    * used to look like.
    */
   public addChange(cursorPosition = [new Position(0, 0)]): void {
-    const newText = this._getDocumentText();
+    // Get the current document's text which is conveniently equals to the zero based offset at the document's end.
+    const documentEnd = cursorPosition[0].getDocumentEnd();
+    const newTextLength = this.vimState.editor.document.offsetAt(documentEnd);
 
-    if (newText === this.oldText) {
+    // Check if the document's text changed.
+    if (newTextLength === this.oldText.length) {
       return;
     }
 
-    // Determine if we should add a new Step.
-
+    // Determine if we should add a new Step. Additionally, If we are on Insert mode and this isn't the first change we made then we should get out of here.
     if (
       this.currentHistoryStepIndex === this.historySteps.length - 1 &&
       this.currentHistoryStep.isFinished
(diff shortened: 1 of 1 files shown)


src/history/historyTracker.ts
```diff
@@ -427,14 +428,16 @@ export class HistoryTracker {
    * used to look like.
    */
   public addChange(cursorPosition = [new Position(0, 0)]): void {
-    const newText = this._getDocumentText();
+    // Get the current document's text which is conveniently equals to the zero based offset at the document's end.
+    const documentEnd = cursorPosition[0].getDocumentEnd();
+    const newTextLength = this.vimState.editor.document.offsetAt(documentEnd);
```
J-Fields: Perhaps we can just enumerate the special cases and deal with them separately, but doesn't this break with `:s/abc/def`, for instance?

src/history/historyTracker.ts
```diff
@@ -427,14 +428,16 @@ export class HistoryTracker {
...
 
-    if (newText === this.oldText) {
+    // Check if the document's text changed.
+    if (newTextLength === this.oldText.length) {
```
J-Fields: Perhaps we can just enumerate the special cases and deal with them separately, but doesn't this break with :s/abc/def, for instance?

CVVPK: I believe you're referring to the case of doing `iabc<Esc>:s/abc/def\nu` which results in `|` when it should be `abc`. It does break that, sorry I hadn't considered that case and it seems we aren't currently testing for it. Can I add a test for it?

J-Fields: Yeah a test for this would be great. If we go in this direction of optimization, I'd like to make sure we cover all the edge cases

CVVPK:
I'm closing this because I messed up my master....

PR #4386, opened 2019-12-15
Implement  _isDocumentTextNeeded()

### LINKED ISSUE #3920: Investigate optimizing HistoryTracker.addChange()

cvaldev:
For reference, see #4380 and #3920 .

As mentioned in #4380 `HistoryTracker.addChange()` performance gets worse the larger the file and the faster we make individual changes, because on each change we would get the entire document's text. We don't use the individual changes, so we really only care about the initial step and the final one.

A new `historyStep` is created when we first enter a new mode and finalized after we exit it. We can keep track of the `currentMode` internally, and when it is different to `vimState.currentMode` we know we have switched. This way we prevent making unnecessary calls to  `TextDocument.getText()` while changes are still in progress.

Fixes #3920 

These changes are passing all current tests and I proposed #4385 to cover the `:s/abc/def` edge case. Since `HistoryTracker.addChange()` remains completely unchanged besides the condition added at the top, we could easily make this an experimental opt-in setting.

**Edit:**
I have added the option to turn this on or off in the config. Leaving it off by default is probably best for now since there could be edge cases that haven't been covered.

If it's off by default, how can we make sure the tests hit it?

**Edit2:**
I don't think this fixes 2021 and 2216 because those issues are tracking way more things than what this covers and wouldn't make sense for this to close them. Also removing the reference I had made to checking the text length because upon review that's not really a viable option.


johnfn:
Wow! If this fixes the lag issue, that would be pretty amazing!

CVVPK:
From my own testing the lag is completely gone. Here's a profile after my last commit:

[testRun.cpuprofile.txt](https://github.com/VSCodeVim/Vim/files/3968345/testRun.cpuprofile.txt)

package.json
```diff
@@ -905,6 +905,11 @@
           "type": "boolean",
           "markdownDescription": "When `true` the commands listed below move the cursor to the first non-blank of the line.  When `false` the cursor is kept in the same column (if possible).  This applies to the commands: `<C-d>`, `<C-u>`, `<C-b>`, `<C-f>`, `G`, `H`, `M`, `L`, `gg`, and to the commands `d`, `<<` and `>>` with a linewise operator.",
           "default": true
+        },
+        "vim.experimentalOptimizations": {
```
J-Fields: Cool, I like the defensiveness here.

J-Fields:
> If it's off by default, how can we make sure the tests hit it?

I think as long as it's behind an experimental setting, it's fine if CI doesn't run tests against it. When it's enabled, the test suite passes, right?

CVVPK:
@J-Fields Yeah it is passing right now.

J-Fields:
Okay, so I'm digging into this a bit more now - sorry for not getting around to it sooner

src/history/historyTracker.ts
```diff
@@ -420,13 +424,45 @@ export class HistoryTracker {
...
+      (this.currentHistoryStepIndex === this.historySteps.length - 1 &&
+        this.currentHistoryStep.isFinished) ||
+      this.currentHistoryStepIndex !== this.historySteps.length - 1;
+
+    if (isModeDiff) {
+      this.currentMode = this.vimState.currentMode;
+    }
+
+    // If all of these are false we can avoid requesting the entire doc.
+    return (isTextLengthDiff && (isNewHistoryStep || isModeDiff)) || isNewHistoryStep || isModeDiff;
```
J-Fields: This seems wrong or at least overcomplicated. Isn't this equivalent to just `isNewHistoryStep || isModeDiff`?

src/history/historyTracker.ts
```diff
@@ -420,13 +424,45 @@ export class HistoryTracker {
...
+    const documentEnd = new Position(0, 0).getDocumentEnd();
+    const newTextLength = this.vimState.editor.document.offsetAt(documentEnd);
+
+    const isTextLengthDiff = this.oldText.length !== newTextLength;
+
+    // Determine if we just switched modes.
+    // This prevents recording steps in between start-end of a historyStep.
+    const isModeDiff = this.currentMode !== this.vimState.currentMode;
+
+    const isNewHistoryStep =
```
J-Fields: I think this logic might need to be a bit more complicated. I think a normal motion like `j` or `<C-d>` without any operator (`cj` for instance) should ideally never trigger a document diff, but currently it does. I'll have to do a bit more thinking about this, but maybe we keep track of the number of history steps and see if that's changed?

J-Fields: Or maybe we just don't call `addChange` in the first place if it's a movement without an operator?

J-Fields: Oh, we need to keep in mind that our definition of "movement" basically means able to be combined with an operator... maybe we just label certain commands as ones that never change document text.

J-Fields: I'm going to play around with that idea and put up another PR. I think that might be an easy win that could work alongside this optimization.

src/history/historyTracker.ts
CVVPK: Yes, you're right!. Thanks for catching it, it completely went over my head. Originally I wanted to just check the text length to avoid getting the whole doc when just moving around, but that doesn't cover cases where an actual change doesn't affect the length of the text. In trying to solve that, I think I got my own logic mixed up, sorry about that.

CVVPK: Checking the text length is  not a viable solution, I'm going to remove that.

src/history/historyTracker.ts
CVVPK: > maybe we just label certain commands as ones that never change document text
I like this idea, we definitely need a different solution to optimizing navigation.

CVVPK:
> Okay, so I'm digging into this a bit more now - sorry for not getting around to it sooner.

@J-Fields It's all good, thank you for taking the time to review it. I've changed it so we don't use the text length, as pointed out that wasn't even doing anything. My initial assessment that this would improve the performance of navigating through files is incorrect. This will, however, still improve performance when editing files because we are only requesting the whole doc on the first and last change.

CVVPK:
Not passing after #4451 .

J-Fields:
> Not passing after #4451

Hmm, what's failing?

CVVPK:
```
 1) Mode Insert
       <C-c> can exit insert:
     Error: Failed to handle key=<copy>. Action '<copy>' changed the document unexpectedly!
      at ModeHandler.<anonymous> (src/mode/modeHandler.ts:310:15)
      at Generator.throw (<anonymous>)
      at rejected (out/src/mode/modeHandler.js:6:65)


  2) Mode Insert
       Can perform <ctrl+o> to exit and perform one command in normal:
     Failed to handle key=<C-o>. Action '<C-o>' changed the document unexpectedly!
      at Suite.<anonymous> (test/mode/modeInsert.test.ts:218:3)
```
I left a comment on that PR as to why I believe this happens. Basically I think it is because `addChange()` will return true after the command `<C-c>` or `<C-o>` switched from Normal to Insert mode, since the `addChange()` called from `ModeHandler.handleKeyEventHelper()` doesn't log the changes before switching modes like it normally would have.

J-Fields:
Oh, I missed your comment on the other PR. Thanks, I'll take a closer look

CVVPK:
No idea why 08ceb00 didn't pass, it's passing on my machine and f92a881 was fine as well.

J-Fields:
Probably just a flake, I wouldn't worry about it for now

cvaldev:
Looks like this got a bit messed up when I changed usernames and we had lost: 8ac6afa, 4d68a43 and f92a881.

J-Fields:
Whoops, that probably happened when I rebased, sorry about that!

J-Fields:
Thanks again @cvaldev!

PR #4389, opened 2019-12-16
Go the dest line directly when execute move up/down with count

CharAct3:
Executing move up/down with count is not efficient. `100j` runs `moveDown` 100 times instead of going to the dest line directly.

First time contributing to this repo, please let me know if there are any problems.


J-Fields:
Doesn't this break `10j` when `foldfix` is enabled? I think we want to check `foldfix` and fall back on the right commands in that case.

Issue #4403, opened 2019-12-18
Unable to remap keys via noremap in vimrc


jjoekoullas:
**Describe the bug**
Very simple vimrc doesn't appear to be getting applied.

.vimrc contents:
```
noremap <Up> <NOP>
noremap <Down> <NOP>
noremap <Left> <NOP>
noremap <Right> <NOP>
```

**To Reproduce**
On windows.
1.  Create file ~\.vimrc with the above contents
2.  Check vscode vim settings - confirm vimrc is enabled, confirm default vimrc path (no vimrc related settings in settings json)
3.  Restart vscode and open a file
4.  Use the arrow keys to navigate around in a file.

**Expected behavior**
Arrow keys are disabled when editing a file.

**Environment (please complete the following information):**
- Extension (VsCodeVim) version: 1.12.1
- VSCode version: 1.41.1 (user setup)
- OS: Windows 10

**Additional context**
I'm a vim newbie - but this vimrc works when I open vim.

J-Fields:
Yeah, we don't support `<NOP>` currently. Should be a pretty straightforward addition

jjoekoullas:
If it's straight forward, I would really like to take a crack at submitting a PR for it.  Any chance you can point me to similar bugs/PRs that I can use as reference?

jjoekoullas:
I'd guess https://github.com/VSCodeVim/Vim/blob/master/test/configuration/vimrcKeyRemappingBuilder.test.ts is a pretty good place to start.

J-Fields:
Yeah a test there would be great. I think we also want a command (`actions.ts`) for `<NOP>` that does nothing. I think a change in `notation.ts` (and possibly elsewhere, I'm not sure off the top of my head) so we parse `<NOP>` as a single keystroke.

jjoekoullas:
@J-Fields thanks for the pointers.

`<NOP>` to work out of the box and the issue looks like it was due to `noremap` not being supported.  See: https://github.com/VSCodeVim/Vim/blob/master/src/configuration/vimrc.ts#L46

PR #4409, opened 2019-12-21
support vimrc map & noremap  

jjoekoullas:
Adds support for `map` and `renoremap` in .vimrc 

#4403 

Had a test failure however it's currently failing on master.

```
  1) Remapper
       leader, c -> closeActiveEditor in visual mode through modehandler:

      AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:

1 !== 0
```

### DIFF
--- a/src/configuration/vimrc.ts
+++ b/src/configuration/vimrc.ts
@@ -42,33 +42,46 @@ class VimrcImpl {
    * Adds a remapping from .vimrc to the given configuration
    */
   private static addRemapToConfig(config: IConfiguration, remap: IVimrcKeyRemapping): void {
-    const remaps = (() => {
+    const mappings = (() => {
       switch (remap.keyRemappingType) {
+        case 'map':
+          return [
+            config.normalModeKeyBindings,
+            config.visualModeKeyBindings,
+          ];
         case 'nmap':
-          return config.normalModeKeyBindings;
+          return [config.normalModeKeyBindings];
         case 'vmap':
-          return config.visualModeKeyBindings;
+          return [config.visualModeKeyBindings];
         case 'imap':
-          return config.insertModeKeyBindings;
+          return [config.insertModeKeyBindings];
         case 'cmap':
-          return config.commandLineModeKeyBindings;
+          return [config.commandLineModeKeyBindings];
+        case 'noremap':
+          return [
+            config.normalModeKeyBindingsNonRecursive,
+            config.visualModeKeyBindingsNonRecursive,
+          ];
         case 'nnoremap':
-          return config.normalModeKeyBindingsNonRecursive;
+          return [config.normalModeKeyBindingsNonRecursive];
         case 'vnoremap':
-          return config.visualModeKeyBindingsNonRecursive;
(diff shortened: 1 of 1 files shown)


J-Fields:
>Had a test failure however it's currently failing on master.
Hmm, what platform are you on?

src/configuration/vimrc.ts
```diff
@@ -42,33 +42,54 @@ class VimrcImpl {
...
         default:
+          console.warn('Encountered an unrecognized mapping type.');
           return undefined;
       }
     })();
 
-    // Don't override a mapping present in settings.json; those are more specific to VSCodeVim.
-    if (remaps && !remaps.some(r => _.isEqual(r.before, remap!.keyRemapping.before))) {
-      remaps.push(remap.keyRemapping);
+    if (!mappings) {
```
J-Fields: This shouldn't be necessary - the `forEach` will be a no-op if `mappings` is empty

src/configuration/vimrc.ts
```diff
@@ -42,33 +42,54 @@ class VimrcImpl {
...
    */
   private static addRemapToConfig(config: IConfiguration, remap: IVimrcKeyRemapping): void {
-    const remaps = (() => {
+    const mappings = (() => {
       switch (remap.keyRemappingType) {
+        case 'map':
+          return [
+            config.normalModeKeyBindings,
+            config.visualModeKeyBindings,
+            config.insertModeKeyBindings,
```
J-Fields: According to [this](https://vim.fandom.com/wiki/Mapping_keys_in_Vim_-_Tutorial_(Part_1)) and a bit of testing, seems `map` doesn't create a mapping for insert/command line modes.

jjoekoullas:
> > Had a test failure however it's currently failing on master.
> 
> Hmm, what platform are you on?

Ubuntu.  I can write up a bug report if you'd like.

src/configuration/vimrc.ts
jjoekoullas: `mappings` could be undefined (not empty) from switch statement.  Also tsc complains that mappings might be null or undefined.  So I think this is necessary to keep the error of an unrecognized mapping type as non-fatal (as it was before).

src/configuration/vimrc.ts
jjoekoullas: Ah, thanks.  Good catch,

Fixed.

src/configuration/vimrc.ts
J-Fields: Oh I see. Could you remove this check and do `mappings?.forEach` instead?

src/configuration/vimrc.ts
J-Fields: Hey @jjoekoullas did you push your changes? I'm still seeing `config.insertModeKeyBindings`

src/configuration/vimrc.ts
jjoekoullas: Done.

src/configuration/vimrc.ts
jjoekoullas: Ah, actually I just missed it.  Fixed now.

PR #4424, opened 2019-12-27
Fix zz updating desired column

schu34:
basically the same as  #4298. That PR was causing a bunch of difficult to debug issues since it updated the default behaviour for all actions. This PR takes mostly the same approach, except it defaults `doesntChangeDesiredColumn` to `false` and just sets it to `true` for `zz`. There are probably other actions that update `desiredColumn` when they shouldn't, and this PR won't address those, but at least now there's an easy route to address those bugs as they're noticed. (just add `doesntChangeDesiredColumn = true`)

I also added a bunch of unit tests that I wrote while trying to debug the last PR. 

#4296

### DIFF
--- a/src/actions/base.ts
+++ b/src/actions/base.ts
@@ -17,8 +17,12 @@ export class BaseAction {
 
   public canBeRepeatedWithDot = false;
 
-  public doesntUpdateDesiredColumn = false;
-
+  /**
+   * Whether we should change desiredColumn in VimState.
+   */
+  public doesntChangeDesiredColumn(): boolean {
+    return false;
+  }
   /**
    * Modes that this action can be run in.
    */
--- a/src/actions/base.ts
+++ b/src/actions/base.ts
@@ -17,6 +17,8 @@ export class BaseAction {
 
   public canBeRepeatedWithDot = false;
 
+  public doesntUpdateDesiredColumn = false;
+
   /**
    * Modes that this action can be run in.
    */
@@ -95,7 +97,7 @@ export class BaseAction {
     const isSingleNumber: RegExp = /^[0-9]$/;
     const isSingleAlpha: RegExp = /^[a-zA-Z]$/;
 
-    for (let i = 0, j = 0; i < one.length; i++, j++) {
+    for (let i = 0, j = 0; i < one.length; i++ , j++) {
       const left = one[i],
         right = two[j];
 
--- a/src/actions/base.ts
+++ b/src/actions/base.ts
@@ -97,7 +97,7 @@ export class BaseAction {
     const isSingleNumber: RegExp = /^[0-9]$/;
     const isSingleAlpha: RegExp = /^[a-zA-Z]$/;
 
-    for (let i = 0, j = 0; i < one.length; i++ , j++) {
+    for (let i = 0, j = 0; i < one.length; i++, j++) {
       const left = one[i],
         right = two[j];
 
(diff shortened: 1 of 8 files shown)


test/mode/modeVisualLine.test.ts
```diff
@@ -470,5 +470,11 @@ suite('Mode Visual Line', () => {
       keysPressed: 'VjjA_',
       end: ['111', '222_|', ' ', '444_', '555'],
     });
+    newTest({
```
J-Fields: nit: newline above this line

src/actions/base.ts
```diff
@@ -95,7 +97,7 @@ export class BaseAction {
     const isSingleNumber: RegExp = /^[0-9]$/;
     const isSingleAlpha: RegExp = /^[a-zA-Z]$/;
 
-    for (let i = 0, j = 0; i < one.length; i++, j++) {
+    for (let i = 0, j = 0; i < one.length; i++ , j++) {
```
J-Fields: I think prettier will handle this - can you run `gulp forceprettier` please?

J-Fields:
Thanks @schu34, going with this less disruptive approach is probably better, even if there's some subtle existing misbehavior we haven't caught.

src/actions/base.ts
schu34: Fixed!

J-Fields:
What about other commands which move the view, but (generally) not the cursor? For instance, `zt`/`zb` or `<C-e>`/<C-y>`?

schu34:
Yeah sure, wasn't aware of all those but easy enough to add. Is there a list of commands that move the view but not the cursor that I could reference?

J-Fields:
Not that I know of. Maybe `H`/`L`/`M` too? I can't right now but before I merge I'll double check that these all do behave this way in vim and we didn't miss anything too obvious.

schu34:
did some testing in neovim. It seems that all the `z` commands as well as `<C-e>` don't update the desiredColumn but `H`/`L`/`M`/`<C-y>` take you to the first non-blank character of the line so they do.

J-Fields:
I think that behavior is dependent on `startofline` being set. Can you double check that the desired column isn't updated when that's unset?

schu34:
Yeah, you're right about `startofline`, in actual vim when it's not set the desired column isn't updated. but it seems like the plugin currently doesn't respect `startofline` for `H`/`L`/`M` anyway unless I'm missing something. In any case, should I update `doesntChangeDesiredColumn` to be a function so it can change values based on config options?

J-Fields:
Thanks for pointing that out, I just created #4428.

Yeah, if you could make `doesntChangeDesiredColumn` a function in preparation for that change, that'd be awesome

schu34:
Alright updated that and also removed it from `BaseMovement` since that inherits from `BaseAction`

J-Fields:
I think you missed `<C-y>`. Other than that, looks good to me. Thanks a lot for the rapid iteration on this!

test/mode/normalModeTests/motions.test.ts
```diff
@@ -804,7 +804,7 @@ suite('Motions in Normal Mode', () => {
     end: ['blah', 'duh', 'a', 'hu|r '],
   });
 
-  suite("doesn't update desiredColumn when it shouldn't", () => {
+  suite.only("doesn't update desiredColumn when it shouldn't", () => {
```
J-Fields: I've done this more than once 😄

J-Fields:
Thanks again! I'll merge once the tests are confirmed to be green

PR #4433, opened 2019-12-29
Stop using timeout in `getCursorsAfterSync()`

### LINKED ISSUE #4432: Is `getCursorsAfterSync()` necessary any more?
It seems like `getCursorsAfterSync` warns that the cursors didn't update *every* time it's called, yet everything still works. I wonder if at some point VSCode changed so that it doesn't give control back until the cursors are updated.

J-Fields:
The issue this function was meant to solve seems to have disappeared.
Fixes #4432


J-Fields:
Works on my machine? No test failures? Sounds like it's gonna cause issues 😎

PR #4435, opened 2019-12-29
Update extensionKind

### LINKED ISSUE #4379: `extensionKind` as a string is now deprecated
See https://code.visualstudio.com/api/advanced-topics/remote-extensions#incorrect-execution-location

CVVPK:
Fixes #4379 

> The old "ui" value (as a string) maps to this type for backwards compatibility, but is considered deprecated.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,10 @@
     "Other",
     "Keymaps"
   ],
-  "extensionKind": "ui",
+  "extensionKind": [
+    "ui",
+    "workspace"
+  ],
   "activationEvents": [
     "*",
     "onCommand:type"


J-Fields:
I *think* `["ui", "workspace"]` is correct, but not 100% sure. Do you have a good explanation for this, or did you choose this only because it's equivalent to what we had before?

CVVPK:
@J-Fields Sorry, I should have cited my quote. I chose it because according to [this](https://code.visualstudio.com/api/advanced-topics/remote-extensions#incorrect-execution-location) it is the equivalent of what we had before.

J-Fields:
Cool, just checking. I think this is right, and if it causes issues for anyone we can always change it. Thanks!

PR #4451, opened 2020-01-01
Improve performance on large files by not checking for changes after commands which never change the document

J-Fields:
Many commands, like movement and typing a search, will never change the document. We waste considerable time diffing the entire document on each of these keystrokes to figure out if we should add an undo step.
I've added a property to BaseAction, `mightChangeDocument`, which determines whether we should check for changes.
This defaults to false for the sake of brevity (most commands actually don't change the document), but this is a bit dangerous - what if I forgot to override it in a command that changes the document?
The planned solution is to release this as a check - we will always run the diff, as we currently do, then make sure a command didn't add a change when we asserted it wouldn't. After this is in the wild for a little while with no reported issues, we'll actually flip the switch to optimize.
In my testing on a ~80,000 line file, this saves maybe 10-15ms per keystroke on average, which does translate into a noticeable improvement in responsiveness.

### DIFF
--- a/src/mode/modeHandler.ts
+++ b/src/mode/modeHandler.ts
@@ -603,7 +603,15 @@ export class ModeHandler implements vscode.Disposable {
         this.vimState.alteredHistory = false;
         vimState.historyTracker.ignoreChange();
       } else {
-        vimState.historyTracker.addChange(this.vimState.cursorsInitialState.map(c => c.stop));
+        const addedChange = vimState.historyTracker.addChange(
+          this.vimState.cursorsInitialState.map(c => c.stop)
+        );
+        const mightChangeDocument = recordedState.hasRunOperator
+          ? recordedState.operator.mightChangeDocument
+          : action.mightChangeDocument;
+        if (addedChange && !mightChangeDocument) {
+          throw new Error(`Action '${action.toString()}' changed the document unexpectedly!`);
+        }
       }
     }
 
--- a/src/actions/base.ts
+++ b/src/actions/base.ts
@@ -47,6 +47,20 @@ export class BaseAction {
   private static readonly isSingleAlpha: RegExp = /^[a-zA-Z]$/;
 
   /**
+   * True if this action has the potential to change the document's text, and therefore trigger an
+   * undo step. This is for optimization purposes - we skip the expensive process of figuring out
+   * what changed to create that undo step if this returns false.
+   *
+   * Most commands do not modify the document, so this is false by default. This means if you add a
+   * command that can modify the document, you MUST set mightChangeDocument = true. Otherwise undo
+   * will not work properly.
+   *
+   * TODO: if all actions were pure, I think this would be unnecessary, as we could deduce it from
+   * vimState.transformations being empty or not.
+   */
+  public mightChangeDocument: boolean = false;
+
+  /**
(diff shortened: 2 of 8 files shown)


src/mode/modeHandler.ts
```diff
@@ -603,7 +603,15 @@ export class ModeHandler implements vscode.Disposable {
...
       } else {
-        vimState.historyTracker.addChange(this.vimState.cursorsInitialState.map(c => c.stop));
+        const addedChange = vimState.historyTracker.addChange(
+          this.vimState.cursorsInitialState.map(c => c.stop)
+        );
+        const mightChangeDocument = recordedState.hasRunOperator
+          ? recordedState.operator.mightChangeDocument
+          : action.mightChangeDocument;
+        if (addedChange && !mightChangeDocument) {
+          throw new Error(`Action '${action.toString()}' changed the document unexpectedly!`);
```
CVVPK: This will always throw with #4386 when switching out of Insert mode using `<C-c>` and `<C-o>` because `_isDocumentTextNeeded()` is expecting modes to be different so that the state of the doc after the last change is recorded. When this `addChange()` gets called we will be in Normal Mode and thus `<C-c>` (key=<copy>) might not change the document, which is correct. Typically we would have recorded the changes just before switching out of Insert mode because the `addChange()` in `ModeHandler.handleKeyEventHelper()` takes care of that before the action is ran.

CVVPK: Maybe have a `action.willSwitchModes` and pass that over to `addChange()` in `HistoryTracker.handleKeyEventHelper`?

J-Fields: I'm tempted to just set `mightChangeDocument` to `true` on these two commands and add to the `mightChangeDocument` comment explaining that since we only want to diff at the end of insert mode, a command that goes from insert mode to normal mode does (sort of) change the document. Thoughts on that?

CVVPK: I think `<C-o>` (CommandOneNormalCommandInInsertMode) should definitely have `mightChangeDocument = true` since that's exiting Insert mode. `<C-c>` (CommandEscInsertMode) already has it properly set, but I believe is failing because it gets overwritten to `<copy>`.

CVVPK: I guess setting `mightChangeDocument = true` on `<copy>` would solve it.

CVVPK: Yeah I like this idea better, would you like me to just commit these changes into #4386?

J-Fields: That'd be great, thanks

sqlkoala:
Hi, I activated this feature to try and test it.
First: results are really noticeable, nice :)

But I found some issues in connection with undo function. Repro:

1.  habe a buffer with "foobar"
2. do ciw on it and chage to "foobaz"
3. undo
4. I end up with "foobarfoobaz"

A similar problem like this:
1. goto insert mode
2. press Backspace or delete a few times
3. leave insert mode
4. undo ( is "messed up")

The second is more of an "uncommon workflow". Thanks for looking into it.

J-Fields:
@sqlkoala Thanks so much for testing this out! I'm not able to reproduce either one; would you mind submitting a new issue so we can track it more easily?

PR #4452, opened 2020-01-01
Ignore vimrc lines that attempt to remap <Plug>

### LINKED ISSUE #4446: Consider ignoring vimrc lines with <Plug>

A line on a vimrc config that looks like this:
```
nmap s <Plug>(easymotion-s2) 
```
Can cause problems by creating this remap:
```
before=s. after=<Plug>,(,e,a,s,y,m,o,t,i,o,n,-,s,2,)
```

It seems `<Plug>` is only used to remap plugins in Vim ([source](https://vim.fandom.com/wiki/Mapping_keys_in_Vim_-_Tutorial_(Part_3)#Use_of_.3CPlug.3E)). Since VSCode/Vim doesn't support these plugins at the moment, it appears it should have no negative side effects to ignore these lines. However, I'm not a vimrc config expert so there could be other better ideas.

CVVPK:
`<Plug>` is used to remap keys to a specific function in a plug in. This is a useful thing in Vim but not in VSCode/Vim since those plug ins are not even supported. Currently lines that include `<Plug>` in them get remapped, and end up causing some weird behaviours like inserting a string or creating recursive actions that end up blocking the editor. 

This PR adds a [negated look ahead](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions#special-negated-look-ahead) to the regex we're using to determine if a line is a remap, If the line includes `<Plug>` it'll fail to match the regex, thus saving us the trouble of having to go through the matches. 

Fixes #4446 

**Special notes for your reviewer:**
My knowledge of vimrc configs is quite limited, so I did some research to make sure `<Plug>` doesn't do something else and these changes end up breaking peoples' vimrc configs. I think it should be pretty safe to implement this, but let me know if that's not the case

Some useful info:
[Reddit post that quickly explains `<Plug>`](https://www.reddit.com/r/vim/comments/78izt4/please_help_understand_how_to_use_plug_mapping/)
[Probably a better explanation on stackoverflow](https://stackoverflow.com/questions/18546533/execute-plug-commands-in-vim)
[My initial source](https://vim.fandom.com/wiki/Mapping_keys_in_Vim_-_Tutorial_(Part_3)#Use_of_.3CPlug.3E)


J-Fields:
Thanks! I think a better long-term solution may be to respect mappings for plugins we support, but this is a good stopgap for the non-obvious misbehavior we've sometimes seen

PR #4456, opened 2020-01-01
Update statusbar with result of ex command from neovim

rsslldnphy:

I noticed that when running certain ex commands with neovim integration enabled, I wasn't seeing any output in the statusbar, regardless of the success or failure of the command. As a result I wasn't sure if commands had been successful or not.

On investigation I realised that there are two cases in which the extension passes an ex command to neovim for execution, but currently only one of them displays the result in the status bar. This commit brings the second case in line with the first so that the result of running the command is displayed:

image

As it was a simple fix I decided to just go ahead and make a PR rather than creating an issue.

As a side-note, watching the tests run was extremely cool. Nice work on those!

### DIFF
--- a/src/cmd_line/commandLine.ts
+++ b/src/cmd_line/commandLine.ts
@@ -88,7 +88,8 @@ class CommandLine {
     } catch (e) {
       if (e instanceof VimError) {
         if (e.code === ErrorCode.E492 && configuration.enableNeovim) {
-          await vimState.nvim.run(vimState, command);
+          const statusBarText = await vimState.nvim.run(vimState, command);
+          StatusBar.setText(vimState, statusBarText);
         } else {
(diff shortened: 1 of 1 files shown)


src/cmd_line/commandLine.ts
```diff
@@ -88,7 +88,8 @@ class CommandLine {
     } catch (e) {
       if (e instanceof VimError) {
         if (e.code === ErrorCode.E492 && configuration.enableNeovim) {
-          await vimState.nvim.run(vimState, command);
+          const statusBarText = await vimState.nvim.run(vimState, command);
+          StatusBar.setText(vimState, statusBarText);
```
jpoon: Should be same statement as line 94 (isError = true); only difference would be the text that is displayed.

PR #4461, opened 2020-01-02
Fix returnToInsertAfterCommand not false after switching to Insert Mode

### LINKED ISSUE #4453: Bad behavior when using C-o in insert mode
When using `C-o` in insert mode, followed by some "enter to insert mode" commands, we enter a buggy "insert mode" which requires to press `<Esc>` twice in order to exit it, instead of exiting it by pressing `<Esc>` only once.

**To Reproduce**
1. Enter insert mode.
2. Type any word.
3. While still in insert mode, press `C-o` followed by either `A`, `a`, `I`, `i`, `O`, `o`.
4. Press `<Esc>` **once**
5. *Still in insert mode* :disappointed:  

**Expected behavior**
After pressing `<Esc>`, the expected behavior is to enter normal mode.

**Environment**

- VSCode version: 1.41.1
- OS: Linux x64

cvaldev:

When we `<C-o>` and then perform a command that would change us into Insert mode (e.g. `i`,`a`,`o`, etc.) there would be an issue where we'd have to press `<Esc>` twice to exit Insert Mode. This happens because `<C-o>` sets `VimState.returnToInsertAfterCommand` to `true` and then we wait until an action is ran to set it back to `false`, the problem with this comes from this particular condition:

```ts
 ranAction = ranAction && vimState.currentMode === Mode.Normal;
```

As you may notice `vimState.currentMode` will not be `Mode.Normal` when we just executed a command that switched us into `Mode.Insert`. This blocks us from checking the state of `VimState.returnToInsertAfterCommand` and increasing the `VimState.actionCount`.

However, when `<Esc>` is executed the condition above will be `true` and thus we will finally be able to to check if `vimState.returnToInsertAfterCommand` is `true`. But, because this action (`<Esc>`) was the first to trigger the condition, we consider it the first action after `<C-o>` even though it is not. Thus, it would take another `<Esc>` to exit Insert Mode.

This PR fixes this issue by setting `VimState.returnToInsertAfterCommand` to `false` as soon as we change into Insert Mode. This way any action that switches to insert mode will be accounted for, i.e. we don't need to "returnToInsertAfterCommand" if we're already in Insert Mode. Additionally, this PR adds a test to check that `<C-o>` behaviour is correct.

Fixes #4453


test/mode/modeInsert.test.ts
```diff
@@ -215,6 +215,13 @@ suite('Mode Insert', () => {
     end: ['|abcd'],
   });
 
+  newTest({
+    title: 'Can <Esc> after entering insert mode from <ctrl+o>',
+    start: ['abc|'],
+    keysPressed: '<C-o>a<Esc>d',
```
J-Fields: Don't you need to be in insert mode before hitting `<C-o>`? Also, use `endMode: ` to assert that we're in normal mode rather than hitting `d` and making sure it's not inserted.

CVVPK: Fixed.

CVVPK: Thanks for the feedback, didn't know about `endMode`. Fixed.

turara:
@cvaldev 
Thank you! I've been really annoyed with this and I couldn't fix it myself.

PR #4468, opened 2020-01-04
Fix getCursorStyleForMode() always returned undefined

### LINKED ISSUE #4355: vim.cursorStylePerMode is being ignored
**Describe the bug**
In my settings.json I have these settings enabled:

```
"vim.cursorStylePerMode.insert": "underline",
"vim.cursorStylePerMode.normal": "underline",
```

**To Reproduce**
1. Copy the settings and see that it's not working

**Expected behavior**
Should respect the settings.json file

**Screenshots**

- Extension (VsCodeVim) version: 1.12.2
- VSCode version: 1.40.2
- OS: macOS

**Additional context**

It's working with 1.11.3 version

cvaldev:

This PR fixes the issue of not changing the cursor style as per the user settings when we switch modes. I've done this by changing `modeName` from being a type `Mode` back to a type `string`, and having the modeHandler resolve `ModeHandler.currentMode` into a `string`. 

Fixes #4355 

The first reason why we cannot have `modeName` be a `Mode` is because it is a [numeric enum](https://www.typescriptlang.org/docs/handbook/enums.html#numeric-enums) and not an [string enum](https://www.typescriptlang.org/docs/handbook/enums.html#string-enums), which means that `modeName.toString()` will simply be the number corresponding to that particular member (e.g. `Mode.Normal.toString() === 0`). The second reason is because it seems we cannot refer to `Mode` inside configuration.ts, i.e. doing `Mode[modeName]` or even checking if `Mode` is `undefined` inside of `getCursorStyleForMode()` breaks the entire extension as it fails to validate with an error:

```
Activating extension 'vscodevim.vim' failed: Cannot read property 'iskeyword' of undefined.
```
I'm unsure why this happens, it seems as if the Mode module doesn't get loaded until the extension has been activated? If you know what may cause this, please let me know.

### DIFF
--- a/src/configuration/configuration.ts
+++ b/src/configuration/configuration.ts
@@ -15,7 +15,6 @@ import {
   IHighlightedYankConfiguration,
   ICamelCaseMotionConfiguration,
 } from './iconfiguration';
-import { Mode } from '../mode/mode';
 
 const packagejson: {
   contributes: {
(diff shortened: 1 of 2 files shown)


notiv-nt:
we need to build a release.
seriously, this is an annoying bug

Issue #1490, opened 2017-04-10
VSCodeVim Undo stack should be synced with VSCode undo stack


casey-speer:
VSCodeVim undo is not synced with VSCode undo. I've provided a couple different issues/test cases below to illustrate what I mean:

**Simple undo**
VSCode and VSCodeVim seem to maintain different undo stacks, such that pressing `u` in VIM adds to the VSCode undo stack so that doing a VSCode undo (`CMD-Z`) undoes the VIM undo. `u` and `CMD-Z` should have the same function and share a common undo stack.

**Search Replace**
When doing a vim-based search and replace, one change with all the replacements is pushed to the VSCodeVim undo stack, whereas one change for each replacement is pushed to the VSCode undo stack. It's easy to get confused about the state of the document if you make a replacement, then press `CMD-Z` thinking you've reversed all the changes, make some other changes, then have to backtrack to revert the rest of the replacements.  

I'm not familiar with the extension architecture for VSCode. I can break these off into separate issues if you'd prefer.

*thumbs-up* 👍 this issue if it personally affects you! You can do this by clicking on the emoji-face on the top right of this post. Issues with more thumbs-up will be prioritized.

-----

### Simple undo
***What did you do?***
Press `i`. Type some text: `PIZZA`. Press escape. Press `i`. Type some more text: ` IS GREAT`. Press the `u` key to undo. Press CMD-Z.

***What did you expect to happen?***
" IS GREAT" should disappear upon pressing `u`. "PIZZA" should disappear upon pressing CMD-Z so we're left with an empty string.

***What happened instead?*** 
" IS GREAT" disappeared and then reappeared upon pressing CMD-Z so we're left with "PIZZA IS GREAT".

### Search and replace
***What did you do?***
Given a 50-line document with the string "Pizza" on each line, type `%s/Pizza/Hot Dogs/` from the VSCodeVim command line. Press `CMD-Z`. 

***What did you expect to happen?***
"Pizza" should appear on every line. 

***What happened instead?***
"Hot Dogs" appears on every line except the last one which shows "Pizza". Similarly, if you first undo with `u` then press `CMD-Z`, "Pizza" appears on every line except the first one which shows "Hot Dogs".

### Technical details:

* VSCode Version: 1.11.1
* VsCodeVim Version: 0.6.14
* OS: MacOS 10.11.6

johnfn:
The basic explanation is this.

[...]

PR #4484, opened 2020-01-09
Reduce side effects in actions

J-Fields:
VSCodeVim's basic architecture is "actions modify the VimState, which is then rendered into VSCode". Some actions, however, modify VSCode's state themselves (for instance, with `TextEditor.insert()`).
I'm attempting to minimize the number of actions like this for several reasons, the main one being that I believe having "pure" actions will make #1490 more achievable.


J-Fields:
Closing for now. This effort should happen at some point, but I probably won't get to it particularly soon

PR #4492, opened 2020-01-13
Implement <Del> handler for insert mode

### LINKED ISSUE #2423: Repeat with . ignores Delete key if first keystroke
Bug report

**What happened**:
Delete keystroke in insert mode is not repeated with . unless another keystroke comes before it

Delete keystroke is repeatable as first keystroke

type:
```
123test
123test
```
place cursor after first 123, change to insert mode and press the delete key 4 times to erase "test" then type "qwert". Exit insert mode, place cursor after second 123 and press . to repeat. "qwert" is inserted but "test" is not deleted.

Everything works normally as long as some other keystroke comes before the delete key.

**Environment**:

- Extension (VsCodeVim) version: 0.11.0
- VSCode version: 1.20.1
- OS version: Windows 7 x64

### LINKED ISSUE #4082: Cannot map <C-d> to <Del> in insert mode
**Describe the bug**
Cannot map `<C-d>` to `<Del>` in insert mode

**To Reproduce**
- Add `{ "before": ["<C-d>"], "after": ["<Del>"] }` into `"vim.insertModeKeyBindingsNonRecursive"`
- Press `<C-d>` in insert mode to delete text

**Expected behavior**
Press `<C-d>` in insert mode then text will be deleted like pressing `<Del>`

- Extension (VsCodeVim) version: 1.10.2
- VSCode version: 1.38.1
- OS: Windows 10

### LINKED ISSUE #4219: Implement <Del> as vim insert command ( like <c-h> )
Hi VScodeVim Team,
let me start with: Thank you for this extension. I just installed it together with vscode and it feels comfortable (to an experienced vim user).
However, I found one missing feature that seems worth implementing:  `del>` as vim insert command ( in a way that is repeatable and a single undo ).

I think `<del>` is not implemented in insert mode. I noticed this because I like to map `<c-l>`to `<del>`.
When pressing Del key, the vscode command "deleteRight" is probably executed. This does delete the char to the right, but it messes up  vim's undo and repeat. Like so: undo does not undo the `del`, it undoes the last vim command.
This extension provides a command: extension.vim_delete but this is not for insert mode, as it appears.

I would suggest to implement this analog to `c-h`. This is from source insert.js. I think a simple copy  + paste and changing position to position + 1 char to the right would to it.
````
let CommandCtrlHInInsertMode = class CommandCtrlHInInsertMode extends actions_1.BaseCommand {
    constructor() {
        super(...arguments);
        this.modes = [mode_1.ModeName.Insert];
        this.keys = ['<C-h>'];
    }
    exec(position, vimState) {
        return __awaiter(this, void 0, void 0, function* () {
                type: 'deleteText',
                position: position,
            });
            return vimState;
        });
    }
};
````

I mapped `<c-l>` to `<c-o>x` and this does the job, but it also "destroys" undo/repeat.

Relates to:
https://github.com/VSCodeVim/Vim/issues/4082

### PR DESCRIPTION
**Pull request by ldm0**
In the insert mode, `<Del>` key are not handled by VSCodeVim. So `<Del>` commands used in insert mode are transparent to the `dot` command and cannot be repeated with it, which is very inconvenient. And another problem is that user cannot remap any key to `<Del>`.

With this pull request, these problems are solved.  
Fixes #2423, fixes #4219, fixes #4082. 

BTW: I think we need some documentation and code annotations. I spent a whole night tracking the missing delete commands and find out that this bug can be easily located even with a little documentation describing the key handling precedure......

### DIFF
--- a/src/actions/commands/insert.ts
+++ b/src/actions/commands/insert.ts
@@ -270,7 +270,7 @@ export class CommandBackspaceInInsertMode extends BaseCommand {
         type: 'deleteRange',
         range: new Range(position.withColumn(desiredLineLength), position.withColumn(line.length)),
       });
-    } else if (position.line !== 0 || position.character !== 0) {
+    } else if (!position.isAtDocumentBegin()) {
       // Otherwise, just delete a character (unless we're at the start of the document)
       vimState.recordedState.transformations.push({
         type: 'deleteText',
--- a/src/actions/commands/insert.ts
+++ b/src/actions/commands/insert.ts
@@ -285,6 +285,32 @@ export class CommandBackspaceInInsertMode extends BaseCommand {
 }
 
 @RegisterAction
+export class CommandDeleteInInsertMode extends BaseCommand {
+  modes = [Mode.Insert];
+  keys = ['<Del>'];
+  mightChangeDocument = true;
+
+  public async exec(position: Position, vimState: VimState): Promise<VimState> {
+    const selection = TextEditor.getSelection();
+
+    if (!selection.isEmpty) {
+      // If a selection is active, delete it
+      vimState.recordedState.transformations.push({
+        type: 'deleteRange',
+        range: new Range(selection.start as Position, selection.end as Position),
+      });
+    } else if (!position.isAtDocumentEnd()) {
+      // Otherwise, just delete a character (unless we're at the end of the document)
+      vimState.recordedState.transformations.push({
+        type: 'deleteText',
+        position: position.getRightThroughLineBreaks(true),
+      });
+    }
+    return vimState;
+  }
+}
+
+@RegisterAction
 export class CommandInsertInInsertMode extends BaseCommand {
   modes = [Mode.Insert];
   keys = ['<character>'];
--- a/src/editorIdentity.ts
+++ b/src/editorIdentity.ts
@@ -1,7 +1,7 @@
 import * as vscode from 'vscode';
 
 /**
- * We consider two editors to be the same iff their EditorIdentities are the same
+ * We consider two editors to be the same if their EditorIdentities are the same
  */
 export class EditorIdentity {
   private _fileName: string;
--- a/src/editorIdentity.ts
+++ b/src/editorIdentity.ts
@@ -1,7 +1,7 @@
 import * as vscode from 'vscode';
 
 /**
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by J-Fields** [comment]
Would you mind writing some tests?

> BTW: I think we need some documentation and code annotations. I spent a whole night tracking the missing delete commands and find out that this bug can be easily located even with a little documentation describing the key handling precedure......

This is something that I've been working on recently. Feel free to reach out to me if you have trouble finding/understanding something in the codebase.

**Comment by ldm0** [comment]
> Feel free to reach out to me if you have trouble finding/understanding something in the codebase.  

Thanks!

And tests were added.

**Comment by ldm0** [comment]
Is there anything blocking on merging of this pull request? I've compiled and use my personal fork for more than a month and haven't seen any bug related to this.

**Comment by J-Fields** [comment]
> Is there anything blocking on merging of this pull request?

Nope, I've just been busy with other things recently. I'll take a look now

File: src/editorIdentity.ts [hunk 1]
```diff
@@ -1,7 +1,7 @@
 import * as vscode from 'vscode';
 
 /**
- * We consider two editors to be the same iff their EditorIdentities are the same
```
J-Fields: "iff" here is actually shorthand for "if and only if", though it's probably worth writing out because otherwise it can look like a typo hahaha

ldm0: OK, I'll leave it here :-P

src/actions/commands/insert.ts
```diff
@@ -284,6 +284,32 @@ export class CommandBackspaceInInsertMode extends BaseCommand {
...
+      // If a selection is active, delete it
+      vimState.recordedState.transformations.push({
+        type: 'deleteRange',
+        range: new Range(selection.start as Position, selection.end as Position),
+      });
+    } else if (!position.isAtDocumentEnd()) {
+      // Otherwise, just delete a character (unless we're at the end of the document)
+      vimState.recordedState.transformations.push({
+        type: 'deleteText',
+        position: position.getRightThroughLineBreaks(true),
```
J-Fields: I was surprised that this is vim's default behavior when I double-checked. I then wondered if it was configurable, which led me to file https://github.com/VSCodeVim/Vim/issues/4573

J-Fields:
LGTM, thanks for the PR!

Issue #4439, opened 2019-12-30
Multiple Cursor bug in html file


amitbisht511:
**Describe the bug**
In the latest vscode (1.41) the vscode team release a new feature called [html mirror cursor](https://code.visualstudio.com/updates/v1_41#_html-mirror-cursor) now when i open a html file and go to a tag in normal mode it also select its closing tag in multi cursor mode but when i move my cursor it does not close the multi cursor mode

**To Reproduce**
create a simple html file with some tag like 
```
<form>
<input type='text'>
<input type='text'>
</form>
```
 move your cursor in normal mode to opening form tag you see that closing form tag also get now move your cursor into next line in input you see that form tag is still selected

**Expected behavior**
the multi cursor should be closed when i am not in html tag

**Environment (please complete the following information):**

- Extension (VsCodeVim) version:
- VSCode version: 1.41
- OS: elementry os 5 (juno)

amitbisht511:
for anyone who is suffering from the same issue you can disable this vscode feature via setting `html.mirrorCursorOnMatchingTag`  just search _mirror cursor on matching tag_ in setting and uncheck the checkbox

piotrpalek:
@amitbisht511 thanks, this was so annoying I temporarily disabled the whole vim extension until I had time to search for a fix

Christopher-Hayes:
omg this was annoying as hell, had no idea why this was happening. Vim normally doesn't even have multi-cursors, was super confused. Was ready to disable Vim until I found this after an hour.

Brade:
HTML mirror cursor is an absolutely horrible feature. Was screwing me up for quite a while until I was finally forced to search about it. "On" by default is just an unbelievably bad choice.

J-Fields:
@Brade to be fair, I think if you're not using VSCodeVim, it's probably quite useful

Brade:
I'm not using that, and I can assure you it's still not useful. I often paste class attributes etc to similar open tags, and it kept pasting it all on the end tags as well. I've turned it off now, but I really disagree with the decision to have it "on" by default.

J-Fields:
Ah interesting; I'd figured it would be smart enough to not copy attributes to the end tag. That is bothersome.

andradei:
In my case, `html.mirrorCursorOnMatchingTag` was off by default but I had the `https://github.com/formulahendry/vscode-auto-rename-tag` extension installed which messed up this vim plugin big time.

Issue #4402, opened 2019-12-18
Dual cursor appears on tag match and causes error when switching to Visual Line Mode in HTML file


mikenap:
**Describe the bug**
When hovering over the beginning HTML tag of a pair, in Normal Mode, a second cursor appears on the closing tag. When switching to Visual Line Mode, with SHIFT + V, VS Code shows the message "ModeHandler: unexpected selection mode. selectionMode=4" and does not switch to Visual Line Mode.

**To Reproduce**
Create an HTML tag pair (i.e opening and closing `div` tags). In Normal Mode, place the cursor on either of the tag pair. A second cursor will appear on the opposite tag.  Try switching to Visual Line Mode with SHIFT + V, the message "ModeHandler: unexpected selection mode. selectionMode=4" will appear and will not switch to Visual Line Mode and highlight the line.

**Expected behavior**
When placing the cursor on a beginning or closing HTML tag element, a second (or dual cursor) should not appear on the opening and closing HTML tags, and switching to Visual Line Mode with SHIFT + V should switch the editor to Visual Line Mode.

**Environment (please complete the following information):**

- Extension (VsCodeVim) version: 1.12.2
- VSCode version: 1.41.0
- OS: macOS Mojave 10.14.6

icemanx7:
I also experienced this on the following **Environment**:

- VsCodeVim version: 1.12.2

- VSCode version: 1.41.0-1

- Arch Linux

CVVPK:
This seems similar to #4378.

The extra cursor you are seeing is not being caused by the Vim extension. vscode added this feature in the November update: [https://code.visualstudio.com/updates/v1_41#_html-mirror-cursor](https://code.visualstudio.com/updates/v1_41#_html-mirror-cursor). You need to set `html.mirrorCursorOnMatchingTag` to `false` in your settings to have the behaviour you were expecting. 

I was able to reproduce the error, on any file type, by just using VisualLine mode while multiple cursors are active.

armoucar:
> You need to set `html.mirrorCursorOnMatchingTag` to `false` in your settings to have the behaviour you were expecting.

That fixes the problem. But now I can't use `%` command for matching tag pairs :(

`> Emmet: Go to matching Pair` works though.

Svtter:
It is `editor.RenameOnType' now.

Issue #4378, opened 2019-12-13
Html tag mirror causes issues with visual line select.


syfenx:
Html tag mirror causes issues with visual line select etc.

**To Reproduce**

1.  Place cursor in html tag, press shift v and both the top/bottom html tags are selected

**Expected behavior**
When I tried to visually select some text to grab it and copy somewhere else, the second cursor on the top tag will select everything I don't want, unrelated to what I intended to select.

- Extension (VsCodeVim) version: 1.12.2
- VSCode version: 1.41.0
- OS: Win 10

**Temporary fix*
html.mirrorCursorOnMatchingTag = false

I don't have a lot of time right now, if this needs more clarification let me know.

zhengbli:
Seeing the same happening, have to suppress error.
OS: Linux
VsCodeVim: 1.12.2

micka190:
Can also confirm that this is an issue even if you revert to a previous VSCodeVim version. 

Seems that the issue was introduced with [VS Code's update that introduced HTML mirror tags](https://code.visualstudio.com/updates/v1_41#_html-mirror-cursor).

> This feature works by adding a multi-cursor to the matching tag when your cursor moves into an HTML tag name range. Just like in multi-cursor mode, you can use word-wise deletion or word-wise selection. The mirrored cursor is removed when you move your cursor outside the tag name range.

Disabling VS Code's `html.mirrorCursorOnMatchingTag` setting seems to do the trick for now.

minawk:
@micka190 Yes it happen also on previous VSCodeVim version.
OS: Linux
VSCode version: 1.41.0
VSCodeVim version: 1.11.3

J-Fields:
I actually just get an error message when I try this. What's the expected behavior, though? Enabling the setting means you want there to be a second cursor when on an HTML tag, so I'm not sure why `V` wouldn't affect both cursors.

syfenx:
> I actually just get an error message when I try this. What's the expected behavior, though? Enabling the setting means you want there to be a second cursor when on an HTML tag, so I'm not sure why `V` wouldn't affect both cursors.

```html
<p>stuff I don't want selected</p>
<div class="hello">
1
2
3
</div>
```
If I start at the closing div and select upwards, the mirrored cursor will select "stuff I don't want selected" because the mirror cursor starts at the top tag and selects upwards from there as well.

micka190:
@syfenx You'll also straight-up get an error message if you have something like:

```
[...]

PR #4498, opened 2020-01-15
Recommend disabling html.mirrorCursorOnMatchingTag in README

goldhand:
Adds note recommending disabling `html.mirrorCursorOnMatchingTag` in README to avoid errors and unexpected behaviors.

The new [VSCode `html.mirrorCursorOnMatchingTag` default setting](https://code.visualstudio.com/updates/v1_41) is incompatible with Vim UX. Since this is a default setting, it would be helpful in the setup instructions.

* https://github.com/VSCodeVim/Vim/issues/4439
* https://github.com/VSCodeVim/Vim/issues/4402
* https://github.com/VSCodeVim/Vim/issues/4378

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,8 @@ VSCodeVim is a Vim emulator for [Visual Studio Code](https://code.visualstudio.c
 
 VSCodeVim is automatically enabled following [installation](https://marketplace.visualstudio.com/items?itemName=vscodevim.vim) and reloading of VS Code.
 
+Vim keybindings are incompatible with [VSCode's HTML mirror cursor feature](https://code.visualstudio.com/updates/v1_41). You will need to set `"html.mirrorCursorOnMatchingTag": false` in `settings.json` to avoid errors and unexpected behavior (for both Mac and Windows).
+
 ### Mac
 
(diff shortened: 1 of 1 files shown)


J-Fields:
This also pertains to Windows users - should we copy it there or put it above?

goldhand:
Sorry for the delay @J-Fields, I worry moving it out above the mac instructions section might be missed by readers who are focused on finding specific instructions for their OS but hopefully that's not an issue because I think that's the best option. If more issues are open on this topic, we can address consider copying it to both sections.

goldhand:
I moved it above, added a note "(for both Mac and Windows)" and rebased the branch

jpoon:
Would putting this in https://github.com/VSCodeVim/Vim#-faq be a more suitable location?

goldhand:
@jpoon I don't think it would be, as it's a required setup step.

J-Fields:
Wouldn't it be better to check this setting on startup, and warn the user if they have it enabled? I suspect many people don't pay much attention to the README before using the extension.

J-Fields:
This setting was deprecated in favor of `editor.renameOnType`, which defaults to false, so I don't think this is much of an issue any more.

Issue #1505, opened 2017-04-13
ctrl+m or ctrl+j to insert <enter> with insert mode.


millionbonus:
### What did you do?
1. In setting.json turn on the ctrl keys support with `"vim.useCtrlKeys": true`
2. In text editor pressed `i` to go to insert mode.
3. pressed `ctrl+m` or `ctrl+j`.

### What did you expect to happen?
Insert a new line break(\<enter\>) in to text editor.

### What happened instead?
Nothing happened.

### Technical details:

* VSCode Version: 1.11.1
* VsCodeVim Version: 0.16.5
* OS: windows 10

johnfn:
okay i mean we can totally add that but i gotta know: why dont you just use enter??

millionbonus:
As a vimer. 
I use hjkl instead of arrow keys.
I use `ctrl+[` instead of `<esc`>. (and keep index finger on `j`) 
I use `ctrl+m` instead of `<enter>`. (just move index finger little bit to `m` then back to `j` very quick)  
Just because those keys are too far for my little finger.

rebornix:
This can be covered by https://github.com/VSCodeVim/Vim/issues/1543

millionbonus:
Can also this behavior apply in '`/`' and '`?`' searching?

millionbonus:
A workaround is that setting a keybinding `<ctrl-m>` to inert snippet `\n` 
```json

    {
        "key": "ctrl+m",
        "command": "editor.action.insertSnippet",
        "when": "editorTextFocus && !inDebugRepl && vim.mode == 'Insert'",
        "args": {
          "snippet": "\n"
        }
    },
    {
        "key": "ctrl+j",
        "command": "editor.action.insertSnippet",
        "when": "editorTextFocus && !inDebugRepl && vim.mode == 'Insert'",
        "args": {
          "snippet": "\n"
        }
    }
```

However, it still can't use those combination in command line.

can3p:
Is there a technical reason why this can't be done? From the first sight it's not, since Ctrl-[ works both in command line and in insert mode

can3p:
Can we resurrect this issue? I've implemented ctrl+m support in a linked pull request

kktt007:
> As a vimer.
> I use hjkl instead of arrow keys.
> I use `ctrl+[` instead of `<esc`>. (and keep index finger on `j`)
[...]

PR #4526, opened 2020-01-30
Add support for <C-m> in insert, search in progress and commandline in progress modes

can3p:
This pull request addresses issue #1505 
Summary: In vim `<C-m>` key binding works in the same way as pressing enter in all modes. This is not very common across editors but is very handy since requires much less movements.

I've made necessary changes only to discover that I cannot really
capture exactly `<C-m>` shortcut, because it's used by vscode "Tab moves focus"
shortcut. I removed it from the settings, but it looks like Vim plugin
still is not able to get it.

As a proof of concept I added another shortcut - `<C-q>` that does what
`<C-m>` should do. This shortcut should be removed before this pull
request is merged.

I have a couple ideas on how to approach the problem:

1. Disable this setting totally in case of vim plugin enabled
2. Temporary disable "tab moves focus" shortcut in all modes concerned.
3. Do changes on vscode to allow this keybinding to work if default shortcut is erased.

1 sounds to drastic, so I think 2 or 3 are the real options, but at the moment I don't have a good enough understanding of vscode code base to understand how to achieve it.

Can anyone give me some pointers to do progress with either option 2 or option 3? Any other ideas are welcome!

### DIFF
--- a/src/actions/commands/commandLine.ts
+++ b/src/actions/commands/commandLine.ts
@@ -259,8 +259,7 @@ class CommandInsertInSearchMode extends BaseCommand {
     ['<C-n>'], // Next
     ['<C-f>'], // Find
     ['<C-u>'], // Delete to beginning
-    // Another way to run search
-    ['<C-m>'],
+    ['<C-m>'], // Another way to run search
     ['<Home>'],
     ['<End>'],
     ['<Del>'],
--- a/src/actions/commands/commandLine.ts
+++ b/src/actions/commands/commandLine.ts
@@ -134,7 +134,7 @@ class CommandTabInCommandline extends BaseCommand {
 @RegisterAction
 class CommandEnterInCommandline extends BaseCommand {
   modes = [Mode.CommandlineInProgress];
-  keys = ['\n'];
+  keys = [['\n'], ['<C-m>']];
   mightChangeDocument = true;
   runsOnceForEveryCursor() {
     return this.keysPressed[0] === '\n';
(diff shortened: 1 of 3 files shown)


jpoon:
> 1. Disable this setting totally in case of vim plugin enabled

We do this for other commands. See https://github.com/VSCodeVim/Vim/blob/master/package.json#L159

can3p:
Completely missed the notification. Thanks you! I'll try to use this option

oshriPP:
> Completely missed the notification. Thanks you! I'll try to use this option

Thanks for the fixed. 
When the fix should be released?
There is support for <C-i> => tab?

can3p:
@jpoon Changing package.json indeed did a trick for me, thanks for the tip!

I copied ctrl+n shortcut since it has to work under same conditions as ctrl+m

Now it works for me in:
- input command mode
- input search mode
- insert mode (new line)

I updated my fork to the latest master, rebased the branch and squashed everything into a single commit, since there is no need for many.

Please let me know if you have any comments or it can be merged as is.

src/actions/commands/commandLine.ts
```diff
@@ -259,6 +259,8 @@ class CommandInsertInSearchMode extends BaseCommand {
     ['<C-n>'], // Next
     ['<C-f>'], // Find
     ['<C-u>'], // Delete to beginning
+    // Another way to run search
+    ['<C-m>'],
```
jpoon: nit: can you put the comment on the same line to follow the pattern established above?

can3p:
Done!

Issue #3461, opened 2019-02-06
Final cursor position after surround is wrong


ineol:
**Describe the bug**
The final cursor position after surround is wrong: in neovim (tpope/vim-surround) it is *on* the bracket, not after.

**To Reproduce**

1. Write `toto tata titi`
2. in visual mode, select `tata`
3. surround with braces using `S }`
4. the final state is: `toto {>t<ata} titi` where the cursor is on the `t`.

**Expected behavior**
The cursor should be on the `{`.

**Environment (please complete the following information):**

- Extension (VsCodeVim) version: 1.0.7
- VSCode version: Code - Insiders 1.31.0-insider
- OS: Ubuntu 18.10

**Additional context**
The behavior I'm advocating for is useful when the part that is being surrounded is meant to be an argument for a function.

jvoegele:
I would love to see this issue resolved. The nice thing about the cursor placement in `vim-surround` is that after surrounding text with brackets, you can hit `i` to insert text _before_ the bracket, or hit `a` to insert text _after_ the bracket. With the cursor placement in the vscode-vim plugin, you can insert text after the bracket with `i`, but to insert text before the bracket requires two keystrokes: `hi`. Not a huge problem, but it is slightly less efficient and works against my muscle memory.

J-Fields:
This should be a quick fix; I'll take a look when I get a chance.

PR #4558, opened 2020-02-15
Fix surround cursor placement bug (#3461)

TransientError:
This makes VSCodeVim behave more like vim-surround, which I think users would find more intuitive.

https://github.com/VSCodeVim/Vim/issues/3461

I've run the npm test. I had to change a few because the surround ones were checking for the old behavior. I checked that the new tests are consistent with vim-surround's behavior though.
I'm not too familiar with either typescript nor this repo, so I'm grateful for the patience and possibly extra support.

### DIFF
--- a/src/actions/plugins/surround.ts
+++ b/src/actions/plugins/surround.ts
@@ -1,6 +1,6 @@
 import { VimState } from '../../state/vimState';
 import { PairMatcher } from './../../common/matching/matcher';
-import { Position } from './../../common/motion/position';
+import { Position, PositionDiff } from './../../common/motion/position';
 import { Range } from './../../common/motion/range';
 import { configuration } from './../../configuration/configuration';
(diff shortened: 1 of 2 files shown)


src/actions/plugins/surround.ts
```diff
@@ -485,6 +485,7 @@ export class CommandSurroundAddToReplacement extends BaseCommand {
         type: 'insertText',
         text: startReplace,
         position: start,
+        diff: new PositionDiff({ character: -startReplace.length }),
```
J-Fields: For posterity's sake, could you please add a comment explaining what this `PositionDiff` is for?

J-Fields:
Seems reasonable other than my request for a bit of documentation. Thanks a bunch for the contribution!

J-Fields:
Thank you!

PR #4569, opened 2020-02-20
syncCursors() will update all internal cursors

### LINKED ISSUE #4553: Multicursor issues on split window
**Describe the bug**
It seems the cursors don't correspond to the current window but instead to that document's last state.

**To Reproduce**

1.  Create multiple cursors on a document
2.  Split it
3.  On the new window make it a single cursor
4.  Switch back to the original window and immediately perform a move or change mode
5. See your multiple cursors disappear

The reverse, originally having a single cursor and then creating multiple cursors, will give you unwanted multiple cursors when you go back to your original single cursor window. This second case also creates an issue where if you insert a character, it will be inserted at invisible cursor positions.

**Expected behavior**
Same as VSCode where the cursors correspond to the current window.

- Issue Type: `Bug`
- Extension Name: `vim`
- Extension Version: `1.12.4`
- OS Version: `Linux x64 4.15.0-76-generic`
- VSCode version: `1.42.0`

cvaldev:
This PR makes it so all `vimState` cursor positions correctly represent VSCode's rather than only updating the main cursor.

Before this PR the extension would update all cursors only after a command was executed which could cause issues when using different number of cursors on a split editor and performing an action immediately after switching windows. 

Fixes #4553 

All feedback is welcomed.

### DIFF
--- a/src/mode/modeHandler.ts
+++ b/src/mode/modeHandler.ts
@@ -79,6 +79,14 @@ export class ModeHandler implements vscode.Disposable {
   public syncCursors() {
     setImmediate(() => {
       if (this.vimState.editor) {
+        let cursors: Range[] = [];
+        for (const { start, end } of this.vimState.editor.selections) {
+          cursors.push(
+            new Range(Position.FromVSCodePosition(start), Position.FromVSCodePosition(end))
+          );
+        }
+        this.vimState.cursors = cursors;
+
         this.vimState.cursorStartPosition = Position.FromVSCodePosition(
           this.vimState.editor.selection.start
         );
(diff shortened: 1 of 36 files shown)


src/mode/modeHandler.ts
```diff
@@ -79,6 +79,14 @@ export class ModeHandler implements vscode.Disposable {
   public syncCursors() {
     setImmediate(() => {
       if (this.vimState.editor) {
+        let cursors: Range[] = [];
+        for (const { start, end } of this.vimState.editor.selections) {
```
J-Fields: Instead of looping + pushing, could you write this by `map`ping over `this.vimState.editor.selections`?

Issue #3277, opened 2018-12-19
<c-o> jumps back to wrong location after 'gd'


js-shi:
**Describe the bug**
When navigating some C++ source code, I found C-o sometimes cannot jump back to the correct position before 'gd'.  On the other hand, Alt + Left Arrow can correctly jump back to the call site.

**To Reproduce**

Just use 'gd' to perform go to symbol on the call site and <c-o> to jump back.  In most cases, it works great but sometimes not.  The bug can be reproduced constantly but I cannot figure out the pattern yet.

**Expected behavior**
Jump back to the correct position.

- Extension (VsCodeVim) version: 0.17.0
- VSCode version: 1.30.0

GrandZhuo:
I have encountered the same problem recently, but I don't remember which version started to have the problem.
And I found that if I use "gd" jump to the definition in the same file, I can't jump back to the correct position using <C-o>. But it works if I jump to the definition in different files.
I hope to fix this bug as soon as possible.

wYuanHpp:
same issues

jpoon:
Thanks @GrandZhuo for describing the behaviour. Really helped track down the issue.

Filed https://github.com/Microsoft/vscode/issues/65809 against VS Code. The API call to go to the definition returns immediately. As a result, when we save the jump location, we are saving the old location before the cursor actually moves to the new location. 

Adding a 1s sleep seems to do the trick.

jpoon:
We can get rid of our hack following https://github.com/Microsoft/vscode/milestone/83.

eddiewang:
I'm currently on v1.31.0 and experiencing the same issue. Just to confirm, when I use `gd` to jump to a different file, I am unable to use `c-o` to go back to that file.

Using `c-p-p` is my current workaround for this.

solomonchild:
The same behavior is observed if you Ctrl+Click instead of `gd`. `C-o` jumps to a wrong location (even if it is the same file).

emmanueltouzery:
I think this issue is still present, the bug probably should be reopened. I'm certainly having issues with C-o myself, the builtin VS code shortcut works fine though.

ihaveadrame:
Now I have a similar problem，go to defined works fine but error when jumping back（ctrl + o）.  I found that this problem is normal in the old version, such as 1.11.3 I use now。

os： win10

troygibb:
[...]

PR #4581, opened 2020-02-24
Fix wrong return in ModeHandlerMap.get

### LINKED ISSUE #4479: <c-o>/<c-t> jump back to wrong location after 'gd'
**Describe the bug**
Pressing <C-o> after 'gd' goes to a different file and not back to the previous location.

Re-reporting #3277 since this bug has not been re-introduced (I'm guessing after the 1s sleep was removed)

**To Reproduce**
Press 'gd'
Press <C-o>
Now in wrong location

**Expected behavior**
Going to the previous location

- VSCode version: 1.41.1
- OS: macOS 10.14.6

**Additional context**

fatanugraha:
Hello, in my setup, this resolves #4479, #3277

From what i can observe, here's what i found to arrive to this conclusion:
- on `vscode.window.onDidChangeActiveTextEditor` i always get `lastClosedModeHandler = undefined` despite that the current file is exists on `ModeHandlerMap`
- due to `lastClosedModeHandler = undefined` jumping from current file doesn't push it to the jump stack

### DIFF
--- a/src/editorIdentity.ts
+++ b/src/editorIdentity.ts
@@ -13,11 +13,15 @@ export class EditorIdentity {
   /**
    * For use in tests
    */
-  public static createRandomEditorIdentity(): EditorIdentity {
+  public static createRandomEditorIdentity(
+    injectName: string | undefined = undefined
+  ): EditorIdentity {
     return new EditorIdentity(
-      Math.random()
-        .toString(36)
-        .substring(7)
+      injectName
+        ? injectName
+        : Math.random()
+            .toString(36)
+            .substring(7)
     );
   }
 
--- a/src/editorIdentity.ts
+++ b/src/editorIdentity.ts
@@ -10,22 +10,7 @@ export class EditorIdentity {
     return new EditorIdentity(textEditor?.document?.fileName ?? '');
   }
 
-  /**
-   * For use in tests
-   */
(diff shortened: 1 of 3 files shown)


src/editorIdentity.ts
```diff
@@ -13,11 +13,15 @@ export class EditorIdentity {
   /**
    * For use in tests
    */
-  public static createRandomEditorIdentity(): EditorIdentity {
+  public static createRandomEditorIdentity(
```
J-Fields: Why not just call `new EditorIdentity` in the test yourself? Seems weird to call `createRandom` and then override the randomness

J-Fields:
Nice catch, thanks for the PR! I'll take a closer look later but this seems reasonable and does fix the issue, I think.

src/editorIdentity.ts
fatanugraha: the constructor for `EditorIdentity` is marked as `private`. Should i made it `public` or create a new method for it?

J-Fields: Just make it public; you're exposing essentially the same API but less clearly by adding a parameter to the create random function.

fatanugraha: Sure, i also removed the `createRandomEditorIdentity` method.

jpoon:
Thanks for the fix!

PR #4592, opened 2020-02-26
`gq` range when cursor is in the middle of a line.

### LINKED ISSUE #4590: gq fails when cursor is in middle of a line
For situation like this:
```
    // pine |apple 
    // SpongeBob Patrick
```
With `gqj`, we should get result like this:
```
    |// pine apple SpongeBob Patrick
```
But currently we get:
```
|    // pine apple
// SpongeBobPatrick
```

ldm0:
**Bugs:**
Incorrect behaviour when cursor is in the middle of a line.(fixed by expanding start and end.)

**Small Refactor:**
Indentation should be got in paragraph reflow rather than being passed in as a parameter.

Fixes #4590

### DIFF
--- a/src/actions/operator.ts
+++ b/src/actions/operator.ts
@@ -906,7 +906,9 @@ class ActionVisualReflowParagraph extends BaseOperator {
     return '';
   }
 
-  public reflowParagraph(s: string, indent: string): string {
+  public reflowParagraph(s: string): string {
+    const indent = this.getIndentation(s);
+
     let indentLevel = 0;
     for (const char of indent) {
       indentLevel += char === '\t' ? configuration.tabstop : 1;
@@ -1086,10 +1088,11 @@ class ActionVisualReflowParagraph extends BaseOperator {
   public async run(vimState: VimState, start: Position, end: Position): Promise<VimState> {
     [start, end] = Position.sorted(start, end);
 
-    let textToReflow = TextEditor.getText(new vscode.Range(start, end));
-    let indent = this.getIndentation(textToReflow);
+    start = start.getLineBegin();
+    end = end.getLineEnd();
 
-    textToReflow = this.reflowParagraph(textToReflow, indent);
+    let textToReflow = TextEditor.getText(new vscode.Range(start, end));
+    textToReflow = this.reflowParagraph(textToReflow);
 
     vimState.recordedState.transformations.push({
       type: 'replaceText',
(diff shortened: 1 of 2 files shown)


test/mode/modeNormal.test.ts
```diff
@@ -1402,7 +1402,7 @@ suite('Mode Normal', () => {
     start: [
       '//    We choose to write a vim extension, not because it is easy, but because it is hard|.',
     ],
-    keysPressed: 'Vgq',
+    keysPressed: 'gq$',
```
hao-flexport: I think we should both cover `Vgq` and `gp$`.

src/actions/operator.ts
```diff
@@ -1094,21 +1096,22 @@ class ActionVisualReflowParagraph extends BaseOperator {
...
 
     vimState.recordedState.transformations.push({
       type: 'replaceText',
       text: textToReflow,
       start: start,
       end: end,
-      // Move cursor to front of line to realign the view
-      diff: PositionDiff.newBOLDiff(),
+      // Move cursor to front of lowest line to realign the view
+      diff: PositionDiff.newBOLDiff(textToReflow.split('\n').length - 1),
```
hao-flexport: Is it safe to just split by `\n`? I mean can this handle `\r\n`.
Hope I am just over thinking.

ldm0: Yes. `"\n".split('\n')` and `"\r\n".split('\n')` will return the same result. You can check that right now by pressing f12. :-P

test/mode/modeNormal.test.ts
ldm0: I also want to cover the "Vgq". But I don't have enough time to work on the bug of `V`. Current implementation of `V` is just just kind of workaround and it's result is not like what we expected. If we insist on testing this `Vgq`, the more important thing will be postponed.

hao-flexport: I think you can keep previous tests, declaring the wrong behaviors with comment/explanation, and create your new ones.
Hope other members would not oppose.

src/actions/operator.ts
```diff
@@ -1086,18 +1088,19 @@ class ActionVisualReflowParagraph extends BaseOperator {
...
```
J-Fields, line 1103: This doesn't work when the cursor is already at the bottom (`gq5k`, for instance)

ldm0: @J-Fields Ah, that's really a problem. :-/ I'll revert it.

PR #4594, opened 2020-02-26
Fix misplaced cursor after `VV`

### LINKED ISSUE #4593: Cursor should stay where it is after two `V`
Just like the title, `VV` will move the cursor to the end.
This seems to be a missing workaround in commit 81614d1e7655efafabab04593b713f7d28139c7e

ldm0:
So simple :-P
Fixes #4593
But currently these are just workaround. We need to simulate a cursor in visualLine mode in the future.

### DIFF
--- a/src/actions/commands/actions.ts
+++ b/src/actions/commands/actions.ts
@@ -2364,6 +2364,12 @@ class CommandExitVisualLineMode extends BaseCommand {
   keys = ['V'];
 
   public async exec(position: Position, vimState: VimState): Promise<VimState> {
+    if (vimState.visualLineStartColumn !== undefined) {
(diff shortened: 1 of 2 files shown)


J-Fields:
Thanks!

> But currently these are just workaround. We need to simulate a cursor in visualLine mode in the future.

Last I checked, you couldn't put a cursor in the middle of a selection - has that changed?

ldm0:
> Thanks!
> 
> > But currently these are just workaround. We need to simulate a cursor in visualLine mode in the future.
> 
> Last I checked, you couldn't put a cursor in the middle of a selection - has that changed?

I haven't check that, but simulating a cursor position which can be moved with hjkl is possible, even if it can't be shown on the screen.

Issue #4610, opened 2020-02-29
Visibility of easymotion markers


mkrptr:
Easymotion markers look misaligned and are sometimes cut off. Also the font makes it difficult to see. Here's how it looks like in my installation:
![Screenshot_2020-02-29_12-50-41](https://user-images.githubusercontent.com/19928439/75604675-b9908e00-5af4-11ea-8a17-d61ee6f4dcb3.png)
I played a bit with the styles and am going to link the pr here later

PR #4611, opened 2020-02-29
Easymotion markers visibility

mkrptr:
Easymotion markers don't look quite right out of the box. The hardcoded font and the way decorations look like they're shifted down make it really hard to see. Especially with the theme I use (Dracula). I made those decorations look slightly more distinguishable. Here's what I ended up with.
Before:
Before
After:
After
I tested it on ~20 themes and it seems to look good on all of them.

Relevant issue: #4610

### DIFF
--- a/src/actions/plugins/easymotion/easymotion.ts
+++ b/src/actions/plugins/easymotion/easymotion.ts
@@ -207,6 +207,13 @@ export class EasyMotion {
     );
   }
 
+  private getEasymotionMarkerHeight() {
+    if (configuration.easymotionMarkerHeight === 0) {
+      return 'auto';
+    }
+    return `${configuration.easymotionMarkerHeight}px`;
+  }
+
   public updateDecorations() {
     this.clearDecorations();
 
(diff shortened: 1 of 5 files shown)


src/actions/plugins/easymotion/easymotion.ts
```diff
@@ -243,22 +232,18 @@ export class EasyMotion {
         after: {
           contentText: keystroke,
           backgroundColor: this.getEasymotionMarkerBackgroundColor(),
-          height: `${configuration.easymotionMarkerHeight}px`,
```
xconverge: If someone specifies a height now it is ignored. I am not "opposed" to this but it makes it a bit tricky to clean up the configuration for these markers, I don't think too many people have set them, but for a few people this could be a regression if they like the way they setup their markers.

mkrptr: Okay, I'll get it back then.

cupcakearmy: Yes ! Love this ❤️

pushqrdx:
hey there @mkrptr you might be interested in #4635 which implements a behavior similar to the official plugin

J-Fields:
Thanks for the contribution, but I think we're going to go with #4635

Issue #3274, opened 2018-12-18
s and S Key Commands Not Working


caseysiebel:
in command mode the s and S key commands don't do anything. I expect that when I press s in command mode, the character under my cursor should be deleted and then I should be in insert mode at the cursor location. But currently it's not doing anything.

https://stackoverflow.com/questions/53821721/vs-code-vim-plugin-in-command-mode-the-s-and-s-key-commands-dont-do-anything

jpoon:
Can't repro. These are most definitely supported commands: https://github.com/VSCodeVim/Vim/blob/master/src/actions/commands/actions.ts#L2345.

jpoon:
Closing this issue, but feel free to re-open if I am mistaken from my above comment.

ELLIOTTCABLE:
For anybody else running across this — I accidentally enabled the `vim.sneak` option when I first installed VSCode-Vim; make sure you didn't too!

Set `"vim.sneak": false` in your `settings.json`.

caseysiebel:
@ELLIOTTCABLE THANK YOU!

rethab:
Ran into this as well, because I just copied the default config from the README. As I suspect that most people don't use sneak, I think it might make sense to not include that in the example.

WDYT @jpoon ? I'd be happy to draft a PR.

jpoon:
Feel free to create a PR @rethab

rethab:
@jpoon : https://github.com/VSCodeVim/Vim/pull/4619

PR #4619, opened 2020-03-04
readme: remove sneak from quick example

rethab:
  * many people seem to copy the 'quick example' from the readme as
    their initial config. The problem with that quick example is that it
    enables 'sneak', which by defaults breaks the S and s commands (see
    issue #3274).

  * by removing that setting from the example, people should have a
    working config out of the box. If someone is interested in sneak,
    other sections of the readme still go into details.

  * #3274 

  * Please see comment by @jpoon : https://github.com/VSCodeVim/Vim/issues/3274#issuecomment-594106691

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,7 +73,6 @@ Below is an example of a [settings.json](https://code.visualstudio.com/Docs/cust
 ```json
 {
   "vim.easymotion": true,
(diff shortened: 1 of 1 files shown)


jpoon:
Thanks for da PR.

PR #4628, opened 2020-03-07
Fix error when count exceeds max number of rows in "[count]Y/J"

### LINKED ISSUE #4625: 10000 Shift-Y  causes error
**Describe the bug**
A common pattern I use to yank a large block of code from a random point in a file to the end of a file is 10000 Shift-Y.   ie, pick a number that is definitely larger than the end of file.

**To Reproduce**

1.  Open any file
2.  Hit 10000 Shift-Y

**Expected behavior**
To yank all lines in the file from the cursor to the end of the file into the buffer.
This works in regular vim.

But the Vim plugin throws the following error:

TaskQueue: Error running task. Failed to handle key=Y. getLineLength() called with out-of-bounds line 9999.

**Screenshots**

- Extension (VsCodeVim) version: v1.13.0
- VSCode version: 1.42.1
- OS: Mac

**Additional context**

lusingander:

An error occurred when executing yank(Shift+y) or join(Shift+j) with a count exceeding the number of lines in the file.

(e.g. execute `1000Y` on the 100 line file.)

So I fixed it to calculate the number of rows to be processed using `Position#getDownByCount`.

Fixes #4625


test/mode/modeNormal.test.ts
```diff
@@ -1805,6 +1805,20 @@ suite('Mode Normal', () => {
...
+    end: ['one', '|one', 'two', 'two', 'three'],
+  });
+
+  newTest({
+    title: 'can do [count]Y if count is larger than EOF',
+    start: ['|one', 'two', 'three'],
+    keysPressed: '100Yp',
+    end: ['one', '|one', 'two', 'three', 'two', 'three'],
+  });
+
```
hao-flexport: Better to add a test case for >100 lines by:
```javascript
lines = new Array(100).fill().map((_, i) => `line-${i}`)
```
Because the issue is `error in "[count]Y/J" with large count`, now we should prove for large count it works for either <=100 or > 100.

test/mode/normalModeTests/commands.test.ts
```diff
@@ -248,6 +248,20 @@ suite('Mode Normal', () => {
...
+    end: ['one two| three', 'four'],
+  });
+
+  newTest({
+    title: "Can handle 'J' with count if count is larger than EOF",
+    start: ['|one', 'two', 'three', 'four'],
+    keysPressed: '100J',
+    end: ['one two three| four'],
+  });
+
```
hao-flexport: So, here for >100 lines again.

test/mode/modeNormal.test.ts
lusingander: Thank you for the review!
I think it's right in a way, but I think that the problem is that the pull request title and commit comments are not right. The root issue is that error occurred when exceeding the max number of rows.

hao-flexport: Ah, make sense.

lusingander: Thank you for confirmation, I changed PR title.

src/actions/commands/actions.ts
```diff
@@ -2145,7 +2145,7 @@ export class CommandYankFullLine extends BaseCommand {
   public async exec(position: Position, vimState: VimState): Promise<VimState> {
     const linesDown = (vimState.recordedState.count || 1) - 1;
     const start = position.getLineBegin();
-    const end = new Position(position.line + linesDown, 0).getLineEnd().getLeft();
+    const end = position.getDownByCount(linesDown).getLeft();
```
J-Fields: Why can we drop the `getLineEnd()`?

lusingander: I thought it is unnecessary because end position is recalculated in the `YankOperator#run`.

```typescript
    let extendedEnd = new Position(end.line, end.character + 1);

    if (vimState.currentRegisterMode === RegisterMode.LineWise) {
      start = start.getLineBegin();
      extendedEnd = extendedEnd.getLineEnd();
    }
```
(operator.ts L284)

However, if so, `getLeft` also seems unnecessary...

src/actions/commands/actions.ts
```diff
@@ -2140,7 +2140,7 @@ export class CommandYankFullLine extends BaseCommand {
   public async exec(position: Position, vimState: VimState): Promise<VimState> {
     const linesDown = (vimState.recordedState.count || 1) - 1;
     const start = position.getLineBegin();
-    const end = new Position(position.line + linesDown, 0).getLineEnd().getLeft();
+    const end = position.getDown(linesDown).getLeft();
```
J-Fields: `getLeft()` might be necessary due to this line in `YankOperator`, which does no bounds-checking:
```let extendedEnd = new Position(end.line, end.character + 1);```

This could be refactored, but the PR currently seems like it's strictly improving matters, so I'll merge. Thank you!

lusingander: I understand, thanks for the confirmation and response!

PR #4641, opened 2020-03-10
Fix `g~`

### LINKED ISSUE #4567: Support g~~
Like `guu` and `gUU`, `g~~` should invert the case of every character on the current line.

`V~` does this, however it also moves the cursor to the beginning of the line

**Additional context**
You can see the documentation on this command [here](http://vimdoc.sourceforge.net/htmldoc/change.html#g~~). As a side note, `gugu`, `gUgU`, and `g~g~` are synonymous commands in vim and are also not supported by this extension, however I don't see this as an issue since I don't know why anyone would be using these longer versions.

lusingander:

Fix two behaviors for `g~` and add some tests on `~`/`g~` in visual mode.

- Support `g~~`
  - It was not implemented.
- Support `g~` in visual block mode
  - It worked in visual mode and visual line mode.
  - `gu` and `gU` also worked in visual block mode.
- Add some tests

Fixes #4567


src/actions/operator.ts
```diff
@@ -771,7 +771,7 @@ class ToggleCaseVisualBlockOperator extends BaseOperator {
 
 @RegisterAction
 class ToggleCaseWithMotion extends ToggleCaseOperator {
-  public keys = ['g', '~'];
+  public keys = [['g', '~']];
```
J-Fields: This change shouldn't be necessary

lusingander: Thanks for the review.
If I don't make this change, I get `ts2416` error.
I suppose this is because `ToggleCaseWithMotion` inherits `ToggleCaseOperator` and`ToggleCaseOperator` implements `keys` of type `string[][]`.
Do you know how to solve this?

J-Fields: Oh I see - nope, that's fine

Issue #4543, opened 2020-02-06
Can we support textobj-parameter plugin please ??


skywind3000:
Can we support [textobj-parameter](https://github.com/sgur/vim-textobj-parameter) ??

It is the best textobj plugin which enables me to modify function arguments and any comma separated list items in a fast way.

Usage
-----

-   `i,` to inner parameter object

    ```vim
    function(param_a, param_b, param_c)
             |<--->|  |<--->|  |<--->|
    ```

-   `a,` to a parameter object including whitespaces and comma

    ```vim
    function(param_a, param_b, param_c)
             |<----->|
    function(param_a, param_b, param_c)
                    |<----->|
    function(param_a, param_b, param_c)
                             |<----->|
    ```

-   In addition, 'i2,' is similar to `a,` except trailing whitespace characters (especially for first parameter)

    ```vim
    function(param_a, param_b, param_c)
             |<---->|
    ```

It is something really hold me back from moving to vscode.

J-Fields:
#2338

J-Fields:
Closing in favor of #2338. There is an open PR for this feature (#4653), so hopefully it'll be in a near-future release.

Issue #2338, opened 2018-01-28
Add support for text-object like motion for arguments


fabrik42:
This is a feature request. 🙏

First of all, thank you for this great extension! I am currently considering switching vom Spacemacs (evil) to VS Code and the Vim extension is crucial for me.

The main thing that is missing for me is the following motion:
[argtextobj.vim](http://www.vim.org/scripts/script.php?script_id=2699) or [evil-args](https://github.com/wcsmith/evil-args).

This is extremely handy during development, especially refactoring. Is there any chance this will get implemented in the Vim extension?

*An example from the evil-args README*

### `evil-inner-arg`/`evil-outer-arg`
Select an inner/outer argument text object.

For example, `cia` transforms:

    function(ar|g1, arg2, arg3)
    function(|, arg2, arg3)

or

    function(arg1, ar|g2, arg3)
    function(arg1, |, arg3)

`daa` transforms:

    function(ar|g1, arg2, arg3)
    function(|arg2, arg3)

or

    function(arg1, ar|g2, arg3)
    function(arg1, |arg3)

---

Let me know what you think!
Best, Christian

johnfn:
I'm a bit confused how `cia` is different than `caw` and how `daa` is different than `df ` (delete find space). They seem the same to me. :) If it's just the keymappings you miss, you can use our remapping feature to map cia to caw and daa to `df `.

fabrik42:
These are just the most simple examples, the movement can to [much more](http://www.vim.org/scripts/script.php?script_id=2699), like handling types and more complex expressions in an argument.

For example deleting an argument like `myfunc(int counter+1)` is supported.

calebeby:
I really think this would be an awesome feature to have, it was probably my most used text object in Vim. Is there any chance this could be implemented?

Thanks!

delaaxe:
I find the arguments plugin to be just as useful as surround!

fabrik42:
@joonro Hi! Is there any chance this will make it into VSCodeVim? If there is general interest, I could also try to contribute and help!

Cheers, Christian

J-Fields:
@fabrik42 None of the maintainers plan on doing this any time soon, I think. Feel free to submit a PR!

skywind3000:
I wish to create a PR, but I haven't written any typescript before.

lmNt:
I have started working on this feature, because it bothers me to not have this movement available.
[...]

Issue #2196, opened 2017-12-05
Support for omap-param plugin


fly1n:
I found omap-param plugin was pretty useful to do parameter level edit. Can you add support for this plugin?

J-Fields:
Could you link to the plugin you're referring to?

fly1n:
https://github.com/LucHermitte/lh-cpp/blob/master/plugin/omap-param.vim

J-Fields:
#2338

PR #4653, opened 2020-03-15
WIP: Textobject argument movement

lmNt:
This PR implements the often requested parameter/argument textobject movement known from many vim plugins, such as [targets.vim](https://github.com/wellle/targets.vim) or [argtextobject.vim](https://www.vim.org/scripts/script.php?script_id=2699).
It is a very useful movement across basically every language and will make editing a lot more comfortable.

#2338  #4543 #2196

This is my first vscode-vim contribution, and, my first attempt in the typescript language - so please let me know which things are inconsistent on the language side of things.

##### Failing tests
Apart from that, this PR will still has 2 failing tests, where I am not quite sure of how to fix them.
These failing tests are only for *failed* movements:
   * "Will fail 'cia' with no delimiters"
   * "Will fail 'cia' with flipped delimiters"

For both these tests the cursor position after the movement is offset `+1`, which I don't really understand, since I return `const failure = { start: position, stop: position, failed: true };` where `position` is the initial position. I would expect that returning `failed: true` should not advance the cursor. Can somebody help me out here? (maybe you @J-Fields? :))

##### Refactor
The `execAction` method got quite long. I was thinking of pulling out the `start/stop` position refinement into a separate method for more readability. Let me know if I should.

##### Missing configuration
The current state of the PR does not yet support configuration. I had a quick look into how the configuration is handled and would aim to implement the following configuration keys:

 * `vim.argumentObjectDelimiters`
 * `vim.argumentObjectOpeningDelimiters`
 * `vim.argumentObjectClosingDelimiters`

all being of type `string[]` and their defaults to be the currently hardcoded delimiters, i.e. `,`, `[]`, `()`.


src/actions/textobject.ts
```diff
@@ -640,3 +640,211 @@ class InsideIndentObjectBoth extends IndentObjectMatch {
...
+    }
+    vimState.cursorStartPosition = start;
+
+    return {
+      start: start,
+      stop: stop,
+    };
+  }
+
+  public static getLeftDelimiter(position: Position): Position | null {
```
J-Fields: Is there a reason this (and `getRightDelimiter`) can't be private?

src/actions/textobject.ts
```diff
@@ -640,3 +640,211 @@ class InsideIndentObjectBoth extends IndentObjectMatch {
...
+      }
+
+      leftWalkPos = leftWalkPos.getLeftThroughLineBreaks(true);
+    }
+
+    return leftDelimiterPosition;
+  }
+
+  public static getRightDelimiter(position: Position): Position | null {
+    let rightDelimiterPosition: Position | null = null;
```
J-Fields: could you use `undefined` instead of `null`?

src/actions/textobject.ts
```diff
@@ -640,3 +640,211 @@ class InsideIndentObjectBoth extends IndentObjectMatch {
...
+  //     have a matching closed one. This way we can ignore pairs
+  //     of parentheses which are part of the current argument.
+  // 1.2 Vice versa for walking right.
+  // 2   Depending on our mode (inner or around), improve the start/stop
+  //     locations for most consistent behaviour, especially in case of
+  //     multi-line statements.
+
+  public async execAction(position: Position, vimState: VimState): Promise<IMovement> {
+    const failure = { start: position, stop: position, failed: true };
+    let leftSearchStartPosition = new Position(
```
J-Fields: Maybe add a `clone` method to `Position`?

J-Fields:
Re: configuration, how does this proposed config compare to that of the plugins we're trying to emulate?

lmNt:
Thanks for your review, and sorry for not yet continuing to work on it. I will get back to this ASAP (this weekend or beginning of next week).

Regarding configuration: the targets.vim plugin (which I am using) has the options `opening`, `closing` and `separator`. And all of the options have the same defaults as I am proposing for this PR.
To be fully consistent, we could shoot for `vim.argumentObjectSeparators` instead of `vim.argumentObjectDelimiters`. What do you think?

J-Fields:
No trouble. I haven't had a chance to really dig into the changes yet, will try to soon.

Let's go with `vim.argumentObjectSeparators` for consistency.

src/actions/textobject.ts
```diff
@@ -640,3 +640,211 @@ class InsideIndentObjectBoth extends IndentObjectMatch {
...
+      stop = rightDelimiterPosition.getLeftThroughLineBreaks(true);
+    }
+
+    // Handle case when cursor is not inside the anticipated movement range
+    if (position.isBefore(start)) {
+      vimState.recordedState.operatorPositionDiff = start.subtract(position);
+    }
+    vimState.cursorStartPosition = start;
+
+    return {
```
J-Fields: This can be
```typescript
return {
  start,
  stop
}
```

src/actions/textobject.ts
```diff
@@ -640,3 +640,211 @@ class InsideIndentObjectBoth extends IndentObjectMatch {
...
+    return leftDelimiterPosition;
+  }
+
+  public static getRightDelimiter(position: Position): Position | null {
+    let rightDelimiterPosition: Position | null = null;
+    let rightWalkPos = position;
+    let openedParensCount = 0;
+
+    while (true) {
+      let char = TextEditor.getCharAt(rightWalkPos);
```
J-Fields: Use `const` wherever possible, please

skywind3000:
awesome !!

src/actions/textobject.ts
lmNt: Turns out, I don't need to construct a new Position and can just use a regular assignment :)

lmNt:
I still want to improve on the **around** movement to delete trailing whitespace(s) after the argument boundary properly, i.e. so that `daa` includes the following whitespace(s)

```
        this whitespace (or possibly more whitespaces) should be included
        V
(al|pha, beta, gamma)
```
so that we end up with `(|beta, gamma)` instead of `(| beta, gamma)`.

---

Then there is one more thing I would like to discuss, that is kind of language-specific and can be a pain to implement. Currently the movement can be confusing to the user, in case there is a string-like argument that includes one of the separator characters.
For example:

```
(foo, "bar, b|az", qux)
```

The user would expect the result to be `(foo,|, qux)`, however it will be `(foo, "bar,|, qux)`. To handle this, one would need to introduce some kind of group of *ignore* characters. However considering all the different language features, such as triple-quoted strings, it can be difficult to anticipate every use case. Any thoughts?

skywind3000:
@lmNt can we introduce something from [textobj-parameter](https://github.com/sgur/vim-textobj-parameter) ??  it can parse strings correctly.

J-Fields:
@lmNt What's the current status, assuming we forego correct string handling for now?

lmNt:
@J-Fields It's done, except the two (minor) test cases that still fail and maybe a final code-review. 
I have mentioned the two failing tests in the description of the PR as well. Here is the relevant part quoted:

> ##### Failing tests
> 
> Apart from that, this PR will still has 2 failing tests, where I am not quite sure of how to fix them.
> These failing tests are only for _failed_ movements:
> 
>     * "Will fail 'cia' with no delimiters"
> 
>     * "Will fail 'cia' with flipped delimiters"
> 
> 
> For both these tests the cursor position after the movement is offset `+1`, which I don't really understand, since I return `const failure = { start: position, stop: position, failed: true };` where `position` is the initial position. I would expect that returning `failed: true` should not advance the cursor. Can somebody help me out here? (maybe you @J-Fields? :))

I still could not figure out, what causes this behavior. Maybe you can shed some light on this or can point me in a direction?

J-Fields:
Thanks @lmNt, I'd missed that. I'll keep it in mind as I review.

Would you mind rebasing onto master and fixing the conflict when you get a chance?

src/actions/textobject.ts
```diff
@@ -640,3 +641,219 @@ class InsideIndentObjectBoth extends IndentObjectMatch {
...
+    }
+
+    const leftArgumentBoundary = SelectInnerArgument.findLeftArgumentBoundary(
+      leftSearchStartPosition
+    );
+    const rightArgumentBoundary = SelectInnerArgument.findRightArgumentBoundary(
+      rightSearchStartPosition
+    );
+
+    if (leftArgumentBoundary === undefined || rightArgumentBoundary === undefined) {
```
J-Fields, line 706: Check `leftArgumentBoundary === undefined` before you calculate `rightArgumentBoundary` to avoid unnecessary computation

src/actions/textobject.ts
```diff
@@ -640,3 +641,219 @@ class InsideIndentObjectBoth extends IndentObjectMatch {
...
+    return configuration.argumentObjectOpeningDelimiters;
+  }
+  private static closingDelimiterCharacters(): string[] {
+    return configuration.argumentObjectClosingDelimiters;
+  }
+  private static separatorCharacters(): string[] {
+    return configuration.argumentObjectSeparators;
+  }
+
+  protected selectAround = false;
```
J-Fields, line 658: Can you add a comment here describing the difference between "inner" and "around"?

J-Fields:
Okay, I've fixed the last two tests. I'm not 100% confident I didn't break anything by doing so, but the rest of the tests are happy, so after one more glance over this I'll merge and see if anyone complains 🙃 

Thanks a ton for implementing this, @lmNt!

skywind3000:
fantastic !! @J-Fields , when can we get this released ??

lmNt:
@J-Fields Thanks a lot for reviewing and fixing the failing tests :)

While using the textobject argument all day today, I've stumbled across two minor improvements I'd like to get out before this goes live in a new release. I will open a new PR for that and let you know.

J-Fields:
@lmNt Sounds great - any idea when you'll have that PR ready?

lmNt:
I'll start working on it right now. I expect a couple of hours, given I don't run into unexpected edge-cases.

J-Fields:
Sweet, no rush!

J-Fields:
@skywind3000 We'll see, but the most likely scenario is a release this weekend

Issue #4643, opened 2020-03-11
Screen scrolls with cursor if code is folded


siddharth-krishna:
**Describe the bug**
With the Vim extension enabled, if I have a file open where I've folded some block of code (that I can see on screen) and my cursor is below this folded block, then moving my cursor up and down lines also makes the screen move. The screen moves so that the cursor is always in the center. This happens in both insert and visual modes. However, this does not happen if my cursor is above the folded block of code, and it stops when the folded block moves out of view.

This causes the screen to jump around, and is quite distracting.

**To Reproduce**

1.  Open any file with blocks of code that can be folded, and that is long enough. For example, open a new file and copy the following block into it 100 times:
```
test
  test test
  test test test
  test test

```
2.  Scroll to the middle of the file, and then fold one of the blocks.
3.  Now place the cursor below the folded block, and try to move the cursor downwards.
4.  See error.

**Expected behavior**
I expect the cursor to move down, but the screen/view to stay the same. Instead, the screen/view scrolls down, and the cursor remains in the center of the screen.

**Environment**

- Extension (VsCodeVim) version: 1.13.0
- VSCode version: 1.43.0
- OS: Windows 10

anhtukhtn:
I have this issue too. I have to switch to use simple Vim when waiting it's fixed

tuledev:
Downgrade Vim to 1.12.4 fix the issue 
https://github.com/VSCodeVim/Vim/issues/4632

Issue #4632, opened 2020-03-08
Cursor is always in center of page


yar00001:
- Related to [this issue](https://github.com/VSCodeVim/Vim/issues/2960#issuecomment-593027992)

> The above commit (4434) seems to behave unexpectedly around heavily folded code. If I'm traversing lines (simple cursor movements like jk) in a non-folded section of code, the cursor works just fine. When I'm doing the same amidst many folds, cursor navigation behaves almost like a scrollOffset is set (keeping the cursor in the center always).

![ezgif com-video-to-gif](https://user-images.githubusercontent.com/2102983/76155398-655f4e00-60a0-11ea-838f-ee9cebbd85e0.gif)

nightire:
Confirmed. Downgrade to 1.12.4 works again.

PaulTalbot-INL:
Also confirmed, and downgrading fixed it.

J-Fields:
This was caused by https://github.com/VSCodeVim/Vim/pull/4434. I'm not sure if VSCode's API gives us enough information to do this correctly, so we may just want to check `vim.foldFix` and revert to the old behavior if that's enabled.

jordankaiser:
Confirmed the issue. Downgrading to 1.12.4 fixes for me as well.

yar00001:
Thank you Jason works as expected now.

J-Fields:
Thank @DianeLooney! 🙂

PR #4662, opened 2020-03-22
Fix cursor movement with large folded regions.

DianeLooney:
With a folded region larger than 15 lines, when the cursor is below the
folded region, it is always 15 lines below the end of the first region,
so the existing logic will incorrectly tell vim to always re-center the
cursor. Instead, we should check if the cursor is either below all
visible regions, or above all visible regions.

Fixes:

https://github.com/VSCodeVim/Vim/issues/4643
https://github.com/VSCodeVim/Vim/issues/4632

### DIFF
--- a/src/mode/modeHandler.ts
+++ b/src/mode/modeHandler.ts
@@ -1336,11 +1336,7 @@ export class ModeHandler implements vscode.Disposable {
 
       const { visibleRanges } = vimState.editor;
       const centerViewportAroundCursor =
-        visibleRanges.reduce(
-          (acc, visibleRange) => acc && isCursorAboveRange(visibleRange),
-          true
-        ) ||
-        visibleRanges.reduce((acc, visibleRange) => acc && isCursorBelowRange(visibleRange), true);
+        visibleRanges.every(isCursorAboveRange) || visibleRanges.every(isCursorBelowRange);
 
       const revealType = centerViewportAroundCursor
         ? vscode.TextEditorRevealType.InCenter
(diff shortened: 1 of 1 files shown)


DianeLooney:
An example file for testing this change: https://gist.github.com/DianeLooney/679af92cf6d2027e56d3f58f606eb347

Repro steps: Enable scrolling, move to line 34, hold down `j` or `down arrow` to observe it incorrectly snapping to the cursor.

src/mode/modeHandler.ts
```diff
@@ -1329,10 +1329,19 @@ export class ModeHandler implements vscode.Disposable {
...
-      const visibleRange = vimState.editor.visibleRanges[0];
-      const centerViewportAroundCursor =
-        visibleRange.start.line - vimState.cursorStopPosition.line >= 15 ||
+      const isCursorAboveRange = (visibleRange: vscode.Range): boolean =>
+        visibleRange.start.line - vimState.cursorStopPosition.line >= 15;
+      const isCursorBelowRange = (visibleRange: vscode.Range): boolean =>
         vimState.cursorStopPosition.line - visibleRange.end.line >= 15;
+
+      const { visibleRanges } = vimState.editor;
+      const centerViewportAroundCursor =
```
J-Fields: I think you can express this more cleanly with [every](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every) than `reduce`

DianeLooney: You're right, will update.

Issue #2207, opened 2017-12-08
Joining multiple lines with `gJ` deletes all lines past 2nd line


jsermeno:
* Click *thumbs-up* 👍 on this issue if you want it!
* Click *confused* 😕 on this issue if not having it makes VSCodeVim unusable. 

The VSCodeVim team prioritizes issues based on reaction count. 

--------

**Is this a BUG REPORT or FEATURE REQUEST?** (choose one):

**Environment**: macOS Sierra 10.12.6

- **VSCode Version**: 1.18.1
- **VsCodeVim Version**: 0.10.5
- **OS**: macOS Sierra 10.12.6

**What happened**: Joining multiple lines with `gJ` deletes lines all but the first two lines. For example, the input
```
a
b
c
d
e
```
with the keys `5gJ` will produce the output `ab`.

**What did you expect to happen**: For the input
```
a
b
c
d
e
```
and the keys `5gJ` I expected the output `abcde`.

**How to reproduce it**: An empty file with the input above and the keys above.

chris-morgan:
This is still an issue, and is not the only problem with `gJ` with count: it also operates upon one too many lines, e.g. `2gJ` should be the same as `1gJ` or `gJ`, but in actual fact causes line three to disappear.

PR #4663, opened 2020-03-22
Support `J` in visual block mode

lusingander:

`J` (join lines) works in visual block mode in original Vim, but this is not supported this extension.

expected:

image

If `J` is executed in the state of this image, expected to be:

```
one two three
four
```

- Cursor position after execution is different from original Vim, but this is the same behavior as visual mode and visual line mode. So I think this should be fixed as another Pull Request.
  - In the above example, cursor should be before `three` after execution, but actually it is before `two`.
- `gJ` doesn't work either, but doesn't fix it. Because `gJ` has a bug like #2207 and it doesn't work properly.

And if necessary, I'd like to fix and create another PR for these as well 👻

### DIFF
--- a/src/actions/commands/actions.ts
+++ b/src/actions/commands/actions.ts
@@ -3282,13 +3282,11 @@ class ActionJoinVisualBlockMode extends BaseCommand {
   mightChangeDocument = true;
 
   public async exec(position: Position, vimState: VimState): Promise<VimState> {
-    let actionJoin = new ActionJoin();
-    const range = vimState.cursors[0];
-    const topLeft = visualBlockGetTopLeftPosition(range.start, range.stop);
-    const bottomRight = visualBlockGetBottomRightPosition(range.start, range.stop);
+    const start = vimState.cursorStartPosition;
+    const end = vimState.cursorStopPosition;
 
     vimState.currentRegisterMode = RegisterMode.CharacterWise;
-    vimState = await actionJoin.execJoinLines(topLeft, bottomRight, vimState, 1);
+    vimState = await new ActionJoin().execJoinLines(start, end, vimState, 1);
     return vimState;
   }
 }
--- a/src/actions/commands/actions.ts
+++ b/src/actions/commands/actions.ts
@@ -3282,11 +3282,13 @@ class ActionJoinVisualBlockMode extends BaseCommand {
   mightChangeDocument = true;
 
   public async exec(position: Position, vimState: VimState): Promise<VimState> {
-    const start = vimState.cursorStartPosition;
-    const end = vimState.cursorStopPosition;
-
     vimState.currentRegisterMode = RegisterMode.CharacterWise;
-    vimState = await new ActionJoin().execJoinLines(start, end, vimState, 1);
+    vimState = await new ActionJoin().execJoinLines(
+      vimState.cursorStartPosition,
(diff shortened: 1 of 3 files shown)


src/actions/commands/actions.ts
```diff
@@ -3270,6 +3275,24 @@ class ActionJoinVisualMode extends BaseCommand {
...
+class ActionJoinVisualBlockMode extends BaseCommand {
+  modes = [Mode.VisualBlock];
+  keys = ['J'];
+  mightChangeDocument = true;
+
+  public async exec(position: Position, vimState: VimState): Promise<VimState> {
+    let actionJoin = new ActionJoin();
+    const range = vimState.cursors[0];
+    const topLeft = visualBlockGetTopLeftPosition(range.start, range.stop);
+    const bottomRight = visualBlockGetBottomRightPosition(range.start, range.stop);
```
J-Fields: Could you use `vimState.cursorStartPosition` and `vimState.cursorStopPosition` instead?

src/actions/commands/actions.ts
```diff
@@ -3270,6 +3275,24 @@ class ActionJoinVisualMode extends BaseCommand {
...
+
+    vimState.currentRegisterMode = RegisterMode.CharacterWise;
+    vimState = await actionJoin.execJoinLines(topLeft, bottomRight, vimState, 1);
```
J-Fields: Just construct the `ActionJoin` inline `await new ActionJoin().execJoinLines(...);`

test/mode/modeVisualBlock.test.ts
```diff
@@ -188,4 +188,12 @@ suite('Mode Visual Block', () => {
     keysPressed: '<C-v>llj"ayGo<C-r>a<Esc>',
     end: ['abcde', '01234', 'abcde', '01234', '123', 'bcd', '|'],
   });
+
+  newTest({
```
J-Fields: Can you write a test where the entire visual block is on the same line? It's a bit of a special case.

J-Fields:
> I'd like to fix and create another PR for these as well

That'd be fantastic, thanks 🙂

lusingander:
@J-Fields thanks for the confirmation, I fixed these!

src/actions/commands/actions.ts
```diff
@@ -3270,6 +3275,22 @@ class ActionJoinVisualMode extends BaseCommand {
...
+  modes = [Mode.VisualBlock];
+  keys = ['J'];
+  mightChangeDocument = true;
+
+  public async exec(position: Position, vimState: VimState): Promise<VimState> {
+    const start = vimState.cursorStartPosition;
+    const end = vimState.cursorStopPosition;
+
+    vimState.currentRegisterMode = RegisterMode.CharacterWise;
+    vimState = await new ActionJoin().execJoinLines(start, end, vimState, 1);
```
J-Fields: Could you condense this like so: `return new ActionJoin().execJoinLines(vimState.cursorStartPosition, vimState.cursorStopPosition, vimState, 1);`?

Sorry to nitpick again; really appreciate the contributions! 🙂

lusingander: No problem, I think this is good, thank you 😀

J-Fields:
Thanks again!

Issue #4513, opened 2020-01-23
<Ctrl-W> o works but isn't documented?


max-sixty:
The `<Ctrl-W> o` chord seems to work, closing all other editor groups. But I didn't see it in the documentation? Is this right? Are there other <Ctrl-W> chords that work but aren't documented? Happy to help add if so.

J-Fields:
I think you're right - a PR would be very welcome!

PR #4668, opened 2020-03-22
Add ctrl-w o to Roadmap

max-sixty:

Fixes https://github.com/VSCodeVim/Vim/issues/4513


max-sixty:
Separately, would you be open to renaming "Roadmap" to something like "Mappings & Roadmap"? It's the best (only?) source of keymappings in the docs. Having a single explicit place with all key mappings would be great for new users.

J-Fields:
Thanks!

> Separately, would you be open to renaming "Roadmap" to something like "Mappings & Roadmap"? It's the best (only?) source of keymappings in the docs. Having a single explicit place with all key mappings would be great for new users.

Yeah, possibly. Maybe the README language needs to make it clearer that the roadmap includes all the supported mappings (or attempts to, any way)

ROADMAP.md
```diff
@@ -524,6 +524,7 @@ Since the list is too long, now we just put those already supported options here
 | :white_check_mark: :star: | :vsp {file}       | Split vertically current window in two.                                 |                                                                                                                        |
 | :white_check_mark:        | &lt;ctrl-w&gt; s  | Split current window in two.                                            |                                                                                                                        |
 | :white_check_mark: :star: | &lt;ctrl-w&gt; v  | Split vertically current window in two.                                 |                                                                                                                        |
+| :white_check_mark: :star: | &lt;ctrl-w&gt; o  | Close other editor groups                                               |                                                                                                                        |
```
J-Fields: Was about to merge, then realized that the other descriptions end in a period. Hate to be that guy, but could you add one?

max-sixty:
```suggestion
| :white_check_mark: :star: | &lt;ctrl-w&gt; o  | Close other editor groups.                                              |                                                                                                                        |
```

max-sixty: Do you squash on merge or do you need it on the same commit?

J-Fields: Squash on merge, don't worry about it. Thanks 🙂

PR #4681, opened 2020-03-26
Optimize HistoryTracker using a cached TextDocument.version

### LINKED ISSUE #4665: Consider using `TextEditor.document.version` rather than `mightChangeDocument`, etc.
Reading through the code in `SearchState._recalculateSearchRanges`, something caught my eye. We're using [TextDocument.version](https://code.visualstudio.com/api/references/vscode-api#TextDocument) to decide whether we need to recalculate the search ranges. Couldn't we just use this instead of messing around with `mightChangeDocument`? (I was unaware that this information was provided by VSCode's API until today). This might give a performance boost (as `mightChangeDocument` can naturally give false positives) while also reducing complexity.

Would love input from @cvaldev 

Relates to #4451

J-Fields:
This replaces `mightChangeDocument`, which sought to do the same sort of optimization in a less effective and far more clumsy way.

Fixes #4665

### DIFF
--- a/package.json
+++ b/package.json
@@ -918,11 +918,6 @@
           "type": "boolean",
           "markdownDescription": "When `true` the commands listed below move the cursor to the first non-blank of the line.  When `false` the cursor is kept in the same column (if possible).  This applies to the commands: `<C-d>`, `<C-u>`, `<C-b>`, `<C-f>`, `G`, `H`, `M`, `L`, `gg`, and to the commands `d`, `<<` and `>>` with a linewise operator.",
           "default": true
-        },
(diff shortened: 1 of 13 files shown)


J-Fields:
This is **much** better than `mightChangeDocument`. In addition to being far less cumbersome and error prone, and not generating any "false positives", it also works [here](https://github.com/VSCodeVim/Vim/blob/historytracker-version/src/mode/modeHandler.ts#L352).

I'm playing around with an ~80,000 line file, and navigation is pretty much buttery smooth.

PR #4690, opened 2020-03-27
Another crack at bundling the extension with webpack

### LINKED ISSUE #3127: Add Webpack Bundling

J-Fields:
This broke the extension last time we tried 🙂
If anyone could try this out and let me know if the extension still works, that would be super cool
Fixes #3127


J-Fields:
Lacking other testers, I'm going to merge this, mark a new release, and see what happens. It's the weekend, so risk is minimal, and despite the fact that we haven't gotten all I intended to done for this release, there hasn't been one in a while, and I think the improvements made already are substantial.

PR #4691, opened 2020-03-27
Fix a bug of `J` command in visual block mode

lusingander:

Sorry, my PR (#4663) contained a bug 🙁

`J` doesn't work properly when the start position of the visual block is below the end position.

e.g.

At first the cursor is on the `|`.
```
one
two
t|hree
four
```

If execute `<C-v>kkJ`, of course expected result is as follows:

```
one two three
four
```

but not.


J-Fields:
Nice catch, thanks!

PR #4695, opened 2020-03-28
Remove old workaround that is no longer needed

### LINKED ISSUE #1962: Multi-cursor append line above (Shift+O) only works for 1 cursor

--------

**Environment**:

- **VSCode Version**: 1.14.2
- **OS**: Windows 10

**What happened**:

1) be in normal mode
2) using Alt+leftclick, create cursors at multiple lines
3) press Shift+O
4) a new line is appended only above the first cursor

4) new lines are appended above all cursors

(**note:** it already works correctly when appending below with O instead of above with Shift+O)

**How to reproduce it**:

see What happened

cvaldev:
The issue this workaround used to solve appears to have been fixed in the upstream and is no longer reproducible. Additionally, because it would update the `activeTextEditor.selection` it caused vscode to loose our selections in multicursor mode, thus `editor.action.insertLineBefore` would only insert a line above the first cursor and subsequently take us out of multicursor mode.

fixes #1962 

A bit unrelated but I'm unsure if I should open an issue for this, I normally do format on save but since prettier was added it re-formats the whole document and would have made this PR a mess. Should I push another commit with the prettier formatting?

### DIFF
--- a/src/actions/commands/actions.ts
+++ b/src/actions/commands/actions.ts
@@ -2635,15 +2635,7 @@ class CommandInsertNewLineAbove extends BaseCommand {
   public async execCount(position: Position, vimState: VimState): Promise<VimState> {
     await vimState.setCurrentMode(Mode.Insert);
     let count = vimState.recordedState.count || 1;
-    // Why do we do this? Who fucking knows??? If the cursor is at the
-    // beginning of the line, then editor.action.insertLineBefore does some
-    // weird things with following paste command. Refer to
-    // https://github.com/VSCodeVim/Vim/pull/1663#issuecomment-300573129 for
-    // more details.
-    const tPos = Position.FromVSCodePosition(
-      vscode.window.activeTextEditor!.selection.active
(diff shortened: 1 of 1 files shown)


J-Fields:
> A bit unrelated but I'm unsure if I should open an issue for this, I normally do format on save but since prettier was added it re-formats the whole document and would have made this PR a mess. Should I push another commit with the prettier formatting?

Is it the line endings that are being changed?

cvaldev:
@J-Fields Sorry I didn't look at what it was doing initially, there are no changes to line endings as far as I can tell but it's removing all the trailing commas.

J-Fields:
Ah! `npm install` should resolve that, I think. The default for `trailingComma` changed from `none` to `es5` in prettier 2.0, so I removed it from our `.prettierrc` (we had it set to `es5` before)

cvaldev:
> Ah! `npm install` should resolve that, I think. 

Thanks! that did it.

PR #4702, opened 2020-03-30
Fix cursor position after executing `J` in visual mode

lusingander:
The cursor position after executing `J` command in visual (line / block) mode is different from expected.

**e.g.**

In the following state, for example,

```
o|ne
two
three
four
five
```

now, if execute `vjjjJ`, we will get the following state in the original Vim.

```
one two three| four
five
```
(The cursor should be in the space in front of the last word.)

but now, it's like this:

```
one| two three four
five
```

the same is true for visual line mode and visual block mode.

Same fix is not made for `gJ` because `[count]gJ` doesn't work properly... (#2207)

### DIFF
--- a/src/actions/commands/actions.ts
+++ b/src/actions/commands/actions.ts
@@ -3158,7 +3158,8 @@ class ActionJoin extends BaseCommand {
           manuallySetCursorPositions: true,
         });
 
-        vimState.cursorStartPosition = vimState.cursorStopPosition = startPosition.withColumn(
+        vimState.cursorStartPosition = vimState.cursorStopPosition = new Position(
+          startPosition.line,
           trimmedLinesContent.length - columnDeltaOffset
         );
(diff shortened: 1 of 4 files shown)


src/actions/commands/actions.ts
```diff
@@ -3158,7 +3158,8 @@ class ActionJoin extends BaseCommand {
           manuallySetCursorPositions: true,
         });
 
-        vimState.cursorStartPosition = vimState.cursorStopPosition = startPosition.withColumn(
```
J-Fields: Could you explain what's changing here?

lusingander: In `withColumn`, the length is rounded to the length of the original text in the `startPosition`, so the value of `trimmedLinesContent.length - columnDeltaOffset` is not used correctly.

Issue #4710, opened 2020-04-01
Display vim markers


gamcoh:
**Is your feature request related to a problem? Please describe.**
My request is not related to a problem, it's something i'd like to use.

**Describe the solution you'd like**
I want to see icons in the gutter where my vim marks are set. Like [Bookmarks](https://marketplace.visualstudio.com/items?itemName=alefragnani.Bookmarks) but for vim markers.

**Describe alternatives you've considered**
A clear alternative is to use the Bookmarks extension but I want to use the markers of vim.

**Additional context**
I also would like to participate, so I am beginning now to develop this feature.

J-Fields:
Duplicate of #3963. Thanks for offering to tackle this; feel free to bug me if you need any guidance!

gamcoh:
Hi, I just created my PR but the CI tells me : `TypeError: vscode.window.createTextEditorDecorationType is not a function`.
Do you have any idea why it works on my machine but not in the CI?

gyzerok:
Hello @J-Fields!

In the changelog to version 1.14.0 it mentions this issue within enhancements. However based on actual behavior, this issue and related issue it seems that nothing was implemented.

Can you help me understand it? I am a bit confused :)

J-Fields:
@gyzerok Sorry for the confusion. The changelog is generated automatically based on the issues/PRs that are closed. This was closed as a duplicate. There is a pending PR for it though, so hopefully it'll be in the next feature release.

gyzerok:
Sure, thanks for explaining!

gamcoh:
Yes, by the way @J-Fields, the changes you requested on the PR are done. everything works well, just waiting for your approval.

J-Fields:
@gamcoh Yep, I will soon. Just been focusing on some other things recently 🙂

gamcoh:
Of course, take your time!

PR #4715, opened 2020-04-02
Added bookmark for vim native marks in the gutter as decoration

gamcoh:
This PR puts bookmarks icons in the gutter when you use vim native marks

#4710 

I used `path` in order to get the icon because I didn't find any other alternatives.


src/history/historyTracker.ts
```diff
@@ -441,6 +442,24 @@ export class HistoryTracker {
...
+  /**
+   * Update marks decorations in the gutter
+   */
+  private updateMarksDecorators(): void {
+    if (!this.vimState.editor) {
+      return;
+    }
+
+    let decorators: vscode.Range[] = [];
+    for (const mark of this.getMarks()) {
```
J-Fields, line 457: nit: can you change this to use `map` rather than a for loop?

src/history/historyTracker.ts
```diff
@@ -441,6 +442,24 @@ export class HistoryTracker {
       editor: isUppercaseMark ? vscode.window.activeTextEditor : undefined,
     };
     this.putMarkInList(newMark);
+    this.updateMarksDecorators();
```
J-Fields, line 445: Let's separate the data updating from the view updating, so that if the marks are updated elsewhere (`:delmarks`, for instance), we don't have to remember to update the view.
https://github.com/VSCodeVim/Vim/blob/3d9718d70debe881751ddf37832995b1100c1ee0/src/mode/modeHandler.ts#L1347

src/configuration/decoration.ts
```diff
@@ -1,11 +1,21 @@
...
+import * as path from 'path';
 import { IConfiguration } from './iconfiguration';
 
 class DecorationImpl {
+  private _mark: vscode.TextEditorDecorationType;
   private _default: vscode.TextEditorDecorationType;
   private _searchHighlight: vscode.TextEditorDecorationType;
   private _easyMotion: vscode.TextEditorDecorationType;
 
+  public set Mark(value: vscode.TextEditorDecorationType) {
```
J-Fields, line 11: Lowercase for new methods, please. Not sure why there's some inconsistency, I'll go through the code base and clean that up some time.

src/history/historyTracker.ts
gamcoh: done

src/history/historyTracker.ts
gamcoh: done

src/configuration/decoration.ts
gamcoh: done

J-Fields:
Closing in favor of #5924. Thanks a lot for providing a starting point!

Issue #4716, opened 2020-04-03
VisualLine mode leaving selection visible after changing to normal mode


berknam:
**Describe the bug**
With the new refactors to the modeHandler.updateView if you go into VisualLine mode, then escape to normal mode, then reselect with 'gv' and escape again, it goes to normal mode but leaves the selection visible.

**To Reproduce**

1.  Go to VisualLine mode by pressing 'V'
2.  Press 'Esc' to go back to normal mode
3.  Press 'gv' to reselect previous selection
4.  Press 'Esc' to go back to normal mode
5. It goes to normal mode but leaves selection visible

**Expected behavior**
Expected the selection to disappear after going back to normal mode.

**Environment (please complete the following information):**

- Extension (VsCodeVim) version: latest commits
- VSCode version: 1.43.2
- OS: Windows 10

**Additional context**
This also happens when performing an easy motion in VisualLine mode. If we go to visualLine mode use easymotion to move and then escape, it leaves the selection visible

PR #4717, opened 2020-04-03
Fix VisualLineMode leaving selection after escape

berknam:
With the new refactors to the modeHandler.updateView if you go into VisualLine mode, then escape to normal mode, then reselect with 'gv' and escape again, it goes to normal mode but leaves the selection visible, because the 'visualLineStartColumn' is undefined.

Fixes issue #4716


src/actions/plugins/easymotion/easymotion.cmd.ts
```diff
@@ -79,6 +79,17 @@ abstract class BaseEasyMotionCommand extends BaseCommand {
       // Search all occurences of the character pressed
       const matches = this.getMatches(position, vimState);
 
+      // If previous mode was visual, restore visual selection
+      if (
```
J-Fields, line 83: Can you use `isVisualMode` here? (Mode.ts)

J-Fields:
LGTM, other than small nitpick

src/actions/plugins/easymotion/easymotion.cmd.ts
berknam: Yes already changed it.

PR #4721, opened 2020-04-06
Tag motion now considers the tag starting on the cursor line

### LINKED ISSUE #4685: Copy the current line tag and its children with the command yat
 I'm always frustrated when I want yank the entire html tag typing `yat` but instead copy the tag that starts on the current line and its children tags, it copy the entire parent html tag.

type `yat` and copy the tag within the current line

```html
<div class="dont-copy-parent">

<!--  yat --> | <div class="want-copy-this">
                  <div>Content</div>
             </div>

</div>
```

Vim and neoVIM do this job

gergelyth:

When executing `yat` or `dat` on a line where a new tag is opened (and there are only whitespaces between the beginning of the line and the opening `<` of the tag and the cursor is either on one of these whitespaces or on the opening `<`), the current version of the extension copies or deletes the outer tag, whereas Vim operates on the inner. This PR corrects the extension's behaviour.

**Example:**
```html
<foo>
 |   <bar> test </bar>
</foo>   
```
Executing `dat` at the given position now results in:
```html
<foo>
</foo>
```

Fixes #4685


src/common/matching/tagMatcher.ts
```diff
@@ -104,6 +132,27 @@ export class TagMatcher {
...
+  /**
+   * Most of the time the relevant tag is the innermost tag, but when Visual mode is active,
+   * the rules are different.
+   * When the cursorStart is standing on the < character of the inner tag, with "at" we must
+   * jump to the outer tag.
+   */
+  determineRelevantTag(
+    tagsSurrounding: MatchedTag[],
+    adjustedStartPosOffset: number,
+    selectionActive: boolean
```
J-Fields, line 144: Is there a reason for using this `selectionActive` flag rather than checking the mode directly?

gergelyth: Hm..you mean just checking whether `vimState.currentMode === Mode.Visual`? If so, the reason is that in Vim if you enter visual mode but don't select anything (i.e. `startCursorPosition === endCursorPosition`), the `at` motion will return the inner tag. If we were to compare modes here instead of cursor positions, `vat` would select the outer tag because of the mode. I apologize, I should have added a comment to specify that edge case.

PR #4725, opened 2020-04-08
Fix: cmd+D now jumps and scrolls window to the last selected word

### LINKED ISSUE #880: cmd + D selection don't jump and scroll window to latest selected text
Please _thumbs-up_ 👍 this issue if it personally affects you! You can do this by clicking on the emoji-face on the top right of this post. Issues with more thumbs-up will be prioritized.

---
### What did you do?

Press cmd + d multiple times to select same words

When the word are not in current view, it should jump and scroll window to the view of the newly selected word.
### What happened instead?

It flashed the view to the selected word but jump back to the original one.
### Technical details:
- VSCode Version: 1.6.0
- VsCodeVim Version: 0.3.5
- OS: macOS Sierra

gergelyth:

There is an issue in the current version, where if you search and select with cmd+D, if the next match is out of view of the current window, the window is flashed to that next match but then it immediately returns to the first result. This PR fixes this issue.

The problem was that while the action inserts multiple cursos into the workspace, the `updateView()` function always tracks the first one (there are also 2 functions which call the `updateView()` right after each other: `handleKeyEventHelper()` and `handleSelectionChange()`, but I got scared of touching the latter function just by reading the comment above it, so I tinkered with `updateView()`).

I'm aware that this rather a patch than a true fix and is utterly inelegant (it checks whether the last action was cmd+D and reacts accordingly - maybe if other actions are found, which would benefit from this solution, we can expand this condition), so I'm open to suggestions. The flipside of this is that the patch is isolated only for this action, so it shouldn't break any previously correct behaviour.

Fixes #880


J-Fields:
Thanks @gergelyth! I'm going to merge this after making a very slight change, because it definitely improves matters. I wonder if it's desirable for the view to jump back to the first cursor after a different action is done - I'll put a comment asking for feedback in the original issue.

Issue #4733, opened 2020-04-12
.vimrc: Cannot map key to a command of an extension


mkrawiec:
**Describe the bug**
The functionality of mapping a key to a command in `.vimrc` works only for builtin commands. When an action is provided by a plugin, instead of executing it, the command is inserted literally. 

**To Reproduce**
This mapping will work:
    
    nnoremap <leader>sa workbench.action.findInFiles

This will not work:

    nnoremap <leader>conf vim.editVimrc
    nnoremap <leader>ren fileutils.renameFile

**Expected behavior**
User should be able to map any action to a key

**Environment (please complete the following information):**

- Extension (VsCodeVim) version: 1.13.1
- VSCode version: 1.44
- OS: Linux

sqlkoala:
Hi, you could create the keybindings you want by adapting 4th example here:
https://github.com/VSCodeVim/Vim/blob/master/README.md#key-remapping
(I know this is more effort than the new, convenient style with .vimrc).

If you look at your example, there is now way to syntactically distinguish the right side "fileutils.renameFile" from a series of vim keystrokes/commands.

PR #4750, opened 2020-04-18
Add support for mapping extension commands from vimrc

mkrawiec:

This adds possibility of adding `.vimrc` mappings to commands of installed extensions ie.

    nnoremap <leader>fn fileutils.newFile

https://github.com/VSCodeVim/Vim/issues/4733

First commit :)


J-Fields:
Cool, thanks! We should update this variable any time an extension is installed/uninstalled, which we can detect with `vscode.extensions.onDidChange`

J-Fields:
Closed in favor of #6694. I think that's the direction we should go, probably deprecating and eventually removing support for "direct" remaps to VSCode commands. Thank you for the contribution!

Issue #4760, opened 2020-04-21
Add setting to change the editor current line background color


marianomms:
**Is your feature request related to a problem? Please describe.**
In VIM I use a different background color in the current line to know what is the VIM mode.

**Describe the solution you'd like**
I would like include something similar in vscode. For instance in current line background color (setting in vscode: `editor.lineHighlightBackground`) we could use the same color than defined in `statusBarColorControl` options.

**Describe alternatives you've considered**
I've used the `statusBarColorControl` setting but for me this is not enough.

**Additional context**
`.vimrc`
```
set cursorcolumn          "Highlight current column
set cursorline            "Highlight current row

" custom colors
function SetLinesDefault()
  hi CursorLine   cterm=NONE ctermbg=88
  hi CursorColumn cterm=NONE ctermbg=88
endfunction
function SetLinesHighlight()
  hi CursorLine   cterm=NONE ctermbg=28 ctermfg=15
  hi CursorColumn cterm=NONE ctermbg=28 ctermfg=15
endfunction

" hi clear CursorColumn
call SetLinesDefault()
autocmd InsertEnter * call SetLinesHighlight()
autocmd InsertLeave * call SetLinesDefault()

" Only highlight current window
augroup CursorLine
  au!
  au VimEnter,WinEnter,BufWinEnter * setlocal cursorline
  au VimEnter,WinEnter,BufWinEnter * setlocal cursorcolumn
  au WinLeave * setlocal nocursorline
  au WinLeave * setlocal nocursorcolumn
augroup END
```

![vim_currentline_color](https://user-images.githubusercontent.com/6368004/79859299-11e75c00-83d1-11ea-9190-5c9da927d698.gif)

PR #4761, opened 2020-04-21
Adding vim.lineHighlightColorControl setting

marianomms:
In VIM I use a different background color in the current line to know what is the VIM mode.

I would like include something similar in vscode. For instance in current line background color (setting in vscode: `editor.lineHighlightBackground`) we could use the same color than defined in `statusBarColorControl` options.

UI in vscode:
vscode_line

fixes https://github.com/VSCodeVim/Vim/issues/4760


J-Fields:
Really sorry for not getting around to this sooner. IMO `statusBarColorControl` was a mistake and at this time I'm reluctant to extend it.

May reconsider this in the future, but at the moment my focus is on consolidating and clearing out old PRs. Thanks for the contribution!

PR #4819, opened 2020-05-08
Feature/vimrc mappings

vtsimbos:
[package.txt](https://github.com/VSCodeVim/Vim/files/4601047/package.txt)

By default vim extension, doesn't recognise <shift+space> keystroke. We need to add

{"key": "shift+space","command": "extension.vim_shift+space", "when": "editorTextFocus && vim.active && vim.use<shift+space> && !inDebugRepl"},

to .vscode\extensions\vscodevim.vim-1.8.0\package.json, to do so.

When the extension gets updated, this gets deleted, and when i press <shift+space> in vscode it says "command 'extension.vim_shift+space' not found". We need to add again the code above.
So I think this command must included in default package.json


J-Fields:
@vtsimbos Could you rebase onto the latest master, please?

PR #4469, opened 2020-01-04
Refactor <C-a> logic

ldm0:
**The reason**
I've noticed the code about `<C-a>` and `<C-x>` is very fragile.
So I refactored it.

**What this PR achieves**
This PR totally fixes #4308, fixes #4457 and more.
Please check newly-added tests to get part of what this PR achieves.
These tests **all fails** without this PR, all of them are the correct behavior of original vim.
Now it's behavior on `<C-a>` and `<C-x>` is 99% similiar to the oiginal Vim.

**To the maintainer**:
PLZ don't ignore my pull request again... :-/
If something went wrong, at least give me a hint.

PR #4825, opened 2020-05-10
Fix numeric length on Ctrl A

dihak:
This will fix the numeric length

Example :
A01 will become A02,

in the previous version the number 0 is missing, it will be A2 instead of A02


J-Fields:
Thanks @dihak! This behavior was also fixed by #4469, so I'm closing this PR

PR #4832, opened 2020-05-11
Implement put ex command

### LINKED ISSUE #4741: add :pu[t] command
I would like to be able to use the ```:pu``` command for 'putting under'. See the vim manual (```:help p```) or this SO issue:

https://stackoverflow.com/questions/1346737/how-to-paste-in-a-new-line-with-vim

tagniam:
This PR ports the `:put` and `:put!` ex commands from Vim, as described by the [official documentation](http://vimdoc.sourceforge.net/htmldoc/change.html#:put). The command is just a way to put text on the line above or under the cursor (always linewise).

Key things this PR adds/modifies:
* Refactors `PutCommand` to take in an object of boolean flags, instead of a bunch of boolean parameters
* Adds two new `PutCommand` flags to force linewise put and force cursor to move to the last line of pasted text
* Creates a new ex command `PutExCommand` with full testing suite

Fixes #4741


src/cmd_line/subparser.ts
```diff
@@ -225,6 +226,11 @@ export const commandParsers = {
     abbrev: 'xa',
     parser: parseWriteQuitAllCommandArgs,
   },
+
+  put: {
```
J-Fields, line 230: This was recently changed to be sorted - could you put this command in the right place please?

tagniam: Fixed!

J-Fields:
`:3855put` gives me

```
CommandLine: Error executing cmd=3855put. err=TypeError: Cannot read property 'type' of undefined.
```

J-Fields:
Nice, thanks. I'll merge pending green tests

J-Fields:
Thanks!

PR #4735, opened 2020-04-13
Refactor Remapper

berknam:
**What this PR does / why we need it**:

Refactor the Remapper and ModeHandler to allow better remapping experience.
It will allow to remap operator keys, motion keys and multiple keys when the first key could be handled.
It also implements the timeoutlen and OperatorPendingMode.

**Which issue(s) this PR fixes**

Should fix the following issues (maybe more):
fixes #5125
fixes #5084
fixes #5067
fixes #5057
fixes #5016
fixes #4991
fixes #4928
fixes #4883
fixes #4756
fixes #4674
fixes #4563
fixes #4353
fixes #4532
fixes #4530
fixes #4464
fixes #4236
fixes #4118
fixes #4057
fixes #3988
fixes #3373
fixes #3768
fixes #3742
fixes #3413
fixes #3171
fixes #3086 (except for the SelectMode, maybe a new issue with this feature request should be opened)
fixes #3082
fixes #2975
fixes #2955
fixes #2897 (although this one will still be missing the language mapping that is in Todo)
fixes #2466
fixes #2234
fixes #2041
fixes #1883
fixes #1870
fixes #1835
fixes #1821
fixes #1579
fixes #1398
fixes #1261

There was an issue with timings after 'v' visual mode, that I don't know exactly what caused but this PR fixed it. Like when you pressed 'v' to enter visual mode and then pressed a remapped key too quickly it wouldn't work.

There is also a lot of old issues (back when there still was a 'otherModesKeyBindings') that would probably be fixed by this (I think some of them are already fixed for a long time but no one closed them).

I'm still looking up the issues list when I have time and I will continue to add to this list.

**Special notes for your reviewer**:

~~Needs more testing. I don't really know how we should implement tests for the timeout part of the remap.~~

I've created a new way of making tests with remaps that allows to create a few tests all with the same remaps by using an array of steps with the keypresses and end result of each step.

If someone could test this out and give me some feedback I would really appreciate. I think this refactor could solve alot of issues and also brings VSCodeVim closer to Vim when it comes to remapping experience.

**Todo**:

- [x] Allow .vimrc mappings with `<silent>, <nowait>, <buffer> and <special>` arguments. Currently the arguments are ignored, but the mapping is done. I think the only arguments that make sense for us is the `<nowait>` and maybe the `<silent>` but I don't think this is needed for now, if someone wants it they can open an issue and then we'll see. The mappings with `<expr> or <Plug>` will not be allowed. (Refs #463 )
[...]

PR #4866, opened 2020-05-16
Change count handling on remapping, vscode commands and operators

### LINKED ISSUE #2244: Multiple insert with <Esc> keybinding only inserts once

--------

**Environment**:

- **VSCode Version**: 1.19.1
- **VsCodeVim Version**: 0.10.6
- **OS**: macOS 10.12.6

**What happened**:

I have the key sequence jk mapped to escape in insert mode, but this causes multiple inserts to only insert once when I exit insert mode using jk. For example, the key sequence `3i=jk` on an empty document produces only:
```
=
```

In vim, and using the plain old escape key, `3i=<Esc>` on an empty document produces:
```
===
```

**How to reproduce it**:

Bind jk to `<Esc>` in insert mode:

```json
    {
        "before": ["j", "k"],
        "after": ["<Esc>"]
    }
],
```

### LINKED ISSUE #4903: 'gj' and 'gk' when mapped move lines one by one instead of immediately
**Describe the bug**
I'm creating this issue in case someone sees in the changelog that this was fixed by this [commit](https://github.com/VSCodeVim/Vim/commit/756a7583321e902b406773c5772d7099c50bf270) (Refs #4162) but they still see the same behavior happening.
 @J-Fields Your commit fixed this when using `gj` and `gk` manually. But this still has the same behavior (of changing line one by one) when using a remapping like `j -> gj` because the remapping is calling each `gj` in a *for* loop, which is wrong and should be fixed by the PR #4866 when merged.

**To Reproduce**

1.  Map `j -> gj`
```
        {
            "before": ["j"],
            "after": ["g", "j"]
        }
]
```
2.  Press `10j`
3.  Cursor moves one line at a time instead of moving immediately to 10 lines below.

**Expected behavior**
Cursor moving immediately to 10 screen lines below

berknam:
Change Count handling when remapping:
- The count value is already stored on RecordedState so we only need to
send the remap keys and the corresponding actions will act accordingly
- The commands weren't respecting the count, if we mapped '<' in visual
mode to the command 'editor.action.outdentLines' we couldn't use it with
count, now we can.
- If you pressed '2d3w' it would delete 23 words, when it should multiply both counts and only delete 6 words. This PR implements an `'operatorCount'` to keep track of the count before the operator to be able to multiply it to the count numbers after it.

fixes #2244
fixes #4903

I've only added a simple test. But I'm trying to create a new simplified way of handling tests with remaps on the PR #4735, after that gets implemented it should be easier to create test with remaps like this one.

### DIFF
--- a/src/actions/commands/actions.ts
+++ b/src/actions/commands/actions.ts
@@ -308,8 +308,31 @@ export class CommandNumber extends BaseCommand {
 
   public async exec(position: Position, vimState: VimState): Promise<VimState> {
     const number = parseInt(this.keysPressed[0], 10);
+    const operatorCount = vimState.recordedState.operatorCount;
 
-    vimState.recordedState.count = vimState.recordedState.count * 10 + number;
+    if (operatorCount > 0) {
+      const lastAction =
+        vimState.recordedState.actionsRun[vimState.recordedState.actionsRun.length - 2];
+      if (!(lastAction instanceof CommandNumber)) {
+        // We have set an operatorCount !== 0 after an operator, but now we got another count
+        // number so we need to multiply them.
+        vimState.recordedState.count = operatorCount * number;
+      } else {
+        // We are now getting another digit which means we need to multiply by 10 and add
+        // the new digit multiplied by operatorCount.
+        //
+        // Example: user presses '2d31w':
+        // - After '2' the number 2 is stored in 'count'
+        // - After 'd' the count (2) is stored in 'operatorCount'
+        // - After '3' the number 3 multiplied by 'operatorCount' (3 x 2 = 6) is stored in 'count'
+        // - After '1' the count is multiplied by 10 and added by number 1 multiplied by 'operatorCount'
+        //   (6 * 10 + 1 * 2 = 62)
+        // The final result will be the deletion of 62 words.
+        vimState.recordedState.count = vimState.recordedState.count * 10 + number * operatorCount;
+      }
+    } else {
+      vimState.recordedState.count = vimState.recordedState.count * 10 + number;
+    }
 
     return vimState;
   }
--- a/src/configuration/remapper.ts
+++ b/src/configuration/remapper.ts
@@ -150,36 +150,39 @@ export class Remapper implements IRemapper {
     vimState.keyHistory = vimState.keyHistory.slice(0, -numCharsToRemove);
 
     if (remapping.after) {
-      const count = vimState.recordedState.count || 1;
-      vimState.recordedState.count = 0;
-      for (let i = 0; i < count; i++) {
-        await modeHandler.handleMultipleKeyEvents(remapping.after);
-      }
+      await modeHandler.handleMultipleKeyEvents(remapping.after);
     }
 
     if (remapping.commands) {
-      for (const command of remapping.commands) {
-        let commandString: string;
-        let commandArgs: string[];
-        if (typeof command === 'string') {
-          commandString = command;
-          commandArgs = [];
-        } else {
-          commandString = command.command;
-          commandArgs = command.args;
-        }
+      const count = vimState.recordedState.count || 1;
+      vimState.recordedState.count = 0;
+      for (let i = 0; i < count; i++) {
+        for (const command of remapping.commands) {
+          let commandString: string;
+          let commandArgs: string[];
+          if (typeof command === 'string') {
+            commandString = command;
+            commandArgs = [];
+          } else {
+            commandString = command.command;
+            commandArgs = command.args;
+          }
 
-        if (commandString.slice(0, 1) === ':') {
-          // Check if this is a vim command by looking for :
-          await commandLine.Run(commandString.slice(1, commandString.length), modeHandler.vimState);
-          await modeHandler.updateView(modeHandler.vimState);
-        } else if (commandArgs) {
-          await vscode.commands.executeCommand(commandString, commandArgs);
-        } else {
-          await vscode.commands.executeCommand(commandString);
-        }
+          if (commandString.slice(0, 1) === ':') {
+            // Check if this is a vim command by looking for :
+            await commandLine.Run(
+              commandString.slice(1, commandString.length),
+              modeHandler.vimState
+            );
+            await modeHandler.updateView(modeHandler.vimState);
+          } else if (commandArgs) {
+            await vscode.commands.executeCommand(commandString, commandArgs);
+          } else {
+            await vscode.commands.executeCommand(commandString);
+          }
 
-        StatusBar.setText(vimState, `${commandString} ${commandArgs}`);
+          StatusBar.setText(vimState, `${commandString} ${commandArgs}`);
+        }
       }
     }
   }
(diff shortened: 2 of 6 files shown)


berknam:
Should commits like this last one 75c37ab0709913c845987941b3cb3c036cfda986 be done in a different PR? In this case since it was part of the same code that was being changed in this PR I kept it here, but if it was better to create a new PR for it tell me so next time I will do that.

vettukal:
I pulled the branch https://github.com/berknam/Vim/tree/remapRefactor and had built it. 
Saw an issue with Visual Mode. If I select some lines and then do a `x` (do it randomly multiple times), sometimes it hangs, then no Vim command responds. Toggle Vim Mode doesn't work. I have to basically close and build it again. 

Else everything looks pretty good. This PR solves 2-3 bugs that I was facing in remap for both normal mode and visual mode. Thanks for working on it.

berknam:
> 
> 
> I pulled the branch https://github.com/berknam/Vim/tree/remapRefactor and had built it.
> Saw an issue with Visual Mode. If I select some lines and then do a `x` (do it randomly multiple times), enter normal mode and press x, sometimes it hangs, then no Vim command responds. (Sorry I can't reproduce it.) Toggle Vim Mode doesn't work. I have to basically close and build it again.
> 
> Else everything looks pretty good. This PR solves 2-3 bugs that I was facing in remap for both normal mode and visual mode. Thanks for working on it.

Thanks @vettukal for the feedback! :+1:

By the way this isn't the right PR for that comment it relates to the PR #4735.

Also that branch is not exactly the "stable-ish" branch like the one in the PR.

Still, I tried to reproduce that issue but can't do it. Can you give some more info? When you say visual mode is it visual mode through 'v' key or visual line mode through 'V' key? You mean it happens when you get into visual mode move to some lines above or below and then press `x`?

From what you say it looks like there is some kind of infinite loop remapping, can you tell what are the remappings you are using?

vettukal:
> By the way this isn't the right PR for that comment it relates to the PR #4735.
My bad. I was intended to reply to #4735 
> Also that branch is not exactly the "stable-ish" branch like the one in the PR.

> Still, I tried to reproduce that issue but can't do it. Can you give some more info? When you say visual mode is it visual mode through 'v' key or visual line mode through 'V' key? You mean it happens when you get into visual mode move to some lines above or below and then press `x`?

It was 'v' key. I get into visual mode and select some lines using remapped j,k and then press `x` to delete them. 

> From what you say it looks like there is some kind of infinite loop remapping, can you tell what are the remappings you are using?
w,a,s,d -> k,h,j,l

src/actions/commands/actions.ts
```diff
@@ -308,8 +308,31 @@ export class CommandNumber extends BaseCommand {
...
   public async exec(position: Position, vimState: VimState): Promise<VimState> {
     const number = parseInt(this.keysPressed[0], 10);
-
-    vimState.recordedState.count = vimState.recordedState.count * 10 + number;
+    const operatorCount = vimState.recordedState.operatorCount;
+
+    if (operatorCount > 0) {
+      const lastAction =
+        vimState.recordedState.actionsRun[vimState.recordedState.actionsRun.length - 2];
+      if (!(lastAction instanceof CommandNumber)) {
```
J-Fields, line 316: Is this branch necessary? Isn't `vimState.recordedState.count` 0 before it's set? (could be wrong about that)

src/configuration/remapper.ts
```diff
@@ -150,36 +150,39 @@ export class Remapper implements IRemapper {
     vimState.keyHistory = vimState.keyHistory.slice(0, -numCharsToRemove);
 
     if (remapping.after) {
+      await modeHandler.handleMultipleKeyEvents(remapping.after);
```
J-Fields, line 153: Can you explain why this is only being done once now?

src/actions/commands/actions.ts
berknam: This is to handle the first number after an operator in a different way from the following ones. For example, if you press '3d12w' the '3' makes `vimState.recordedState.count = 3`, then 'd' makes `vimState.recordedState.operatorCount = 3`, then when pressing '1' this branch sees that the action previous to itself was not a `CommandNumber` which means it was the operator so it makes `vimState.recordedState.count = 3 * 1`, then the last number '2' goes to the other branch and makes `vimState.recordedState.count= 3 * 10 + 2 * 3` which makes the final count 36 (3*12).

src/configuration/remapper.ts
berknam: All the actions already handle the case when there is a count. Some of them have a special handle that applies the entire counts in one go others have the `for` loop on the `BaseAction.execCount`. This is what the modeHandler always calls even if `recordedState.count` is 0. Previously on remaps we were setting the count to 0 and then using a `for` loop to call the remaps 'count' number of times but that means that the actions would always be run with `execCount(0)` so if you had a remap to `2V` instead of selecting two lines like the `execCount(2)` would've done we are call `V` twice meaning we go to VisualLine mode and then we leave VisualLine mode.
This might be from an older version were the actions didn't handle count properly, but now they do so this was unnecessary. If you map `v -> V` and you pressed `2v` it would do nothing because it would make `recordedState.count=0` and handled 'VV', but what you want is `2v` to become `2V` after remap, and when you get to the `v` remap the '2' was already handled and `recordedState.count=2` so when handling the remap all you need to do is handling the `V` without changing the count.

src/actions/commands/actions.ts
J-Fields: Got it 👍

berknam: Also the `operatorCount` is only set when we get an Operator and `recordedState.count > 0`. So when `operatorCount` is higher than 0 we know that at least we had a number followed by an operator key followed by another number (example `2d3`). So this first branch handles only the first number after an operator that had a count before, the branch below handles all the following numbers and the last branch handles all the numbers before an operator and the numbers after an operator that didn't had a count before.

src/configuration/remapper.ts
J-Fields: Bleh. I remember taking a look at this stuff a while back and thinking it needs a refactor. Thanks for the explanation!

berknam: Also in the situation where you did `3i=jk` with `jk -> <Esc>` in insert mode this would result in `3i=<Esc><Esc><Esc>`. But when handling the first <Esc> the `recordedState.count` was set to 0 so the `CommandEscInsertMode` which is responsible to handle the `[count]i` to repeat the insertions would see that `recordedState.count` was 0 and it wouldn't repeat like it should.

berknam: > 
> 
> Bleh. I remember taking a look at this stuff a while back and thinking it needs a refactor. Thanks for the explanation!

If you ever need any help understanding something from the `handleKeyEvent`, `handleMultipleKeyEvents`, `handleKeyEventHelper` or even some parts from `runAction` from modeHandler or anything from `remapper` you can always ask me since I think I'm starting to understand it all a little bit better now. 😄

PR #4878, opened 2020-05-19
Fix recording macro not showing after 'setText'

### LINKED ISSUE #4753: confusing macro recording
When recording macro, I notice that the state being displayed at the bottom bar can be quite confusing. It looks like the information ```Recording @qr``` can easily be masked. For example:

- perform a write ```:w```
- then start a macro recording: ```qr```
- in this case nothing gets displayed on the bottom bar about the macro (I still see only '"space_age.cpp" 48L 1462C Written' displayed, nothing about the macro status).

By contrast:

- starting from the last previous point:
- go to insert mode: ```i```
- then the displayed information becomes '--INSERT-- Recording @qr' 

So the macro was well being recorded, but the info message was being hidden.

A fix for this is, each time one wants to record a macro:

- press ```i```
- press ```ESC```
- then press ```q```+CHAR, and the info is well displayed.

But still, think this may be confusing, at least this confused me for 10 minutes or so until I understand this :) .

berknam:
Fix recording macro not showing after 'setText'
- If 'setText' was called somewhere else like when undoing or when ':w'
the Recording info would be lost
- This commit makes it so that after the next key press we will get the
info back, this is the same behavior as Vim

Fixes #4753

The `Recording @` message was being erased because every time the `StatusBar.setText` is called it replaces the existing text with the new message.

This happens, for example, when the `:write`  command calls `setText` or when you undo something, which might happen in the middle of the macro recording and erases the recording message.

The message would only update again with the `Recording @` after a mode change. This makes it update after the next key press, which means if you undo you get the undo message until you press some key which replaces that message with the clear text that includes the macro recording.

This is the Vim behavior but I think Neovim always shows it. We could make it always show up by appending it to `text` on the `StatusBar.setText(text)` function as well. Some users might prefer it that way.

### DIFF
--- a/src/mode/modeHandler.ts
+++ b/src/mode/modeHandler.ts
@@ -330,7 +330,8 @@ export class ModeHandler implements vscode.Disposable {
 
     // We don't want to immediately erase any message that resulted from the action just performed
     if (StatusBar.getText() === oldStatusBarText) {
-      // Clear the status bar of high priority messages if the mode has changed or the view has scrolled
+      // Clear the status bar of high priority messages if the mode has changed, the view has scrolled
+      // or it is recording a Macro
       const forceClearStatusBar =
         (this.vimState.currentMode !== oldMode && this.vimState.currentMode !== Mode.Normal) ||
         this.vimState.editor.visibleRanges[0] !== oldVisibleRange ||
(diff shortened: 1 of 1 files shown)


src/mode/modeHandler.ts
```diff
@@ -333,7 +333,8 @@ export class ModeHandler implements vscode.Disposable {
       // Clear the status bar of high priority messages if the mode has changed or the view has scrolled
```
J-Fields, line 333: Can you update this comment please?

J-Fields:
> This is the Vim behavior but I think Neovim always shows it

I think I have an old version of neovim installed, but this doesn't seem to be the case for me

J-Fields:
Testing this out again in neovim, it seems like there's just a timeout. I like this behavior, though, and it's a quick fix, so let's go with it at least for now.

PR #4901, opened 2020-05-26
Swap workspace and ui in extensionKind to enable WSL neovim use

### LINKED ISSUE #4614: Cannot use neovim on WSL 2 remote workspace (Windows 10)
**Describe the bug**
I'm running VSCode on Windows 10 and have the new Windows Subsystem for Linux 2 set up, with most of my projects in my Linux environment. When I enable the _Vim: Enable neovim_ option and set the `vim.neovimPath` JSON setting in my remote settings file, it doesn't work. The following error is displayed in a popup:

```
Configuration: Invalid neovimPath. spawn `/usr/bin/nvim` ENOENT.
```

**To Reproduce**

1.  Install WSL 2 on Windows 10 and a suitable OS
2.  Create a project within the WSL environment
3.  Install neovim in the WSL environment
4.  Configure VSCodeVim to use neovim
5.  See error

**Expected behavior**
VSCodeVim should use the neovim installed in the WSL environment

OS: Windows 10 (Build 19041)

```
$ which nvim
/usr/bin/nvim
```

```
nvim --version
NVIM v0.2.2
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
Compilation: /usr/bin/cc -g -O2 -fdebug-prefix-map=/build/neovim-_mzR4z/neovim-0.2.2=. -fstack-protector-strong -Wformat -Werror=format-security -Wdate-time -D_FORTIFY_SOURCE=2 -DDISABLE_LOG -Wdate-time -D_FORTIFY_SOURCE=2 -Wconversion -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=1 -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -O2 -g -DMIN_LOG_LEVEL=3 -Og -g -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wimplicit-fallthrough -Wvla -fstack-protector-strong -fdiagnostics-color=auto -DINCLUDE_GENERATED_DECLARATIONS -D_GNU_SOURCE -I/build/neovim-_mzR4z/neovim-0.2.2/build/config -I/build/neovim-_mzR4z/neovim-0.2.2/src -I/usr/include -I/usr/include -I/usr/include -I/usr/include -I/usr/include -I/usr/include -I/usr/include -I/build/neovim-_mzR4z/neovim-0.2.2/build/src/nvim/auto -I/build/neovim-_mzR4z/neovim-0.2.2/build/include
Compiled by pkg-vim-maintainers@lists.alioth.debian.org

Features: +acl +iconv +jemalloc +tui 
See ":help feature-compile"

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "/usr/share/nvim"

Run :checkhealth for more info
```

mhchen:
**Note: this PR is a proposal. Please read the caveats below about remote development. Would love to hear thoughts.**

I'm running WSL on Windows. When I try to use Neovim, VSCodeVim can't find it. This change prefers having me install the extension on WSL instead, where it can find my copy of Neovim and everything is great.

I frustratingly ran into #4614 and did a bunch of research on how to resolve this. I believe this is the most straightforward approach. The fact that this extension is trying to spawn a process using a file means that it (I think?) it's unequivocally a workspace extension. See https://code.visualstudio.com/api/advanced-topics/remote-extensions#architecture-and-extension-kinds and let me know if you agree. Looking at [the original PR](https://github.com/vsCodeVim/Vim/pull/4435) it looks like we weren't quite sure at the time. Maybe the neovim integration didn't exist back then.

I've been running this extension as a workspace extension for a few days without any negative effects, so I don't think VSCodeVim actually requires any UI capabilities. I also set this to workspace on my Mac which does not do any remote development and it works fine.

**The main caveat** is that I'm not sure if `ui` would've allowed VS Code over [Remote SSH](https://code.visualstudio.com/docs/remote/ssh) to work with the local copy of Neovim. Since the `neovimPath` is `machine-overridable` anyway though, I'm not convinced this was working in the first place, but swapping the two in `extensionKind` might actually make it impossible.

As an alternative, we can also take the [approach vscode-neovim took](https://github.com/asvetliakov/vscode-neovim/blob/a563847641e4d2cd7003cd5a26c60069be3ca716/src/controller.ts#L295) and specify a setting that will launch nvim directly using wsl.exe. It feels a bit more heavy-handed and using `workspace` feels cleaner to me though.

Fixes #4614

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,8 +31,8 @@
     "Keymaps"
   ],
   "extensionKind": [
-    "ui",
-    "workspace"
+    "workspace",
+    "ui"
   ],
   "activationEvents": [
     "*",


J-Fields:
> Looking at the original PR it looks like we weren't quite sure at the time. Maybe the neovim integration didn't exist back then.

Yeah I remember researching this and finding that the documentation was seriously lacking. Seems it's improved somewhat since.

> The main caveat is that I'm not sure if ui would've allowed VS Code over Remote SSH to work with the local copy of Neovim.

I think this is true, however the documentation you linked suggests to me that `"remote.extensionKind"` can override this. If someone wants to use a local nvim while connecting to a remote machine, hopefully they can do that. Connecting to WSL and using its nvim seems like a more likely workflow, so I'm fine prioritizing that.

> As an alternative, we can also take the approach vscode-neovim took and specify a setting that will launch nvim directly using wsl.exe

Bleh. Unless someone can convince me that accommodates a use that we can't otherwise, I'm not at all inclined to do that

J-Fields:
I'm actually reverting this because VSCodeVim definitely prefers to run on your local machine (`UI`) - you can put this into your settings to force it to run remotely:

```
"remote.extensionKind": {
  "vscodevim.vim": ["workspace"]
},
```

gemhowie:
> I'm actually reverting this because VSCodeVim definitely prefers to run on your local machine (`UI`) - you can put this into your settings to force it to run remotely:
> 
> ```
> "remote.extensionKind": {
>   "vscodevim.vim": ["workspace"]
> },
> ```

After I set these options, whether neovim-related options like "vim.enableNeovim" and "vim.neovimPath" should be placed in workspace's settings.json file, because the path of neovim is determined by the OS where the workspace is located.

Issue #4913, opened 2020-05-29
Undo after inserting keys and pressing Ctrl+u is broken


mhchen:
**Describe the bug**
In insert mode, when I type and then press Ctrl+u (to delete to beginning of line), then try to undo, it only moves my cursor without bringing back the deleted text.

**To Reproduce**

1.  Enter insert mode in the middle of a line
2.  Type some text
3.  `<C-u>`
4. Exit insert mode
5. `Press u`

**Expected behavior**
Pre-existing text (prior to inserting characters) is ignored.

**Screenshots**
![Kapture 2020-05-28 at 23 07 09](https://user-images.githubusercontent.com/1325549/83227085-32c17080-a138-11ea-8af5-227c1e1addc9.gif)

**Environment (please complete the following information):**

- Extension (VsCodeVim) version: 1.14.5
- VSCode version: 1.46.0-insiders
- OS: OS X 10.15.4

kevinw:
Undo and insert mode is super broken! This makes the editor almost unusable for me daily.

sethlivingston:
I'm so thankful for this plugin, but if I fat-finger a few keys and accidentally delete some characters and jump somewhere, then I'm hosed because undo doesn't work. This is a big problem for me.

bitshiftnetau:
I agree with @kevinw I have just turned this plugin off after the broken undo feature undid a whole swathe of text, rather than just one character at a time. Unusable.

berknam:
@bitshiftnetau  Were you by any chance recording a macro before that resulted in error?

PR #4914, opened 2020-05-29
Add `<C-u>` insert mode stop point at column where cursor started

### LINKED ISSUE #4875: CTRL + U
**Describe the bug**
Using the special key on insert mode (CTRL + U) is not only deleting the current characters that just entered but also deletes the beginning of the line from the cursor.

**Expected behavior**
Delete only the entered characters

- Issue Type: `Bug`
- Extension Name: `vim`
- Extension Version: `1.14.4`
- OS Version: `Windows_NT x64 10.0.18363`
- VSCode version: `1.45.1`

mhchen:
NOTE: I also discovered #4913 while debugging this issue. The bug was pre-existing but seems pretty important to fix (undo bugs seem really potentially catastrophic).

Gets us closer to functionality parity with Vim's Ctrl+U, which has a separate stop point if you are to the right of the original column in insert mode.

Fixes #4875

### DIFF
--- a/src/actions/commands/insert.ts
+++ b/src/actions/commands/insert.ts
@@ -563,7 +563,10 @@ class CommandCtrlUInInsertMode extends BaseCommand {
         ? position.getLineBegin()
         : position.getLineBeginRespectingIndent();
     }
-    await TextEditor.delete(new vscode.Range(start, position));
+    vimState.recordedState.transformations.push({
+      type: 'deleteRange',
+      range: new Range(start, position),
+    });
     vimState.cursorStopPosition = start;
     vimState.cursorStartPosition = start;
     return vimState;
--- a/src/actions/commands/insert.ts
+++ b/src/actions/commands/insert.ts
@@ -551,9 +551,18 @@ class CommandCtrlUInInsertMode extends BaseCommand {
   keys = ['<C-u>'];
 
   public async exec(position: Position, vimState: VimState): Promise<VimState> {
-    const start = position.isInLeadingWhitespace()
-      ? position.getLineBegin()
-      : position.getLineBeginRespectingIndent();
+    let start: Position;
+
+    if (
+      vimState.insertModeCursorCtrlUStopPosition?.line === position.line &&
+      vimState.insertModeCursorCtrlUStopPosition?.isBefore(position)
+    ) {
+      start = vimState.insertModeCursorCtrlUStopPosition;
+    } else {
+      start = position.isInLeadingWhitespace()
+        ? position.getLineBegin()
+        : position.getLineBeginRespectingIndent();
+    }
     await TextEditor.delete(new vscode.Range(start, position));
     vimState.cursorStopPosition = start;
     vimState.cursorStartPosition = start;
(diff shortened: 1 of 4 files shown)


src/actions/commands/insert.ts
```diff
@@ -551,9 +551,18 @@ class CommandCtrlUInInsertMode extends BaseCommand {
...
+      vimState.insertModeCursorCtrlUStopPosition?.line === position.line &&
+      vimState.insertModeCursorCtrlUStopPosition?.isBefore(position)
+    ) {
+      start = vimState.insertModeCursorCtrlUStopPosition;
+    } else {
+      start = position.isInLeadingWhitespace()
+        ? position.getLineBegin()
+        : position.getLineBeginRespectingIndent();
+    }
     await TextEditor.delete(new vscode.Range(start, position));
```
berknam, line 566: Since you're changing this action, maybe you could also change it to make use of transformations like `deleteRange` instead of changing the TextEditor directly. I'm not entirely sure, but I think this is preferred. Maybe @J-Fields can give some input here?!

mhchen: For my education, what's the advantage of the proposed approach? I'm open to changing it as long as @J-Fields agrees.

berknam: Currently it won't change much. But the idea I think is to not have changes to editor on the actions but instead on the 'modeHandler' through transformations. If later someone tries to add some kind of compressing of changes to macros like @J-Fields did to the 'DocumentContentChangeAction' it will be harder if the actions are doing the changes themselves, instead of giving the transformation that needs to be done. But this is just my opinion.

J-Fields: This design decision was made before I got involved, so I don't have the full story, but I think the idea is basically to have all the view changes in one place so the system as a whole is easier to reason about. It would probably make things like adopting VSCode's undo stack easier.

In theory I like the idea a lot, but I've found in practice it sometimes makes things *harder* to reason about, but I think that's mostly because `ModeHandler` is stuffed full of poorly documented accidental complexity.

So if it's simple to switch to a transformation, that'd be cool! If not, don't worry about it for now - once we have `ModeHandler` under more control, we can get to it (or possibly decide it was never a good idea in the first place and abandon it).

mhchen: Sounds good, I'll take a shot at using `deleteRange` instead

mhchen:
Updated @berknam @J-Fields please take another look when you get the chance

berknam:
There is currently an issue that if you go into insert mode with `a` instead of `i` it deletes just one character instead of deleting to the start of line. Also if you do `a<right><C-u>` it deletes the two previous characters instead of just one.
If you move up or down and then move to the right (`i<up><right><C-u>`) it deletes just one character, but in Vim it deletes the whole line.
If you are in the middle of a line and do `Atest<C-u>` it deletes until the middle of the line instead of deleting just 'test'. And the worst is that after doing this, if you do `<Esc>u` it doesn't undo.

berknam:
Also you should test the behavior when using `<C-a>`, `<C-o>` and `<C-t>` in insert mode.
Examples:
- if you do `atest<Esc>a<C-a><C-u>` it should only delete the second test but it deletes everything
- if you do `atest<Esc>a<C-a>2<C-u>` it should only delete 'test2' but it deletes everything
- if you do `atest<Esc>bli<C-t><C-u>` it should only delete the 't' but it deletes until column 2

These are some examples I though of trying but maybe there's more.

J-Fields:
I'm trying to clear out old PRs - this is still incomplete and dormant. If you want to revive this, feel free and let me know if I can help. Thanks!

PR #4935, opened 2020-06-09
Fix docker tests

berknam:
Currently when you run `npx gulp` to build and run the tests it creates a Docker image from Dockerfile but it uses the extension root path as the context to build the image even though it is not needed. This means that every time we run the tests we are sending over 400MB to the daemon just to build the image and we don't even use anything from that context.

I changed the context to the `./build/` folder that contains the Dockerfile.

Then we create a container from that image but we don't give it the `--rm` flag which means after a while we have docker full of stopped vscodevim containers. I have added that flag to remove the container when done with the tests.

Also the `runTests.ts` file had the wrong path to 'extensionDevelopmentPath' because the file is in `/app/src/test/` and is built to folder `/app/out/test/` but we were resolving the path with `../../../` which meant we were setting the extensionDevelopmentPath to `/` (root). I changed it to `../../` so that it resolves to `/app` as it should.

The file `index.ts` was setting the 'testsRoot' to `/out/` because the file is inside `/out/test/` and we were resolving the path with `..`. I changed it to just `.` so that the 'testsRoot' folder is set to `/out/test/`. This wasn't a problem because the only `**/**.test.js` files on `/out/` were the ones inside `/out/test/` but there is no need to be looking files in the wrong folder.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -213,7 +213,7 @@ gulp.task('run-test', function (done) {
   console.log('Building container...');
   var dockerBuildCmd = spawn(
     'docker',
-    ['build', '-f', './build/Dockerfile', '.', '-t', dockerTag],
+    ['build', '-f', './build/Dockerfile', './build/', '-t', dockerTag],
     {
       cwd: process.cwd(),
       stdio: 'inherit',
@@ -232,6 +232,7 @@ gulp.task('run-test', function (done) {
     const dockerRunArgs = [
       'run',
       '-it',
+      '--rm',
       '--env',
       `MOCHA_GREP=${options.grep}`,
       '-v',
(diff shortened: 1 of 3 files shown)


J-Fields:
I always run the tests from the VSCode debugger, and know next to nothing about docker, so wasn't able to fix this. Thanks a bunch! 🙃

berknam:
No problem. I started looking into because all of a sudden my docker tests started failing randomly and when I after a while I figured the problem was low memory and swap allocated to docker. And the reason it worked before and then stopped was because I had more than 50 stopped vscodevim containers 😄.

You should try using it though! You only need to have docker installed and if you then run `npx gulp` it builds and runs the tests in docker which allows you to keep working while the tests run.

PR #4952, opened 2020-06-17
Add `init.vim` support for vimrc

vegerot:
This PR adds support for `init.vim` for automatic detection in `vim.vimrc.enable`.  We need this because as neovim becomes more and more popular, more people will be using `init.vim` files instead of `.vimrc` files.  Now, many people use both, and tell their `init.vim` to source `.vimrc`.  I put the `init.vim` as the lowest priority; favoring `.vimrc` before `init.vim`.

I was thinking of making an issue for this this morning, when I figured I could probably do it myself.  Should I make an issue too?

I put the `init.vim` as the lowest priority; favoring `.vimrc` before `init.vim`.  I looked through [vimrc.test](https://github.com/VSCodeVim/Vim/blob/master/test/configuration/vimrc.test.ts) and didn't see any tests that needed to be modified.

### DIFF
--- a/package.json
+++ b/package.json
@@ -787,7 +787,7 @@
         },
         "vim.vimrc.path": {
           "type": "string",
-          "description": "Path to a Vim configuration file. If unset, it will check for $HOME/.vimrc, $HOME/_vimrc, or $HOME/config/nvim/init.vim, in that order."
+          "description": "Path to a Vim configuration file. If unset, it will check for $HOME/.vimrc, $HOME/_vimrc, or $HOME/.config/nvim/init.vim, in that order."
         },
         "vim.substituteGlobalFlag": {
           "type": "boolean",
--- a/package.json
+++ b/package.json
@@ -787,7 +787,7 @@
         },
         "vim.vimrc.path": {
           "type": "string",
-          "description": "Path to a Vim configuration file. If unset, it will check for $HOME/.vimrc or $HOME/_vimrc."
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -787,7 +787,7 @@
         },
         "vim.vimrc.path": {
           "type": "string",
-          "description": "Path to a Vim configuration file. If unset, it will check for $HOME/.vimrc or $HOME/_vimrc."
+          "description": "Path to a Vim configuration file. If unset, it will check for $HOME/.vimrc, $HOME/_vimrc, or $HOME/config/nvim/init.vim, in that order."
```
J-Fields, line 790: `.config`, not `config`, right?

vegerot: Thanks for the catch!

package.json
```diff
@@ -787,7 +787,7 @@
         },
         "vim.vimrc.path": {
           "type": "string",
-          "description": "Path to a Vim configuration file. If unset, it will check for $HOME/.vimrc or $HOME/_vimrc."
+          "description": "Path to a Vim configuration file. If unset, it will check for $HOME/.vimrc, $HOME/_vimrc, or $HOME/.config/nvim/init.vim, in that order."
```
J-Fields, line 790: Sorry, this should also be `and`, not `or`

vegerot: Kinda confused by that.  If `$HOME/.vimrc` exists it WON'T check for `init.vim`.  So how is it "and"?

vegerot: On second thought, I don't really care lol

vegerot:
No, thank YOU @J-Fields !!

PR #4967, opened 2020-06-23
Add mapping for opening definition to the side.

matthughes:

Adds a key mapping for go to definition to the side.  This is the same as if you CMD+Option+Click on a definition, it opens the definition in a side panel.

### DIFF
--- a/src/actions/commands/actions.ts
+++ b/src/actions/commands/actions.ts
@@ -2543,6 +2543,24 @@ class CommandGoToDefinition extends BaseCommand {
 }
 
 @RegisterAction
+class CommandGoToDefinitionAside extends BaseCommand {
+  modes = [Mode.Normal];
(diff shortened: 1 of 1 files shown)


J-Fields:
Any reason `nnoremap gs editor.action.revealDefinitionAside` isn't sufficient?

matthughes:
I’m not familiar with that command.  Where/how would I configure that in VsCode?

J-Fields:
That would be via `.vimrc.` Equivalently, in `settings.json`:
```json
  "vim.normalModeKeyBindingsNonRecursive": [{
    "before": ["g", "s"],
    "commands": ["editor.action.revealDefinitionAside"]
  }]
```

This seems to work fine, and this isn't a standard vim keybinding, so I'm gonna close this PR. Thanks though!

Issue #3197, opened 2018-11-06
:u should undo


mchong-teal:
**Is your feature request related to a problem? Please describe.**
I use :u to undo changes in vim:
http://vimhelp.appspot.com/undo.txt.html#undo

Entering the command ':u' does not undo, and instead prints the message 'E492 Not an editor command. u'.  I have not been able to find clear documentation on how to change that setting, and it appears in the most recent update the 'otherModesKeyBindings' setting was removed.

The first time this happened to me I immediately uninstalled the extension.  It makes me consider uninstalling every time I try and perform an undo action.

**Describe the solution you'd like**
Entering the command :u[ndo] will undo one change.  For bonus points, :u[ndo] {N} jumps to after change number {N}.

**Describe alternatives you've considered**
Add clear documentation for binding the undo command to :u in command line mode now that 'otherModesKeyBindings' is not available.

**Additional context**
None

mchong-teal:
- Open new file (in normal mode)
- press `i` (to insert mode)
- type something (or do the thing to be undone)
- press `esc` (back to normal mode)
- type `:`, press `u`, press `enter` (should undo)

^ doing the above in vim undoes the action (expected behaviour), but displays the above error in vscode.

mchong-teal:
This is on v0.17.3 (2018-12-30), with vscode Version: 1.30.1.

mchong-teal:
![vscode-vim-undo](https://user-images.githubusercontent.com/43995143/61823061-fc84d780-ae28-11e9-8447-6e242e9e322f.gif)
![vim-undo](https://user-images.githubusercontent.com/43995143/61823102-145c5b80-ae29-11e9-8bdc-d412da7ba221.gif)

Issue #4963, opened 2020-06-23
Undo command doesn't accept [count]


berknam:
**Describe the bug**
Undo command doesn't accept [count]

**To Reproduce**

1.  Do `aa<Esc>` + `ab<Esc>` + `ac<Esc>`
2.  You should get 'abc'
3.  Now do `3u` to undo the last 3 actions
4.  You get 'ab' instead of nothing, only one undo is performed

**Expected behavior**
Should undo the [count] number of actions

sql-koala:
@berknam ,
I did take at look at this, since I just added a count to dot. There are two ways to do this. One: provide an execCount-Method.

Another way is this, just add line 4 (in action.ts).
```
export class CommandUndo extends BaseCommand {
  modes = [Mode.Normal];
  keys = ['u'];
  runsOnceForEachCountPrefix = true;
  runsOnceForEveryCursor() {
```

Since you are a dev, you can try that easy.

berknam:
@sql-koala I hadn't looked into this yet because I want to look into all 'undo' logic one day. After I finish all the remapper stuff and some other things I'm working right now I intend to have a look at the possibility of changing the entire 'undo' logic to the vscode undo stack.

@J-Fields I've seen that you mentioned you were looking into this at some point. Is there any sort of prototype or initial attempt that I can lookup? I'm probably not going to do anything yet but just in case I decide to give it a try would be nice to know what has been tested already.

J-Fields:
>Is there any sort of prototype or initial attempt that I can lookup?

No, I'd meant to get a proof of concept working but haven't gotten around to it. I'll try to soon - let me know if/when you start working on this. I have some thoughts that I'd like to clarify by working on a proof of concept but if you start playing with it before I get around to it, I'll just relay them to you.

berknam:
> 
> No, I'd meant to get a proof of concept working but haven't gotten around to it. I'll try to soon - let me know if/when you start working on this. I have some thoughts that I'd like to clarify by working on a proof of concept but if you start playing with it before I get around to it, I'll just relay them to you.

I'll let you know if/when I start working on it. One thing to note is I will probably do it on top of the remapper refactor because one of the issues with using VSCode undo that always pops up in conversations is the problem with undoing the `jj -> <Esc>` in insert mode on the remapper, but that will no longer be an issue with that refactor! 👍

PR #4980, opened 2020-06-28
fixes issue #3197: Adding :u[ndo] cmd and appropriate test

adamjhawley:
This PR implements the :u[ndo] command (a requested feature which is available in Vim)

Issue #3197 

First contribution, so very open to feedback/suggestions! Also I will continue to work on the undo command to add the :undo n argument to undo n number of steps.

### DIFF
--- a/src/cmd_line/commands/undo.ts
+++ b/src/cmd_line/commands/undo.ts
@@ -1,0 +1,37 @@
+import * as node from '../node';
+import { Logger } from '../../util/logger';
+import { StatusBar } from '../../statusBar';
+import { VimState } from '../../state/vimState';
+import { Range } from '../../common/motion/range';
+
+//
+//  Implements :u[ndo]
+//  http://vimdoc.sourceforge.net/htmldoc/undo.html
+//
+export class UndoCommand extends node.CommandBase {
+  protected _arguments: {};
+  private readonly _logger = Logger.get('Undo');
+
+  constructor(args: {}) {
+    super();
+    this._arguments = args;
+  }
+
+  get arguments(): {} {
+    return this._arguments;
+  }
+
+  async execute(vimState: VimState): Promise<void> {
+    const newPositions = await vimState.historyTracker.goBackHistoryStep();
+
+    if (newPositions === undefined) {
+      StatusBar.setText(vimState, 'Already at oldest change');
+    } else {
+      vimState.cursors = newPositions.map((x) => new Range(x, x));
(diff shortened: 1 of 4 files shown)


src/cmd_line/commands/undo.ts
```diff
@@ -0,0 +1,37 @@
...
+    super();
+    this._arguments = args;
+  }
+
+  get arguments(): {} {
+    return this._arguments;
+  }
+
+  async execute(vimState: VimState): Promise<void> {
+    const newPositions = await vimState.historyTracker.goBackHistoryStep();
```
J-Fields, line 25: Rather than copying this from `CommandUndo`, could you just call it (`await new CommandUndo().exec(...);`)?

test/cmd_line/undo.test.ts
```diff
@@ -0,0 +1,25 @@
...
+    modeHandler = await getAndUpdateModeHandler();
+    vimState = modeHandler.vimState;
+  });
+
+  teardown(cleanUpWorkspace);
+
+  test('undoes last action after insert mode', async () => {
+    await modeHandler.handleMultipleKeyEvents(['i', 'a', '<Esc>']);
+    await modeHandler.handleMultipleKeyEvents(['i', 'b', '<Esc>']);
+    await commandLine.Run('undo', vimState);
```
J-Fields, line 22: Just refer to `modeHandler.vimState` here, rather than getting it in `setup`

src/cmd_line/commands/undo.ts
adamjhawley: Wasn't sure whether it would have been better with or without the binding. I settled on copying since I was planning on adding the option to add the argument described in the issue #3197 to undo n steps. Which approach do you think would be best with that in mind?

test/cmd_line/undo.test.ts
adamjhawley: Will do 👍

src/cmd_line/commands/undo.ts
J-Fields: Call `exec` for now, then when you add the ability to undo multiple, implement it as `execWithCount` in `CommandUndo` and call that from the ex command. That'll solve #4963 as well

adamjhawley: Great, I will amend accordingly then. 👍

J-Fields:
Thanks!

PR #4981, opened 2020-06-28
prepration for running extension in nodeless environment

rebornix:

VS Code is adding support for running extensions in nodeless environment. This PR tries to refactor the code a bit to allow later on replacing node APIs (`fs`, `child_process`) with shims. 

This PR should not change any functionality for running Vim in desktop.

Once we can run Vim in web worker, we can also think about running Vim in its own worker, which at the end will help with the performance issue: [Overriding the default 'type' command and then calling the default 'type' command results in significantly slower execution time ](https://github.com/microsoft/vscode/issues/65876)


src/configuration/vimrc.ts
```diff
@@ -50,12 +50,16 @@ class VimrcImpl {
...
-          VimrcImpl.addRemapToConfig(config, remap);
+      try {
+        const lines = (await fs.readFileAsync(this.vimrcPath, 'utf8')).split(/\r?\n/);
+        for (const line of lines) {
+          const remap = await vimrcKeyRemappingBuilder.build(line);
+          if (remap) {
+            VimrcImpl.addRemapToConfig(config, remap);
+          }
         }
+      } finally {
```
J-Fields, line 61: Shouldn't this be `catch`?

src/state/vimState.ts
```diff
@@ -25,6 +24,7 @@ import { ReplaceState } from './../state/replaceState';
  */
 export class VimState implements vscode.Disposable {
   private readonly logger = Logger.get('VimState');
+  private _disposables: vscode.Disposable[] = [];
```
J-Fields, line 27: Is this needed if nothing `VimState` holds is disposable (after you pulled `nvim` out)?

extension.ts
```diff
@@ -3,528 +3,27 @@
...
-
-    if (previousActiveEditorId) {
-      const prevHandler = ModeHandlerMap.get(previousActiveEditorId);
-      prevHandler!.vimState.focusChanged = true;
-    }
-  }
-
-  return curHandler;
-}
+import './src/actions/include-main';
```
J-Fields, line 6: What's the rationale for splitting this file into `extension.ts` and `extensionBase.ts`?

src/cmd_line/commandLine.ts
```diff
@@ -33,13 +39,17 @@ class CommandLine {
...
   public previousMode = Mode.Normal;
 
   constructor() {
     this._history = new CommandLineHistory();
   }
 
-  public async load(): Promise<void> {
+  public async load(context: vscode.ExtensionContext): Promise<void> {
+    const m = await import('../neovim/neovim');
+    this._nvim = new m.NeovimWrapper();
```
J-Fields, line 52: Is this getting disposed anywhere?

extension.ts
rebornix: The names are not intuitive and at the beginning I call them `extensionNodeEntry.ts`, `extensionBase.ts` and then we later on we introduce `extensionWebEntry.ts`. The idea is that we will have different entry files for node and web, and the major difference between the entry files are the modules they load and the default flag/options based on the environment. The `extensionBase` will be platform/env agnostic ultimately.

 I didn't make it so now as there are quite a few places (package.json, test, etc) rely on `extension.js` so I try to keep the change minimal.

J-Fields: Got it 👍

rebornix:
@J-Fields addressed your feedback, thanks for them!

Based on our code structure, to allow us ship bundles for different platforms (node/web), we need to do multiple different refactorings (not all included in this PR)

* static contributions if possible. `Actions` are the first set of modules we made static, each `*action` module is *plug-and-play*. In this PR, I made configuration validators static as well so in Web, all we need to do is not including `import './src/configuration/validators/neovimValidator';` to get validators work.
* Make `extensionBase` platform agnostic if possible. For example, in Web, `context.globalStoragePath` might not be available so we should not depend on it directly. We can move most platform specific configurations into entry files, e.g., `extension.ts`, `extensionWeb.ts`
* Lastly, and also the dirty part. We have a lot of code directly calling `fs`, `child_process`, we need to carefully shim them. The work behind the this will come in another PR.

rebornix:
@J-Fields thanks for merging it with master ;)

Issue #4973, opened 2020-06-25
Can't use the surround 'S' in visual mode, always defaults to ActionChangeLineVisualMode


berknam:
**Describe the bug**
After commit d43a6e1ef49da86a4e6978760299f5196d137431 you can no longer use the surround 'S' in visual mode. It always gets the action 'ActionChangeLineVisualMode' first.

**To Reproduce**

1.  With text `|hello world`
2.  Press `viw` to select 'hello' in visual mode
3.  Press 'S' to surround
4.  The line gets deleted and you go to insert mode

**Expected behavior**
After pressing 'S' it should wait for the surround key and then after pressing that key it should surround the word.

**Additional context**
This is another case where remapping the plugin actions would fix this, like mentioned in #4916

PR #4985, opened 2020-06-30
Fix surround alias using shortcut version with space instead of the ones without

### LINKED ISSUE #4965: extra space in Surround
**Describe the bug**

The simulated vim-surround plugin adds extra space for `b`. 

**To Reproduce**

`foo bar` (cursor on a) => (type `ysiwb`) => `foo ( bar )`

According to the document, `b` is the alias for `)` so I'm expecting `foo (bar)`.

**Expected behavior**
The expected result is `foo (bar)`. 

- Extension (VsCodeVim) version: 1.14.5
- VSCode version: 1.46.1
- OS: macOS Catalina 10.15.4

EDIT: Sorry for posting it early while I was still editing.

berknam:
Fix surround alias using wrong shortcut version
- Shortcut alias were using the version with space but they are supposed
to use the version without space
fixes #4965
- Add 'a' shortcut alias to '>'
- Add tests for alias
- Add some tests for visual and visual line mode (they are not extensive
but should be enough for now) but they are currently being skipped
because of issue #4973

fixes #4965

I've added the alias 'a' to '\>'. I've also added some visual mode tests, they're not very extensive but should be enough for now just to have something. However, those visual tests are currently being skipped because of issue #4973

### DIFF
--- a/src/actions/plugins/surround.ts
+++ b/src/actions/plugins/surround.ts
@@ -294,13 +294,16 @@ export class CommandSurroundAddToReplacement extends BaseCommand {
     // Convert a few shortcuts to the correct surround characters when NOT entering a tag
     if (vimState.surround.replacement.length === 0) {
       if (stringToAdd === 'b') {
-        stringToAdd = '(';
+        stringToAdd = ')';
       }
       if (stringToAdd === 'B') {
-        stringToAdd = '{';
+        stringToAdd = '}';
       }
       if (stringToAdd === 'r') {
-        stringToAdd = '[';
+        stringToAdd = ']';
+      }
+      if (stringToAdd === 'a') {
+        stringToAdd = '>';
       }
     }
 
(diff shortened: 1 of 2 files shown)


J-Fields:
Just made a quick fix to #4973, can you update this accordingly?

berknam:
Done. Also I've made this comment on your commit:
> To fix the test errors I think the best thing to do is not have the 'testConfiguration' have surround enabled by default and then > enable it only on surround tests.
> I had to do this on my plugin refactor. Look [here](https://github.com/berknam/Vim/pull/1/files#diff-e0df06845c1731321fcc4683646037e3) and [here](https://github.com/berknam/Vim/pull/1/files#diff-efd31f886a37854d3a92af292d427897).

I've encounter that issue as well on my plugin refactor. And I fixed it the way mentioned above. If you want I can create a PR for that?

J-Fields:
Thanks for the heads up - just fixed it

J-Fields:
Thanks!

PR #4986, opened 2020-07-01
Fix marks behaviour

### LINKED ISSUE #4955: In a line with indentation, jumping to mark does not work after inserting a new line above
**Describe the bug**
In a line with indentation, jumping to mark (with backtick `` ` ``) does not work after inserting a new line above

**To Reproduce**

1.  In a line with 4-space indentation, `mo` somewhere in the middle of this line to set a mark at `o`
1.  `O` to create a new line above and enter insert mode
1.  `Esc` to go back to normal mode
1.  `` `o `` to go back to mark `o`. It will not go back to mark `o`, instead it will go to a position 4-space to the right of mark `o` in the same line.

**Expected behavior**
It should go back to the position of mark `o`

- Extension (VsCodeVim) version: v1.14.5
- VSCode version: 1.46.0
- OS: Windows 10 Pro

ph0ngp:
This PR fixes #4955 and a bunch of other marks behavior. See the gifs below. (Recording start with `F7` and I use `Caps Lock` as `Escape` so it shows `CapitalEscape`)

- Handle CRLF linebreak
  - Old behavior: mark `a` does not point to the same character
1crlf_old
  - New behavior: mark `a` now points to the same character
1crlf_new

- Insert newline right before a mark
  - Old behavior: mark `a` does not point to the same character
2new_line_before_mark_old
  - New behavior: mark `a` now points to the same character
2new_line_before_mark_new

- Join a line containing a mark
  - Old behavior: mark `a` does not point to the same character
2ajoin_prev_line_old
  - New behavior: mark `a` now points to the same character
2ajoin_prev_line_new

- If a line containing a mark is deleted, that mark should also be deleted
  - Old behavior: mark `a` is not deleted
3delete_mark_old
  - New behavior: mark `a` is deleted
3delete_mark_new

- Update global mark position
  - Old behavior: after edit, there are multiple mark `A`, jumping back to mark A does not respect the edit
4update_global_mark_position_old
  - New behavior: there is always a single mark `A` and its position respects the edit
4update_global_mark_position_new

- Fix bug about changing tabs:
  - Old behavior: after changing tabs back and forth, mark `A` stops responding to the edit
5global_marks_stop_respond_after_changing_tab_old
  - New behavior: after changing tabs back and forth, mark `A` still responds to the edit
5global_marks_stop_respond_after_changing_tab_new

- Jump to a global mark in another file
  - Old behavior: you must jump twice to get to the mark
6jump_to_global_mark_other_file_old
  - New behavior: you only need to jump once
6jump_to_global_mark_other_file_new

- Fix `:marks` command bug (temporarily):
  - Old behavior: `:marks` command tries to show all global marks but fails if there is a global mark from a longer file
7marks_command_global_mark_old
  - New behavior: `:marks` now shows only local and global marks targeting the current document. I know this is not the correct behavior in Vim but at least for now people can partially use this command without error for the currently active file.
7marks_command_global_mark_new

#4955 and the ones listed above

- There remains 2 problems related to marks behavior that I haven't found a way to fix. Check the `TODO` in my code.
- I'm not quite sure if I have missed something. Please correct me and let me know if there is any problem.


J-Fields:
@phphong Would you mind adding a few test cases for the behavior you fixed?

phphong:
@J-Fields yes I will add some test cases

phphong:
@J-Fields please check my commits. I merged the test cases from `modeNormal.test.ts` and `historyTracker.test.ts` and added new test cases into `marks.test.ts`

burgreen:
Before this PR is merged, can you ensure proper vi behavior for the backtick operator?

The proper vi function of `o will return to the same line and COLUMN (not the same letter) of the original mark o. I am requesting that this proper vi function of the backtick be implemented. I have a favorite key binding that critically depends on this standard behavior.

Try this in a Linux vi editor.
On the line given by:
abcde
Place a mark o on the letter d.
Insert two x letters at the front of the line, giving:
xxabcde
`o will place the cursor on the letter b because the line and COLUMN are preserved by the mark o. The cursor will not be on letter d from `o. That would be non-standard vi.

See: https://vim.fandom.com/wiki/Using_marks in the "Using marks" subsection.

phphong:
> The proper vi function of `o will return to the same line and COLUMN (not the same character) of the original mark o

This is valid for changes **in the same line**. Regarding proper vim function, in addition to your example, I have another example where if you insert a newline right before `a` (move cursor to `a`, press `i` and `<enter>`)

```

abcde
```

now press `` `o ``, it will jump to the first line (I will refer to this as behavior A - marks returning to the same line and column)

However, it does not hold true anymore if you **insert/delete/join lines before the line** containing the original mark `o`. In those cases, the proper vim function of `` `o `` is that it will follow **the same character** instead of returning to the same line and column. For example. If you insert a new line above `abcde` by pressing `O` (instead of `i` and `<enter>`):

```

abcde
```

now press `` `o ``, it will jump to `d` (I will refer to this as behavior B - marks returning to the same character)

In this PR I have fixed behavior B, but haven't been able to fix behavior A, as you noted. It's difficult to get both of this behavior right in VSCodeVim for now because **VSCodeVim currently mixes these two types of changes together**. In cases like #4955 where there are, for example, two whitespaces at the beginning of both lines.

```
  1234
  abcde
```

If you set mark `o` at `d` and then you insert a newline above that line by pressing `O`

```
  1234
  |
  abcde
```

then at https://github.com/VSCodeVim/Vim/blob/41dbba15b7dd97185917b4c78261b71c3304d9ad/src/history/historyTracker.ts#L321 VSCodeVim will register the changes as `\n<space><space>`, starting from line 2, col 2, as if you have just moved the cursor to character `a` and then press `i` and `<enter>` and insert two whitespaces. So, to get behavior B right, I have to "sacrifice" behavior A.

I haven't looked at why VSCodeVim registers the `currentHistoryStep.changes` like that. The correct behavior for it would be to register the change as `\n<space><space>`, starting from line 1, col 6 (after character `4`). If that is fixed, both behavior A and behavior B can get right.

burgreen:
Behavior B that tracks on a character is not observed in vim.

When one puts a mark on character d in line:
`abcde`
it puts the mark in that line at COLUMN 4.

When one opens a line above this line, \`o will return to character `d` because the mark is in the marked line, and character `d` still resides in COLUMN 4. 

In vim, make any changes you wish to or around the line originally containing abcde, and `o will ALWAYS return to the originally marked line and COLUMN 4. Period. This IS proper vim behavior. Try it and see for yourself. 

I understand the desire to add cool new multi-line editing, but the proper behavior of vim should not be co-opted to attain that. Please choose some other key binding, but let vim behave as vim.

phphong:
> Behavior B that tracks on a character is not observed in vim, and I do not see Behavior A that returns to the original absolute file line observed either in actual vim. And maybe we are actually saying and seeing the same behavior, but describing it differently.

I have added GIFs to my previous comment to ensure that you don't misunderstand me.

> In vim, make any changes you wish to or around the line originally containing abcde, and `o will ALWAYS return to COLUMN 4 of the originally marked line that now contains some modified remnant of abcde. This IS proper vim behavior. Try it and see for yourself.

That is not correct. `` `o `` does not always return to column 4. Please see this example.

b2

I understand your request and I know it's right but I haven't yet found a way to fix it without affecting other proper vim function. I haven't had the time, but if somebody could take a look at `currentHistoryStep.changes` then it would be possible to fix this.

burgreen:
I see your example, but that behavior is because it is a column based mark.

For instance, if you make your joined line look like:
1234 xabcde
by adding an x before the a, \`o will go to the c character because it is a 4th column based mark.

And if you make the joined line look like:
1234 xxxxx
there is no longer a d character, but \`o will go to the 4th x, again because it is 4th column based mark, not character based mark.

J-Fields:
Thanks for the PR! At this point, there are lots of conflicts due to refactors and fixes that have happened. If you're still interested in getting this merged, please rebase and re-submit the PR, but for now I'm going to close it to help de-clutter the pull requests page.

Issue #4871, opened 2020-05-17
`yk` should move your cursor up


J-Fields:
From the paragraph after `:help blockwise-register`:

> Note that after a characterwise yank command, Vim leaves the cursor on the
> first yanked character that is closest to the start of the buffer.  This means
> that "yl" doesn't move the cursor, but "yh" moves the cursor one character
> left.
> Rationale:	In Vi the "y" command followed by a backwards motion would
> 		sometimes not move the cursor to the first yanked character,
> 		because redisplaying was skipped.  In Vim it always moves to
> 		the first character, as specified by Posix.
> With a linewise yank command the cursor is put in the first line, but the
> column is unmodified, thus it may not be on the first yanked character.

PR #4988, opened 2020-07-02
Fix multicursor yank motion

### LINKED ISSUE #4562: Bug: Multicursor - normal mode - yank word
**Describe the bug**
There is a small bug in normal Mode for multicursor, in combination of yank a word (ye, yw). These keys exit multicursor mode.

**To Reproduce**

1.
Have a text like this:
```
foo bar baz
foo bar baz
```

2. create two cursors on any word
3. exit visual / selection using `v`, so you are in normal mode with multicursor
4. goto begin of a word and press `ye` or `yw`
5. multicursor mode is exited

**Expected behavior**
Mode should stay normal with multicursor and word correctly yanked.

- VSCode version: latest
- OS: win7 64

There is a workaround: (use yiw) and it is only with yanking (ce or de work), so it is not a big deal imho.
Can another user try a repro, please?

### LINKED ISSUE #5239: Failed to handle key=p. v.split is not a function
**To Reproduce**

1.  Use multiple selection (cmd+D)
2.  press ESC to enter Visual Mode
3.  press yy to copy the hole line
4.  See error

**VSCodeVim version**: 1.17.0

```
TypeError: Failed to handle key=p. v.split is not a function
	at n.exec (/Users/carles.ballester/.vscode/extensions/vscodevim.vim-1.17.0/out/extension.js:9:409622)
	at runMicrotasks (<anonymous>)
	at async n.execCount (/Users/carles.ballester/.vscode/extensions/vscodevim.vim-1.17.0/out/extension.js:1:37418)
	at async n.execCount (/Users/carles.ballester/.vscode/extensions/vscodevim.vim-1.17.0/out/extension.js:9:411621)
	at async F.runAction (/Users/carles.ballester/.vscode/extensions/vscodevim.vim-1.17.0/out/extension.js:9:475314)
	at async F.handleKeyAsAnAction (/Users/carles.ballester/.vscode/extensions/vscodevim.vim-1.17.0/out/extension.js:9:474707)
	at async F.handleKeyEvent (/Users/carles.ballester/.vscode/extensions/vscodevim.vim-1.17.0/out/extension.js:9:471516)
	at async Object.promise (/Users/carles.ballester/.vscode/extensions/vscodevim.vim-1.17.0/out/extension.js:9:235653)
	at async Object.runTasks (/Users/carles.ballester/.vscode/extensions/vscodevim.vim-1.17.0/out/extension.js:9:621827)
```

</details>

### LINKED ISSUE #5460: Paste (single cursor) after multicursor yank is broken. Error: Failed to handle key=p. v.split is not a function
**Describe the bug**
Ok, today I hit this bug, too. So I could get to the bottom of it.

**To Reproduce**
1.  create multi cursor
2.  yank (either from visual mode or with a textobject).
3.  remove multi cursor with Esc
4.  paste

This will result in the above error.

**Expected behavior**
D'oh. ;-)

**Additional context**
The problem is that there is no handling for this situation in the code ( checking if register.text is an array or not). Pasting in insert mode (C-r") does this, so it is easy to adapt and fix the bug. I have update one of my PR.

sql-koala:
Fixes a bug for multicursor + yank operator. I reset this to just the bugfix. That is easier to review.

fixes #4562
fixes #5239
fixes #5460
(and duplicates)


sql-koala:
This is harder than I first thought.
This is the line that breaks multicursor:
`      vimState.cursors = vimState.cursorsInitialState;`
I can only guess that cursorsInitialState is not updated enough.

What works with multicursor is this assignment:
`      vimState.cursorStopPosition = start;`

But that generates correct cursor positioning only for forward movements (w, e, ...).  Problem are 
* that I dont know how to distinguish forward from backward movements
* and I dont know where to get an unchanged initial position for all cursors.

So the best I can come up with is this, like the first commit:
```
    if (originalMode === Mode.Normal && !moveCursor && !vimState.isMultiCursor) {
      vimState.cursors = vimState.cursorsInitialState;
    } else {
      vimState.cursorStopPosition = start;
    }
```
This passes tests (does not change single cursor), improves multicursor situation, and is correct for yanking forward motions (which is the more relevant usecase imo).
Thoughts, Feedback?

sql-koala:
Still WIP ;-)

sql-koala:
Hi @J-Fields this is not fully ready yet, but it is getting there. I can add tests for the new cursor positioning later, but I would need some feedback/help first.

sql-koala:
I have reset this to just the bugfix. This does not change any property of yank operator, that users could be used to.
If we want to address #4871, we can also provide #2397

src/actions/operator.ts
```diff
@@ -321,7 +321,14 @@ export class YankOperator extends BaseOperator {
     }
 
     if (originalMode === Mode.Normal && !moveCursor) {
-      vimState.cursors = vimState.cursorsInitialState;
+      // the reset in else would destroy multicursor
+      // so we set StopPosition (of this operator) to the stop position of our multicursor
+      // result: not moving the cursor
+      if (vimState.isMultiCursor && this.multicursorIndex) {
```
duffytilleman, line 327: Do you maybe mean `this.multicursorIndex !== undefined`?  As it stands this will evaluate false for the first cursor index (`0`)

sql-koala: Thanks for looking into this. I think you are right.
Funny thing: if I change the comparison like you propose, multicursor yank behaves different from when I submitted the PR. Need to look into this.

sql-koala: Thanks for the review. I did not know that this evaluates to false for cursor 0. But in practice, this is what works best, because the else branch works for the first cursor. But not for secondary cursors.

J-Fields (approved):
For some reason I'm still unable to reproduce, but this seems reasonable. Thanks!

PR #4989, opened 2020-07-03
Implement ":!" bang command

### LINKED ISSUE #3069: filter the highlighted lines through {command}
Taken from https://github.com/VSCodeVim/Vim/blob/master/ROADMAP.md

I use the `:%!` etc a lot. For instance I use `sort -u` on my import statements in python, or I use a bash script to generate commands etc. Adding this feature would be very nice :+1:

Thanks in advance,
Kfir.

### LINKED ISSUE #3136: Support "bang" operations on selected code blocks
With standard Vim I regularly use block-select to pick a group of lines and then pass them to a piped Unix filter, e.g. `:'<,'>!sort|uniq`.  As best I can tell, this isn't possible with the VSC Vim plugin today.

Support "bang" (`!`) operations when in command mode

I don't know another way to do this.  Right now if it is important enough I cut and paste the content into an empty document in Vim, do the operation there, then cut and paste the content back in to Visual Studio Code.

### LINKED ISSUE #3154: External Filter support
`:{range}!` is a common idiom for running an external filter like `awk` or `jq` for processing data.  Currently all commands fail with "Not an editor command"

- `:! foo` would execute `foo` and pipe the result to the Output tab
- `:<range>!foo ` would execute `foo`, using the specified lines as stdin, and replacing those lines in the file with the output from the command.  stderr would be piped to Output.

An alternative is to run a parallel instance of vim and perform those operations on the vim instance, relying on the fact that both vscode and vim will detect if the file changed, but that's not exactly a great solution 

**Additional context**
See [`:help filter`](http://vimhelp.appspot.com/change.txt.html#filter) for more info on the family of commands

### LINKED ISSUE #4920: Unable to use :!<command>
Unable to use :!<command>

Would be great to support this. I very frequently do things like :!git grep bla | grep -v bla | ...

tagniam:
This PR ports the [`:!{cmd}`](http://vimdoc.sourceforge.net/htmldoc/various.html#:!) bang command from Vim, allowing external shell commands to be run directly from Vim's command mode.

Key features:
* `:!{cmd}` executes the shell command
* `:{range}!{cmd}` executes the shell command, pipes the line range to `stdin`, and replaces the line range with the output from `stdout` and `stderr`
* Embedded bangs are replaced with the previous bang command (e.g. `:!dir` => `:!echo !` will execute `echo dir`)
* Cross-platform, supports Unix and Windows shells

Fixes #3154, fixes #3136, fixes #3069, fixes #4920

Some implementation details:
* Shell execution is all done through the new `ExternalCommand` utility class, which uses `child_process.exec()` for cross-platform exec
* Piping to stdin and piping from stdout/stderr is all done with temporary files (same as how Vim does it)
* Added a special case for `!` in the lexer, since `:! {cmd}` was getting parsed correctly but `:!{cmd}` wasn't


zimio:
Does this use the integrated terminal on vscode?

tagniam:
No, but I suppose you could use it for `:!{cmd}`. However, it wouldn't work for `:{range}!{cmd}`, since there is no way to capture command output from an integrated terminal due to the very limited vscode terminal api.

src/util/externalCommand.ts
```diff
@@ -0,0 +1,121 @@
+import { exec } from 'child_process';
```
J-Fields, line 1: node APIs should be isolated so they can be replaced at a later date (see #4981 and let me know if you've got any questions about that)

sjackman:
@tagniam Thank you so much for implementing my feature request! `!` was my number one missing feature in VSCodeVim.
I'm very much looking forward to the next stable release. Soon, I hope?

J-Fields:
> I'm very much looking forward to the next stable release. Soon, I hope?

I make no promises, but probably this weekend

sjackman:
No promises needed. I wouldn't be using VSCode if VSCodeVim didn't exist. Thank you, Jason!

sjackman:
Commented at https://github.com/VSCodeVim/Vim/issues/3136#issuecomment-660280988

PR #4990, opened 2020-07-03
Include "add surround" example in Readme

marcinbiegun:
### What this PR does / why we need it

I was struggling to find "add" case for surround feature. There were only "change" and "delete" examples provided in Readme. This PR adds "add" case to Readme.

### Which issue(s) this PR fixes

No issue linked.

### Special notes for your reviewer

No code changes, only readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -468,8 +468,9 @@ Based on [surround.vim](https://github.com/tpope/vim-surround), the plugin is us
 
 | Surround Command                     | Description                                                           |
 | ------------------------------------ | --------------------------------------------------------------------- |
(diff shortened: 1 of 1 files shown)


berknam:
The line `y s <motion> <desired char>` is the same that you added. But the existing description could definitely be better. Maybe put your line as an example of the existing one, where `iw` is the `<motion>`. The existing description should probably be something like 'Add desired surround around text defined by the \<motion\>'.

J-Fields:
5ad1c8063248bf6fbd1385102e22367c7ab4ffd8

PR #4995, opened 2020-07-07
Remap refactor get leader key original action and handle plugins through mappings

### LINKED ISSUE #2027: Cannot rebind alternate key to run leader key's original function

--------

**Environment**:

- **VSCode Version**: 1.16.1
- **OS**: Arch Linux, 4.12.13-1-ARCH

**What happened**:

These are the relevant key bindings/settings I have defined:
```
	{
		"before": ["t", "n"],
		"after": ["<Esc>"]
	}
],
	{
		"before": ["m"],
		"after": ["n"]
	}
],
"vim.easymotion": true,
"vim.leader": "n",
```

In a file I search for text by typing `/text<Enter>` and then press `m` and nothing happens. When I remove the `"vim.leader": "n",` line from my settings, this text searching works correctly.

I want my leader key to be `n`, and my "find next occurrence of search term" (normally this is `n`) to be `m`.

**How to reproduce it**:
See above.

**Other**:
There is always a chance this is already possible and I missed something with my config. It makes sense to me why this bug is occurring, although there are some strange interactions. When I press `n` in normal mode, I see `<leader>` in the VSCode footer, indicating that I pressed the leader key. When I press `m` in normal mode, I do not see this visual indicator. However, there are some strange inconsistencies here:

| Combination | Easymotion? |
|---|---|
|`nnb`| :heavy_check_mark: | 
|`nmb`| |
|`mnb`| :heavy_check_mark: | 
|`mmb`| |

So `m` is not _really_ bound to the leader key (but it sort of is) but it is certainly not bound to the original function of `n`, which is what I want.

### LINKED ISSUE #4909: How to go prev in f search when comma is the leader key?
How to go prev in f search when comma is the leader key?
the timeout seems doesn't work for me.

### LINKED ISSUE #4916: [Discussion] Getting original action of <leader> key
[@J-Fields This is a discussion issue to know what is the best way to proceed]

**The Issue:**
Currently the key you choose for your Leader key can't be used anymore. This is mostly ok when you use `<Space>` as the leader key because it's default action is exactly the same as `l`. But if you use some other key as the leader like `,` for example you can't get the original action of the `,` key which is MoveRepeatReversed. Even if you try to remap some other key to comma it will always end up remapped to `<leader>`.
I'm trying to solve this in my PR #4735 but I encountered some problems.

**The Cause:**
This happens because we have some Actions that use the `<leader>` key. The Actions in question are the ones from the Easymotion and CamelCaseMotion plugins. This plugins register some actions that start with `<leader>` key, which makes those actions behave like Vim commands.
In Vim, commands that have multiple characters like `'gg', 'gqap', 'f<character>', 'zf', 'q<register>', '"<register>'` after you get the first character of those commands the following characters can't be remapped (except the ones that accept a `<character>` like 'f', that can have language mappings but that is not important here). I have already implemented this in my PR. But because there are actions that use the `<leader>` key it makes this key behave like a vim multiple key command.

In Vim, all the plugins create functions and then create mappings that call those functions. For example, the CamelCaseMotion plugin creates a function like `CamelCaseMotion_w` and then you `map <leader>w <Plug>CamelCaseMotion_w`. So when you first press the `<leader>` key it sees there is a possible remap and waits until timeoutlen finishes, if there is no other key it performs the original key like `<Space>`.

In my PR implementation this would actually work like this if the plugins didn't exist! Or if the plugins implemented the 'couldActionApply' method to see if they are enabled first this would work if the user disabled the plugins. 😄
But because of those two plugins it waits for another key like if you had pressed 'g'.

**Solutions:**
1. I have a somewhat working solution that makes it so every remapping that is NonRecursive allows for the original action to trigger the 'doesActionApply'. This way you can remap in non recursive map `,n -> ,` when the leader is comma and if you press `,n` you get the original behavior of ','.
The problem with this implementation is if you want to remap a plugin key in non recursive like `vnoremap s <leader><leader>s` it won't work, you would have to make a recursive map like this `vmap s <leader><leader>s` which wouldn't loop because after the first `<leader>` key the others wouldn't be allowed to remap anyway.
And if you press the `<leader>` key it still waits for another key like the 'g' key does.
2. Implement a new timeout just for this key and for this specific situation (I don't like this, it is really hacky and stupid because we already have the correct behavior implemented. I'm just putting this here because it is an option)
3. (This I think should be the best option) Change the plugins to not use any key combination. Instead, we would create some new SpecialKeys like `<CamelCaseMotion_w>` or `<EasyMotion_SearchCharacter>`, and then we would create on configuration the default mappings to this keys like this:
```json
"normalModeKeyBindings": [
    {
        "before": ["<leader>", "w"],
        "after": ["<CamelCaseMotion_w>"]
    }
]
"operatorPendingModeKeyBindings": [
    {
        "before": ["<leader>", "w"],
        "after": ["<CamelCaseMotion_w>"]
    },
    {
        "before": ["<leader>", "i", "w"],
        "after": ["<CamelCaseMotion_iw>"]
    }
]
```
This would solve everything and also make it easier to rebind the plugin keys. And if someone decides to create another new plugin, they just have to create a new key for its action and create the default remappings on the default configuration. **(I'm happy to make this change in my PR if you think its worth it)**

**Issues Related:**
References #2027, #4909 issues
References PR #4735

### LINKED ISSUE #4973: Can't use the surround 'S' in visual mode, always defaults to ActionChangeLineVisualMode

berknam:
It creates a new way of handling plugins. Now the plugin's main actions are registered through the new `@RegisterPluginAction(pluginName)` which takes the new `BasePluginAction` with the new property `pluginActionDefaultKeys: string[]` which holds the keys that will be used to create the default remappings. The `keys` property now needs to have the plugin action name (the same as used by that plugin in Vim) as the first key, but can have more keys.

Examples:
*Plugin:* CamelCaseMotion
*Action:* Move to "camel word" begin
In Vim you would do `map <leader>w <Plug>CamelCaseMotion_w`, here we have:
```typescript
@RegisterPluginAction('camelcasemotion')
class MoveCamelCaseWordBegin extends CamelCaseBaseMovement {
  pluginActionDefaultKeys = ['<leader>', 'w'];
  keys = ['<Plug>CamelCaseMotion_w'];

```
Which when registering will create the default remapping `<leader>w` -> `<Plug>CamelCaseMotion_w`, here we consider `<Plug>CamelCaseMotion_w` as a single key.

*Plugin:* Sneak
*Action:* find 2 chars
In Vim you would do `nmap s <Plug>Sneak_s`, here we have:
```typescript
@RegisterPluginAction('sneak')
class SneakForwardNormalAndVisualMode extends SneakForward {
  keys = ['<Plug>Sneak_s', '<character>', '<character>'];
  pluginActionDefaultKeys = ['s'];
  modes = [Mode.Normal, Mode.Visual, Mode.VisualLine, Mode.VisualBlock];
}
```
Here even though we have 3 keys for the action we only map the `pluginActionDefaultKeys` to the first key of keys, so we get a remap of `s -> <Plug>Sneak_s` which when it gets to the action handler waits for two more keys of type \<character\>.

This was implemented on all plugins and since I was changing them I also added some features to them (this method actually makes it easier to add more features from these plugins) like the `<Plug>Sneak_t`, `<Plug>Sneak_T` from sneak and the keys 'cS', 'yS' (operator), 'ySs', 'ySS' (same as 'ySs') and 'gS' (visual mode) from surround plugin.

On register we create the default mappings but we add them to the mappings Maps when loading config depending on whether the plugin is active or not, we also don't add the plugin default mapping if the user already has a map for that key. Example: if the user has this map: `w` -> `<Plug>CamelCaseMotion_w` we don't create the mapping for `<leader>w`. This is a behavior that a lot of plugins do in Vim as well.

The result of all this is not only do we have a better implementation of plugins, we also get the original action of the \<leader\> key back. Now if you press the leader key it waits for timeout when it has potential mappings, but after timeout finishes it executes its original action. If you don't want to wait, you can now create a different NonRecursive mapping to the leader key. For example if you set your leader to `,`, you can do:
```json
    {
        "before": ["\\"],
        "after": [","]
    }
]
```
And this way you get the original action of `,` with the `\` key.

The next step with this is to be able to read the mappings from .vimrc that use `<Plug>`. I intend to do this by storing all the `<Plug>name` that we implement on the `RegisterPluginAction`, then the vimrc builder can use those stored implemented plugin actions to regex the vimrc for compatible plugin actions. This would mean that we would be able to read the following remaps from vimrc:
```vim
map w <Plug>CamelCaseMotion_w
omap iw <Plug>CamelCaseMotion_iw
nnoremap <leader>s" v<Plug>CamelCaseMotion_iw<Plug>VSurround"
```

fixes #4916
fixes #2027
fixes #4909
fixes #4973

This PR is based off of the PR #4735. When I tried to base this PR off of the other one it created this PR on my fork. But since I think this PR needs to get attention so people can give their feedback I'm adding it here too.
**This PR should not be merged before the PR #4735 is merged!**


berknam:
This PR doesn't change much for the user except the ability to remap plugin keys and getting the original leader key action back. But under the hood this PR makes plugins easier to maintain and implement. The next step would be to find a way of implementing a `getChar` function that the plugins could use to get a key from the user so that we don't need to create specific input modes like `SurroundInputMode` or `EasyMotionInputMode`. I've already made a sort of `getChar` with the Sneak plugin where it needs for other keys to come after the `<Plug>Sneak_s` key. This would probably work for the Surround plugin as well but not for EasyMotion.

This will also allow to create mappings for our implemented plugins by reading them from the user .vimrc file, which I intend to implement after this gets merged.

And if in the future there is any need to create custom plugins for VSCodeVim by creating other vscode extensions that implement our `@RegisterPluginAction(pluginName)` this PR lays work that might make that feasible, because the only way that will ever be possible is probably with remaps like Vim does it and like this PR does it.

sql-koala:
Hi @berknam 
I tried this branch with my cx plugin, but I could not get it to work. What I did:
* merge this branch
* change my code like so:
```
@RegisterPluginAction('Exchange')
export class ExchangeOperator extends BaseOperator {
  pluginActionDefaultKeys = ['c', 'x'];
  keys = ['<Plug>CExchange'];
  public modes = [Mode.Normal];
...
}
```
* when typing "cxw" the operator is not executed
* logging from remapper is this:
```
Remapper: debug: trying to find matching remap. keys=c. mode=Normal. keybindings=normalModeKeyBindingsMap.
Remapper: verbose: key=c. keySlice=c.
Remapper: debug: normalModeKeyBindingsMap. potential remap found. waiting for other key or timeout to finish.
Remapper: debug: trying to find matching remap. keys=c,x. mode=Normal. keybindings=normalModeKeyBindingsMap.
Remapper: verbose: key=c,x. keySlice=cx.
Remapper: verbose: key=c,x. keySlice=x.
Remapper: debug: normalModeKeyBindingsMap. potential remap broken. resending keys without allowing a potential remap on first key. keys=c,x
Remapper: debug: trying to find matching remap. keys=c. mode=Normal. keybindings=normalModeKeyBindingsMap.
Remapper: verbose: key=c. keySlice=c.
Remapper: debug: trying to find matching remap. keys=x. mode=Normal. keybindings=operatorPendingModeKeyBindingsMap.
Remapper: verbose: key=x. keySlice=x.
Remapper: debug: trying to find matching remap. keys=w. mode=Normal. keybindings=normalModeKeyBindingsMap.
Remapper: verbose: key=w. keySlice=w.
Extension Startup: debug: Selections: Ignoring selection: [28, 34; 28, 34], Count left: 0
```
* I have no other maps with cx or even c.

Any ideas?

berknam:
@sql-koala You also need to add the plugin name on the fil `src/configuration/validators/remappingValidator.ts ` on the following function:
```ts
  private isPluginActive(config: IConfiguration, pluginName: string | undefined) {
    if (!pluginName) {
      return true;
    } else {
      switch (pluginName) {
        case 'camelcasemotion':
          return config.camelCaseMotion.enable;
        case 'easymotion':
          return config.easymotion;
        case 'replacewithregister':
          return config.replaceWithRegister;
        case 'sneak':
          return config.sneak;
        case 'surround':
          return config.surround;
        // ----------------------------
        // This is what you need to add:
        case 'Exchange':
          return config.exchange;
        // ----------------------------
        default:
          return false;
      }
    }
  }
```
I really need to make this automatic as well, when registering the actions for a given plugin if it doesn't know about it yet it should create its configuration enabler. But I'm still thinking how to do this, because it would require that all plugins are enabled/disabled the same way and currently that's not true. I could create a `configuration.enabledPlugins: string[]` with the names of the plugins you want to enable, but that would require the user knowing their names which is not feasible, or we could make it so that the enabler setting was always something like `vim.plugin.pluginName.enable: true` where `pluginName` had to be exactly the same name as the one used when registering the plugin actions.

But for now it has to be done manually.

berknam:
Also if you want to be able to repeat your operator with `cxx` you need to add the `Mode.OperatorPendingMode` to the operator `modes` like so:
```ts
@RegisterPluginAction('Exchange')
export class ExchangeOperator extends BaseOperator {
  pluginActionDefaultKeys = ['c', 'x'];
  keys = ['<Plug>CExchange'];
  public modes = [Mode.Normal, Mode.OperatorPendingMode];
...
}
```

sql-koala:
Thanks @berknam  for the quick help; that worked.

berknam:
@J-Fields I've already rebased this to master, so it is ready to review and merge! But maybe we should wait for the feedback from #4735 once it gets released before going to this one.

PR #5010, opened 2020-07-10
Add support for smartRelativeLine

aymanbagabas:
Adds support for `smartRelativeLine` which basically toggles VSCode's lineNumbers between `on` and `relative` depending on insert mode.

https://github.com/VSCodeVim/Vim/issues/3020

### DIFF
--- a/package.json
+++ b/package.json
@@ -644,6 +644,11 @@
           "markdownDescription": "Enable the [ReplaceWithRegister](https://github.com/vim-scripts/ReplaceWithRegister) plugin for Vim.",
           "default": false
         },
+        "vim.smartRelativeLine": {
+          "type": "boolean",
+          "markdownDescription": "`#editor.lineNumbers#` is determined by the active Vim mode, absolute when in insert mode, relative otherwise.",
+          "default": false
+        },
         "vim.sneak": {
           "type": "boolean",
           "markdownDescription": "Enable the [Sneak](https://github.com/justinmk/vim-sneak) plugin for Vim.",
--- a/src/configuration/configuration.ts
+++ b/src/configuration/configuration.ts
@@ -195,6 +195,8 @@ class Configuration implements IConfiguration {
 
   replaceWithRegister = false;
 
+  smartRelativeLine = false;
+
   sneak = false;
   sneakUseIgnorecaseAndSmartcase = false;
   sneakReplacesF = false;
--- a/src/configuration/iconfiguration.ts
+++ b/src/configuration/iconfiguration.ts
@@ -137,6 +137,11 @@ export interface IConfiguration {
   replaceWithRegister: boolean;
 
   /**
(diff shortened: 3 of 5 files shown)


ahmedkhalf:
Hello. When is this going to be pushed? I am in great need of this feature.

revgum:
👍  Would love to see this merged!

revgum:
I gotta say, I've been using the zipped version of the extension with this feature for the past month(ish) and it's been a really nice addition.

ersel:
can we get this merged please? 🙏

ersel:
This is not something new. It matches the popular `vim-numbertoggle` plugin behaviour.

https://github.com/jeffkreeftmeijer/vim-numbertoggle

Since Vim 7.4 it is also possible to have this hybrid approach natively.

https://jeffkreeftmeijer.com/vim-number/

aymanbagabas:
> Thanks for the PR! Sorry for not getting to this sooner, I've been quite busy recently.
> 
> My only hesitation is a temptation to generalize. Can anyone think of other uses for a setting which would change some config when you switch modes?

Vim has an event system that recognized many [events](http://vimdoc.sourceforge.net/htmldoc/autocmd.html#{event}) including `InsertEnter` and `InsertLeave` which is what we are doing here. I would say the ideal way to generalize this is by emulating the event system that Vim has. Obviously, this is out of the scope of this PR.

J-Fields:
> This is not something new. It matches the popular vim-numbertoggle plugin behaviour.

Yup, I realize that! My inclination is just always to think about whether configuration should be more generic than the immediate use-case, so we don't end up creating bloat in configuration the next time we want a similar feature. Probably it's not worth it in this case.

> Vim has an event system that recognized many events including InsertEnter and InsertLeave which is what we are doing here. I would say the ideal way to generalize this is by emulating the event system that Vim has. Obviously, this is out of the scope of this PR.

Yeah, thanks for bringing this up! I agree that this is the correct long-term implementation goal, and also that it's too big a lift to worry about for now.

package.json
```diff
@@ -644,6 +644,11 @@
           "markdownDescription": "Enable the [ReplaceWithRegister](https://github.com/vim-scripts/ReplaceWithRegister) plugin for Vim.",
           "default": false
         },
+        "vim.smartRelativeLine": {
+          "type": "boolean",
+          "markdownDescription": "`#editor.lineNumbers#` is determined by the active Vim mode.",
```
J-Fields, line 649: Could you be a bit more specific in the description, to make it clear that it'll be absolute in Insert mode and relative otherwise?

aymanbagabas: Rebased and updated.

J-Fields:
Awesome. Thank you @aymanbagabas!

revgum:
Thanks @J-Fields and @aymanbagabas for getting this shipped!

PR #5021, opened 2020-07-14
Allows :ju and :jumps command to work. 

zimio:

Allows the parsing of the jump command. This opens a small dialog where the user can select which place to jump to.

None that I'm aware of.


src/cmd_line/subparser.ts
```diff
@@ -83,7 +82,11 @@ export const commandParsers = {
 
   jumps: {
     abbrev: 'ju',
-    parser: parseJumpsCommandArgs,
+    parser: () => new JumpsCommand(),
+  },
+
+  jump: {
```
J-Fields, line 88: This shouldn't be necessary - the `abbrev` above should make everything between `ju` and `jumps` parse to `JumpsCommand`

zimio: Vim handles three aliases: ju, jump and jumps for the same command. In order to follow vim as close as possible I added 'jump'.

J-Fields: You'll find that `jum` works as well 🙂 Using `abbrev` makes everything between `abbrev` and the full name work, so without this entry into the map, `jump` will work (as will `jum`)

zimio: Oh, I didn't know that. That's pretty great. I have removed it now.

src/cmd_line/subparser.ts
```diff
@@ -79,6 +80,15 @@ export const commandParsers = {
     parser: parseHistoryCommandArgs,
   },
 
+  jumps: {
+    abbrev: 'ju',
+    parser: () => new JumpsCommand(),
+  },
+
+  clearjumps: {
```
J-Fields, line 88: This should have an `abbrev` of `cle`

src/cmd_line/commands/jumps.ts
```diff
@@ -0,0 +1,55 @@
...
+import { VimState } from '../../state/vimState';
+import { globalState } from '../../state/globalState';
+import { Jump } from '../../jumps/jump';
+import { Range } from '../../common/motion/range';
+
+class JumpPickItem implements QuickPickItem {
+  jump: Jump;
+
+  label: string;
+  description?: string | undefined;
```
J-Fields, line 13: These `| undefined`s are unnecessary because you've got the `?`s

src/cmd_line/subparser.ts
zimio: That is true. I have added it.

src/cmd_line/commands/jumps.ts
zimio: I have remove them.

PR #5023, opened 2020-07-14
Fix textobject: a paragraph ending position (when paragraph is single line)

### LINKED ISSUE #5012: `dap` deletes more than one paragraph
**Describe the bug**
`dap` deletes more than one paragraph

**Expected behavior**
`dap` deletes one paragraph

**Screenshots**

- Extension Name: `vim`
- Extension Version: `1.14.5`
- OS Version: `Windows_NT x64 10.0.19041`
- VSCode version: `1.48.0-insider`

sql-koala:

This PR fixes a small bug, also described in the linked issue.
Right now, when you operate on "ap"  (cap, dap, vap, ...) and the paragraph is a single line, vscodevim will operate till EOF.

Fixes #5012 

I think this small bug is pretty obvious (thats why I could fix it ;-)
The while-loop, that adds empty lines at the end of a paragraph checks against position "start" while update/moving stop. So it is an "infinite" loop, that will always go to EOF (other condition that ends while).
The original condition is always true (for a single line paragr) because then, the next line is empty.
`      TextEditor.getLineAt(start.getDown()).isEmptyOrWhitespace`


J-Fields:
Thanks @sql-koala! Would you mind writing a test case for this?

sql-koala:
@J-Fields I kind of expected that ;-)
The right file should be:
test\mode\modeNormal.test.ts

I can adapt 1,2 test from there.

I have problems getting tests to work (as a part of building on my machine), so I would need to use the PR&travis for it.

J-Fields:
What's the problem with the tests? In the meantime, feel free to just push what you think is right and Travis will correct you

sql-koala:
I found the same issue/line of code with "dip". Will fix that tomorrow.

J-Fields:
Sweet sounds good. I'll hold off on merging til then

sql-koala:
Technically, it is tomorrow in my time zone ;-)
That should be the fix for dip.

On that test build, only the second test "Can handle dip on empty lines" was supposed to fail (that test for bug we fix), for the other test I had to correct the cursor position.
https://travis-ci.org/github/VSCodeVim/Vim/jobs/708147264

PR #5040, opened 2020-07-17
add :sh command 

zimio:

Add :sh command by allowing the creation of a new external terminaling the built-in terminal of vscode.

None that I'm aware of.

Very simple change which emulates the behavior of vim.


src/cmd_line/subparser.ts
```diff
@@ -202,6 +203,10 @@ export const commandParsers = {
     parser: parseOptionsCommandArgs,
   },
 
+  sh: {
```
J-Fields, line 206: This should be `shell` with `abbrev: 'sh'`

zimio: Yes, this is true. I have changed it now.

Issue #4806, opened 2020-05-05
`d<count>gg` should be linewise


J-Fields:
Actually, according to `:help`, it seems that the only difference between `<count>gg` and `<count>G` is that the latter is a jump-motion.

PR #5046, opened 2020-07-19
Fix gg as linewise operation

sql-koala:
This makes gg a linewise operation.

Fixes: #4806

Changed one existing test, because it probably does not fail if operation is not linewise.
Since you mentioned it in #4806: here, gg is also a jump. I did not change that.


J-Fields:
> Since you mentioned it in #4806: here, gg is also a jump. I did not change that.

`:help` mentions that `G` is a jump command, and does not mention that for `gg`, but testing it in neovim, they act the same as far as I can tell 🤷‍♂️

J-Fields:
> Changed one existing test, because it probably does not fail if operation is not linewise.

I think it has to do with `startofline`, actually

PR #5047, opened 2020-07-19
Fix bug on issue #5029

### LINKED ISSUE #5029: Visual mode cannot be navigated with up/down arrow keys
**Describe the bug**
Visual mode cannot be navigated with up/down arrow keys. I guess this is a good time to get used to h/j/k/l :sweat_smile: 

**To Reproduce**

1.  Press v to enter visual mode
2.  Move left/right with arrow keys, works. 
3.  Move up/down with arrow keys, exits visual mode into normal mode

- Extension (VsCodeVim) version: 1.15.0
- VSCode version: 1.47.1
- OS: Ubuntu

berknam:
Change the up/down arrow keys to check only if the 'parameterHints' or 'suggestWidget' are visible because the `parameterHintsMultipleSignatures` and the `suggestWidgetMultipleSuggestions` stay true even when the widgets aren't visible.

fixes #5029

### DIFF
--- a/package.json
+++ b/package.json
@@ -113,12 +113,12 @@
       {
         "key": "up",
         "command": "extension.vim_up",
-        "when": "editorTextFocus && vim.active && !inDebugRepl && !suggestWidgetVisible && !suggestWidgetMultipleSuggestions && !parameterHintsMultipleSignatures"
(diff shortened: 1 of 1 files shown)


J-Fields:
Gah, that's dumb. Thanks for tracking this one down 🙃

PR #5049, opened 2020-07-20
Fix visual mode not preserving desiredColumn

### LINKED ISSUE #5048: Visual mode doesn't keep desiredColumn correctly
**Describe the bug**
If you enter visual mode and then move up and down it keeps the desiredColumn correctly. But if you move horizontally and then move up/down the cursor always goes one to the right of the expected desiredColumn.

**To Reproduce**

Start text:
```json
|one
two
```
'|' is the cursor

1.  Press 'v' to go to visual mode
2.  Press `l` to move right
3.  Press 'j' to move down
4.  It moves down and right

End result:
```json
one
tw|o
```
'|' is the cursor

**Expected behavior**
Desired column should have been kept.

Expected result:
```json
one
t|wo
```
'|' is the cursor

berknam:
Fix visual mode not preserving desiredColumn
- When moving horizontally on visual mode the desiredColumn wasn't being kept correctly, because we were updating the desiredColumn after correcting our cursors by shifting the stop to the right so that vscode included our stop on its selection.
- Add test for this.

fixes #5048 

Why are we changing our cursors just to make vscode selection include the last character? Couldn't we just set the editor selections with the adjusted positions on 'updateView' without changing our cursors?
Also we are doing this adjust on 'runAction' when the start is before stop but then on 'updateView' we also make another adjust when start is after stop (changing our cursors once more).

### DIFF
--- a/src/mode/modeHandler.ts
+++ b/src/mode/modeHandler.ts
@@ -477,6 +477,10 @@ export class ModeHandler implements vscode.Disposable {
     }
     */
 
+    // We handle the end of selections different to VSCode. In order for VSCode to select
+    // including the last character we will at the end of 'runAction' shift our stop position
+    // to the right. So here we shift it back by one so that our actions have our correct
+    // position instead of the position sent to VSCode.
     if (vimState.currentMode === Mode.Visual) {
       vimState.cursors = vimState.cursors.map((c) =>
         c.start.isBefore(c.stop) ? c.withNewStop(c.stop.getLeftThroughLineBreaks(true)) : c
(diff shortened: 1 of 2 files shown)


J-Fields:
Thanks!

> Why are we changing our cursors just to make vscode selection include the last character? Couldn't we just set the editor selections with the adjusted positions on 'updateView' without changing our cursors?

Yep, that should be doable. This is one of my pet peeves about VSCodeVim - that adjustment is made in a number of places all over the code base, so it's not trivial to change the representation of visual cursors without being sure that nothing else is broken, but it would definitely get rid of some technical debt.

PR #5050, opened 2020-07-20
Update cursors with 'keyboard' kind Selections

### LINKED ISSUE #2094: Selection removed on ctrl+shift+right
The problem is that with vim mode enabled on ctrl + shift + right selection appears for a moment and then disappears. 
I tested it without vim plugin enabled and it works fine then so its not vscode problem. 
I have attached gifs bellow one is showing the problem second one is done in SublimeText and displays how it should work.
**Environment**:

- **VSCode Version**: 1.17.2
- **VsCodeVim Version**: 0.10.2
- **OS**: Windows 10 Pro 1709 x 16299.19

**What happened**:
vscode issue multicursor

vscode issue

Thanks.

berknam:
- Selection change events not triggered by us of type 'Keyboard' will
now enter visual mode (if they are not empty selections).
- This allows entering visual mode with shift+arrowKeys or shift+end,
shift+home even when on insert mode and with multi cursors.

fixes #2094
related to #1806

~~This PR is built on top and depends on PR #5015 and should not be merged before that on is merged first.~~ This has now been rebased to master and is ready to merge.


J-Fields:
@berknam Could you resolve the conflicts when you have a chance?

berknam:
I had to rebase this correctly. Since github uses the squash and merge it no longer recognizes the previous commits. So I had to rebase it to master. But should be good now.

berknam:
There was a bug when shift selecting to left/up and then making other movement without shift that I've now fixed. I had previously fixed it and then commented that out because the way I fixed it was by using the `ignoreIntermediateSelections` flag, which I didn't like much. Now I'm calling the `updateView` function letting it draw the selection. This way any resulting selection change will be properly ignored. I think it is a better way to deal with that.
I've also added tests for this situation, including multicursor tests.

stevenguh:
I maintain the VSpaceCdoe extension which try to mimic spacemacs experience in vscode. The extension has a binding to call smart selection grow with a key binding of `<spc> v`. With #5015, it fixes the issue with visual mode when using smart grow selection that's called by the extension, however, it seems to have a regression with the change in this PR.

Screen Recording

The bug doesn't occur when it is triggered by shortcut directly. Is there any way I can help debug this?

berknam:
@stevenguh Have you tried it with the latest version from this PR? Because I tried using the same remap just now and it works fine. There was a bug that I fixed on the last commit that might have been creating the issue you're seeing.

stevenguh:
I think I figured it out. This is because in PR changed 
https://github.com/VSCodeVim/Vim/blob/10701cf5ac66d93056632ee0dc9437992616948a/src/mode/modeHandler.ts#L242-L245
from 
```
await this.updateView(this.vimState, { drawSelection: false, revealRange: false }); `
```

The issue is more involved. VSpaceCode is the a command that binds to `vim.normalModeKeyBindingsNonRecursive` and `vim.visualModeKeyBindingsNonRecursive`. The command return a Promise which awaits until the QuickPick menu exists. Because the command is executed by vim's key bindings, vim will "freeze" until the QuickPick menu from VSpaceCode exits. Therefore, all the commands like smart select grow executed in VSpaceCode's QuickPick menu are not reflected until the menu exits. When the menu exits, the selection operations are send to Vim to handle.

One fix I can do on my side is to not return a promise, so vim is not blocked because of keybinding execution.

berknam:
That change doesn't seem to be the issue. I've tested this with your extension and that part of the code is never hit until you exit the QuickPick menu. The selection change events are being enqueued but they are not called until you exit the quickpick menu.

I've tested with the code from PR #5015 and the behavior from that PR and this PR are the same for me.

stevenguh:
That's super weird. Changing that line definitely fixed it for me, and I am on 10701cf.

My steps to reproduce are
1. `<spc> v` to do smart grow
2. `v` to grow again
3. `esc`
EXPECTED: the correct highlighting
ACTUAL: the first selection is highlighted

This is a highlighting problem for me, any operation after that will result in correct highlighting like move the cursor to the left `h` or yank `y`.

berknam:
I see the problem. When creating the visual selection we need to change vscode selection so we need to call that line allowing the draw selection so that we ignore any resulting selection change events properly. Then when updating the selection since we are not changing it there was no need to call the `updateView` with draw selection. But in your case, all the events are coming at the end all in row, so what happens is we create the visual selection and change the vscode selection to the one we want (the initial selection) the on the following events we update our selection but don't change the vscode selection leaving it at the initial one. All we need to do is update vscode selection when updating our selection as well.

I'll push a fix for that in a moment.

berknam:
@stevenguh That should be fixed now. At least the correct selection is shown after you exit the QuickPick menu, but you should probably not block until the user exits the QuickPick menu. Why do you need to await that promise?

stevenguh:
That makes sense. Thank you for looking into this. I will update the extension later today so it's not blocking.

berknam:
> 
> 
> That makes sense. Thank you for looking into this. I will update the extension later today so it's not blocking.
> 
> EDIT: The extension (`vscode-which-key`) which hosts the command VSpaceCdoe calls to create the QuickPick menu is updated to `v0.7.4` so the menu is not blocking

Nice. That looks better.

@stevenguh By the way just a suggestion, but if you want the selection to look better (currently in some themes the inactive selection is barely noticeable) you can try to change the setting `"editor.inactiveSelectionBackground"` from `"workbench.colorCustomizations"` to `ThemeColor('editor.selectionBackground')` when opening the QuickPick menu and then change it back to whatever the user had it set or undefined if it wasn't set when closing the QuickPick menu. I haven't tested this but it should work.

stevenguh:
Thank you for the suggestion. I created an issue in `vscode-which-key` to track the work 👍 so we don't distract the convo in this PR.

stevenguh:
Sorry to piggyback on this. With the latest vim update, the smart select from normal to visual mode works sometimes, and doesn't work robustly. I am wondering is this what we needed to have a robust smart selection to visual mode? Or should I be opening a new bug to track that.

berknam:
> 
> 
> Rough repo:
> 
>     1. Open a C# file
>     2. Smart select from normal to visual mode still works at this point
>     3. Type something to trigger an error in syntax
>     4. Back to normal mode
>     5. Smart select failed to select in visual mode

I can't reproduce this. When this happens check the developer tools console to see if you have something like this:
```
[Extension Host] Extension Startup: debug: Selections: Ignoring selection: [37, 56; 37, 56], Count left: 0
```
That "Count left: 0" is the important part, it is supposed to always get back to 0, if it is always higher than 0 then something wrong happened and smartSelect won't work (including some other weird issues will show up). Now this usually happens when working with multicursors or snippets that create multicursors. Perhaps you typed some snippet? If that wasn't the case and you actually find a way to reproduce this without using multicursors or snippets please tell me!

stevenguh:
This is weird. I can't consistently reproduce with those steps today, but I can still hit bug from time to time. I am not using multi cursor or snippets. I also checked out the log when I hit the bug, and the diagnostic array is empty. Also I can't find a way to recover it (meaning to have the smart selection working again for that editor) once it hit the bug. Even closing the file and reopening won't fix it. I have to restart the window. I will keep your posted if I found a consistent way to repo this.

berknam:
> 
> 
> I also checked out the log when I hit the bug, and the diagnostic array is empty.

Confirm if you have the setting `"vim.debug.loggingLevelForConsole": "debug"`

stevenguh:
I just turned that option on, it is still empty when I am in the bug. I realized that it also happened in other type of files as well not just C# editor. I don't have a consistent repo yet.

J-Fields:
@berknam Could you please resolve the merge conflict when you get a chance?

berknam:
> 
> 
> @berknam Could you please resolve the merge conflict when you get a chance?

I'm looking into this, but it wasn't a simple fix the conflict as I expected. We've made some other changes with selections required by some other issue that showed up in mean time and I need to see the best way to not create that issue again.

berknam:
@J-Fields I've been thinking about this and I don't believe this is the right approach to this issue. I think it would be better to implement actions for the `<S-right>`, `<S-left>`, `<S-up>`, `<S-down>`, `<S-home>`, `<S-end>` keys that would behave according to 'keymodel' (this config would need to be implemented as well).

If we had 'keymodel' empty the shift right and left keys would behave like `w` and `b` respectively and the shifted home and end keys would behave like normal home and end keys.
If we had 'keymodel=startsel', then using this keys would start a selection and go to visual mode.
If we had 'keymodel=startsel,stopsel', then using this keys would start a selection and go to visual mode, but using the unshifted version of this keys would get out of visual mode.

This way it would make this ready for when we implement SelectMode, since then we would only have to implement the 'selectmode' config and if that one included 'key' while 'keymodel' included 'startsel' the those keys would start a selection in SelectMode instead of VisualMode.

This would also have the benefit that since we would start capturing these keys, then they will be available for remaps as well.

If you agree that this is the better way, then I can change this PR or create a new one. If not, I can try to fix this PR as it is just to get it working.

berknam:
Another benefit would be removing this logic from the 'handleSelectionChange' function and reduce its complexity! 😄

J-Fields:
Ah nice, I wasn't aware of `keymodel`. This does sound like the right way to go about it - feel free to update this PR or scrap it and make a new one.

> Another benefit would be removing this logic from the 'handleSelectionChange' function and reduce its complexity! 😄

Because we'd handle all these through custom commands and be able to ignore `keyboard` selection events?

berknam:
> Because we'd handle all these through custom commands and be able to ignore `keyboard` selection events?

Mostly yes! I would have to look still, because there are some instances that are marked as 'keyboard' selection events (like calling a `cursorMove` command I think). But when we are the ones calling those commands the corresponding selection change events are already ignored, so if we get one of those it will probably be either from another extension or from a custom user keybinding and in those cases it makes sense to overwrite our selection according to the new selection like we are doing now with the 'command' type selection events (like the `editor.action.smartSelection.grow` command for example).

I do have currently a potential fix for this PR that would make it work, but it is a little bit complex and for it to work well with multicursors I would have to copy that complexity to the multicursor logic as well. Also it doesn't keep the desired column when moving up/down and doesn't have any configuration tied in to let users opt out of it. Since it is partially done, if you want I can finish that and push it here, and then create a new PR for the other option. *(I'm proposing this because I don't have any idea of how long it might take me to make the other option)*

J-Fields:
Closing as my understanding is that this is superseded by #5842. Please let me know if I'm mistaken, @berknam

Issue #2349, opened 2018-01-31
New feature request: Support for special marks


npbool:
**Is this a BUG REPORT or FEATURE REQUEST?** (choose one): FEATURE REQUEST

**What happened**:
\`], \`[, \`>, \`< don't work
**What did you expect to happen**:
As stated in http://vim.wikia.com/wiki/Using_marks,
\`[ or \`] should jump to beginning/end of previously changed or yanked text,
\`< or \`> should jump to beginning/end of last visual selection.

**How to reproduce it (as minimally and precisely as possible)**:
yank some text and press \`] or \`[. Visual select some text and deselect it, then press \`< or \`>

**Environment**:

- Extension (VsCodeVim) version: 0.10.13
- VSCode version: 1.19.3
- OS version: macOS High Sierra 10.13.2

tomdl89:
I would also like to see this feature!

J-Fields:
Marking as duplicate of #4734, as I think that's the missing gap at this point.

PR #5056, opened 2020-07-20
Support visual mark movements

### LINKED ISSUE #2690: Support for `'>` and ``>` to go to end of last selected region

Feature request.

**What happened**:

I pressed `'>` or `BACKTICK>` in Visual Studio Code with the VsCodeVim plugin installed after I've previously selected something in visual mode.

I expected to be moved to the end of the last visual region that I selected.

* Select something in visual mode.
* Exit visual mode (e.g. press escape)
* Move around a bit
* Press `'>` or `BACKTICK>`
* Nothing happens

**Background**:

I'm really missing the `'>` (and it's associated `BACKTICK>`) command in VSCodeVim.

This is from the vim manual (`:h '>`):
```
							*'>* *`>*
'>  `>			To the last line or character of the last selected
			Visual area in the current buffer.  For block mode it
			may also be the first character of the last line (to
			be able to define the block).  Note that 'selection'
			applies, the position may be just after the Visual
			area.
```

For example, if you have this code in the editor:

```rust
fn some_func() {
    42
}
```

then you can type `vipy'>p` (view-in-paragraph, yank-visual-area, go-to-end-of-last-visual-area, paste) to make a copy of it:

```rust
fn some_func() {
    42
}
fn some_func() {
    42
}
```

**Environment**:

- Extension (VsCodeVim) version: v0.12.0
- VSCode version: April 2018 (version 1.23)
- OS version: MacOS High Sierra

sql-koala:
This adds support for the visual mark motions: '<, '>. Character variants, too.

Fixes #2690
related to #2349
But does not fix it completely, as the issue is also about ']

Not much, we "calculate" the new position to jump to from vimState.lastVisualSelection on the fly.
From my perspective, this is good to go.


src/actions/motion.ts
```diff
@@ -447,6 +447,73 @@ class CommandPreviousSearchMatch extends BaseMovement {
...
+  private startOrEnd(lastVisualSelection): Position {
+    // marks from vimstate are sorted by direction of selection (moving forward vs backwards).
+    // must sort to document order
+    let [start, end] = sorted(lastVisualSelection.start, lastVisualSelection.end);
+    return this.mark === VisualMark.SelectionStart ? start : end;
+  }
+
+  private inLineCorrection(position: Position): Position {
+    // for ' mark, we must go to BOL.
+    // for `> mark, we must correct by one char left
```
sql-koala, line 469: I don't like this nested ?, but that is what the linter does.

src/actions/motion.ts
```diff
@@ -447,6 +447,73 @@ class CommandPreviousSearchMatch extends BaseMovement {
...
+    return this.registerMode === RegisterMode.LineWise
+      ? position.getLineBeginRespectingIndent()
+      : this.mark === VisualMark.SelectionStart
+      ? position
+      : position.getLeft();
+  }
```
sql-koala, line 475: For end of visual selection, we have to correct one char to the left.
Is this linked to
https://github.com/VSCodeVim/Vim/pull/5049#issue-452966423 ?

sql-koala:
Btw, this is also affected by this: (but the existing motions, too).
https://github.com/VSCodeVim/Vim/issues/5057

src/actions/operator.ts
```diff
@@ -304,11 +304,6 @@ export class YankOperator extends BaseOperator {
 
     Register.put(text, vimState, this.multicursorIndex);
 
-    if (vimState.currentMode === Mode.Visual || vimState.currentMode === Mode.VisualLine) {
-      vimState.historyTracker.addMark(start, '<');
-      vimState.historyTracker.addMark(end, '>');
-    }
-
     await vimState.setCurrentMode(Mode.Normal);
```
sql-koala, line 307: This is probably old code. lastVisualSelection is tracked in vimState and motions must access this, we dont need/sould delete this.

src/actions/motion.ts
J-Fields: Yeah I think so. It's fine for now

J-Fields:
The column doesn't seem to be right for visual line mode, but this looks good other than that. Thanks @sql-koala!

sql-koala:
Yay, merged 👍

PR #5066, opened 2020-07-22
Support % jump for < and >

### LINKED ISSUE #4855: Extend '%' function's pairings
I've been looking for a way to extend the % function to include '<' / '>', and potentially others. So far I haven't been able to figure out how.

Perhaps an option like vim.addPairings or whatever name? I think it'd just be making this part of code accessible through the settings, but I'm unsure about how much work it'd actually take.

I tried enabling the .vimrc settings and adding `set matchpairs+=<:>` but it didn't work. I'm fairly new to vim tinkering so I might've missed something.

Is this feasible or is it too complicated? Perhaps if there's another way of achieveing it, I'd appreciate the pointers.

Thanks in advance!

sql-koala:
This adds support for % to jump between "<" and ">".
It add's a configuration vim.matchpairs that you would need to set to "<:>" for this.
Default % is not changed.

Fixes #4855

It was relatively straightforward ,-)


src/actions/motion.ts
```diff
@@ -1486,8 +1486,23 @@ class MoveToMatchingBracket extends BaseMovement {
...
+
+      // we need to check pairing, because with text: bla |bla < blub > blub
+      // this for loop will walk over bla and check for a pairing till it finds <
+      if (pairing) {
+        // construct string for matchpair: <:>
+        const direction1 = currentChar.concat(':').concat(pairing.match);
+        const direction2 = pairing.match.concat(':').concat(currentChar);
+
+        matchesWithPercentageMotion =
+          pairing.matchesWithPercentageMotion ||
```
sql-koala, line 1501: I changed this block of code mostly because I  have to query configuration here. I did not get it to work in the static Pairing class.

src/actions/motion.ts
```diff
@@ -1486,8 +1486,23 @@ class MoveToMatchingBracket extends BaseMovement {
...
+          configuration.matchpairs.includes(direction1) ||
+          configuration.matchpairs.includes(direction2);
+      }
```
sql-koala, line 1504: I changed this block of code mostly because I  have to query configuration here. I did not get it to work in the static Pairing class.

sql-koala:
I think its a lot better now.

package.json
```diff
@@ -543,6 +543,10 @@
           "markdownDescription": "Override the `ignorecase` option if the search pattern contains upper case characters.",
           "default": true
         },
+        "vim.matchpairs": {
+          "type": "string",
+          "markdownDescription": "additional chars to match as pairs. currently only <:>"
```
J-Fields, line 548: This should match what you get from `:help matchpairs`:

> Characters that form pairs. The `%` command jumps from one to the other.

src/configuration/configuration.ts
```diff
@@ -189,6 +189,8 @@ class Configuration implements IConfiguration {
 
   autoindent = true;
 
+  matchpairs = '';
```
J-Fields, line 192: Add `['mps', 'matchpairs']` to `optionAliases` above

sql-koala: I don't understand, sorry. I do not find any example of an alias (like hls <-> hlsearch) in the repo.

sql-koala: ok, I see it now. this was not present in master when i created this branch.

yunti:
Anything preventing merging this as it would be great to have.  Thanks

J-Fields:
Thanks @sql-koala!

PR #5068, opened 2020-07-23
Improve digraph support.

adlerd:
The most important change in this PR is to support digraphs in commandline and search, which vim 8.2 supports. To make digraphs more user-friendly, also report them in the 'ga' action (character code reporting) like vim. Finally, as a side-effect match digraphs in what seems to me a more intuitive order where user config always overrides defaults, and reverse digraphs are only used as a last resort.

No issue opened.

I have literally never written Javascript/Typescript before, let alone changed a Code extension.

Full disclosure, I also think it would be better to precompute user shadowing of defaults and (less importantly) reverse digraphs, but (a) it's not clear if it's possible to generally react to config changes; (b) it makes the big picture of digraph support harder to understand, even though it would entirely eliminate the growing complexity of individual functions. So for a first PR I avoided this.

### DIFF
--- a/src/actions/commands/actions.ts
+++ b/src/actions/commands/actions.ts
@@ -34,6 +34,8 @@ import { globalState } from '../../state/globalState';
 import { VimError, ErrorCode } from '../../error';
 import { SpecialKeys } from '../../util/specialKeys';
 import _ = require('lodash');
+import { digraphsForChar } from './digraphs';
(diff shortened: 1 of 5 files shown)


J-Fields:
@adlerd Would you mind resolving merge conflicts when you have a chance?

adlerd:
Unfortunately I'm no longer able to build the extension according to the given instructions, so unless someone can champion this PR themselves I feel I have to withdraw it...

J-Fields:
Note to self: this would fix #6035

PR #5071, opened 2020-07-23
Neovim config support

### LINKED ISSUE #5045: Surround and other plugins not working with Neovim Ex commands
**Describe the bug**
Ex commands that use surround don't seem to work from VSCode.
For example, if I try to surround a word with quotes using the command line I would try:
`:norm ysiw"`
This command works as expected when I use it in nvim itself. I have the surround plugin active in VSCode and installed in Neovim.
I do have Neovim enabled in VSCodeVim and most other things seem to be working great.

Not sure how plugins are loaded in Neovim etc. (or how they work at all really). Could this have something to do with plugins not being active in whatever manner VSCode is executing Neovim?

Also I just tried using the sneak plugin in the same way to see what would happen.
I have the same keybindings set up in VSCode and Neovim, so `s` starts sneak rather than substitute in both editors from normal mode.
If I use `:norm sva` in nvim, it sneaks to the letters 'va'. If I use the same command from VSCode it starts substituting and then types the characters 'va', as it would if I didn't have sneak installed.

By the way, other :norm commands seem to work fine from VSCode i.e. `:norm d2w` does delete two words.

Thanks for your work everyone

**Environment:**

- Extension (VsCodeVim) version: 1.15.0
- VSCode version: 1.48.0-insider
- OS: Windows 10
- Neovim version: 0.5.0-594-ga02a267f8

austinclem1:
Allows Neovim to optionally use its own config file, whether that file is specified by the user in VSCode settings, or the user lets Neovim look for a config in its default search path.

fixes #5045

I'm pretty new to this so let me know if anything needs to change.

I added code in `src\configuration\validators\neovimValidator.ts` that determines whether the user's specified Neovim config file exists. It issues a warning and sets the path to empty string if the file doesn't exist.
If that code shouldn't be put in `neovimValidator.ts` let me know if there's a better spot.
Also there could be issues if that file DOES exist, but it's invalid and Neovim fails to load it.

Also there are failing tests, but I think they're the same tests that failed prior to adding this feature.
**Edit:** Just realized those tests fail on windows but not on my linux setup


J-Fields (approved):
LGTM. Thanks a lot!

PR #1939, opened 2017-07-22
Makes all tests pass on Windows

philipmat:
Fixes #1914 

https://ci.appveyor.com/project/philipmat/vim/build/1.0.10

Issue #1914, opened 2017-07-09
Tests failing on master on Windows 


philipmat:
I'm creating this issue to track these four tests failing on Windows 10 on `master`.
I'll try working on them one at a time.

```
 1) read Can read shell command output:

      AssertionError: Content does not match; Expected=hey. Actual=hey
      + expected - actual

      -hey
      +hey

      at Object.assertEqualLines (C:\dev\vscodevim\Vim\test\testUtils.ts:28:12)
      at Object.<anonymous> (C:\dev\vscodevim\Vim\test\cmd_line\read.test.ts:20:5)
      at Generator.next (<anonymous>)
      at fulfilled (C:\dev\vscodevim\Vim\out\test\cmd_line\read.test.js:4:58)

  2) Basic sort Sort whole file, dsc:

      AssertionError: Line count does not match.
      + expected - actual

      -4
      +3

      at Object.assertEqualLines (C:\dev\vscodevim\Vim\test\testUtils.ts:35:10)
      at Object.<anonymous> (C:\dev\vscodevim\Vim\test\cmd_line\sort.test.ts:47:5)
      at Generator.next (<anonymous>)
      at fulfilled (C:\dev\vscodevim\Vim\out\test\cmd_line\sort.test.js:4:58)

  3) Basic sort Sort range, dsc:

      AssertionError: Line count does not match.
      + expected - actual

      -5
      +4

      at Object.assertEqualLines (C:\dev\vscodevim\Vim\test\testUtils.ts:35:10)
      at Object.<anonymous> (C:\dev\vscodevim\Vim\test\cmd_line\sort.test.ts:87:5)
      at Generator.next (<anonymous>)
      at fulfilled (C:\dev\vscodevim\Vim\out\test\cmd_line\sort.test.js:4:58)


  4) Mode Normal / can search through multiple newlines:

      Cursor LINE position is wrong.
      + expected - actual

      -0
      +1

      at Suite.suite (C:\dev\vscodevim\Vim\test\mode\modeNormal.test.ts:1492:3)
```

- [x] 1) read Can read shell command output - 
  *NOTE*: with neovim enabled the second line is "hey " <- extra space.  
  Without neovim it doesn't get the first empty line
- [x] 2) Basic sort Sort whole file, dsc:
  *NOTE*: with neovim there's a fourth empty line
- [x] 3) Basic sort Sort range, dsc:
  *NOTE*: probably same as 2)
[...]

PR #5075, opened 2020-07-24
Fixed neovim newlines for Windows, and related cursor position bug

austinclem1:
Related to issue #1914 and PR #1939
PR #1939 removed dangling \r's after the text buffer was returned from neovim, but I think it makes more sense to remove them on the way out from VSCode. Now neovim works with line commands as expected.
Also after finishing a neovim ex command, the new cursor position was being read from neovim but it wasn't actually being set in VSCode. Maybe it was using a deprecated method for setting the cursor. Also the cursor would have been off because the code didn't account for the fact that lines and columns start at 1 in neovim.

### DIFF
--- a/src/neovim/neovim.ts
+++ b/src/neovim/neovim.ts
@@ -111,7 +111,10 @@ export class NeovimWrapper implements vscode.Disposable {
     }
 
     await this.nvim.setOption('gdefault', configuration.gdefault === true);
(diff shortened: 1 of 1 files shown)


J-Fields:
@austinclem1 Would you mind resolving the merge conflicts when you get a chance?

austinclem1:
Sure thing, let me know if I made any noob mistakes with this merge conflict haha

PR #5077, opened 2020-07-25
Fix gb with count

### LINKED ISSUE #4075: Repeating gb once when selecting Multi-Cursors means subsequent gb are repeated that amount
**Describe the bug**
If you use `gb` with a number, subsequent uses of gb are repeated that same amount. Eg `10gb gb` would result in 20 cursors, not 11.

Similarly, if I do `gb`, `10gb`, `gb`, I get 21 cursors instead of 12.

**To Reproduce**

Lets say I have 10 lines, each with the same single word

```
vim
vim
...
vim
vim
```

Scenario A) I put a cursor on the first line, and in normal mode do `5gb` then `gb`.

Scenario B) I put a cursor on the first line, and in normal mode do `gb`, `5gb`, `gb`.

**Expected behavior**
Scenario A) I would expect 6 cursors but there are 10.

Scenario B) I would expect 7 cursors but there are 10 (there would be 11 if there were more lines).

**Screenshots**

Reproduction Steps Screenshot

- Issue Type: `Bug`
- Extension Name: `vim`
- Extension Version: `1.10.2`
- OS Version: `Linux x64 5.2.14-arch2-1-ARCH`
- VSCode version: `1.38.0`

**Additional context**
N/A

sql-koala:
This fixes a bug where a count to gb, eg "3gb" is carried over, so next gb selects 3 cursor again.

fixes #4075

The bug/problem itself is in modeHandler. It has to do with the rather unique properties of "gb", that it is a complete action but changes to visual mode. This causes a flag to be set, which prevents a fresh recordedState. And that is why recordedState.count is carried over.

That one was quiet a challenge to track down :>


sql-koala:
I actually found more commands with this property: gv, gn, gN.
If you give them a count (which does not make sense) and execute gb afterwards you get the same bug, more cursors :D

sql-koala:
closing this in favor of the approach in #5842

PR #5078, opened 2020-07-26
Fix the behavior of `ga` with respect to digraph

lusingander:

As the following help document, `ga` print the ascii value of the character under the cursor.
And, if the character can be inserted as a digraph, should print the two characters that can be used to create.

But this is not implemented.

```
:as[cii]	or					*ga* *:as* *:ascii*
ga			Print the ascii value of the character under the
			cursor in decimal, hexadecimal and octal.
			Mnemonic: Get Ascii value.

			For example, when the cursor is on a 'R':
				<R>  82,  Hex 52,  Octal 122
			When the character is a non-standard ASCII character,
			but printable according to the 'isprint' option, the
			non-printable version is also given.
			
			When the character is larger than 127, the <M-x> form
			is also printed.  For example:
				<~A>  <M-^A>  129,  Hex 81,  Octal 201
				<p>  <|~>  <M-~>  254,  Hex fe,  Octal 376
			(where <p> is a special character)

			The <Nul> character in a file is stored internally as
			<NL>, but it will be shown as:
				<^@>  0,  Hex 00,  Octal 000

			If the character has composing characters these are
			also shown.  The value of 'maxcombine' doesn't matter.

			If the character can be inserted as a digraph, also
			output the two characters that can be used to create
			the character:
			    <ö> 246, Hex 00f6, Oct 366, Digr o:
			This shows you can type CTRL-K o : to insert ö.
```

And also, since there was no test for `ga`, I added it.

The format of the `Hex` is partly different from the help, but this seems to be another problem originally exists.

### DIFF
--- a/src/actions/commands/actions.ts
+++ b/src/actions/commands/actions.ts
@@ -40,6 +40,7 @@ import { reportLinesChanged, reportFileInfo, reportSearch } from '../../util/sta
 import { globalState } from '../../state/globalState';
 import { VimError, ErrorCode } from '../../error';
 import _ = require('lodash');
+import { DefaultDigraphs } from './digraphs';
 
 export class DocumentContentChangeAction extends BaseAction {
   private contentChanges: vscode.TextDocumentContentChangeEvent[] = [];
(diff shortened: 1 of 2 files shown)


J-Fields:
How does this solution compare to #5068?

lusingander:
Sorry, I didn't notice #5068.

PR #5080, opened 2020-07-26
Make 'gj' and 'gk' preserve desiredColumn

### LINKED ISSUE #1323: Issue with cursor position using visual line movement (gj and gk)

### LINKED ISSUE #4441: `gj` and `gk` should not changed `desiredColumn`

berknam:
Make 'gj' and 'gk' preserve desiredColumn

fixes #4441 
fixes #1323

I had this stashed for some time now. From what I remember this was working properly and only needed some organization of the code.

There is still the issue that if you are on a wrapped line (which is not its first visual line) and you move horizontally it will set the desiredColumn to a value that is not the correct one.
I have another stash with a possible fix for that as well, but it implies changing the cursor once every time you set the desired column. I know this isn't good, but considering the existing vscode API, I believe it is the only possible way to be able to set the desired column correctly even when inside a wrapped line.
(related to comment https://github.com/VSCodeVim/Vim/issues/4441#issuecomment-657651598)


berknam:
This is currently failing because of this https://github.com/VSCodeVim/Vim/commit/c54465dce242d0cf8f7b9d7e73b6ae59e48bb18e#commitcomment-40913575

src/actions/motion.ts
```diff
@@ -120,69 +120,144 @@ class MoveDownByScreenLine extends MoveByScreenLine {
...
+            : 0
+          : indent === 'indent'
+          ? indentSpace + tabSize > previousLineColumn
+            ? indentSpace + tabSize - previousLineColumn
+            : 0
+          : indent === 'deepIndent'
+          ? indentSpace + 2 * tabSize > previousLineColumn
+            ? indentSpace + 2 * tabSize - previousLineColumn
+            : 0
+          : 0;
```
robclancy, line 218: wtf is this

berknam: 😅 Like I said this PR needs some organization of the code. But this specific part of the code is needed because in order to know if we need to move horizontally when we enter a wrapped line we need to know what was our previous column and what is the `editor.wrappingIndent` configuration because this configuration changes the behavior of wrapped lines indentation, when it is `none` we have something like this:
```
level0
    level1: imagine this is a long line that wraps right here
continuing on the next visual line with wrapping indent set to 'none'
level0: en|d
level0: this is where we started
```
where `|` is the cursor that currently is on position (line 2, column 10), but lets assume we started on line 3 on the last `e` of the word `where` which is column 20 (`desiredColumn is 20`) . If you then move up you will get this:

```
level0
    level1: imagine this is a long line that wraps right here 
continuing| on the next visual line with wrapping indent set to 'none'
level0: end
level0: this is where we started
```
And position is now (1, 72). In this case we only need to move `desiredColumn - previousLineColumn` to the right to get to the desiredCoumn which would be on letter `x` of the word `next`. But when you have `editor.wrappingIndent` set to `same` you get this:
```
level0
    level1: imagine this is a long line that wraps right here
    continuing on the next visual line with wrapping indent set to 'same'
level0: en|d
level0: this is where we started
```
Then moving up gets you this:
```
level0
    level1: imagine this is a long line that wraps right here 
    contin|uing on the next visual line with wrapping indent set to 'same'
level0: end
level0: this is where we started
```
But the position is now (1, 68). Now you also need to move `desiredColumn - previousLineColumn` to the right to get to the desiredCoumn which would be on letter `e` of the word `the`. But if the third line was only `end` then it would be smaller than the indent space of the line above but when moving up we would go to the start of the wrapped line like if the indentation didn't exist, like this:
```
level0
    level1: imagine this is a long line that wraps right here
    continuing on the next visual line with wrapping indent set to 'same'
en|d
level0: this is where we started
```
Then moving up gets you this:
```
level0
    level1: imagine this is a long line that wraps right here 
    |continuing on the next visual line with wrapping indent set to 'same'
end
level0: this is where we started
```
Now we not only moved up but we also moved right so now to go to the desiredColumn we need  to move `desiredColumn - previousLineColumn - diff` to the right to get to letter `e` of the word `the`. Where `diff` is the number of characters we moved to the right already which is the difference between the indent space and the previous line column (`indentSpace - previousLineColumn`).

When setting `editor.wrappingIndent` to `indent` the wrapped line is indented on more level in relation to its previous visual line and `deepIndent` goes to levels like this:
```json
level0
    level1: imagine this is a long line that wraps right here
        continuing on the next visual line with wrapping indent set to 'indent'
            continuing on the next visual line with wrapping indent set to 'deepIndent'
```

Sometimes when the indentation is big we might move to the right past the desiredColumn but in those cases the result of the calculated move to the right would be a negative number which means we will actually move to the left to the right position.

I've tested this a lot, I know this isn't very intuitive but since we don't have a better API for this from vscode this is the best solution that I could find. I'm sorry for taking so long to answer you but I've been trying to write this answer in a simpler manner but couldn't. Hope you can understand now, if not please tell me and I'll try to give some more examples.

src/actions/motion.ts
```diff
@@ -120,69 +120,144 @@ class MoveDownByScreenLine extends MoveByScreenLine {
...
+    let activePos = Position.FromVSCodePosition(vimState.editor.selection.active);
+    if (activePos.character < desiredColumn && activePos.line !== prevLine) {
+      activePos = activePos.withColumn(desiredColumn);
+    } else if (previousLineColumn < desiredColumn) {
+      // If the active position is after desiredColumn it means we moved to a wrapped line.
+      // In that case if the position on previous line is less then desiredColumn we need to
+      // move right by the difference of desiredColumn and the previous line column.
+      const indent = vscode.workspace.getConfiguration('editor').get('wrappingIndent');
+      const indentSpace = TextEditor.getFirstNonWhitespaceCharOnLine(activePos.line).character;
+      const tabSize = vscode.workspace.getConfiguration('editor').get<number>('tabSize')!;
```
J-Fields, line 202: I believe you want to use `vimState.editor.options.tabSize` instead

src/actions/motion.ts
J-Fields: I think something like this would be clearer:

```typescript
let diff: number;
switch (indent) {
  case 'none':
    diff = 0;
    break;
  case 'indent':
    diff = Math.max(0, indentSpace + tabSize - previousLineColumn);
    break;
  case 'deepIndent':
    diff = Math.max(0, indentSpace + 2 * tabSize - previousLineColumn);
    break;
  default:
    throw new Error(`Unknown value for 'editor.wrappingIndent': '${indent}'`);
}
```

berknam: Yes it would 😄 When I made this branch I was mainly focused on getting it to work since it took quite a few iterations to get there. I do need to refactor some of the code to make it clearer like your suggestion. I hope to get more time next week and I'll try to do that. Unless you really want to get this on the next release? Then I can see if I can get to it sooner.

J-Fields: It'd be nice to get in the next release, but not critical. Get to it when you're able to, no rush 🙂

src/actions/motion.ts
berknam: Yes you're correct, nice catch!

src/actions/motion.ts
berknam: I'll see what I can do. I just noticed that this still has some issues when working with tabs. I think I have a workaround for that but haven't tried it yet. This PR will be a PITA to try to comment and explain every little weird behavior. I'm sorry for once again creating a PR that will be hard to review. I promise I'm not making these on purpose! 😅

kino-ma:
What is the current status of this?

mogelbrod:
Would also love to see this merged!

kolakode:
If there is a working, testable fix for this rather annoying, persistent issue, could someone PLEASE get it implemented and merged? Thank you.

micimize:
@J-Fields how far is this from merge-able? No promises (None, really) but I may have time to help if we know what needs to be done.

It seems to be the only major gap in softwrap navigation (esp. with visual block mode) & is around the top 10 issues sorted by :+1: while maybe being the lowest hanging fruit

PR #4280, opened 2019-11-19
#4275 support multiple count of MoveInside* without surroundings

hao-flexport:
fix VSCodeVim#4275: support multiple count of MoveInside* without surroundings

- [x] Commit messages has a short & issue references when necessary
- [x] Each commit does a logical chunk of work.
- [x] It builds and tests pass (e.g `gulp`)

**What this PR does / why we need it**:
fix commands for:
```
d3i{
c3i(
v3i<
```

**Which issue(s) this PR fixes**
https://github.com/VSCodeVim/Vim/issues/4275

**Special notes for your reviewer**:
* It's now a concept prove, please help me improve it.
* Added several new test.

Issue #4275, opened 2019-11-18
simple BaseMovement.execActionWithCount for command `d2i{` not working properly


hao-flexport:
**Describe the bug**
`BaseMovement.execActionWithCount` is used to do repeating actions for the canonical form: 
`action` - `count` - `modifier` - `boundary` (`action` and `count` can be switched)
`v`|`d`|`c` - ` `|`1`|`2`|`3`|... - `a`|`i` - `(`|`{`|`[`|`<`|...

It is working for modifiers `a`,
```
d2a{
c3a(
2da{
3ca(
```
but for `i` modifier with `count > 1`, it is not working.
```
d3i{ // simple the same as di{, with `count` not working
3di{
```

**To Reproduce**
The reason is the dilemma, the `d3i{` does not simply equal to `di{` by 3 times, for example:
```
{
  {
    {
      | <-- cursor here
    }
  }
}
```
After any times of `di{`, the cursor is still there.

**Expected behavior**
N/A.

**Screenshots**
N/A.

**Environment (please complete the following information):**
No matter.

**Additional context**
When `BaseMovement.execActionWithCount` is repeating invoking `MoveInsideCharacter.execAction`, extra context like `currentIndex` of `count` can be passed in for improving `Range` computation.

I am interested in fixing this issue, but don't know whether this could be the best practice. I will submit my PR for your suggestions.

hao-flexport:
And I think `lastIteration` is for it, is it true?
https://github.com/VSCodeVim/Vim/blob/38c268a26dadbd0fed1ea297923e7640f358097c/src/actions/baseMotion.ts#L124
https://github.com/VSCodeVim/Vim/blob/38c268a26dadbd0fed1ea297923e7640f358097c/src/actions/baseMotion.ts#L163

hao-flexport:
Both `MoveInsideCharacter` and `MoveTagMatch` have this issue.

hao-flexport:
@J-Fields Anyone would respond to my PR? Need your help. Thanks!

hao-flexport:
I think the PR has fixed this issue. Could anyone take a look and merge? @tail @LeonB @mikew @positron

Xadoy:
> I think the PR has fixed this issue. Could anyone take a look and merge? @tail @LeonB @mikew @positron
[...]

PR #5093, opened 2020-07-31
Fix MoveI* when inner content is selected #4275

Xadoy:

Fixes a bug where MoveInsideCharacter does not work if the inner content is already selected. Although pull request #4280 tried to fix the same issue, in my opinion, it is a work around and it did not fix the root cause. In comparison, this pull request also covers the case that a user in visual mode and tries to select content in two layers of curly braces by hitting "v" followed by "i{" then "i{".

#4275 

Found another issue when I was tracking this issue, I will create a new issue soon.


J-Fields:
@Xadoy Would you mind adding test cases?

Xadoy:
> @Xadoy Would you mind adding test cases?

Of course, I will add the test cases. But in this case, #4280 has added some test cases, how should I process this. Sorry I am kinda new to open source contribution.

J-Fields:
Feel free to borrow those test cases and add any that your fix covers that aren't already there

Xadoy:
@J-Fields I borrowed and modified some test cases. In total, 32 new test cases are added. Let me know if you think this is too much. It covers all brackets, all cases with starting bracket or ending bracket. Also, multiple line scenarios are covered......

J-Fields:
Thanks @Xadoy for adding tests and having patience with my sporadic open source schedule 🙂

asinbow:
Super cool! Thanks!

PR #4718, opened 2020-04-03
Sneak highlighting

gergelyth:
**What this PR does / why we need it**:

Vim-Sneak highlighting has been missing from the extension and this PR adds that.
It also sets the **isJump** flag, same as #4697 does (although that has tests as well). 

sneak_highlighting

**Which issue(s) this PR fixes**

Fixes #3363
Fixes #4696

PR #5095, opened 2020-07-31
Enable to highlight f key searches by a option

Yuuki77:

Often when I'm using f to search for characters in the current line, I'll run into more occurrences of the character than I expected so highlighting each search match would be nice.

Therefore I added an option called `hlsearchF `

No

* When you test it, please `Vim: Sneak Replaces F` to false. Ideally it works with Sneak search.
* This Also, the highlighted characters will be the characters between the beginning and the end of the line, ignoring the current cursor position. This means that the result will be the same as a search from the beginning of the line.
- This changes the kind of behaviour of `f` command. I though this is convenient, but if you disagree, I can start the search from where the current cursor position is  at on the line

This is my first pr to VSCodeVim so If you have any suggestions, please let me know

Aug-01-2020 01-52-59


Yuuki77:
@J-Fields What do you think about this feature?
I would love to hear your opinion 😄

J-Fields:
I like the concept. Haven't gotten a chance to dig into the specifics of this PR yet, but I think it might duplicate the functionality of #4718?

Yuuki77:
@J-Fields Thanks I did now know there is a pr for Sneak highlighting.
Most users use Sneak highlighting, so that's better!
I am going to close this pr.

J-Fields:
Thanks for the contribution!

ronakg:
I'd love this highlighting for f command. I don't use sneak, maybe there are more users who are not using either sneak or easy-motion?

Yuuki77:
@ronakg @J-Fields 
Thank you for the feedback
I guess This feature would be useful for some users.
What do you think?
I am going to reopen this pr

Yuuki77:
@J-Fields 😄

PR #5106, opened 2020-08-03
fix horizontal scroll

corona6:

To fix horizontal scroll key correctly.

http://vimdoc.sourceforge.net/htmldoc/scroll.html#scroll-horizontal

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -207,12 +207,12 @@ The following are all marked low priority because VSCode has very good support f
 
 These only work when 'wrap' is off:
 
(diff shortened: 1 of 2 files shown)


J-Fields:
I'm a bit confused - why is the existing implementation wrong? It's very limited because we can't force the view to scroll horizontally, but if the cursor is on the edge of the screen, it does the right thing - `zh` moves the cursor to the left, forcing the view to the right (and vice versa for `zl`).

J-Fields:
I think the existing implementation (while definitely wrong, due to API limitations) is more correct, so I'm closing this. Thank you!

PR #5130, opened 2020-08-17
Nodeless entry for Vim extension

rebornix:

Add a new entry for running Vim in nodeless environment and load platform specific modules based on target.

This PR includes changes to

* `fs`. In node, it's node's native `fs`; in nodeless, it uses `vscode.workspace.fs`.
* `historyFile`. In node, it stores the history in file system; in nodeless, it stores in memory.
* `logger`. In node, it uses winson; in nodeless, it uses browser `console`.
* lastly, it relies on Webpack to tree shake unwanted code paths (for example, remove `nvim` related code paths from the bundle in nodeless environment).


rebornix:
Now the all tests pass and no linter errors, the change is ready for review @J-Fields . With this change, you should be able to run the Vim extension in web worker in VS Code by adding following config:

```json
    "remote.extensionKind": {
        "vscodevim.vim": [
            "web"
        ]
    }
```

The goal of this PR is adding multiple entry for bundling (one for desktop and one for web) without breaking existing features. The `platform/browser` code still needs more polish later on once we have some real use of it in production.

J-Fields:
Great, thanks @rebornix! I'll take a look as soon as I have a chance.

> once we have some real use of it in production

Just curious, is there any rough timeline for that?

rebornix:
@J-Fields my plan is giving it some testing and adding docs about which features are missing in the nodeless version within next two weeks. Also once this PR is merged, I'll update the publish CD to include the nodeless bits to the marketplace. If it goes well, we can ask users who are suffering from the performance issue to try this first.

rebornix:
@J-Fields do the changes look reasonable to you?

J-Fields:
Sorry for the delay @rebornix, I haven't had much time recently to work on VSCodeVim. This all looks fine to me, thanks a bunch!

J-Fields (approved):

univerz:
> With this change, you should be able to run the Vim extension in web worker in VS Code by adding following config

should this work now?

J-Fields:
> should this work now?

Yes I believe so

rebornix:
@univerz if it doesn't work, try add another setting `"extensions.webWorker": true`

univerz:
> if it doesn't work, try add another setting `"extensions.webWorker": true`

thank you, it works with this setting :)

Issue #5167, opened 2020-09-06
Folding: closed folds with zM, zC will reopen


sql-koala:
**Describe the bug**
When closing folds with zM or zC, they will reopen (unless cursor is on the top/first line of the folded code).

**To Reproduce**

1.  open any document with folding, vsc source will do ;-)
2.  place cursor on a line "inside" a fold, e.g. inside if
3.  press "zM" or "zC"
4.  text will fold for a short moment, then reopen

If you use the Fold Command from vscode, it will work.

**Additional context**
The reason is that cursor position is not updated for zM, zC. I will have a PR for this.

PR #5169, opened 2020-09-06
Refactor folding.

### LINKED ISSUE #5168: Folding: zc and zo handle a count wrong
**Describe the bug**
see topic ;-)

**To Reproduce**

1.  open any document with at least 4 levels of folding, vim source will do ;-)
2.  place cursor on a line deeply nested (have all folds open now)
3.  press "2zc"
4.  this will close 4 levels of folds, not 2.

**Additional context**
The reason is that the operators run for each count, so it multiplies. I will have a PR for this.

sql-koala:
This does a refactor the folding classes a bit and fixes two bugs.

fixes #5168 
fices #5167  

this refactors exec to do these steps (for all derived Commands):
1. call vscode with folding command
2. sync cursor
3. set mode to Normal (orig vim does this too)

Bugs fixed:
- we need to sync cursors after zM, zC
- zo, zc must not run for each count


src/actions/commands/actions.ts
```diff
@@ -1183,7 +1185,16 @@ abstract class CommandFold extends BaseCommand {
...
-    await vscode.commands.executeCommand(this.commandName);
+    let timesToRepeat = vimState.recordedState.count || 1;
+    if (this.direction !== '') {
+      await vscode.commands.executeCommand(this.commandName, {
+        levels: timesToRepeat,
+        direction: this.direction,
+      });
+    } else {
+      await vscode.commands.executeCommand(this.commandName);
+    }
```
sql-koala, line 1196: would be interested if this can be done with a neat one liner?

J-Fields:
```typescript
const args = this.direction != '' ? { levels: timesToRepeat, direction: this.direction} : undefined;
await vscode.commands.executeCommand(this.commandName, args);
```
seems like the nicest way of writing this

src/actions/commands/actions.ts
```diff
@@ -1171,9 +1171,11 @@ class CommandRepeatSubstitution extends BaseCommand {
   }
 }
 
+type FoldDirection = 'up' | 'down' | '';
```
J-Fields, line 1174: Any reason to use `''` over `undefined`?

src/actions/commands/actions.ts
sql-koala: linter turned that into 4 lines :D

J-Fields:
Thanks!

Issue #5116, opened 2020-08-06
GitHub Sponsors / Patreon


max-sixty:
I use VSCode / Vim for a lot of my day, and it makes a material impact on my productivity.

I can't really contribute given my Javascript skills (I have done a couple of tiny doc PRs). I'd be very keen to subscribe to a GitHub Sponsors / Patreon to support its continued development. 

IIRC, @J-Fields has said that having a GH Sponsors could be unfair because others have built so much. But the funds (I imagine fairly token, at least initially) would be for future development — there are 849 open issues, so there's plenty to do! — and having something seems fairer than having nothing. 

That said, I recognize that OSS + $ can be complex, so feel free to close if it's a non-starter.

Taeluf:
You could also work issues.
- Find duplicates & leave links, perhaps recommending OP to close the issue. 
- Look for things that might have been resolved, but not closed.

Taeluf:
[OpenCollective](https://opencollective.com/) might be a good platform?
Maybe a Not For Profit could be formed around VSCodeVim, but that's a whole other headache too.

max-sixty:
I agree those are useful tasks @Taeluf and that's a good outlet for some people to contribute back.

> Maybe a Not For Profit could be formed around VSCodeVim, but that's a whole other headache too.

Agree re the effort for something like a non-profit. I for one would be very happy giving directly with GH Sponsors — the work is fairly centralized already, and (unfortunately) it's not going to be enough money to cause adverse incentives...

I'm happy to help with the setup if @J-Fields gives the go-ahead!

J-Fields:
I do now have GitHub Sponsors set up, if anyone insists on contributing.

max-sixty:
> I do now have GitHub Sponsors set up, if anyone insists on contributing.

Just joined as your first!

(Maybe set up some tiers / suggested amounts?)

J-Fields:
Much appreciated!

PR #5227, opened 2020-09-23
Added "Improve Existing Issues" notes

Taeluf:
For the "duplicate" link, I provided a sample search for "cursor" which returns 200+ issues. Many of which probably aren't duplicates, but I thought that might be a good starting point. MIght be better to link to a blank search for open issues.

For the "old" link, I gave page 25. Currently there are 34 pages, so I thought that would be sufficient insulation to still be "old" but not get shwooped off later

I saw a few folks claim that VSCodeVIM Team gives precedence to issues with more thumbs up, so I included that recommendation. I didn't want to say that so explicitly though, since I don't actually know.

**What this PR does / why we need it**: Some folks may want to contribute outside of code.

**Which issue(s) this PR fixes** I didn't create an issue for this, though https://github.com/VSCodeVim/Vim/issues/5116 is what gave me the idea to add this PR

**Special notes for your reviewer**: I didn't run any tests or anything, cause it's just an .md file.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -14,6 +14,11 @@ When creating a new bug report do:
 - Search against existing issues to check if somebody else has already reported your problem or requested your idea
 - Fill out the issue template.
 
+### Improve Existing Issues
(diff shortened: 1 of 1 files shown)


J-Fields:
Thanks!

PR #5245, opened 2020-09-25
make getConfiguration use the document languageId (fixes #4503)

### LINKED ISSUE #4503: Iskeyword does not respect language specific editor overrides
**Describe the bug**
Suppose we have
```hack
$foo + $bar
^
|
cursor is here
```
When I press `w` I expect my cursor to go to
```hack
$foo + $bar
     ^
     |
```
Instead, it goes to
```hack
$foo + $bar
 ^
 |
```
In my `setting.json`, I have the following override for Hack (read: PHP):
```json
{
  "[hack]": {
    "editor.wordSeparators": "`~!@#%^&*()-=+[{]}\\|;:'\",.<>/?"
  }
}
```
Note the absence of `$`. So it seems that `vim.Iskeyword` doesn't respect language overrides for `editor.wordSeparators`. A workaround could be to set the global `editor.wordSeparators` or `vim.Iskeyword`, but neither is ideal since I work in other languages as well where `$` should be included.

**To Reproduce**

1.  Open a PHP/Hack file
2. Edit the language specific settings for PHP/Hack to exclude `$` in `editor.wordSeparators`
3. Move cursor to the beginning of a variable
4. Press `w` to move cursor past the variable

**Expected behavior**
Cursor moves past variable

- VSCode version: 1.41.1579026862
- OS: Mac OS 10.15.2

dalev:
Fixes #4503 by passing `languageId` in `getConfiguration`.  This required updating to `@types/vscode` 1.42, which appears to be the earliest version that supports this API.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -279,9 +279,9 @@
       }
     },
     "@types/vscode": {
-      "version": "1.37.0",
-      "resolved": "https://registry.npmjs.org/@types/vscode/-/vscode-1.37.0.tgz",
-      "integrity": "sha512-PRfeuqYuzk3vjf+puzxltIUWC+AhEGYpFX29/37w30DQSQnpf5AgMVf7GDBAdmTbWTBou+EMFz/Ne6XCM/KxzQ==",
+      "version": "1.42.0",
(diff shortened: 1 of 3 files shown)


J-Fields:
Nice, thanks!

J-Fields (approved):

PR #5250, opened 2020-09-25
Fix multicursor selections visual mode

### LINKED ISSUE #5212: Mouse clicked position isn't registered after navigating only with `hjkl`
**Describe the bug**
The cursos position isn't update when clicking with the mouse after sometime navigating using `hjkl`.

After some navigation with `hjkl` it stop to update the cursor position when clicking with the mouse.
If I click anywhere to navigate with motions (`w`, `b`, `e`)  or edit from there it will use the last position I navigated using keyboard, it only keep the clicked position if I use `hjkl` again.

If I open a file and just keep clicking to navigate or edit, or even navigate using motions it's ok, the mouse update the cursor position.

**To Reproduce**

1.  Go to any file
2.  Navigate using `h`, `j`, `k` and `l`
3.  Click anywhere else in the file and try any action like `a`, `i`, `w`, `b`
4.  You should see the action starting from last position you navigated wtih `hjkl`

**Expected behavior**
Expected to start the action from the clicked position.

- Extension (VsCodeVim) version: `v1.17.0`
- VSCode version:  `Version: 1.49.1` and `1.50.0-insider`
- OS: macOS Catalina 10.15.6

**Additional context**

I could reproduce this issue in stable and insider version. It started just after I updated the Vim plugin.
On Insider I don't have any other extension installed beside theme and markdown docs, but I also tried with them disabled.

If there is any info or debug I can do please let me know.

### LINKED ISSUE #5221: Cursor location doesn't update on VSCode search / click on line
**Describe the bug**
VSCodeVim doesn't always update cursor location when user changes it using VSCode (by clicking on a line / cycling through VSCode search results).

NOTE 1: This started with v1.17.0, I tested v1.16.0 and it works fine.
NOTE 2: This happens to me about 60% of the time, a few times I tried it worked fine, still really easy to reproduce 

**To Reproduce**

1. Go to a file
2. Search something using VSCode search (⌘ + F), cycle results to a result found in a different part of the file than the current cursor line
3. Move cursor using VSCodeVim (J/K/L/M/e/any other motion really)
4. Cursor jumps back to the previous cursor location (before search)

Or:
1. Go to a file
2. Place cursor on some line 
3. Click on another line (put cursor there)
4. Move cursor using VSCodeVim (J/K/L/M/e/any other motion really)
5. Cursor jumps back to the previous cursor location (before clicking on a line)

**Expected behavior**
Cursor location representation in VSCodeVim should be updated when it's moved to a different location using VSCode tools

- Extension (VsCodeVim) version: 1.17.0
- VSCode version: 1.49.1
- OS: macOS Catalina 10.15.5

**Additional context**
As mentioned above, this started with 1.17.0 (1.16.0 is fine). Something appears to be wrong with this version

### LINKED ISSUE #5248: 1.17.0 clicking causes cursor to jump when using vim hotkeys
**Describe the bug**
Clicking doesn't always properly update cursor position in conjunction with vim hotkeys

**To Reproduce**

1.  Be typing normal navigating with vim hotkeys
2.  Click somewhere else in editor window
3.  Use vim hotkey to move curor
4.  Cursor jumps to previous position ignoring click

Example: if I am currently editing line 20, click on line 50 and then press j the cursor goes to line 21, instead of 51 as it previously did.

**Expected behavior**
Clicking updates the cursor position in such a way that vim recognizes it

**Screenshots**
I'll try to get a screen recording of it happening and update issue with it

- Extension (VsCodeVim) version: 1.17.0
- VSCode version: 1.49.2
- OS: Ubuntu 18.04 and macOS Mojave

**Additional context**
Interestingly, it doesn't happen 100% of the time and I can not track down why it occasionally works as expected

Closed issue on VSCode: https://github.com/microsoft/vscode/issues/107403

### LINKED ISSUE #5268: Vim-Find dialog last cursor position is off
## Issue Type: <b>Bug</b>

1.  put the cursor at start of document (gg)
2.  activate VSCode Find dialog (Cmd+F)
3.  search for something; hit ENTER (one or more times) to step cursor through search results: make sure cursor is no longer on the first line
4.  hit ESC to exit Find dialog: cursor is correctly on last found position
5.  any consecutive VIM command (e.g. ESC) will jump to/be applied to line 1 (or whatever line the cursor was at before stepping through Find search results)

## Expected behavior
- Vim should update cursor position
- Bug was likely introduced in v1.17.0, not reproducible in v1.16.
- Seems to have something to do with the Find dialog, since searching with vim ('/') works as expected

## environment
Extension version: 1.17.1
VS Code version: Code 1.49.2 (e5e9e69aed6e1984f7499b7af85b3d05f9a6883a, 2020-09-24T16:23:52.277Z)
OS version: Darwin x64 19.6.0

### LINKED ISSUE #5380: Cursor stay on initial position after CTRL+arrows
**Describe the bug**
When you use `CTRL + arrows` in **"normal"** or **"insert"** mode (I have `vim.CtrlKeys` disabled for now, but it's independent, I've tested with *enabled* state too)

**To Reproduce**
And that is a problem, it's fluid, some files, for example just plain text, do not affected, but some like TypeScript or HTML in Angular project has this issue, but not from start, and 100% after several seconds of working with them.
Can't catch concrete action that can produce this issue, and just hope that maybe someone already see this issue previously... 

*Forgot to mention*: After toggle `VIM: Toggle VIM Mode` and back, **issue has gone, but** again, after several seconds it arise 

**Expected behavior**
To not have any issue on using CTRL+arrows navigation

**Screenshots**
screencast

- Extension (VsCodeVim) version: : **1.17.1**
- VSCode version:
Version: **1.50.0**
Commit: 93c2f0fbf16c5a4b10e4d5f89737d9c2c25488a3
Date: 2020-10-07T06:01:33.073Z
Electron: 9.2.1
Chrome: 83.0.4103.122
Node.js: 12.14.1
V8: 8.3.110.13-electron.0
OS: Linux x64 5.7.0-997-generic

berknam:
Currently when you have selections with multicursors if you move in a way that those cursors would combine would then result in a situation where our 'updateView' function would think that there was a different selection even when there wasn't, because it was still thinking it should have 2 selections.

This would result in our selections to be ignored count increasing and there would never be a selection change to clear that count, so any future selection changes would be ignored until the document was closed and reopened or vscode window was reloaded or restarted.

This seems to happen a lot of times when using snippets, because now when inserting a snippet it turns the selections to visual mode multicursor and sometimes if the user moves up/down without noticing that it is in visual mode (since previously it would still be in insert mode, but this is a different issue that I'm currently working on trying to make snippets not go into visual mode) then in those case it can create the situation I mentioned above. And the user will probably not even notice it until they click somewhere and try to move.

I think this might fix the issues bellow, but I'm not sure. It would help if I could get some logs, since these selection changes are logged the same way as the remapper ones.
fixes #5212 
fixes #5248 
fixes #5221 
fixes #5268
fixes #5380
***Edit: I marked this issues as fixed. If they're not fixed by this PR, then they can be reopened later***

The fix that @rebornix did on commit https://github.com/VSCodeVim/Vim/commit/bc54d447ea6690f90763439e569246a1e98b1547 works around the issue for mouse selection changes, but doesn't really fix the issue (it still happens if you use something like the `editor.action.smartSelect.grow`)

This fixes the situation caused by the multicursor selections combining, the thing is I don't know if there are other situations that can cause our selections change to be ignored count to not go back to 0. One way to fix this for any future situation I can't think of right now would be to check that if our `vimState.selectionsChanged.ourSelections.length` was greater than 0 for more than a few seconds after the last selection change event, then probably some selection change was wrongfully added and we could reset `vimState.selectionsChanged.ourSelections` make sure that any future selection change events would then be caught again.

The problem with that is that this would probably make quite a few tests fail. If we still get issues related to this after this PR is merged I will try to go with this approach then.


rebornix:
@berknam thanks for the fix. I don't fully understand the initial change and also this one, so I'd like to push the workaround to marketplace real quick to unblock users, and then review this PR.

berknam:
> 
> 
> @berknam thanks for the fix. I don't fully understand the initial change and also this one, so I'd like to push the workaround to marketplace real quick to unblock users, and then review this PR.

Ok. If you need any help understanding this tell me.

rebornix:
@berknam now I understand the intention of the initial change https://github.com/VSCodeVim/Vim/pull/5015#issue-448180259, which is great. There are two potential bugs with how currently `vimState.selectionsChanged.ourSelections` are manipulated (which at least leads to #5212, didn't check others)

1. `selectionsChanged.ourSelections` may contain duplicates. We didn't check if the `selectionHash` already exists in `selectionsChanged.ourSelections` in https://github.com/VSCodeVim/Vim/blob/8cbc79100a1fd930529a2b552a9ebaa466d37fd0/src/mode/modeHandler.ts#L1493 . When removing `selectionHash`, we will only remove the first match we find in `selectionsChanged.ourSelections ` https://github.com/VSCodeVim/Vim/blob/8cbc79100a1fd930529a2b552a9ebaa466d37fd0/extensionBase.ts#L295 . At the end, `selectionsChanged.ourSelections` contains zombie data and from that moment, selection changes from mouse will never take effect.
2. The only chance for `selectionsChanged.ourSelections` to be cleared is we find a match in https://github.com/VSCodeVim/Vim/blob/8cbc79100a1fd930529a2b552a9ebaa466d37fd0/extensionBase.ts#L295. Although #5212 is fixed if the first issue (above) is resolved but it doesn't mean we won't run into similar bugs in the future. I'm wondering if we can clear `selectionsChanged.ourSelections` when `ignoreIntermediateSelections` is set to `false`. It seems we will wrap all actions between `ignoreIntermediateSelections = true; ...; ignoreIntermediateSelections = false;` it can reduce  as many unnecessary selection change events as possible.

berknam:
> 

>     1. `selectionsChanged.ourSelections` may contain duplicates. We didn't check if the `selectionHash` already exists in `selectionsChanged.ourSelections` 
>         . When removing `selectionHash`, we will only remove the first match we find in `selectionsChanged.ourSelections ` 
>         . At the end, `selectionsChanged.ourSelections` contains zombie data and from that moment, selection changes from mouse will never take effect.

I don't think we should prevent having duplicates because you might get situations where you get a selection change like [1,1][1,1] then another one [1,2][1,2] and then a third one which is [1,1][1,1] again.

> 
>     2. The only chance for `selectionsChanged.ourSelections` to be cleared is we find a match 

Yes. This here is the main issue I think. I thought about creating a timeout (of something between 500ms to 1 or 2s) after we get a selection change event and `selectionsChanged.ourSelections.length` is higher than or equal to 1 and keep clearing this timeout and restarting it every time we get a selection change event until our selections are cleared. If that timeout finishes it would mean then that we were still expecting some selection change that never came, meaning this expected change is probably zombie data so we could be confident enough to clear it.

>        . Although #5212 is fixed if the first issue (above) is resolved but it doesn't mean we won't run into similar bugs in the future. I'm wondering if we can clear `selectionsChanged.ourSelections` when `ignoreIntermediateSelections` is set to `false`. It seems we will wrap all actions between `ignoreIntermediateSelections = true; ...; ignoreIntermediateSelections = false;` it can reduce  as many unnecessary selection change events as possible.

I don't think this would work because selection change events are weird and unpredictable. Consider this case, you have a remap that does a bunch of different actions. Since those actions will be called in succession from code they will hit much quicker than if the user had pressed themselves the corresponding keys. Sometimes, the intermediate selection events don't trigger when running the action, so we get to 'updateView' without them triggering (that's why I used that ignore slipped selections). Some times they bunch quite a few together and then you get something like 5 selection change events in row only after you ran through all your actions. This happens a lot when running tests! There the instructions come so fast one after the other that most times all the selection change events trigger only at the end of the test.

This all seems very unpredictable. The only thing I think we can count on is that if some time has passed without any selection change events and without us adding to our selections then that "probably" means that there will not other event coming, so if we're expecting one maybe we should reset it. That why I though of using a timeout.

I really wanted to keep selection change events enable during tests as well because I think this will help catching some future issues in testing in the future. I know I did catch quite a few when implementing this change!

berknam:
Also this PR is dealing with one situation I know for sure was creating zombie data. The selections we were considering when using multicursors were different than vscode (because vscode combines multicursor selections and we weren't doing that). This would result in 'updateView' thinking that vim's selections would trigger a selection change event when in reality they wouldn't.

J-Fields:
> The problem with that is that this would probably make quite a few tests fail.

Why's that, if you set the buffer time conservatively (say, 3 seconds)? Another option would be to just show a popup telling the user this happened (and they should report a bug; pretty please check for duplicates & tell us what you did)

src/mode/modeHandler.ts
```diff
@@ -1472,6 +1483,47 @@ export class ModeHandler implements vscode.Disposable {
...
       }
 
+      // Combine instersected selections - When we have multiple cursors
+      // sometimes those cursors selections intersect and combine, we need
+      // to check that here so that we know if our currents cursors will
+      // trigger a selectionChangeEvent or not. If we didn't check for this
+      // vscode might already have the resulting combined selection selected
+      // but since that wouldn't be the same as our selections we would think
+      // there would be a selectionChangeEvent when there wouldn't be any.
+      let combinedSelections: vscode.Selection[] = [];
```
J-Fields, line 1493: Split this into a separate function (inline if you like) to localize `combinedSelections`

src/actions/motion.ts
```diff
@@ -53,15 +53,12 @@ abstract class MoveByScreenLine extends BaseMovement {
...
-      vimState.cursorStopPosition.isAfterOrEqual(vimState.cursorStartPosition)
-    ) {
+    if (vimState.currentMode === Mode.Visual) {
       // The selection is on the right side of the cursor, while our representation considers the cursor to be the left edge
-      vimState.editor.selection = new vscode.Selection(
-        vimState.cursorStartPosition,
-        vimState.cursorStopPosition
-      );
+      const multicursorIndex = this.multicursorIndex ?? 0;
+      let selections = vimState.editor.selections;
```
J-Fields, line 59: Any reason `vimState.editor.selections[multicursorIndex] = ...` doesn't work?

src/mode/modeHandler.ts
```diff
@@ -1472,6 +1483,47 @@ export class ModeHandler implements vscode.Disposable {
...
+      selections.forEach((s, i) => {
+        if (i > 0) {
+          const previousSelection = combinedSelections[i - 1];
+          const overlap = s.intersection(previousSelection);
+          if (overlap) {
+            // If anchor is after active we have a backwards selection and in that case we want
+            // the anchor that is lower and/or to the right and the active that is up and/or to
+            // the left. Otherwise we want the anchor that is upper and/or to the left and the
+            // active that is lower and/or to the right.
+            const anchor = s.anchor.isBeforeOrEqual(s.active)
```
J-Fields, line 1503: I'm not a big fan of nested ternaries - since the first if/else (forward/backward selection) is the same for `anchor` and `active`, can you pull that out into a single real `if`/`else`?

MacMaru:
Possibly related #5289 ?

berknam:
> 
> 
> > The problem with that is that this would probably make quite a few tests fail.
> 
> Why's that, if you set the buffer time conservatively (say, 3 seconds)? Another option would be to just show a popup telling the user this happened (and they should report a bug; pretty please check for duplicates & tell us what you did)

Actually it might not be a problem at all. At some point when iterating through multiple tries I thought about using a timeout in this way but back then with the approach I was using it wouldn't work. But the current approach is different and it should be fine.

src/actions/motion.ts
berknam: I remember that when I did this I've made like you propose and it didn't work. But now that I'm looking more into this I believe this all "MoveByScreenLine" has quite a few bug and needs some change. I'm working on this.

berknam: This 'cursorMove' command moves the cursor on all multicursors at the same time, so we can only run once on the main cursor (index 0). Then on the other cursors we should only get the updated position. However sometimes the selections combine and then we'll have more cursors than vscode has selections, so we can't update a position of a cursor so I'm removing those cursors.
Example: If we have two cursors with two selections, if we move them in a way that they combine, vscode will have only one selection. But our 'MoveByScreenLine' will still be called for 'multicursorIndex' 1, when we see that it doesn't exist anymore, we return 'IMovement' with a new info `removed: true` so that we can at the end of 'executeMovement' remove the cursors that no longer exist.

If tests are ok I'll push this fix soon.

rebornix:
@berknam

> I don't think we should prevent having duplicates because you might get situations where you get a selection change like [1,1][1,1] then another one [1,2][1,2] and then a third one which is [1,1][1,1] again.

I misunderstood how `ourSelection` is used. Previously I thought it was closer to a map but it's actually the history of us setting `editor.selections` and then we try to ignore them when receiving `textEditorSelectionChangeEvent`, right?

I think we took an assumption here: `editor.selections = selections` will always trigger a selection change event (we check if `editor.selections === selections` before setting them) but it's possible others (for example, one extension from another extension host) modify the selections before our selection change request land in the UI thread.


> I don't think this would work because selection change events are weird and unpredictable. Consider this case, you have a remap that does a bunch of different actions. Since those actions will be called in succession from code they will hit much quicker than if the user had pressed themselves the corresponding keys. Sometimes, the intermediate selection events don't trigger when running the action, so we get to 'updateView' without them triggering (that's why I used that ignore slipped selections). Some times they bunch quite a few together and then you get something like 5 selection change events in row only after you ran through all your actions.

from extensions' angel, yes selection change event is not predictable. As long as we use async/await in the implementation, there is no guarantee when selection change events will come. For example, if we run five simple actions which are all synchronous (and also no other roles modify the editor selections), the selection change events will come only after the five actions are executed. If there are any async code, then some selection changes come in next tick.

---

IMHO the safe thing we did here is wrapping the action executions between `ignoreIntermediateSelections=true;`/`ignoreIntermediateSelections=false;`. When users press a key in UI, the `type` event is redirected to the extension host, it's guaranteed that there are no pending `TextEditorSelectionChangeEvent`. Then we flip `ignoreIntermediateSelections`, before we reset `ignoreIntermediateSelections` to `false`, all selection change events can be ignored. However if `ignoreIntermediateSelections` is already `false`, we don't really know if a selection change event is from us (Vim), or user (UI), or another extension.

Thus following code path is not safe in an async world. Clearing it after some timeouts is fixing the state after it goes wrong instead of preventing it from happening.

```ts
const idx = mh.vimState.selectionsChanged.ourSelections.indexOf(selectionsHash);
if (idx > -1) {
  logger.debug(
    `Selections: Ignoring selection: ${selectionsHash}, Count left: ${
      mh.vimState.selectionsChanged.ourSelections.length - 1
    }`
  );
  mh.vimState.selectionsChanged.ourSelections.splice(idx, 1);
  return;
}
```

J-Fields:
@rebornix Is the extension host's architecture documented anywhere? I'd like to have a better understanding of how these sorts of interactions between VSCode and an extension work

rebornix:
@J-Fields pinged you in Slack.

berknam:
> 
> I misunderstood how `ourSelection` is used. Previously I thought it was closer to a map but it's actually the history of us setting `editor.selections` and then we try to ignore them when receiving `textEditorSelectionChangeEvent`, right?

Correct! It's exactly that.

> 
> I think we took an assumption here: `editor.selections = selections` will always trigger a selection change event (we check if `editor.selections === selections` before setting them) but it's possible others (for example, one extension from another extension host) modify the selections before our selection change request land in the UI thread.
> 
> However if `ignoreIntermediateSelections` is already `false`, we don't really know if a selection change event is from us (Vim), or user (UI), or another extension.
> 

From my testing `editor.selections = selections`  only triggers a selection change event if, and only if the resulting final `selections` is different from `editor.selections`. Now the issue that was happening is that when we have multiple selections we weren't combining them, but when setting them to `editor.selections` if they intersected vscode would combine them. Then next time we would look into our selections (that weren't combined) and see they were different from vscode (even though if combined they would be equal) so we stored them in `ourSelections` and set them to `editor.selections`, but then vscode would noticed that they were intersecting, would combine them and see that the resulting selections were the same so it wouldn't trigger the selection change we were hoping for, thus creating zombie data.

In regards to other parties creating selection change events that shouldn't be an issue, because the worst that could happen is a third party creating a selection change equal to one we were expecting before ours was triggered, we would clear it from history thinking it was ours, and then when ours actually hit we would send it to `handleSelectionChange` which would happen any way if the order was the other way!
We do want to handle selection changes from third parties! We just don't need to/can't be handling ours.

berknam:
@J-Fields I've added a commit that fixes some bugs with the combining of selections. Still haven't done your review suggestions but I'll try to get to them later today.

I do believe that most of the issues we are seeing related to selections are due to this problem that we're fixing with this PR. It happens when using multicursors, but it can also happen when alt+clicking, using 'gb' or 'cmd+d' and the one I think is the major culprit is by using snippets that create multiple selections and goes to visual mode. Something that people aren't expecting because before it would stay in insert mode, so if the move they get instantaneous zombie data.

I propose for now, after making the review changes, we push this one to the beta version and see if it behaves better. Then we can either try the timeout approach to prevent future mishaps or think of something else.

J-Fields:
> I've added a commit that fixes some bugs with the combining of selections. Still haven't done your review suggestions but I'll try to get to them later today.

Sweet, I'll take another look when I have the chance. My review is mostly style nitpicking, if you disagree with anything feel free to override me

src/mode/modeHandler.ts
berknam: What do you mean here exactly with localize `combinedSelections`?

src/mode/modeHandler.ts
berknam: I don't like them either, is just that they seem easier to quickly write and test and I always forget to refactor them 😁

rebornix:
@berknam thanks for your explanation. Your current approach looks good to me for now before we come up with better solutions.

berknam:
@J-Fields Can we get this one merged and released to a beta version 1.17.2? It would really help to know if this was the only case or the main case creating these zombie data that will leave Vim in this unusable state or if there is more situations that lead into this.

J-Fields:
Hey @berknam, would you be interested in becoming a maintainer of VSCodeVim? For personal reasons I haven't had as much time to dedicate to the project recently as I'd like. I've been consistently impressed with your work and you definitely have a better understanding of certain parts of the project than anyone else does.

berknam:
Hey @J-Fields, first of all thank you for the offer. When you say 'maintainer' what would that entail? What 'obligations/responsibilities' would there be? I'm only asking because I've been doing my collaborations in my free spare time, and I'm more than happy to give some of my spare time to this project, but I don't always have the time for it.

Also, there are quite a few things that I don't have any experience with, like the back end stuff like CI/CD setup, performance diagnostics and most of the things related to the extension configuration like npm dependencies are a bit over my head so I would probably not be able to help much with those.

However, I do feel like this project should have more than just one maintainer. It is too much work for just one person. So if you're talking about helping you track and close some of the issues, reviewing PR's and maybe merge them (maybe just the simple ones at first), then yes I'm happy to do so!

J-Fields:
> What 'obligations/responsibilities' would there be?

None, really. Even as a maintainer, you can contribute when you have the time/motivation and not worry about it when you don't! I'd just like to remove some friction in that such as having to go through me for trivial changes or closing an issue.

> there are quite a few things that I don't have any experience with

Same 🙂 you can always pick those things up or stay in your comfort zone if you prefer.

> I do feel like this project should have more than just one maintainer. It is too much work for just one person

Yeah 😵 when I have a lot of time to put into the project, I feel it works well, but sometimes I have very little and things fall behind.

> So if you're talking about helping you track and close some of the issues, reviewing PR's and maybe merge them (maybe just the simple ones at first), then yes I'm happy to do so!

Awesome! I'll add you to the project now.

J-Fields:
Oh, nevermind - I actually don't have permission 🙂
@rebornix would you mind either adding @berknam as a member or elevating me to owner so I can do it?

J-Fields:
@jpoon @johnfn @xconverge Could any of you please add @berknam as a maintainer? I'm not a project owner so I'm unable to do it.

PR #5259, opened 2020-09-26
ExecuteOperator doesn't need to change selections

### LINKED ISSUE #5257: Bug: cursor jumps back to start of visual select after gv is used in mapping followed by paste
**TL;DR:**

Sorry, seems like you introduced a pesky bug recently. **Indeed, I just tried these exact same settings after downgrading to Vscode.Vim v.1.16.0 and it seemed to work again fine!**

**Details**

I often like to duplicate JS objects in an array. For example, to go from this:

```
[
    {
        A:1
    },
    {
        B:2
    },
    ...
]
```
... to this:

```
[
    {
        A:1
    },
    {
        B:2
    },
    {
        C:3
    },
    ...
]
```

... I typically visually select the comma before an object, move down to the closing bracket, then type "Yp" with these bindings:

```
{
        {
            "before": ["Y"],
            "after": [ "\"", "*", "y", "g", "v", "<ESC>"]
        }
    ],
        {      
            "before": ["p"],
            "after": ["\"", "*", "p"]
        }
    ]
}
```

... in order to duplicate the comma+object AFTER the comma+object, and then edit that duplicated object.

This worked fine *until very recently* (and still works fine in my terminal vim). However, what started happening recently (without me knowingly changing anything)  is that pressing 'p' after 'Y' will first cause the cursor to jump back to the beginning of the visual selection block, and *then* perform the paste which, of course, is not what I want (the whole point is to remain at the end of the object that I want to duplicate, and then paste after that object; also, this creates a double comma syntax error).

This problem only occurs when I try to use these _mappings_; it does not occur when I type the keys explicitly that define the mappings.

Sorry, seems like you introduced a pesky bug recently. **Indeed, I just tried these exact same settings after downgrading to Vscode.Vim v.1.16.0 and it seemed to work again fine!**

Thanks for your work on this very-helpful extension!

**1.17.1**
- VSCode version:
**1.49.2**
- OS:
**Mac OS 10.15.6**

berknam:
Stops the 'executeOperator' function from changing the editor selections.
- These selections will be changed on 'updateView' using the logic to
ignore them if needed
- This was in some cases creating a selection change event after the end
of the action and after 'updateView' had already finished. This event
would then wrongfully update the cursors.

fixes #5257 

I think I had previously removed this already, but it might have leaked through again in some merge conflict without me noticing it.

### DIFF
--- a/src/mode/modeHandler.ts
+++ b/src/mode/modeHandler.ts
@@ -1035,9 +1035,6 @@ export class ModeHandler implements vscode.Disposable {
     } else {
       // Keep track of all cursors (in the case of multi-cursor).
       this.vimState.cursors = resultingCursors;
-      this.vimState.editor.selections = this.vimState.cursors.map(
-        (cursor: Range) => new vscode.Selection(cursor.start, cursor.stop)
-      );
     }
   }
 


berknam:
@J-Fields @rebornix 
This is an important fix. There might be a lot of people facing issues with this one. It is also quite simple. Can any of you review this one?

J-Fields (approved):
LGTM, thanks @berknam! Always glad when we can simplify ModeHandler a bit 🙂

Issue #5188, opened 2020-09-15
`<<` substantially slower than `>>`


J-Fields:
I hadn't noticed it before, but it's very obvious now (might be some configuration I changed recently?)

PR #5287, opened 2020-09-28
Fix remap bug with '<' and special keys remaps

berknam:
Fix remap bug with '<' and special keys remaps
- Fix remaps confusing '<' as potential remap to special keys like '\<C-x\>'
- Add test for this situation
- Prevent showing special keys like '\<C-x\>' with virtual key decoration

I don't think anyone has reported this yet

### DIFF
--- a/src/configuration/remapper.ts
+++ b/src/configuration/remapper.ts
@@ -594,9 +594,15 @@ export class Remapper implements IRemapper {
     countRemapAsPotential: boolean = false
   ): boolean {
     const keysAsString = keys.join('');
+    const re = /^<([^>]+)>/;
     if (keysAsString !== '') {
       for (let remap of remappings.keys()) {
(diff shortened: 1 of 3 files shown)


J-Fields:
I'm sure this is the cause of #5188

berknam:
Do you have any remap with Ctrl+something, like `<C-e>` in normal mode?

berknam:
I'm also noticing that you reported that even before version 1.17 was out? Were you using the remap overhaul already?

J-Fields:
> Do you have any remap with Ctrl+something, like <C-e> in normal mode?

Yep I've got a few remaps like that

> I'm also noticing that you reported that even before version 1.17 was out? Were you using the remap overhaul already?

Yeah I generally dogfood with master

berknam:
Then yeah it might have been this.

J-Fields (approved):
Thanks!

PR #5290, opened 2020-09-29
Add vim.useSystemClipboard description to README

NuclearFej:
This setting is quite useful, and documenting it would help others in the same situation I was in.

I actually can't replicate it using the keybindings settings for some reason, this is the only way I know to bind the system clipboard to the default yank/delete/paste.

None, although I might have opened one if I hadn't found this setting outside GitHub (I see now that it's in the example, but that's easy to miss, which I obviously did)


README.md
```diff
@@ -118,6 +118,7 @@ These settings are specific to VSCodeVim.
...
 | vim.overrideCopy                 | Override VS Code's copy command with our own, which works correctly with VSCodeVim. If cmd-c/ctrl-c is giving you issues, set this to false and complain [here](https://github.com/Microsoft/vscode/issues/217).                                                                                                                                                                                                                    | Boolean | false                                 |
+| vim.useSystemClipboard           | Replace the unnamed register with the system clipboard. Pressing d/y/p alone in normal mode will use VS Code's native cut/copy/paste functionality instead of a Vim register.                                                                                                                                                                                                                                                       | Boolean | false                                 |
```
J-Fields, line 121: The second bit isn't quite right - we don't use VSCode's native cut/copy/paste, just the system clipboard

NuclearFej: Would it be more accurate to say that it behaves identically to VS Code's cut/copy/paste, or is there a difference I'm missing?

(I interpreted the calls to `vscode.env.clipboard` in https://github.com/VSCodeVim/Vim/blob/master/src/util/clipboard.ts to be equivalent to the default user-facing controls.)

NuclearFej: Should I open a new PR to fix the merge conflicts cleanly? with wording like
> Replace the unnamed register with the system clipboard. Pressing d/y/p alone in normal mode will use the system clipboard instead of a Vim register.

J-Fields:
Made this change myself, thanks for bringing it to my attention!

PR #5314, opened 2020-10-01
Remove hardcoded values for visual mode cursor colors and make cursor 'disappear' on command line modes

### LINKED ISSUE #5198: Cursor should disappear in command mode
In VIM once you type ':' the cursor moves to the command line and is hidden from the document. This is a visual cue I depend on all the time to know if the command mode was really triggered or not. 

In VSCodeVim the cursor keeps showing and blinking and this signals me that the command mode hasn't been triggered yet, so I automatically keep typing ':' after ':'.

It would be acceptable to show a special cursor.

Hot to reproduce:

Make sure you start in Normal mode

1)press ':'

Expected result: the command line gets the focus and the cursor hides from the document.

Actual result: the cursor remains.

### LINKED ISSUE #5294: Bad coloring of last character in search mode
**Describe the bug**
Since vscode-vim 1.17, with some color themes, the cursor has a bad coloring in search mode, and is barely readable.

**To Reproduce**
Use a color theme like dark+ or monokai (the issue happens with both). Search for something with `/`. The coloring of the last character is barely readable.
image

**Expected behavior**
The last character should appear correctly. As per 1.16.0:
image

VS Code version: Code 1.49.2 (e5e9e69aed6e1984f7499b7af85b3d05f9a6883a, 2020-09-24T16:29:41.983Z)
OS version: Windows_NT x64 10.0.19042
Remote OS version: Linux x64 3.10.0-1127.18.2.el7.x86_64

berknam:
- Removes the hard-coded cursor colors used in visual mode, now it uses the CSS `mix-blend-mode: difference` property to make the color be the opposite of the block cursor color. This is the same thing vscode does, except they actually calculate the value instead of using the CSS property. The result however is the same color as the normal vscode block cursor.

- Change the cursor in both command line modes to be 'UnderlineThin'. I don't think we can actually make it disappear so this is the closest to it. It at least allows the user to have a feedback that they are now typing on a different place.

fixes #5294 
fixes #5198 

Maybe this should have been separate but since they were so simple I kept them together any way. If you think it's better to separate them, tell me.


J-Fields:
Yeah sorry to be a bother, but could you pull these apart into two PRs? Thanks 🙂

berknam:
> 
> 
> Yeah sorry to be a bother, but could you pull these apart into two PRs? Thanks 🙂

I've splitted the underlineThin part to PR #5674.

J-Fields:
The text under the cursor looks thicker in visual mode with this PR. I tried messing with `fontWeight` which didn't appear to help. Any idea what could be going on?

J-Fields:
Maybe [this](https://stackoverflow.com/questions/45883243/thicker-text-after-applying-mix-blend-mode-difference)?

berknam:
> 
> 
> Maybe [this](https://stackoverflow.com/questions/45883243/thicker-text-after-applying-mix-blend-mode-difference)?

Yes I had already looked to that exact same post. The issue is not that it is thicker because it isn't. But because of the way it is rendered it "appears" to be thicker. I couldn't find a fix for that, in the end I thought it wasn't that big a deal. If you look closely you'll see that vscode's native blinking block cursor makes the letter appear thinner, at least it does for me!

cursorThinnerLetter1
The gif doesn't show it very well by I can see it.

Do you think this will be a problem?!

dagood:
The letters are small enough and the difference is hard enough to see that I think this might just be the subpixel rendering being different with the different colors:

image

To eliminate some variables, it might be useful to try a much higher font size (and/or high DPI monitor), turn off subpixel rendering, and test with pure white on black. But even then, there could still be perceptual differences. 😄 

That said, for me, https://github.com/VSCodeVim/Vim/issues/5294 is pretty annoying, and it would be nice to have this fix soon even if it needs to be hidden behind a config flag until it gets some polish.

catskul:
Any update on this? Any reason this can't land?

J-Fields:
Superceded by #7140

PR #5504, opened 2020-10-27
Fixes #5442 - [X]p does not work in visual line mode

### LINKED ISSUE #5442: Glitch after paste multiple lines
**Describe the bug**

When trying paste multiple lines, vscode starts to glitch and paste doesn't occur.

**To Reproduce**

1.  yank any line (**yy**)
2. select any line (**V**)
3. Paste multiple lines (**10p**, for example)

jopa

**Expected behavior**
I expect the specified number of lines to be paste

**Environment:**
- VSCode version: 1.50.0
- OS: 10.15.7

rickythefox:
 [X]p did not work in visual line mode. I've split PutCommandVisual and created a separate command for putting in visual line mode. This works by using the move count to temporarily replace the register content with a repeated version of itself. This is then used in the regular PutCommandVisual and the register is restored.

Fixes #5442

The split between the PutCommandVisual and the new PutCommandVisualLine should be more clean as there is logic related to putting in visual line mode present in PutCommandVisual. I felt that this PR is a step in the right direction (a separate put command for the visual line mode). It also fixes the immediate issue.


J-Fields:
Thanks @rickythefox - would you mind adding a few test cases?

rickythefox:
@J-Fields added some tests ;) note that `gp` is still broken. Will try to fix it.

rickythefox:
`gp` and `[x]gp` should now work too.

J-Fields:
Thanks! Put is horribly overcomplicated at the moment, splitting this out into its own function is probably the right move.

rickythefox:
@J-Fields the transformation implementation changed from under me. Fixed now.

J-Fields (approved):
Thanks!

PR #5543, opened 2020-10-30
Update README.md

joeynguyen:
The previous implementation throws this error:
`Failed to handle key=p. Failed to handle key=p. No multi-cursor index when calling PutCommand#getText`
in VS Code when I use multi cursor mode and have selected multiple selections.  This one works better IMHO.  It sends the selected code to the black hole register and then pastes the code in your clipboard in its place.

Improves on an existing suggested implementation.

Not sure if there's a current Issue for this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -345,12 +345,12 @@ Custom remappings are defined on a per-mode basis.
                 "p",
             ],
             "after": [
-                "p",
-                "g",
-                "v",
-                "y"
+                "\"",
+                "_",
+                "d",
+                "P"
             ]
-        }
+        },
     ],
 ```
 


berknam:
This change breaks the point that the documentation is trying to make. The thing here is for users to understand that when using a non recursive remap the "after" part of the remap won't try to remap again. That's why you can map `p` to an "after" that includes `p` without getting an infinite recursive error.

The error you mention is a bug that is being currently worked on #5713. So when it is fixed this example should work properly.

J-Fields:
👍🏻 thanks for clearing that up

joeynguyen:
@berknam thanks for the clarification.  Good to know about that other issue and that it's being worked on

PR #5587, opened 2020-11-05
Implement normalization for 'Shift+Tab' key combo

### LINKED ISSUE #4719: Mapping `<shift-tab>` does not work, need `<shift+tab>`
**Describe the bug**
First let me thank you for this amazing extension. It made me switch from neovim to VS Code. I use it pretty much every day and only had minor issues with multiple-cursors (which I think should be expected and totally fine).

Today I found something a bit more vexing: To be able to use `<shift-tab>` in mappings, instead of `<shift+tab>`. Noticed the dash in the first, and the plus sign in the second mapping.

**To Reproduce**
Given an config like this:

```json
{
        {
            "before": [
                "Tab"
            ],
            "commands": [
                "editor.action.indentLines"
            ]
        },
        {
            "before": [
                "<shift+tab>"
            ],
            "commands": [
                "editor.action.outdentLines"
            ]
        },
        {
            "before": [
                "<C-a>"
            ],
            "commands": [
                "cursorLineStart"
            ]
        },
        {
            "before": [
                "<C-e>"
            ],
            "commands": [
                "cursorLineEnd"
            ]
        }
    ]
}
```

Pressing `Shift+Tab` works like expected. But if you change it like so:

```json
            "before": [
                "<shift-tab>"
            ],
```

It stops working, without giving some sort of warning like if you miss the `s` in `commands`. Also not that `<C-?>` style mappings work fine and vanilla vim uses `<S-?>`/<`Shift-?>`.

**Expected behavior**
To be able to use dash in all mappings, or perhaps both dash and plus sign.

- Extension (VsCodeVim) version: 1.13.1
- VSCode version: 1.43.2
- OS: MacOS X Catalina

**Additional context**
This really bit me and took a lot of time to debug. Ended up sprinkling `this._logger` calls in the extension before realizing my mistake/the bug.

berknam:
Implement normalization for 'Shift+Tab' key combo
- \<Shift+Tab\> key was being captured but had no normalization
 implemented, so only '\<shift+tab\>' would work.
- This commit, makes '\<s-tab\>' work as well

This basically makes it possible to use `<S-tab>` in remaps as well instead of only `<shift+tab>`.

fixes #4719

I've implemented only the `<S-tab>` notation since that is what normal Vim works and `<shift+tab>` to keep the same philosophy that was implemented for `<ctrl+...>` and `<cmd+...>`. This should allow for the normalization of any `shift+...` key combination, however currently we only capture `<Shift+Tab>` and `<Shift+BS>`.
I didn't implement `<Shift-Tab>` as suggested in #4719 because that doesn't work in normal Vim either.

### DIFF
--- a/src/configuration/notation.ts
+++ b/src/configuration/notation.ts
@@ -5,6 +5,7 @@ export class Notation {
   private static readonly _notationMap: Array<[RegExp, string]> = [
     [/ctrl\+|c\-/gi, 'C-'],
     [/cmd\+|d\-/gi, 'D-'],
+    [/shift\+|s\-/gi, 'S-'],
     [/escape|esc/gi, 'Esc'],
     [/backspace|bs/gi, 'BS'],
     [/delete|del/gi, 'Del'],
@@ -32,6 +33,7 @@ export class Notation {
       this.isSurroundedByAngleBrackets(key) &&
       key !== '<BS>' &&
       key !== '<SHIFT+BS>' &&
+      key !== '<S-BS>' &&
       key !== '<TAB>'
     );
   }
--- a/src/configuration/notation.ts
+++ b/src/configuration/notation.ts
@@ -30,11 +30,7 @@ export class Notation {
   public static IsControlKey(key: string): boolean {
     key = key.toLocaleUpperCase();
     return (
-      this.isSurroundedByAngleBrackets(key) &&
-      key !== '<BS>' &&
-      key !== '<SHIFT+BS>' &&
-      key !== '<S-BS>' &&
-      key !== '<TAB>'
+      this.isSurroundedByAngleBrackets(key) && key !== '<BS>' && key !== '<S-BS>' && key !== '<TAB>'
     );
   }
 
--- a/src/actions/commands/actions.ts
+++ b/src/actions/commands/actions.ts
@@ -2594,11 +2594,11 @@ class ActionReplaceCharacter extends BaseCommand {
     const toReplace = this.keysPressed[1];
 
     /**
-     * <character> includes <BS>, <SHIFT+BS> and <TAB> but not any control keys,
+     * <character> includes <BS>, <S-BS> and <TAB> but not any control keys,
      * so we ignore the former two keys and have a special handle for <tab>.
      */
 
-    if (['<BS>', '<SHIFT+BS>'].includes(toReplace.toUpperCase())) {
+    if (['<BS>', '<S-BS>'].includes(toReplace.toUpperCase())) {
       return;
     }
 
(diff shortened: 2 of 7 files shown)


J-Fields:
I pushed a small change - let me know if you think it's wrong

src/configuration/notation.ts
```diff
@@ -32,6 +33,7 @@ export class Notation {
       this.isSurroundedByAngleBrackets(key) &&
       key !== '<BS>' &&
       key !== '<SHIFT+BS>' &&
```
J-Fields, line 35: Why are both `<Shift+BS>` and `<S-BS>` here? Can't we normalize to the latter?

berknam: Well when I made this change I didn't really had much time to see if it would mess up anything by removing that, so I left it there. But from what I've looked now I think it can be removed. It might create some issues with existing tests but we can update the tests. I'll try to look into it.

J-Fields: I think I broke the tests in the same way, but haven't had time to go back and fix it. I'd love if we could just normalize on the way in, have a canonical representation (`<S-___>` in this case) internally, and not worry about it.

berknam: That is usually what happens. But in tests sometimes we call the `modeHandler.handleKeyEvent('someKey')` directly and that doesn't go through normalization since that is done on the 'configuration' when loading the boundKeys or when handling remaps on the 'remappingValidator'.

J-Fields: Right, I think we should just make sure we're using the normalized representations in the test suite

berknam: I've just pushed a fix that changes all occurrences of `<shift+tab>` and `<shift+BS>` to `<S-tab>` and `<S-bs>`, respectively.

J-Fields (approved):
Thanks!

J-Fields:
Pretty annoying how long these builds take to make it through the Travis queue recently - I'll have to look into that and maybe switch to GitHub Actions or something if I can't resolve it...

berknam:
> 
> 
> Pretty annoying how long these builds take to make it through the Travis queue recently - I'll have to look into that and maybe switch to GitHub Actions or something if I can't resolve it...

At first I thought it was just some weird sync issue that it was showing me that is was in queue but that it was already running, but then it started to happen in all of them... Do you think it might be related to the 'extensionWeb' recently added? Because I've noticed that the building is taking longer as well and it is creating that extensionWeb with around 5MB? Is that ran through the webpack as well?

J-Fields:
I don't think that's the issue - it seems like Travis is taking a while to start the build, the build itself is pretty quick, but worth looking into.

PR #5597, opened 2020-11-07
Add Rust's INNER_LINE_DOC to reflow list

lf-:
See https://doc.rust-lang.org/reference/comments.html

Currently there's a problem where if you have:

```rust
//! A library for reading in Verilog $readmemb/$readmemh files formatted per IEEE Std. 1364-2005, §17.2.9
```

and the user goes onto the long line and does `gqq`, it will split it into two lines like this:

```rust
//! A library for reading in Verilog $readmemb/$readmemh files formatted per
//IEEE Std. 1364-2005, §17.2.9
```

This is incorrect. It should do this:

```
//! A library for reading in Verilog $readmemb/$readmemh files formatted per
//! IEEE Std. 1364-2005, §17.2.9
```


J-Fields:
Thanks!

PR #5674, opened 2020-11-17
Simulate cursor "disappearing" on CommandLineMode

### LINKED ISSUE #5198: Cursor should disappear in command mode

berknam:
Simulate cursor "disappearing" on CommandLineMode
- Change the cursor in both command line modes to be 'UnderlineThin'.
I don't think we can actually make it disappear so this is the closest
to it. It at least allows the user to have a feedback that they are now
typing on a different place.

fixes #5198 

### DIFF
--- a/src/mode/mode.ts
+++ b/src/mode/mode.ts
@@ -27,6 +27,7 @@ export enum VSCodeVimCursorType {
   Underline,
   TextDecoration,
   Native,
+  UnderlineThin,
 }
 
 /**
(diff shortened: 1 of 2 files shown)


J-Fields (approved):
LGTM. This should probably be configurable, but this is fine for now.

PR #5678, opened 2020-11-17
Fix n with matches at eol

### LINKED ISSUE #5591: Reverse search for `{` stops after first `n`
**Describe the bug**
When reverse searching (`?`) for `{`, `n` no longer advances to the next match after the first `n`.

**To Reproduce**

1.  Go to a file with multiple left curly brackets (`{`)
2. Go somewhere in the middle of those brackets
3. Type `?{<ENTER>nnnnn`

**Expected behavior**
`n` should continue to go to next matches

**Actual behavior**
After the first `n`, subsequent `n`s become no-ops.

> Note: `N` (go to previous match) works as intended (as shown in video below)

**Screenshots**
https://streamable.com/brdys8

- VSCode version: 1.51.0
- OS: macos 10.15.6

sql-koala:
This improves handling special cases where a search match is at EOL when jumping to next match using "n".

fixes #5591

Since start, we have the following special rule for n in code: when cursor is at EOL, search next match from after getRight. This is needed when searching for $ or \n. Otherwise we cant jump to next match.
But this creates a bug, when we search backwards and  our match is at EOL, because we are stuck on our match.
I fixed this for n and did an analog fix for N (because it has the same bug).
I checked the tests locally:
- the first fails, when removing getRight()
- the second fails without this PR or when removing "&& searchForward"


J-Fields:
Thanks!

Issue #4585, opened 2020-02-25
We need a `desiredColumn` per cursor


J-Fields:
Put the following into a new document, with `|` representing a cursor on lines 1 and 3:
```
|shortword
looooooonnnnnggggggword
|looooooonnnnnggggggword
looooooonnnnnggggggword
```

Press `e` then `j`; observe that the bottom cursor is in line with the top one

J-Fields:
We pass around `Range`s as "cursors". I'm beginning to think we should make a `Cursor` class that's passed around and encapsulates this sort of data...

PR #5713, opened 2020-11-20
Fix multi cursor visual paste (in part)

### LINKED ISSUE #5208: Operation with multi-cursor doesn't work properly
**Describe the bug**
Whenever there's multi-cursor involved, the following operation on multiple targets behaves weirdly...

**To Reproduce**

1.  Yank something using `yiw`
2.  Select multiple occurrences using `<C-d>` or `gb`
3.  Paste it using `p`
4.  Only the first occurrence was pasted, the rest of the occurrences were left untouched

**Expected behavior**
All occurrences are replaced by the pasted item

**Screenshots**

- VSCode version:
- OS: Win10 / MacOS High Sierra

**Additional context**
If I don't use yank and paste. Instead, I just use the normal ctrl c + ctrl v, then the multi-cursor are behaving properly. Just wondering how would people do operations like this in a Vim-ninja fashion considering multi-cursor may not be a thing in native Vim.

### LINKED ISSUE #5448: Failed to handle key=p. No multi-cursor index when calling PutCommand
Multi-line copy-paste action failed in Windows 10.

Is this possible?

### LINKED ISSUE #5493: Failed to handle key=P. Cannot read property 'split' of undefined
**To Reproduce**

1.  Go to '...'
2.  Click on '....'
3.  Scroll down to '....'
4.  See error

**VSCodeVim version**: 1.17.1

```
TypeError: Failed to handle key=P. Cannot read property 'split' of undefined
    at n.exec (c:\Users\Asus\.vscode-insiders\extensions\vscodevim.vim-1.17.1\out\extension.js:9:409709)
    at runMicrotasks (<anonymous>)
    at y.exec (c:\Users\Asus\.vscode-insiders\extensions\vscodevim.vim-1.17.1\out\extension.js:9:413539)
    at y.execCount (c:\Users\Asus\.vscode-insiders\extensions\vscodevim.vim-1.17.1\out\extension.js:1:37395)
    at F.runAction (c:\Users\Asus\.vscode-insiders\extensions\vscodevim.vim-1.17.1\out\extension.js:9:475401)
    at F.handleKeyAsAnAction (c:\Users\Asus\.vscode-insiders\extensions\vscodevim.vim-1.17.1\out\extension.js:9:474794)
    at F.handleKeyEvent (c:\Users\Asus\.vscode-insiders\extensions\vscodevim.vim-1.17.1\out\extension.js:9:471603)
    at Object.promise (c:\Users\Asus\.vscode-insiders\extensions\vscodevim.vim-1.17.1\out\extension.js:9:235740)
    at Object.runTasks (c:\Users\Asus\.vscode-insiders\extensions\vscodevim.vim-1.17.1\out\extension.js:9:621914)
```

</details>

sql-koala:
This improves visual paste for multicursor. It does the text change and stops the crash.
It does so, by disabling the yank part (for multi cursor only), so the replaced text is not copied to default register.

fixes #5493 
fixes #5208 
fixes #5448 

See:
https://github.com/VSCodeVim/Vim/issues/5493#issuecomment-731147687
https://github.com/VSCodeVim/Vim/issues/5208#issuecomment-700243223

So this is not a complete fix, but an acceptable workaround for now, imo,  (a complete fix looks too hard for me to try ).
If you disagree, feel free to not use this. Then, at least, we have this and the analysis as a starting point.


berknam:
@sql-koala Have you looked into how VsVim (from VisualStudio) does this? We could do something like that here too.

You can look that up [here](https://github.com/VsVim/VsVim/blob/master/Documentation/Multiple%20Selections.md#the-unnamaed-register). It basically creates multiple Registers per cursor.

J-Fields:
A long-term refactor I've slowly been working toward is #4585. I think storing registers on that new `Cursor` class is the right call, which sounds like what you're suggesting @berknam

sql-koala:
To explain this better, we can look at the complete workflow of pasting (over visual selections) with multicursor:

- generic execActionWithCount calls - for each cursor,  in sequence - PutCommandVisual.exec method
- this decomposes further to calling a sequence of operators:
1.  Putting the new text
2. Yanking the to be replaced/old text
3. Deleting old text
- the register state is not really isolated here and there are more complications: 
  * yanking one word/line and put it over multiple selections vs.
  *  yanking and putting with same number multiple selections

So it is no surprise that something goes wrong here. I agree with long term solutions like #4585 and #5856. But I cant do that ;-)
The idea for this quickfix is "better to not break than break" and it works by narrowly disabling the side effect of yanking the old text (which makes the operations per cursor isolated, I think).
If you want to use a "quickfix" like this, is a judgement call not for me make. But this pops up quite frequently.

Would appreciate feedback @J-Fields ;-)

src/actions/commands/put.ts
```diff
@@ -417,27 +421,29 @@ class PutCommandVisual extends BaseCommand {
       if (replaceRegisterName === deletedRegisterName) {
         Register.putByKey(deletedRegister.text, deletedRegisterName, deletedRegister.registerMode);
       }
-      return;
```
sql-koala, line 420: stylewise, I am not 100% sure to remove return here, but I am quite sure code for the alternative case should be in else {...}.

J-Fields: Yep seems reasonable to me

J-Fields:
Yeah this isn't an ideal solution but the yank/put logic is so fucking convoluted. This is definitely a good safe guard. Thanks!

Kadrian:
Hey, everyone who worked on this: thank you so much for getting this in, it's such a pleasure to use now!!

sql-koala:
@Kadrian Freut mich ;-)

PR #5725, opened 2020-11-21
Fix broken table in README

kidonng:

The table is broken:

image

Fix the table by removing unneeded characters.


J-Fields:
Thanks - not sure how that happened!

gyzerok:
Similar thing also happened to easymotion section

PR #5731, opened 2020-11-22
Fix bug where <C-i> <C-o> do not work after easy motion movement #5730

### LINKED ISSUE #5730: <C-o> and <C-i> do not work after EasyMotion movement
**Describe the bug**
`Ctrl + o` and `Ctrl + i` do not work correctly after vim easy motion movement

**To Reproduce**

1.  Open a document.
2.  Navigate to another place in the document using a vim easy motion command e.g. `<leader> <leader> 2 s`
3.  do `<C-o>` to go back to previous position
4.  Notice that it doesn't go back to correctly

**Expected behavior**
Should go back to previous position within the document

rndware:

Fixes bug where `<C-o>` and `<C-i>` do not work after an easy motion movement.

Fixes #5730

The bug is due to isJump not being set on the BaseCommand class which means it's not recorded in the movement history.

Really great plugin. I use it every day. Thanks for all the hard work and effort put into this.


J-Fields (approved):
Thanks!

PR #5814, opened 2020-12-03
Remove zombie data creator

### LINKED ISSUE #5813: <C-o> <C-i> will mess up cursor movement with left and right keys
**Describe the bug**
After using `<C-o>` `<C-i>` jumps, movements with `<left>` and `<right>` key will always bring the cursor back to an old position.

**To Reproduce**
(please also see video below)

1.  Open several files with VSCode
2.  Use `<C-o>` to jump to an old position, and then `<C-i>` to jump back (at 10s in the video).
3.  Use `<up>` or `<down>` to move to another line.
4.  Now press `<left>` or `<right>`, the cursor will jump back to old line.

**Screenshots**

To demonstrate, I've recorded a video: https://youtu.be/b6fV2dUVMF0

- Extension (VsCodeVim) version: 1.17.2
- VSCode version: 1.51.1
- Commit: e5a624b788d92b8d34d1392e4c4d9789406efe8f
- Date: 2020-11-11T01:11:34.018Z
- Electron: 9.3.3
- Chrome: 83.0.4103.122
- Node.js: 12.14.1
- V8: 8.3.110.13-electron.0
- OS: Darwin x64 19.6.0

berknam:
Remove zombie data creator
- When we perform a file jump (with \<C-o\> or \<C-i\> for example), the 'activeTextEditorChangedEvent' will already take care of getting the correct cursors to that new editor's `modeHandler`. Then we would call this 'getCursorsAfterSync' that would get the cursors of that new editor and apply them to the previous editor's `modeHandler` (which makes no sense)
- In this case we don't need to change cursors at all.

This would add an expected selection change that would never come (because it was from another file), thus leaving that "zombie" data). All further selection changes would then be ignored.

fixes #5813 
Might fix a bunch of the other issues we have with cursors failing or being invalid. Because if we were getting the cursors from one file and trying to apply them to other different file, issues are bound to happen when they don't have the same dimensions!

@J-Fields Do you think this sort of fixes could be pushed directly upstream or do you prefer that I create a PR first?

### DIFF
--- a/src/jumps/jumpTracker.ts
+++ b/src/jumps/jumpTracker.ts
@@ -4,7 +4,6 @@ import { FileCommand } from './../cmd_line/commands/file';
 import { VimState } from '../state/vimState';
 
 import { Jump } from './jump';
-import { getCursorsAfterSync } from '../util/util';
 import { existsAsync } from 'platform/fs';
 import { Position } from 'vscode';
 
@@ -169,7 +168,6 @@ export class JumpTracker {
 
     if (jumpedFiles) {
       await this.performFileJump(jump, vimState);
-      vimState.cursors = getCursorsAfterSync();
     } else {
       vimState.cursorStopPosition = jump.position;
     }


J-Fields (approved):

J-Fields:
> Might fix a bunch of the other issues we have with cursors failing or being invalid. Because if we were getting the cursors from one file and trying to apply them to other different file, issues are bound to happen when they don't have the same dimensions!

God, I hope so. This is fundamentally the issue with the `getLineLength` out of bounds issues which clog my inbox. I've been working to reduce our reliance on `activeTextEditor` (mostly via the static `TextEditor` utility class, which I think should be scrapped eventually), as `vimState.editor`/`vimState.document` should always be the thing we interact with within an `Action`. Hopefully this also reduces these sorts of issues.

> Do you think this sort of fixes could be pushed directly upstream or do you prefer that I create a PR first?

Up to you. If you're confident in a change and it's not massive, feel free to push directly. I'll probably take at least a cursory look at the commit anyway, just to keep aware of what's changing.

PR #5842, opened 2020-12-07
SelectMode implementation with 'keymodel' and 'selectmode' settings

### LINKED ISSUE #3260: Clicking end of line with mouse sometimes selects empty character
**Describe the bug**
When using a mouse to select a line, an empty character is sometimes selected at the end of the line instead of the last character.

**To Reproduce**
Click far right of lines until selected character is empty

**Expected behavior**
Cursor is at last character

**Screenshots**

vscode vim eol mouse example

- Extension (VsCodeVim) version: 0.16.14
- VSCode version: 1.29.1
- OS: MacOS 10.14.1

**Additional context**
Only happens with mouse, I'm unable to reproduce with touchpad.

### LINKED ISSUE #4039: [End of line] - mouse selection skip last character (edge case)
**Describe the bug**
When in vim 'Normal' mode, if the cursor is on the last character of a line, selecting with with a mouse omits the last character (the one before the newline).

**To Reproduce**

1.  Enter 'Normal' mode
2.  Navigate to last character of a line of text
3.  Select that line with the mouse
4.  Last character (before the newline) will not be in the selection

**Screenshots**

VSCodeVim 1.10.2
VSCode 1.37.1
MacOSX 10.14.6 (Mojave)

[Link to related issue](https://github.com/VSCodeVim/Vim/issues/1034)

### LINKED ISSUE #4075: Repeating gb once when selecting Multi-Cursors means subsequent gb are repeated that amount

### LINKED ISSUE #5716: Select Mode
See documentation for [Select Mode](https://vimhelp.org/visual.txt.html#Select-mode) and the [selectmode](https://vimhelp.org/options.txt.html#%27selectmode%27) option.

### LINKED ISSUE #5719: Changing to normal mode when clicking is not clearing RecordedState
**Describe the bug**
Currently when vscode creates a selection (like when using the "Find" widget) we go to visual mode. If the user then presses escape on the widget or closes the widget we are still in visual mode and if we press escape or use an operator everything works fine. However if the user clicks somewhere to get out of the widget instead of pressing escape or closing the widget, then the `handleSelectionChange` function changes the mode to normal, however since there was no action executed the `RecordedState` doesn't change and if we were on insert mode when calling the "Find" widget, the `RecordedState` will still have the action that was used to go to insert mode on the `actionsRun`.

**To Reproduce**

1.  Go to insert mode by pressing `i` or `a`
2.  Press `<C-f>` or call the "Find" widget from the command pallete
3.  Write a word to search (this step is probably not needed)
4.  Click somewhere to get the focus out of the "Find" widget (I usually press escape, that's why I never noticed this... 😄 )
5.  It goes into normal mode
6.  Press `i` or `a` to go to insert mode again and it fails...

**Expected behavior**
The recorded state should be cleared when we change to normal mode after a mouse click.

**Additional Context**
@J-Fields We could simply clear the `vimState.recordedState` when changing mode after the click or we could add that logic to the `VimState.setMode` function. Which one do you think it's better?

berknam:
## **What this PR does / why we need it:**  
This started with the implementation of 'keymodel' and the shifted special keys to replace the PR #5050. However it grew larger than expected! 😅 (sorry @J-Fields )

### Here is what is included now (new implementations):  
- Implementation of shifted special keys on all modes (`<S-arrowKeys>`, `<S-Home>`, `<S-End>`, `<S-PageUp>`, `<S-PageDown>`
- Implementation of 'keymodel' setting:
```
                                            'keymodel' 'km'
'keymodel' 'km'         string  (default "")
                                  global
        List of comma separated words, which enable special things that keys
        can do.  These values can be used:
           startsel     Using a shifted special key starts selection (either
                        Select mode or Visual mode, depending on "key" being
                        present in 'selectmode').
           stopsel      Using a not-shifted special key stops selection.
        Special keys in this context are the cursor keys, <End>, <Home>,
        <PageUp> and <PageDown>.
```
- Implement \<S-BS\> and \<C-BS\> on all modes
- Create insert/replace normal pseudo-mode (this happens when you press `<C-o>` on insert or replace mode, it is used just to give the same 'showmode' as normal VIM with `-- (insert) --` or `-- (replace) --`)
- Implement special insert/replace visual pseudo-modes. These pseudo-modes can be entered either by entering visual mode after using `<C-o>` on insert/replace mode or by using the shifted keys in insert/replace mode with `startsel` set on 'keymodel', or by doing a mouse drag selection on insert/replace mode. They show as `-- (insert) VISUAL --` or `-- (replace) VISUAL LINE --` for all possible combinations. The big difference of these modes is that after escaping (`<Esc>`) them or leaving the mode with an action like a delete or yank operator that would usually go to normal mode, now it will go to the mode before entering visual mode (insert or replace). (More information on this modes can be found on `:help operator-pending-mode`, it has info on all pseudo-modes)
- Implement `<C-o>` in replace mode (same as in insert mode, but it reverts back to replace mode)
- Implement `<LeftMouse>`:
    - This creates a new motion action `MoveLeftMouse` that runs when left clicking somewhere
    - Fixes #5719
    - When we get a single click we send a special key (`<LeftMouse>`) to `modeHandler.handleKeyEvent` which triggers a motion action `MoveLeftMouse` that then updates our `cursorStopPosition` according to `vimState.editor.selection.active`. This not only appears to work pretty well so far but it also brings with it some great improvements like:
        1. You can use left mouse click as a motion like in normal VIM. You can press `d` and then click somewhere and it will delete from `cursorStartPosition` *(where you were when pressing '`d`')* until the click position (exclusive motion like in VIM)
        2. You can remap `<LeftMouse>`! I don't know if this is useful at all, but it is something you can do just like you can in normal VIM. If you remap it, then our cursor won't be updated and vscode selection is changed back to our cursor.
- Implement SelectMode
    - Implements all Select modes (`SELECT`, `SELECT VISUAL`, `SELECT BLOCK` and the pseudo-modes like `-- (insert) SELECT --`)
    - Implement `<C-g>` to swap between Visual and Select modes
    - Implement SelectMode remaps:
        - Implement new `selectModeKeyBindings[NonRecursive]` for only select mode and `allVisualKeyBindings[NonRecursive]` that work for visual and select modes.
        - This `allVisual` remaps are the same as `vmap` on Vim and our existing `visualModeKeyBindings[NonRecursive]` are the same as `xmap` on Vim.
        - Implement correct .vimrc interpretation for the new mapping types
        - Implement special case for remaps of type `vmap` (`allVisualModeKeyBindings`) when ran in Select modes. In this situation vim temporarily changes to visual mode so that the behavior of the remap is the same regardless of the mode you're in: (`:help select-mode-mapping`)
        ```
        When mappings and menus are defined with the |:vmap| or |:vmenu| command they
        work both in Visual mode and in Select mode.  When these are used in Select
        mode Vim automatically switches to Visual mode, so that the same behavior as
        in Visual mode is effective.  If you don't want this use |:xmap| or |:smap|.
        (...)
        After the mapping or menu finishes, the selection is enabled again and Select
        mode entered, unless the selected area was deleted, another buffer became
        the current one or the window layout was changed.
        ```
        **TODO: currently we check if the selected area was deleted by checking if the text in that area is the same. This might not be the best way to do so, but I couldn't think of a better way and couldn't find that piece of code in VIM to figure out how they do it! (ideas are welcome!)**
    - Implement 'selectmode' setting:
    ```
                                                'selectmode' 'slm'
    'selectmode' 'slm'      string  (default "")
                            global
            This is a comma separated list of words, which specifies when to start
            Select mode instead of Visual mode, when a selection is started.
            Possible values:
               mouse        when using the mouse
               key          when using shifted special keys
               cmd          when using "v", "V" or CTRL-V
            See Select-mode.
    ```
    **TODO: we have a special situation that doesn't exist in VIM which is that we may get a visual selection from vscode, either from native commands like the "Find" widget or from 3rd party extensions. We need to figure out what to do in those cases. Currently I have it going to Visual mode instead of Select. We could add another possible value to the 'selectmode' setting like `vscode` that would make those situations always go to Select mode with that setting, or (my favorite option) we can do it on a per mode basis, so if we get one of those selections and we are in normal mode we would go to Visual, but if we were on insert/replace mode we would go to Select, this would be more similar to the experience the users have been getting since selections in insert mode currently behave just like Select mode.**

### Small bug fixes included:  
- Make \<BS\> work on all visual modes
- Update 'wrapping' `shouldWrapKey` to include all visual modes and replace mode (TODO: this function should be used in all places that are affected by `whichwrap`, currently it is only applied on `h` and `l` and the rest either don't have it correctly implemented or re-implement something similar to this function)
- Add count handling on Visual and Visual Block modes. This makes 'v' and '<C-v>' with count move 'count' to the right, if you press `2v` it will select 2 characters.
- Fix mouse selection not selecting last character of line (fixes #4039, fixes #3260)
- Fix go to insert mode actions not updating start position. This wasn't a problem before but it turned out to be a problem after implementing Select Mode.
- Fix `gv` (`ReselectVisual`) going left when mode is different from visual mode. It should only update left on visual mode.
- Fix `IncrementDecrementNumberAction`:
    - Fix <C-a>, <C-x> (Increment/Decrement number) in visual mode changing the number twice (it was getting the ranges duplicated)
    - fix <C-a>, <C-x> (Increment/Decrement number) in visual mode updating outside the range (in visual mode if you select '1' on the number '14' it should increase to '24' but it was going to '15')
- Fix `recordedState.count` not being cleared when it should. Make `recordedState.count` be cleared after `executeOperator` and after running a `BaseCommand` with `isCompleteAction = true`, also made the enter visual mode command `v` be a complete action. (why was it not a complete action???)
    - Change the `BaseCommand.isCompleteAction` to `false` by default so that we know that when setting it to true we need to either handle the count properly or have the `runsOnceForEachCountPrefix` set to true. This makes the `isCompleteAction` actually behave like a flag that states that the action ran was complete. Fixes #4075 and can replace #5077.

## **Which issue(s) this PR fixes**  
Some of them are state above. Others were never created and there are quite a few that should be fixed by this but I just haven't had the time yet to look for them (I'll try to add them later).
Fixes #5716 

## **Special notes for your reviewer**:  
This again is quite an overhaul, but I think it had to be in order to properly implement SelectMode. I tried to split the fixes on their own commits before the SelectMode commit so that we could cherry-pick if needed (except for the last fix that made changes to pretty much all actions so I made that one after SelectMode to include the new actions as well)

Please check out the `TODO`s mentioned above and give me your opinion. There is also the following todos that I would like an opinion in some of them:
- [ ] implement <C-o> in select/s-line/s-block mode you can use <C-o> to do one visual/v-line/v-block command
- [ ] nested <C-o> (you can use <C-o> in insert mode to do one normal command, then run 'gh', 'gH' or 'g^V' to
go to one of the select modes and then use <C-o> again to run one visual/v-line/v-block command, after this
you go back to select mode, then if you escape you go back to insert mode)
- [ ] implement <C-n>, <C-p> in all visual and select modes (they are the same as <up>, <down>)
- [ ] CommandCmdA (<D-a>) -> selectAll - should it apply in selectMode as well? I don't know if it does in vim?!
- [ ] ActionOverrideCmdD (<D-d>) -> addToSelection - should it apply in selectMode as well?
- [ ] ActionOverrideCmdAltDown (<D-alt+down>/<C-alt+down>) -> insertCursorBelow - should it apply in selectMode as well?
- [ ] ActionOverrideCmdAltUp (<D-alt+up>/<C-alt+up>) -> insertCursorAbove - should it apply in selectMode as well?
- [ ] ActionTriggerHover is now shadowed by CommandSelectMode that enters select mode, we need to get another shortcut
for it, ideas?!

**Some issues I have:**
- *There is currently an "issue" with \<S-Home\> in insert mode: it selects the character on the right of the cursor like in normal VIM, vscode users that are used to <S-Home> not being handled by VSCodeVim will probably not going to like this change... 😞 Altough this can be fixed with a remap like `inoremap <S-Home> <S-Home><C-g>o<left>o<C-g>` (the `<C-g>` are only needed if 'selectmode' has 'key')*
- *As stated in `:help select-mode-mapping`, the printable keys pressed in select mode go through the insert mode remapper. In order to implement this I needed to have access to the modeHandler (using `await getAndUpdateModeHandler()`) inside the action `CommandInsertInSelectMode` in order to resend the key, so that it goes through the insert mode remapper. I really don't like this, but I couldn't think of another way!*


berknam:
If anyone can give a try to this PR and report back I would really appreciate! From what I've tested this looks and feels really good. I didn't know I wanted SelectedMode so much before having it! 😄 

I didn't change the behavior of snippets, they still stay in insert mode with a vscode selection, but that could (probably should) be changed, so that after a snippet it would go into SelectMode, this way it would look and feel the same, however you would have all SelectMode remaps and all SelectMode actions available. The only reason I didn't implement this is that it would require not capturing the TAB key like it happens in insertMode, but maybe it's better to do so.

package.json
```diff
@@ -849,6 +965,23 @@
...
+            "startsel",
+            "stopsel",
+            "startsel,stopsel"
+          ],
+          "default": "startsel"
+        },
+        "vim.selectmode": {
+          "type": "string",
+          "markdownDescription": "This is a comma separated list of words, which specifies when to start Select mode instead of Visual mode, when a selection is started.\nPossible values:\n\t `mouse`\t -> when using the mouse\n\t `key`\t -> when using shifted special keys\n\t `cmd`\t -> when using 'v', 'V' or 'CTRL-V'",
+          "pattern": "^(?:mouse(?:,\\s*|\\b)|key(?:,\\s*|\\b)|cmd(?:,\\s*|\\b)){1,3}$",
```
stevenguh, line 982: This pattern doesn't seem to match an empty string. I am wondering is empty string allowed, so basically I want to disable select mode.

package.json
```diff
@@ -849,6 +965,23 @@
           "description": "If enabled, dragging with the mouse activates Visual mode.",
```
stevenguh, line 965: nit: I was confused when I was looking at the config. The description should probably be updated since this is depended on `vim.selectmode`. Or maybe remove completely since we have selection mode?

stevenguh:
I tried out some simple usages of select mode as I am not too familiar with select mode, and it works great! That's some amazing work. Thank you :)

package.json
berknam: Nice catch! Yes empty string should be allowed!

package.json
berknam: I was thinking of making it "deprecated" and while it is deprecated it would change the defaults of 'selectmode' to be:
```jsonc
"vim.selectmode": "key", // <- it removes the 'mouse' option so it goes to Visual Mode instead of Select Mode
```

berknam: Actually this isn't correct. Because this setting is the equivalent of having at least `mouse=nv` or `mouse=a`. Because this is the setting that allows a mouse selection to start visual mode. The `selectmode=mouse` makes the mouse selection start Select mode. So in order to replace that setting we would have to implement the `mouse` setting. The problem is that in VIM if you have `mouse=` the mouse doesn't do anything, it doesn't even move the cursor, and even though this might be possible to do, it probably won't look good, since you'll have the cursor changing place and then go back again to where it was...

package.json
berknam: This is now fixed. Thank you!

package.json
berknam: Basically this setting on/off is the equivalent to having `on` -> `mouse=a` or `off` -> `mouse=ni`, except that when `off` the vscode selection is still kept, even though VSCodeVim doesn't know about it...

stevenguh: > `on` -> `mouse=a` or `off` -> `mouse=ni` the vscode selection is still kept, even though VSCodeVim doesn't know about it

until user clicks `<shift+right>` for example. As long as we are not implementing `mouse` option, I think we just need to clarify what `vim.mouseSelectionGoesIntoVisualMode` does? Also, I don't know how hard it is to implement `mouse` option?

Since it doesn't work exactly like the `mouse` option, maybe updating the description to something like following would help?
> If enabled, dragging with the mouse activates Visual/Select mode depending on `vim.selectmode`.

berknam: Yes the description needs to be updated. And so does the README file, I think we need to create a "Selections" section on the README to explain all things related to selections, but I really didn't want to add to the already big confusion that is the README file. We need to create an organized Wiki...

J-Fields:
How simple would it be to split the following out into their own commits/PRs?

- Implementation of shifted special keys on all modes (<S-arrowKeys>, <S-Home>, <S-End>, <S-PageUp>, <S-PageDown>
- Implement <S-BS> and <C-BS> on all modes
- Create insert/replace normal pseudo-mode (this happens when you press <C-o> on insert or replace mode, it is used just to give the same 'showmode' as normal VIM with -- (insert) -- or -- (replace) --)
- Implement <C-o> in replace mode (same as in insert mode, but it reverts back to replace mode)

As well as some of the bug fixes unrelelated to select mode?

package.json
```diff
@@ -119,6 +129,96 @@
...
+        "command": "extension.vim_ctrl+shift+up",
+        "when": "editorTextFocus && vim.active && !inDebugRepl && !suggestWidgetVisible && !parameterHintsVisible"
+      },
+      {
+        "key": "ctrl+shift+down",
+        "command": "extension.vim_ctrl+shift+down",
+        "when": "editorTextFocus && vim.active && !inDebugRepl && !suggestWidgetVisible && !parameterHintsVisible"
+      },
+      {
+        "key": "ctrl+Home",
```
J-Fields, line 203: I assume it doesn't matter, but is there a reason `Home` and `End` are capitalized here but not below?

package.json
```diff
@@ -487,6 +587,22 @@
           "type": "array",
           "markdownDescription": "Non-recursive keybinding overrides to use for Visual mode. Allows mapping to Vim commands or VS Code actions. See [README](https://github.com/VSCodeVim/Vim/#key-remapping) for details."
         },
+        "vim.allVisualModeKeyBindings": {
+          "type": "array",
+          "markdownDescription": "Remapped keys in Visual and Select mode. When using these mappings in Select mode, vim temporarily switches to Visual mode, so that the same behavior as in Visual mode is effective. Allows mapping to Vim commands or VS Code actions. See [README](https://github.com/VSCodeVim/Vim/#key-remapping) for details."
```
J-Fields, line 592: `vim` should be capitalized here and below

package.json
```diff
@@ -845,6 +961,23 @@
           "description": "If enabled, dragging with the mouse activates Visual mode.",
           "default": true
         },
+        "vim.keymodel": {
+          "type": "string",
+          "markdownDescription": "Defines the behavior of shifted arrow keys and List of comma separated words, which enable special things that keys can do. These values can be used:\n\t `startsel` -> Using a shifted special key starts selection (either Select mode or Visual mode, depending on 'key' being present in 'selectmode').\n\t `stopsel` -> Using a not-shifted special key stops selection.\nSpecial keys in this context are the cursor keys, `<End>`, `<Home>`, `<PageUp>` and `<PageDown>`.",
```
J-Fields, line 966: Since this is markdown, maybe using bullet points is better than `\n\t`?

src/actions/base.ts
```diff
@@ -174,9 +174,10 @@ export abstract class BaseAction {
 export abstract class BaseCommand extends BaseAction {
   /**
    * If isCompleteAction is true, then triggering this command is a complete action -
-   * that means that we'll go and try to run it.
+   * that means that after the command is ran we can clear recordedState.count because
```
J-Fields, line 177: nit: "command is run", not "command is ran"

src/actions/base.ts
```diff
@@ -174,9 +174,10 @@ export abstract class BaseAction {
...
+   * it will have been handled already.
    */
-  isCompleteAction = true;
+  isCompleteAction = false;
```
J-Fields, line 180: Why change the default value of `isCompleteAction`?

src/configuration/configuration.ts
```diff
@@ -129,6 +129,17 @@ class Configuration implements IConfiguration {
       this.wrapKeys[wrapKey] = true;
     }
 
+    // set keymodel options
+    const km = this.keymodel.split(',');
+    this.keymodelStartsSelection = km.includes('startsel');
+    this.keymodelStopsSelection = km.includes('stopsel');
+
+    // set selectmode options
+    const sm = this.selectmode.split(',');
```
J-Fields, line 138: I think is better done by defining a getter for `selectmode` which returns a value like this:

```typescript
{
  mouse: boolean,
  key: boolean,
  cmd: boolean,
}
```

(and similar for `keymodel`). Or maybe a separate method (not a real getter) to avoid messing up things like `:set selectmode?`

src/configuration/iconfiguration.ts
```diff
@@ -296,6 +296,29 @@ export interface IConfiguration {
...
+   * Special keys in this context are the cursor keys, <End>, <Home>,
+   * <PageUp> and <PageDown>.
+   */
+  keymodel: string;
+
+  /**
+   * This is a comma separated list of words, which specifies when to start
+   * Select mode instead of Visual mode, when a selection is started.
+   * Possible values:
+   *   mouse	when using the mouse
```
J-Fields, line 316: Indentation

src/actions/commands/actions.ts
```diff
@@ -1588,36 +1804,150 @@ class CommandClearLine extends BaseCommand {
...
+        newMode = Mode.Visual;
+        break;
+      case Mode.SelectLine:
+        newMode = Mode.VisualLine;
+        break;
+      case Mode.SelectBlock:
+        newMode = Mode.VisualBlock;
+        break;
+      default:
+        // This shouldn't happen! Should we throw an error here???
```
J-Fields, line 1923: Yeah invalid state like this definitely deserves throwing an exception

src/configuration/remapper.ts
```diff
@@ -480,6 +483,21 @@ export class Remapper implements IRemapper {
 
     vimState.recordedState.resetCommandList();
     if (remapping.after) {
+      let changeToVisualForRemap: Mode | undefined;
+      let selectedRange: vscode.Range | undefined;
+      let selectedText: string | undefined;
+      if (this._configKey === 'allVisualModeKeyBindingsMap' && isSelectMode(vimState.currentMode)) {
+        // Remaps from 'allVisualModeKeyBindingsMap' (VIM's vmap) when ran in select modes temporarely
```
J-Fields, line 490: spelling: `temporarely`

src/configuration/remapper.ts
```diff
@@ -480,6 +483,21 @@ export class Remapper implements IRemapper {
...
+        // change to visual mode so that the execution is similar to the visual mode
+        // (:help select-mode-mapping)
+        changeToVisualForRemap = vimState.currentMode;
+        selectedRange = new vscode.Range(
```
J-Fields, line 494: nit: I think this can just be `selectedRange = vimState.editor.selection;` since IIRC `Selection` is a subclass of `Range`

src/mode/mode.ts
```diff
@@ -10,14 +10,34 @@ export enum Mode {
...
   CommandlineInProgress,
   Replace,
   EasyMotionMode,
   EasyMotionInputMode,
   SurroundInputMode,
-  OperatorPendingMode, // Pseudo-Mode, used only when remapping. DON'T SET TO THIS MODE
   Disabled,
+  // The following modes are Pseudo-Modes, used only when remapping or for 'showmode'
+  // to give feedback to user.
+  // DON'T SET TO THIS MODES!!!
```
J-Fields, line 25: spelling: s/THIS/THESE

src/mode/mode.ts
```diff
@@ -10,14 +10,34 @@ export enum Mode {
...
+  SelectLine,
   SearchInProgressMode,
   CommandlineInProgress,
   Replace,
   EasyMotionMode,
   EasyMotionInputMode,
   SurroundInputMode,
-  OperatorPendingMode, // Pseudo-Mode, used only when remapping. DON'T SET TO THIS MODE
   Disabled,
+  // The following modes are Pseudo-Modes, used only when remapping or for 'showmode'
```
J-Fields, line 23: I wonder if there's a better way of encoding this information, rather than an ever-growing list of modes. Any thoughts?

package.json
berknam: This was because it was copied from the `"key": "Home",` above. I'll make them all lowercase.

src/actions/base.ts
berknam: Because of the explanation above. When it `isCompleteAction` the action is supposed to handle the `recordedState.count` (either by the default way, which calls the `exec` a total of `recordedState.count` times or by handling it in a custom way on `execCount`).

This was supposed to remember people that if the action isComplete than you have to set that to `true` and then handle `execCount` or let it be handled by default.

If you think this was better the other way, I can change it...

src/configuration/configuration.ts
berknam: This would definitely look better, but I'm afraid that it won't be as readable and as simple to understand:
```ts
if (configuration.selectModeMouse) {
```
vs
```ts
if (configuration.selectMode().mouse) {
```

I could even change the current implementation adding `By` like this, which could be even more readable:
```ts
if (configuration.selectModeByMouse) {
```
```ts
if (configuration.selectModeByKey) {
```
```ts
if (configuration.selectModeByCmd) {
```

src/configuration/remapper.ts
berknam: Didn't notice that. 😄

src/mode/mode.ts
berknam: We could do something like this:
```ts
export enum DefaultMode {
  Normal,
  Insert,
  Visual,
  VisualBlock,
  VisualLine,
  Select,
  SelectBlock,
  SelectLine,
  SearchInProgressMode,
  CommandlineInProgress,
  Replace,
  Disabled,
}

export enum PluginMode {
  EasyMotionMode,
  EasyMotionInputMode,
  SurroundInputMode,
}

export enum PseudoMode {
  // The following modes are Pseudo-Modes, used only when remapping or for 'showmode'
  // to give feedback to user.
  // DON'T SET TO THESE MODES!!!
  OperatorPendingMode,
  InsertNormal,
  ReplaceNormal,
  InsertVisual,
  InsertVisualBlock,
  InsertVisualLine,
  ReplaceVisual,
  ReplaceVisualBlock,
  ReplaceVisualLine,
  InsertSelect,
  InsertSelectBlock,
  InsertSelectLine,
  ReplaceSelect,
  ReplaceSelectBlock,
  ReplaceSelectLine,
}

export type Mode = DefaultMode | PluginMode | PseudoMode;
export const Mode = { ...DefaultMode, ...PluginMode, ...PseudoMode };

```

This way we could separate the modes and everything else should still work the same. Then on functions like `setCurrentMode(mode : Mode)` we would change it to `setCurrentMode(mode : DefaultMode | PluginMode)` this way you could never set to a pseudo-mode.

The only issue with this approach would be that all pieces of code like this:
```ts
if ([Mode.Insert, Mode.Replace].includes(vimState.currentMode))
```
would have to be changed to:
```ts
if (([Mode.Insert, Mode.Replace] as Mode[]).includes(vimState.currentMode))
```

berknam:
> 
> 
> How simple would it be to split the following out into their own commits/PRs?
> 
>     * Implementation of shifted special keys on all modes (, , , ,
> 
>     * Implement  and  on all modes
> 
>     * Create insert/replace normal pseudo-mode (this happens when you press  on insert or replace mode, it is used just to give the same 'showmode' as normal VIM with -- (insert) -- or -- (replace) --)
> 
>     * Implement  in replace mode (same as in insert mode, but it reverts back to replace mode)
> 
> 
> As well as some of the bug fixes unrelelated to select mode?

It might be doable, but it might not be that easy because some of the things are related to each other and depend on each other. I tried to make the commits in a way were this things are separated.

If you wanted this because of being easier to pinpoint issues, then I can just push the commits on this PR directly to master, once we feel the PR is ready, instead of merging this PR. This way all commits will be in history and should be possible to pinpoint any issues created by any of this commits.

src/mode/mode.ts
J-Fields: That seems slightly better to me, but still not ideal. I think this deserves more thought; let's keep it as is for now.

J-Fields:
> > How simple would it be to split the following out into their own commits/PRs?
> > ```
> > * Implementation of shifted special keys on all modes (, , , ,
> > 
> > * Implement  and  on all modes
> > 
> > * Create insert/replace normal pseudo-mode (this happens when you press  on insert or replace mode, it is used just to give the same 'showmode' as normal VIM with -- (insert) -- or -- (replace) --)
> > 
> > * Implement  in replace mode (same as in insert mode, but it reverts back to replace mode)
> > ```
> > 
> > 
> > As well as some of the bug fixes unrelelated to select mode?
> 
> It might be doable, but it might not be that easy because some of the things are related to each other and depend on each other. I tried to make the commits in a way were this things are separated.
> 
> If you wanted this because of being easier to pinpoint issues, then I can just push the commits on this PR directly to master, once we feel the PR is ready, instead of merging this PR. This way all commits will be in history and should be possible to pinpoint any issues created by any of this commits.

Makes total sense. I think the best way to do this is with a real merge commit, rather than the squash-merge we usually do.

package.json
J-Fields: https://github.com/VSCodeVim/Vim/wiki/Selections

If you feel so inclined, dump stuff here - even if it's not polished, better than the current state of documentation.

J-Fields:
@berknam Thanks again for the excellent work!

I don't understand every nook and cranny of this, but it all seems reasonable to me and I trust you've figured out the details well enough, so I'm ready to merge this when you are. If you think any of the TODOs you listed are blockers, let me know; otherwise, could you migrate them into their own issue(s) so we can discuss & track them separately?

berknam:
@J-Fields I'm sorry I haven't had any time to work on this project lately. I hope that by the end of this month the situation should get better and I should be able to get back to this project.

Miigon:
This is awesome! I'm constantly bothered by the fact that using shift+arrowKeys to select code will leave me in NORMAL mode instead of putting me into VISUAL mode.(yes, I am a newbie who uses arrowKeys a lot😂) Hope this gets merged soon!

J-Fields:
Hey @berknam - do you still plan on getting back to this? I'm very happy to help fix merge conflicts and split this up so we can start merging it in

xRoker:
@J-Fields would it be possible to merge at least some of the commits from this PR? Specifically, https://github.com/VSCodeVim/Vim/pull/5842/commits/1e0984bc5b4785e1aa2211e2f56690b0a496ab2d as it resolves https://github.com/VSCodeVim/Vim/issues/4039 (a very annoying bug).

J-Fields:
1e0984b merged (a7f71a1a53b1519cddf8d9a625f189192063ff86)

v3ss0n:
Can we get this merge , this would make vscode VIM experience much better. ( Or i would end up switching to neovim soon lol)

Issue #5306, opened 2020-09-30
Cannot read property 'after' of undefined


ns11:
**To Reproduce**

1.  Go to '...'
2.  Click on '....'
3.  Scroll down to '....'
4.  See error

**VSCodeVim version**: 1.17.1

<details><summary>Stack trace</summary>

```
TypeError: Cannot read property 'after' of undefined
	at Object.promise (/Users/***/.vscode/extensions/vscodevim.vim-1.17.1/out/extension.js:9:236829)
	at runMicrotasks (<anonymous>)
	at processTicksAndRejections (internal/process/task_queues.js:94:5)
	at async Object.runTasks (/Users/***/.vscode/extensions/vscodevim.vim-1.17.1/out/extension.js:9:621914)
```

</details>

J-Fields:
What were you trying to do when this happened?

berknam:
@ns11 Where you trying to run the "Vim: Remap..." command from command palette?

ckoparir:
Ok thanks, can be closed.

berknam:
@J-Fields Are you considering this a duplicate of #5841? Are we sure that they are the same?

J-Fields:
Yeah I was considering them duplicates. @ckoparir could you clarify - did you try to run the command `Vim: Remap any key combination that VS Code supports to Vim motions/operators/ExCommands/macro.`?

ckoparir:
Sorry, I can't repeat the same issue or clarify it now, because I've been working completely different OS and IDE. Actually I can't just remember the details about the issue that I've faced before. Thank you for your time and interest.

Issue #5186, opened 2020-09-14
Can't map keybinding to: "Ctrl+\"


ProgrammingLife:
Trying to map keybinding "Ctrl+\" to quit from insert mode. My config looks like this:
```
"vim.insertModeKeyBindings": [
	{
		"before": [ "<C-\\>" ],
		"after": [ "<Esc>" ]
	}
]
```
It doesn't work this way. Also it's not possible to set "C-\". How to fix it?

VsCodeVim: 1.16.0
VSCode: 1.48.2

sql-koala:
Hi, take a look at #5152 . It should be possible to adapt.

berknam:
@ProgrammingLife @stu0292 @sql-koala @J-Fields 

Just so you all know and for future reference, you can pretty much remap any key that isn't handled by VSCodeVim like this (on 'keybindings.json'):
```jsonc
    {
        "key": "ctrl+oem_5",
        "command": "vim.remap",
        "when": "editorTextFocus && vim.active && vim.mode == 'Insert' && !inDebugRepl"
        "args": {
            "after": ["<Esc>"]
        }
    },
```
For me the key `\` is `oem_5`.

You can also change in which mode to apply said remap with the 'when' clause using `vim.mode == 'Insert'`

It can also remap to commands, however currently if it has an "after" it will just remap to that "after" keys and ignore the commands (maybe we should change this...). So if you wanted to do the above keybinding while at the same time formatting the current line you could do it this way:
```jsonc
    {
        "key": "ctrl+oem_5",
        "command": "vim.remap",
        "when": "editorTextFocus && vim.active && vim.mode == 'Insert' && !inDebugRepl"
        "args": {
            "commands": [
                {
                    "command": "extension.vim_escape",
                    "args": []
                },
                {
                    "command": "editor.action.format",
                    "args": []
                },
            ]
        }
    },
```
*Notice that in this case I couldn't use "after" and had to use the `extension.vim_escape` command*

The nice thing of this approach is that you can map to multiple commands without having to use an extension like "multi-command"

PR #5843, opened 2020-12-07
Remove "vim.remap" from the command palette

### LINKED ISSUE #5841: Cannot read property 'after' of undefined
**To Reproduce**

I'm running VSCode Insiders on Mac:

```
Version: 1.52.0-insider
Commit: 532ba43c78c82b24bdf94f9edde30cdf9b9bfa12
Date: 2020-12-05T01:08:42.214Z
Electron: 9.3.5
Chrome: 83.0.4103.122
Node.js: 12.14.1
V8: 8.3.110.13-electron.0
OS: Darwin x64 19.6.0
```

1.  Run the action "Vim: Remap key combination  that VS Code supports as Vim ..." (vim.remap)
  - I did this because of my request for the feature for #5822 where it was said that C-f and C-b should be working as page up/down - but I'm not seeing that. I then tried to do this remap, and saw the error - I'm wondering if that is related then.
2.  See error

**VSCodeVim version**: 1.18.0

```
TypeError: Cannot read property 'after' of undefined
    at Object.promise (/Users/darrenk/.vscode-insiders/extensions/vscodevim.vim-1.18.0/out/extension.js:9:229185)
    at Object.runTasks (/Users/darrenk/.vscode-insiders/extensions/vscodevim.vim-1.18.0/out/extension.js:11:285853)
```

</details>

berknam:
Remove "vim.remap" from the command palette
- This command is supposed to be used in keybindings, it can't be used
directly from the command palette

- fixes #5841

We should add an entry for this command explaining its use on the README file. I just don't know were or if we should even do it at all. I'm afraid it might create more confusion and future issues.

We really need a Wiki! 😄 Then we could add an advanced section inside a Remappings section where we could add this.

I also think this might be the cause of the issues related to #5306. But I'm not sure because no one as yet explained what they were doing...

### DIFF
--- a/extensionBase.ts
+++ b/extensionBase.ts
@@ -411,11 +411,16 @@ export async function activate(context: vscode.ExtensionContext, handleLocal: bo
         return;
       }
 
+      if (!args) {
+        throw new Error(
+          "'args' is undefined. For this remap to work it needs to have 'args' with an '\"after\": string[]' and/or a '\"commands\": { command: string; args: any[] }[]'"
+        );
+      }
+
       if (args.after) {
         for (const key of args.after) {
           await mh.handleKeyEvent(Notation.NormalizeKey(key, configuration.leader));
         }
-        return;
       }
 
       if (args.commands) {
--- a/package.json
+++ b/package.json
@@ -46,10 +46,6 @@
   "contributes": {
     "commands": [
       {
-        "command": "vim.remap",
-        "title": "Vim: Remap any key combination that VS Code supports to Vim motions/operators/ExCommands/macro."
-      },
-      {
         "command": "toggleVim",
         "title": "Vim: Toggle Vim Mode"
       },


J-Fields:
Hmm, I wasn't aware that you could provide and execute commands that aren't exposed in `contributes.commands`, but looks like that's fine.

I wonder if anyone even uses this - I was only vaguely aware of it. Is the point to allow incorporating `when` clauses to make remaps conditional on context?

J-Fields:
> We really need a Wiki! 😄

Agreed 100%. I'd like to break the README up and migrate the pieces to the wiki along with more detailed info. Just haven't found the time yet.

J-Fields:
In addition to this, we should check the existence/structure of the argument to `vim.remap` and throw a more helpful error

berknam:
> 
> 
> Hmm, I wasn't aware that you could provide and execute commands that aren't exposed in `contributes.commands`, but looks like that's fine.

Me neither. Just found out now as well.

> I wonder if anyone even uses this - I was only vaguely aware of it. Is the point to allow incorporating `when` clauses to make remaps conditional on context?

That is one possibility yes, but I think the main advantage of "vim.remap" is that it allows to remap keys that are not captured by Vim. Like mentioned here https://github.com/VSCodeVim/Vim/issues/5186#issuecomment-699168965, this allows to map keys from different keyboard layouts that are not captured by Vim.

berknam:
There is also the issue that currently if you have some `after` and `commands` as well, then `commands` will never run, because "vim.remap" checks for `after` and then returns, only if it `after` is undefined will it look for commands. This is not the same behavior as it happens in our remaps.

berknam:
@J-Fields Does this last commit make it any better, at least for now?

PR #5848, opened 2020-12-08
Fix broken table in README's vim-easymotion section

kidonng:

Fix broken table in README's vim-easymotion section mentioned in https://github.com/VSCodeVim/Vim/pull/5725#issuecomment-739733173.

I'm not sure if the extra characters are part of the content, so correct me if the regex is not correct.


J-Fields (approved):
Not sure how these got messed up. Thanks!

PR #5892, opened 2020-12-16
Fix gq to preserve blank lines like Vim does (fix #2393)

### LINKED ISSUE #2393: `gqap` deletes whitespace in LaTeX
BUG REPORT

**What happened**:
Ran `gqap` to wrap a paragraph of text in a tex file.  While the text was wrapped correctly, whitespace up until the next paragraph was incorrectly deleted.

Paragraph is wrapped at 80 characters and contents outside the paragraph are unmodified.

Create a tex file with a line longer than 80 characters, several blank lines, and then more text.  Move the cursor to the first line/paragraph and run `gqap`.

**Environment**:

- Extension Name: vim
- Extension Version: 0.10.13
- OS Version: Darwin x64 17.4.0
- VSCode version: 1.20.1

edemaine:
Fixes the `gq` operator to not discard blank lines at the beginning or end of a chunk, like Vim does. This turned out to be a one-line change: blank lines at the beginning and end of a chunk were getting accidentally removed via an extra `.trim()` call. There was already code to properly deal with blank lines (which previously handled cases like reformatting two paragraphs separated by blank lines, but now also handle initial and final blank lines).

#2393 shows examples of the bad behavior.  Open any text document with multiple paragraphs (separated by blank lines), such as the repository's `README.md`.  Type `gqap` on a long-line paragraph (e.g. `VSCodeVim is a Vim emulator...`) and the newline(s) after that paragraph will disappear with the old code, but not disappear with this PR (as in Vim).

**Which issue(s) this PR fixes**: Fixes #2393

* I confirmed that the current implementation of `ap` (a-paragraph) motion matches that of Vim. In particular, `yap` (yank a-paragraph) *does* yank the blank lines that follow a paragraph. So the issue was with `gq`, not `ap`.
* In a second commit, I added some minor optimization that seemed to be intended, replacing repeated calls to `line.trim()` with an already existing `trimmedLine`.
* I'm new to this repository (and VSCode development in general), and didn't succeed in running the automatic tests. I did test `gqap` by hand using a built `.vsix`. (It would have been nice to have a `CONTRIBUTING.md` file to get me started [`yarn`, `yarn run build`, `yarn run vsce package --web`]. Happy to write one assuming there's interest.)

### DIFF
--- a/src/actions/operator.ts
+++ b/src/actions/operator.ts
@@ -923,7 +923,7 @@ class ActionVisualReflowParagraph extends BaseOperator {
       let commentType: CommentType | undefined;
 
       for (const type of ActionVisualReflowParagraph.CommentTypes) {
-        if (line.trim().startsWith(type.start)) {
+        if (trimmedLine.startsWith(type.start)) {
           commentType = type;
 
           break;
@@ -931,13 +931,13 @@ class ActionVisualReflowParagraph extends BaseOperator {
 
         // If they're currently in a multiline comment, see if they continued it.
         if (lastChunk && type.start === lastChunk.commentType.start && !type.singleLine) {
-          if (line.trim().startsWith(type.inner)) {
+          if (trimmedLine.startsWith(type.inner)) {
             commentType = type;
 
             break;
           }
 
-          if (line.trim().endsWith(type.final)) {
+          if (trimmedLine.endsWith(type.final)) {
             commentType = type;
 
             break;
--- a/src/actions/operator.ts
+++ b/src/actions/operator.ts
@@ -1008,7 +1008,7 @@ class ActionVisualReflowParagraph extends BaseOperator {
       // don't want to add an extra space. In addition, when there's a blank
       // line, this needs to be reset.
       let curIndex = 0;
-      for (const line of content.trim().split('\n')) {
+      for (const line of content.split('\n')) {
         // Preserve newlines.
 
         if (line.trim() === '') {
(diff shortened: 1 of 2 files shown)


J-Fields:
Thanks! This doesn't seem to behave correctly when multiple blank lines are selected - it adds n-1 blank lines instead of leaving them as-is.

We do have a `CONTRIBUTING.md` - our documentation is a bit lacking generally, so let me know if there are any specific issues you have with it and I can adapt as needed.

edemaine:
@J-Fields Thanks for testing this out!  Indeed, there turned out to be a lot more lines of logic that needed adjusting to properly preserve blank lines.  I believe the new code does the trick, and I added some tests to solidify this.  Overall, I think the code is actually cleaner than before, which is nice, though the changes are more extensive.

Along the way, I also ran into various existing bugs with formatting comments.  Formatting `/* foo */` didn't work well, as well as other variations where `/*`, ` *`, and/or `*/` appeared on the same line.  Also, two multiline comments like `/* foo */` got merged, which Vim doesn't do; there was no code to notice that a multiline comment had been ended.  I fixed all such issues that I could find.

One exception: Vim preserves the whitespace on blank lines, while before and after this PR, such whitespace is removed.  I couldn't find an easy way to fix that, so left it for future work.

Let me know if you can think of further tests to add.

J-Fields:
LGTM. Thanks a bunch!

PR #5895, opened 2020-12-16
Support 'set joinspaces' in join and reformat paragraph (fix #2505)

### LINKED ISSUE #2505: Support 'joinspaces' option
**Is this a BUG REPORT or FEATURE REQUEST?** (choose one): Feature request

**What happened**: When formatting text with <kbd>gq</kbd> or joining lines with <kbd>J</kbd>, French spacing (one space after punctuation) is used.

**What did you expect to happen**: Non-French spacing (two spaces after punctuation) to be used (as is the default in Vim), and/or support for the `joinspaces` option that controls this functionality in Vim.

Create a file with the text:

    a.
    b

Place your cursor on the first line and press <kbd>J</kbd> or <kbd>gqj</kbd>. VSCodeVim produces the result:

    a. b

While Vim (with default options) produces the result:

    a.  b

When typing `:set joinspaces` in VSCodeVim, an error message appears:

> Vim: Error: Unsupported option. joinspaces

**Environment**:

- Extension Name: vim
- Extension Version: 0.11.3
- OS Version: Linux x64 4.13.0-37-generic
- VSCode version: 1.21.1

edemaine:
This PR implements the `joinspaces` configuration option, set via `set joinspaces` or `set js` and unset via `set nojoinspaces` or `set nojs`, and configured in VSCode settings by `vim.joinspaces`, as requested in #2505 (marked `help wanted`).

As described in #2505, [`joinspaces` is a simple Vim option](http://vimdoc.sourceforge.net/htmldoc/options.html#'joinspaces') that affects the behavior of two other features. When set:
* Action 'J' (join line) will separate joined lines by two spaces instead of one when the first line ends with `.`, `?`, or `!`.
* Operator `gq` (reformat paragraph) will similarly separate two words by two spaces instead of one when the first word ends with `.`, `?`, or `!`.

I also fixed a small bug in the behavior of 'J' (join line). Consider the following document where the second line is blank:
```
hello

world
```
Currently in VSCodeVim, if the cursor is on the first line, pressing <kbd>J</kbd> appends a space to the first line (and consumes the second line). But Vim doesn't behave that way: it just consumes the second line, without adding any space. I've fixed the code to behave like Vim in this situation, which also makes the join code substantially shorter.

Fixes #2505

* Should `joinspaces` be `true` or `false` by default? In Vim, the default is `true`, so this would make sense for Vim compatibility (coming from Vim, I found it a bit shocking when joining a line for the first time in VSCodeVim and not getting the expected behavior), and this is arguably fixing a bug in Vim emulation. But if you're worried about backward compatibility with past VSCodeVim versions, `false` would make it behave like before until users change the setting.
* ~~There will be a slight conflict with #5892 because I had to edit an adjacent line. I'm happy to resolve the conflict once one of the PRs gets merged.~~ Now resolved.
* Vim supports a Vi-compatibility mode [`set cpoptions=j`](http://vimdoc.sourceforge.net/htmldoc/options.html#'cpoptions') that changes `joinspaces`'s behavior to only use two spaces after `.` (not `?` or `!`). I don't personally find this important: Vim compatibility seems desirable, but Vi compatibility doesn't seem like a priority, even though I am old enough to have used Vi before Vim was released. 😄 But if you'd like me to implement, I probably can.  (Note that `cpoptions` currently isn't supported at all in VSCodeVim; supporting all of it would probably be beyond my available time, but `cpoptions=j` would be easy enough.)


edemaine:
I fixed the (expected) merge conflict with #5892, so this is now ready for review.

J-Fields:
Thanks @edemaine!

kainino0x:
I don't know if it was due to this patch or a later one, but I recently found that e.g. `gqq` replaces all whitespace after a `.` with an exact number of spaces (1 or 2), throughout the paragraph. This differs from Vim, which appears to preserve all internal spaces (unless they get turned into a line break), using `joinspaces` only to determine whether a `\n` should be replaced with one space or two spaces.

J-Fields:
@kainino0x Almost definitely this one. Can you please open an issue with repro steps and link to this PR?

kainino0x:
Done, thanks for the quick response!

nomad-software:
Can this be defaulted to false please? This is breaking behaviour for all users. Now everyone will have to add a config option to revert back.

edemaine:
For what it's worth, I don't think this PR caused the issue.  I'm still running a version of the Vim pluging that doesn't have the `set [no]joinspaces` feature, and it has the same problem.  But I'll take a look to see whether I can fix it.

nomad-software:
I've opened another issue to change the default: https://github.com/VSCodeVim/Vim/issues/6572

PR #5924, opened 2020-12-23
Show Marks in Gutter

### LINKED ISSUE #3963: Re-enable editor decorations for marks under a config setting
Seeing marks, ideally in the gutter, but maybe as a highlight or something, would be helpful. This apparently existed at one point (https://github.com/vscodevim/Vim//commit/713f30a1a8f4923dfc2e7b0cbceffce78ab83c01#diff-a7723f32fe6b05e135838f84e5c23c0fR870), but decoration options have gotten richer and we can put it under a setting.

### LINKED ISSUE #5119: Ability to view marks set in the file.
Marks are not visible and there is no way to confirm where marks are present in the file.

Extension https://github.com/kshenoy/vim-signature implementation

zstix:
## What this PR does / why we need it:
Adds the ability to see what marks you have set in the gutter.

This PR expands on the great work @gamcoh has in #4715 and adds the mark name to the "bookmark" icon (see screenshots below). Since this might not be everyone's cup of tea, I've added a new setting to enable / disable this (defaults to disabled).

This is my first PR in this project; let me know if I missed anything!

* Closes #3963
* Closes #5119
* Relates to #4710

## Special notes for your reviewer:
Screen Shot 2020-12-23 at 10 22 24 AM
_NOTE: the default is currently set to **off**_
Screen Shot 2020-12-23 at 10 22 35 AM

### DIFF
--- a/src/configuration/decoration.ts
+++ b/src/configuration/decoration.ts
@@ -1,30 +1,34 @@
 import * as vscode from 'vscode';
 import { IConfiguration } from './iconfiguration';
 
-// TODO: move this to a helper file?
-const getMarkUri = (text: string) => {
-  const svg = [
-    '<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 30 30" width="30px" height="30px">',
-    '<style>text { font-family: sans-serif; font-size: 0.8em; }</style>',
-    '<circle cx="15" cy="15" r="9" fill="rgb(3,102,214)" />',
-    `<text x="50%" y="50%" fill="rgb(255,255,255)" text-anchor="middle" dominant-baseline="middle">${text}</text>`,
-    '</svg>',
-  ].join('');
-
-  const uri = `data:image/svg+xml;utf8,${encodeURI(svg)}`;
-
-  return vscode.Uri.parse(uri, true);
-};
+interface IMarks {
+  [key: string]: vscode.TextEditorDecorationType;
+}
 
 class DecorationImpl {
   private _default: vscode.TextEditorDecorationType;
-  private _mark: vscode.TextEditorDecorationType;
+  private _marks: IMarks;
   private _searchHighlight: vscode.TextEditorDecorationType;
   private _easyMotionIncSearch: vscode.TextEditorDecorationType;
   private _easyMotionDimIncSearch: vscode.TextEditorDecorationType;
   private _insertModeVirtualCharacter: vscode.TextEditorDecorationType;
   private _operatorPendingModeCursor: vscode.TextEditorDecorationType;
   private _operatorPendingModeCursorChar: vscode.TextEditorDecorationType;
+
+  private _markNames = [...'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ'];
+
+  private getMarkUri(text: string) {
+    const svg = [
+      '<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 30 30" width="30px" height="30px">',
+      '<style>text { font-family: sans-serif; font-size: 0.8em; }</style>',
+      '<circle cx="15" cy="15" r="9" fill="rgb(3,102,214)" />',
+      `<text x="50%" y="50%" fill="rgb(255,255,255)" text-anchor="middle" dominant-baseline="middle">${text}</text>`,
+      '</svg>',
+    ].join('');
+    const uri = `data:image/svg+xml;utf8,${encodeURI(svg)}`;
+
+    return vscode.Uri.parse(uri, true);
+  }
 
   public set default(value: vscode.TextEditorDecorationType) {
     if (this._default) {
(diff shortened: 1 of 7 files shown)


gamcoh:
That's great! Hope it gets merge soon

gamcoh (approved):

zstix:
@J-Fields Let me know if I missed anything with this update. Thanks!

src/configuration/decoration.ts
```diff
@@ -1,15 +1,35 @@
...
+    const svg = [
+      '<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 30 30" width="30px" height="30px">',
+      '<style>text { font-family: sans-serif; font-size: 0.8em; }</style>',
+      '<path fill="rgb(3,102,214)" d="M23,27l-8-7l-8,7V5c0-1.105,0.895-2,2-2h12c1.105,0,2,0.895,2,2V27z"/>',
+      `<text x="50%" y="40%" fill="rgb(200,200,200)" text-anchor="middle" dominant-baseline="middle">${text}</text>`,
+      '</svg>',
+    ].join('');
+    const uri = `data:image/svg+xml;utf8,${encodeURI(svg)}`;
+
+    return vscode.Uri.parse(uri, true);
```
J-Fields, line 30: Do you think it's worth caching these parsed `Uri`s?

src/configuration/decoration.ts
```diff
@@ -102,6 +130,18 @@ class DecorationImpl {
       borderWidth: '1px',
     });
 
+    this.marks = this._markNames.reduce(
+      (marks, name) => ({
+        ...marks,
+        [name]: vscode.window.createTextEditorDecorationType({
```
J-Fields, line 136: Probably it makes more sense to cache the `TextEditorDecorationType`s. IIRC, these take a non-negligible time to construct if you're doing it often.

src/mode/modeHandler.ts
```diff
@@ -1510,7 +1510,17 @@ export class ModeHandler implements vscode.Disposable {
...
     );
 
-    // TODO: draw marks (#3963)
+    if (configuration.showMarksInGutter) {
+      for (const { position, name } of this.vimState.historyTracker.getMarks()) {
+        const markLine = position.getLineBegin();
+        const markRange = new vscode.Range(markLine, markLine);
+        const markDecoration = decoration.marks[name];
+
+        if (this.vimState.editor.hasOwnProperty('setDecorations')) {
```
J-Fields, line 1519: Why are you checking whether `editor.setDecorations` exists? We used it just above, so if it doesn't, we've presumably already thrown. Is there any reason to suspect it may not exist?

src/mode/modeHandler.ts
```diff
@@ -1510,7 +1510,17 @@ export class ModeHandler implements vscode.Disposable {
       opCursorCharDecorations
     );
 
-    // TODO: draw marks (#3963)
+    if (configuration.showMarksInGutter) {
```
J-Fields, line 1513: If this is `false`, we should probably clear the decorations by calling `editor.setDecorations(decorationType, [])` for each `decorationType` we've constructed (this will only matter if the user disables it at runtime).

src/configuration/iconfiguration.ts
```diff
@@ -398,4 +398,9 @@ export interface IConfiguration {
    * and `>>` with a linewise operator.
    */
   startofline: boolean;
+
+  /**
+   * Enable showing mark(s) in the gutter.
```
J-Fields, line 403: This should match the description in `package.json` ("Show the currently set mark(s) in the gutter.")

src/mode/modeHandler.ts
J-Fields: This check actually fails for me (Maybe `TextEditor` is a subclass of whatever defines `setDecorations`?) so the feature doesn't work unless I remove it

src/configuration/decoration.ts
```diff
@@ -21,6 +41,14 @@ class DecorationImpl {
     return this._default;
   }
 
+  public set marks(value: IMarks) {
```
J-Fields, line 44: What's the purpose of this getter/setter pair?

J-Fields (changes requested):
Left a few comments, but overall this seems good. Nice job on the icons, they look really good (personally I'd like an option for no bookmark background, just text, and slightly larger - maybe I'll add that at some point).

I did notice that the icons don't update position as you type (adding lines above, for instance). This is no big deal for now, just something to note.

src/mode/modeHandler.ts
zstix: When I ran the test suite without this condition, I ran into a bunch of errors about `editor` not having the key `setDecoration`. I'm still fairly unfamiliar with the project and I wasn't quite sure what was going on. I can try to dig into it further.

zstix: Just pushed up fixes to the other feedback you provided (thank you!).

We'll see if the tests pass with those updates 🤞

zstix:
@J-Fields Thanks for the great feedback. I've incorporated those changes and it looks like the tests are now passing.

🎁

J-Fields:
Made a small edit so that the DecorationTypes wouldn't be created unless the feature is enabled. Should be ready to merge pending green tests. Thanks a bunch @zstix!

J-Fields (approved):

zstix:
Thanks @J-Fields I really appreciate the assistance! 🎉

simplast:
Is there any way to show it in minimap?

Nikita0x:
why is it hidden by default though?

PR #5941, opened 2020-12-29
Fixes for `viw` and whitespace at the beginning of the line

### LINKED ISSUE #5572: `viw` results in wrong selection in whitespace at beginning of line
Scenario 1:
```python
if x is None:
  | y = 0
```
`viw` should select all of whitespace before `y`, but not the newline or anything on the first line. Our current implementation selects up to and including the `:`.

Scenario 2:
```python
  | y = 0
```
This is the first line in the file. Same as above, `viw` should select all the whitespace. Here our current implementation misses the first space.

zstix:
## What this PR does / why we need it:
VSCodeVim was not behaving like vim when the user selects the inner word (`viw`) and they are on leading space on a line.

### Scenario:

```python
if x is None:
 | y = 0
```

Typing `viwx` would result in the following:

```python
if x is None:y = 0
```

This PR ensures that VSCodeVim behaves like vim,  by _just_ selecting the leading space on the line, resulting in this:

```python
if x is None:
y = 0
```

* Closes #5572 

## Special notes for your reviewer:
I've added two extra tests to cover the problems presented in the linked issue. I recognize that this PR interacts with a fairly core part of the extension, so let me know if I missed something!

### DIFF
--- a/src/textobject/textobject.ts
+++ b/src/textobject/textobject.ts
@@ -273,7 +273,7 @@ export class SelectInnerWord extends TextObjectMovement {
     const currentChar = vimState.document.lineAt(position).text[position.character];
 
     if (/\s/.test(currentChar)) {
-      start = position.getLastWordEnd().getRight();
+      start = position.getWordLeft(true);
       stop = position.getWordRight().getLeftThroughLineBreaks();
     } else {
       start = position.getWordLeft(true);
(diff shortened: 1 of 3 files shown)


zstix:
Hmm, I'm not quite sure why the prettier step is failing. It seems like there is an issue with the branch not being defined? I could use some help figuring out what's going on there.

Either way, it doesn't _seem_ to be an issue with the content of the PR.

J-Fields:
`gulp forceprettier` fixed it. I wonder if github actions can't modify the branch on your repo?

J-Fields:
Hey @zstix this breaks a handful of tests, and I suspect the fix is a bit more complicated than you'd expected. Should I close this PR or do you plan on coming back to it?

zstix:
@J-Fields feel free to close it, I haven't had the opportunity to look into the tests. I agree, this is probably more complicated than I originally thought.

J-Fields:
LMK if you do want to give it another go; I'm gonna close for now just bc I'm trying to reduce number of open PRs

PR #5946, opened 2020-12-30
Document build/test process better

edemaine:
1. `yarn test` runs `node ./out/test/runTest.js` which doesn't work because `out/test/runTest.js` does not exist yet.  Add `npx gulp prepare-test` step.
2. Mention how to build `.vsix` extension.

Mentioned as a side note in #5892


J-Fields (approved):
Thanks!

PR #5271, opened 2020-09-27
Fix remaps when entering chinese input mode (composition state)

berknam:
**What this PR does / why we need it**:
When entering composition state to enter Chinese characters we now don't send the characters to Vim even in insert mode, we only send them after composition end. This should also allow to use Chinese characters in remaps on insert mode.

**Which issue(s) this PR fixes**
fixes #4815 
fixes #5237 

**Special notes for your reviewer**:

Issue #1287, opened 2017-02-09
Korean characters overlap each other


Anongh:
* Click *thumbs-up* 👍 on this issue if you want it!
* Click *confused* 😕 on this issue if not having it makes VSCodeVim unusable. 

The VSCodeVim team prioritizes issues based on reaction count. 

--------

### What did you do?
I entered Korean characters in insert mode.

### What did you expect to happen?
I expected a situation where the cursor would move to the right without overlapping when entering the letters as usual.

without Vim;
![without vim](https://cloud.githubusercontent.com/assets/22952514/22779832/adbac226-eeff-11e6-91f1-9b7c6b113215.gif)

### What happened instead?
During typing, the letters overlap each other.

with Vim:
![with vim gif](https://cloud.githubusercontent.com/assets/22952514/22779835/b027a2fe-eeff-11e6-9c86-add5dfb68704.gif)

### Technical details:

* VSCode Version: 1.9.1 (The same problem has occurred since 1.8)
* VsCodeVim Version: 0.5.2
* OS: Windows 8.1

xconverge:
@rebornix any ideas on this one?

misoguy:
I'm interested in fixing this issue and have been digging around.
It seems like when a type is made in Korean(Chinese), the following code is invoked
>registerCommand(context, 'compositionStart', async (args) => {
and listens to 'replacePreviousChar' to compose the texts and changes the state when 'compositionEnd' is invoked.
When those handlers are deleted it works like the screenshot above without vim but it comes with some side effects...(The one that I could see is bugs in normal mode)
I'll try to dig more but any guide or hint would be much appreciated!

rebornix:
@misoguy thanks for taking a look at this issue. Please note that we bind our specific handlers for composition to fix problems with CJK IME in Normal Mode. So whatever modification we make to this piece of code, we don't want to have regression.

5pecia1:
The same issue occurred.

VSCode Version: 1.52.0 
VsCodeVim Version: 1.18.5
OS: Mac Darwin x64 19.6.0

PR #5951, opened 2020-12-31
Potential fix for non latin chars issue

### LINKED ISSUE #5900: When Korean keyboard layout (Hangul) is active, Entering a space does not work properly.
When Korean keyboard layout (Hangul) is active
Entering a space does not work properly.

If I enter a space twice, it works as one space.

When the extension downgraded to version 17.x, works properly.

Please help!

**To Reproduce**

1. Set keyboard layout to Korean.
2. Enter a word, enter a space once, and then enter the word again.
3. You can see that there is no space.

berknam:
This issue was introduced with #5271 while fixing another issue with chinese input. The problem is that chinese input creates a popup box that stays open until the user presses `<Esc>` (to cancel), `<space>` (to accept the first suggestion, it doesn't insert any space), `<Enter>`  (to input the actual latin pressed characters instead of chinese characters) or the number of the pretended suggestion (to insert that suggestion). The `CompositionEnd` event is triggered when the popup box closes, so we are storing the composition state until the `CompositionEnd` and then handle the resulting keys one by one.

The problem is that with korean input, there is no popux box, there is a different cursor that shows the changed/replaced characters with each key press and the `CompositionEnd` event is triggered either by `<Esc>` (to cancel), `<space>` (to accept the current character and insert a space), `<Enter>`  (to accept the current character and insert a new line) or when pressing a character key that can't change the existing one (that accepts the current character, ends composition and then starts composition again with the new character).

The `<space>` and `<enter>` is already fixed on master, the problem is when you press a character that can't change the current one so it accepts the current one and starts a new composition. This happens on vscode side straight away and that means that it starts the new composition state way before we have the chance to handle and insert the accepted character, so the new composition will always start where we will insert the character and after we insert the character it doesn't update its position so it stays on top of our character. (this was mentioned on this comment https://github.com/VSCodeVim/Vim/issues/1287#issuecomment-744963840)

To fix this, I'm inserting the first character after 'CompositionStart' with `default:type` and letting the 'ReplacePreviousChar' event work as default while also keeping track of the `composingText`. And when we get 'CompositionEnd' I call `default:replacePreviousChar` to replace the previous char with '' (empty) to effectively remove the previous char and the handle the `composingText` key by key.

It still bugs me a bit, because my fix is basically doing the exact same thing as vscode does (calling the defaults on the `CompositionStart` and `type` (after composition started) events) in order to keep track of the inserted text and then remove that text and handle it through Vim which will end up inserting it again if it is not remapped. This works well and doesn't reintroduce the issue that #5271 fixed. But this introduces a visual cursor flicker that is barely noticeable if the extensionHost isn't busy but can be noticeable if it is busy.

**Current master:**
koreanInputCurrent

**With fix:**
koreanInput

*Notice that on current master the next key composition starts on top of the one that was just inserted but we can't see it until we introduce another one or a space, that might lead to confusion*

I couldn't find another way to fix this, because we need to let vscode insert the character so that the composition cursor moves correctly but we also need to handle the resulting keys in order to check for remaps, or in the case of chinese for actions, since with chinese pressing `<Enter>` results on the initially pressed latin characters.

fixes #5900 (the issue with `<space>` was already fixed, however the other mentioned issue on the comments wasn't)

@J-Fields Do you have any idea for a way that we can create tests for this sort of issues with different keyboard layouts?


jinkyuhan:
Korean want it

J-Fields:
> @J-Fields Do you have any idea for a way that we can create tests for this sort of issues with different keyboard layouts?

Nope, this is far outside my wheelhouse.

If we get a few people to test this solution out and verify it's an improvement, I'm happy to merge.

tbm98:
Hi. any update for this PR ?

J-Fields:
Gonna merge because this seems to improve matters, and we'll see if someone complains. Thanks @berknam!

PR #5970, opened 2021-01-07
Add Erlang style comments

maxnordlund:
Allows formatting of comments in Erlang, and any other language that uses `%` as comment marker.

Not sure how to handle [EDoc](https://erlang.org/doc/apps/edoc/chapter.html) style comments. Right now I've added them as a block comment, even tough they are just a number of single line ones. The difference is that doc comments are two percent, `%%`, while normal ones are just one.


J-Fields:
> Not sure how to handle EDoc style comments. Right now I've added them as a block comment, even tough they are just a number of single line ones. The difference is that doc comments are two percent, %%, while normal ones are just one.

I think these should be single line - try that and let me know if anything doesn't work as expected

J-Fields:
Does it work without including `%%`? I'd expect doc comments to be reflowed as regular comments as it stands now

J-Fields:
Gonna merge this in since it improves matters. LMK if you think we should add something to support doc comments

maxnordlund:
Great thanks, and sorry for not responding. Let let you know/make a PR if
needed.

Den tors 11 mars 2021 02:30Jason Fields ***@***.***> skrev:

PR #6004, opened 2021-01-16
Multicursor: Support skip and undo last selection

### LINKED ISSUE #2385: How do I skip word entries found by gb command? Skip option is needed.

**What happened**: There's no ability to skip word entries when using a multi-cursor

**What did you expect to happen**: When hitting gb there must me some key binding to skip the selection of the word entry.

Open any text file containing multiple entries of the same words. Try to edit only chosen part of them, having an option to skip some entries.

**Environment**:
- Extension (VsCodeVim) version: v0.10.13
- VSCode version: 1.20.1
- OS version: 10.0.16299.0

### LINKED ISSUE #5975: Multi-cursor: How to undo the last selection <C-d>? Similar to default Ctrl + U.
Couldn't find it anywhere on google, most the places only teaches the <C-d>, but never how to undo the last multi cursor selection. Is there a way to achieve this?

sql-koala:
This supports two commands that are useful when working with MC, one already expected by at least some users.
Luckily, there are existing vscode commands, we just need to call them.

fixes #2385
fixes #5975 

Question is: what keys to assign to. I chose gs to skip and a "user defined" option for undo (because C-u is already used in vim). This requires users to add a mapping (this is possible with berknams new remapper):
`vnoremap <c-u> <cursorundo>`

As an alternativ we could restrict this to normal mode:
```
class CommandMoveHalfPageUp extends CommandScrollAndMoveCursor {
  keys = ['<C-u>'];
```
and assign C-u in visual to the new cursorUndo. I prefer that.


rbutera:
This PR looks ready to merge from my end

J-Fields:
Is there a reason these can't be mapped manually? I'm a bit hesitant to provide mappings for VSCode commands that don't correspond to mappings in Vim unless there's a compelling reason to do so.

sql-koala:
@J-Fields the implementation of "gb"/Cmd-D has these line around the vscode command, so I thought it is needed.
I did a short test with just the mappings, that also worked. We can close the PR if mappings alone work fine.

```
    vimState.cursors = getCursorsAfterSync();

    // If this is the first cursor, select 1 character less
    // so that only the word is selected, no extra character
    vimState.cursors = vimState.cursors.map((x) => x.withNewStop(x.stop.getLeft()));

util.ts:
/**
 * We used to have an issue where we would do something like execute a VSCode
 * command, and would encounter race conditions because the cursor positions
 * wouldn't yet be updated. So we waited for a selection change event, but
 * this doesn't seem to be necessary any more.
 */
export function getCursorsAfterSync(): Range[] {
```

Acumane:
Has this been abandoned?

PR #6032, opened 2021-01-25
Remove stray console.log statements

edemaine:
Thanks for reviewing and merging #5892!  I was just getting #5895 ready for review, and I noticed some leftover debugging statements made it into #5892. Very sorry about this!


J-Fields:
No trouble at all, I missed that as well 🙂

PR #6033, opened 2021-01-25
Fix 'cw' behavior on whitespace

edemaine:
When `changeWordIncludesWhitespace` is `false`, 'cw' is supposed to act like 'ce' when the current character is a non-blank.  However, it is supposed to act like a regular 'w' motion when activated on a blank character.  This fixes a small typo for correctly detecting this case.  I also added a test to cover this case.

(haven't opened one, can do so if you wish)

* I'm surprised TypeScript allowed this bug.  I realized via `console.log` that `line[position.character]` was `undefined`.

### DIFF
--- a/src/actions/motion.ts
+++ b/src/actions/motion.ts
@@ -1287,7 +1287,7 @@ export class MoveWordBegin extends BaseMovement {
         return position;
       }
 
-      const char = line[position.character];
+      const char = line.text[position.character];
 
       /*
       From the Vim manual:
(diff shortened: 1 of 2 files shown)


J-Fields:
> I'm surprised TypeScript allowed this bug

I was too. Turns out we need to enable a flag in `tsconfig` to get this stricter behavior - I filed #6034 to do this. Thanks for bringing this to my attention!

edemaine:
Cool, thanks for figuring that out, and for the quick review!

Issue #6177, opened 2021-02-20
multiple circular dependencies between TypeScript files in the VSCodeVim code base


inspirer:
I tried to build VSCodeVim (for web) using a JS bundler that [does not allow](https://github.com/theseanl/tscc#rules-imposed-by-closure-compiler-and-tsickle) circular references between TypeScript files, and it turned out that there are plenty of cycles in the code base. A few of them require a certain initialization order, which make the extension rely on the current order produced by webpack.

The most prominent cycles:
logger <-> configuration (via vscodeContext)
vimState/globalState  <-> implementations of commands
RecordedState <-> Transformer (via execute)

I managed to resolve all of them by introducing a few interfaces and splitting up a couple of files, and I have a working prototype in a local fork. I'm now wondering if I should try to clean this all up and send out as PRs. Would you be interested in reviewing a ~dozen of cleanup commits that resolve the dependency cycles? If yes, what would be the best way to proceed? IMO most changes are not too controversial, while some have multiple solutions and might need to be discussed first.

**To Reproduce**

1.  $ yarn add circular-dependency-plugin -D
2.  add the basic snippet from https://github.com/aackerman/circular-dependency-plugin to webpack.config.js
3.  $ npx gulp build
4.  See multiple error messages in the form of:
```
Circular dependency detected:
src/mode/modeHandler.ts -> src/mode/modeHandlerMap.ts -> src/mode/modeHandler.ts
```

J-Fields:
Thanks for bringing this up, @inspirer. I was under the impression that ES6 modules can handle circular dependencies just fine - is that wrong? In any case, feel free to open a PR and I'll take a look.

inspirer:
Circular dependencies are fine as long as modules just declare things and don't attempt to use each other during initialization. The problem is that sometimes innocent things that look rather declarative require a certain initialization order (such as extending a class). See, for example, the third example in [this article](https://medium.com/content-uneditable/circular-dependencies-in-javascript-a-k-a-coding-is-not-a-rock-paper-scissors-game-9c2a9eccd4bc).

Let me put together some PRs and see where this brings us.

inspirer:
[...]

PR #6208, opened 2021-02-27
Fix several dependency cycles

inspirer:
This PR starts cleaning up dependency cycles, which make code sensitive to the module initialization order.

#6177 (partially)

A very general plan is to make central types, such as vimState, globalState, recordedState and configuration to not depend on actions/commands and their parsers. Instead they will deal with commands/motions via interfaces.


J-Fields:
Looks reasonable to me. Thanks!

PR #6243, opened 2021-03-07
Fix even more dependency cycles

inspirer:
This PR continues the import cycles cleanup started in #6208.

#6177 (mostly)

We do several things here:
1. Sort out the direction of dependencies between recordState, vimState, and EasyMotion.
2. Fix loggers reading from the default configuration before it has loaded (was not a problem in practice but relatively hard to reason about).
3. Split Transformer into two parts - a smaller interface for all kinds of actions and a heavier one for modeHandler.

I'll need a 3rd PR to finish this work, which will either name some commands or introduce categories of commands (to avoid using instanceof in a few places that are relatively central and cannot have a reverse dependency on commands).


J-Fields (approved):
Thanks!

PR #6364, opened 2021-03-28
More consistent behavior of C-d

artagnon:
**What this PR does / why we need it**: <kbd>C-d</kbd> behavior is inconsistent with vim.use.

### DIFF
--- a/package.json
+++ b/package.json
@@ -177,7 +177,7 @@
       {
         "key": "ctrl+d",
         "command": "extension.vim_ctrl+d",
-        "when": "editorTextFocus && vim.active && !inDebugRepl"
+        "when": "editorTextFocus && vim.active && vim.use<C-d> && !inDebugRepl"
       },
       {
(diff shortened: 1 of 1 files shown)


stevenguh:
Why are we removing `ctrl+d` for `listFocus && !inputFocus`?

artagnon:
... because it overrides the default <kbd>C-d</kbd> on macOS which does a delete-character-forward? As far as I can tell, the setting I'm removing makes <kbd>C-d</kbd> jump to the end of the file, which is very annoying.

stevenguh:
> As far as I can tell, the setting I'm removing makes C-d jump to the end of the file, which is very annoying.

The settings you removed only works on a list and non input, I am failed to see how removing that can makes `C-d` not jump to the end of file.

artagnon:
Oops, sorry. Reverted.

J-Fields (approved):
Thanks!

PR #6373, opened 2021-03-29
Consistent keybindings for <C-{n,p,f,b}> and arrows (fixes #6363)

### LINKED ISSUE #6363: Visual mode + <C-n>/<C-p>/<C-f>/<C-b> doesn't work as expected
In visual mode, hitting either of the four default macOS keys for moving the cursor creates a secondary cursor, without marking lines/blocks visual. When exiting from Visual mode, the secondary cursor disappears, and we're back to the original position.

**To Reproduce**

I have `vim.useCtrlKeys` set to `false`.

**Expected behavior**

It should work exactly like the arrow keys.

- Extension (VsCodeVim) version: 1.19.2
- VSCode version: 1.55.0-insider
- OS: macOS Big Sur

artagnon:


artagnon:
None of the keybindings seem to work for some reason.

PR #6393, opened 2021-04-01
Add Python specific motions (fixes #6213)

### LINKED ISSUE #6213: Python Object Motions
The extension does not support Python Object Motions i.e. using `[[`, `[m`, etc. does not work with Python's indentation based specification of function, class, and method bodies. This is functionality that is included in Vim 8.0 by default.

Add native support for Python Object Motions so that `[[`, `[m`, etc. work in Python files. A good reference is the [vim-pythonsense](https://github.com/jeetsukumaran/vim-pythonsense) native vim plugin which goes further than Vim 8.0 and has semantic motions.

N/A

abid-mujtaba:
- Map ([|])(m|M) to motions to the prev|next start|end of Python
functions.
- Map ([|])([|]) to motions to the prev|next end|start of Python
classes.
- Add tests for these motions.

Adds semantic Python object motions (move to function/class start/end) which are supported by native Vim 8.0.

#6213 

Since motions already exist for `[|]+[|]` I had to add a conditional to switch when the `languageId` says Python. If there is a better mechanism to add/modify motions based on language it should be straight-forward to incorporate in this PR.


abid-mujtaba:
@J-Fields, first time contributor here. What steps do I need to take to get this PR merged?

Thanks.

src/actions/languages/python/motion.ts
```diff
@@ -0,0 +1,256 @@
...
+import { BaseMovement } from '../../baseMotion';
+import { Position, TextDocument } from 'vscode';
+
+interface LineInfo {
+  line: number;
+  indentation: number;
+  text: string;
+}
+
+interface MinimalPosition {
```
J-Fields, line 12: This seems unnecessary - why not just use `vscode.Position`?

src/actions/languages/python/motion.ts
```diff
@@ -0,0 +1,256 @@
...
+    }
+
+    return structure;
+  }
+
+  /*
+   * Filter function that returns true if an element is strictly ahead (false if equal)
+   * of the specified (first arg) position.
+   */
+  static isAhead(position: MinimalPosition, elementPosition: MinimalPosition) {
```
J-Fields, line 130: `vscode.Position.isAfter` (and `vscode.Position.isBefore` instead of `isBehind`)

src/actions/languages/python/motion.ts
```diff
@@ -0,0 +1,256 @@
...
+
+        structure.push({
+          type,
+          start: {
+            line: info.line,
+            character: info.indentation,
+          },
+          end: {
+            line: endLine.line,
+            character: endLine.text.search(/(?<=\S)\s*$/) - 1, // Calculate position of last non-white character
```
J-Fields, line 117: I believe inline regexes are parsed each time they're used - consider making this a private static property to avoid incurring that cost for each line

src/actions/languages/python/motion.ts
```diff
@@ -0,0 +1,256 @@
...
+  }
+}
+
+type Type = 'function' | 'class';
+type Edge = 'start' | 'end';
+type Direction = 'next' | 'prev';
+
+// Uses the specified findFunction to execute the motion coupled to the shortcut (keys)
+abstract class BasePythonMovement extends BaseMovement {
+  type: Type;
```
J-Fields, line 207: Make these properties abstract

src/actions/languages/python/motion.ts
```diff
@@ -0,0 +1,256 @@
...
+      default:
+        return position;
+    }
+  }
+}
+
+@RegisterAction
+class MovePythonNextFunctionStart extends BasePythonMovement {
+  keys = [']', 'm'];
+  type = 'function' as Type;
```
J-Fields, line 229: Are these type assertions necessary?

src/actions/languages/python/motion.ts
```diff
@@ -0,0 +1,256 @@
...
+      // otherwise return the last element
+      const index = direction === 'next' ? 0 : elements.length - 1;
+      const element = elements[index];
+      const pos = element[edge];
+
+      // execAction MUST return a fully realized Position object created using new
+      return new Position(pos.line, pos.character);
+    }
+
+    return null;
```
J-Fields, line 184: Use `undefined` instead of `null`

J-Fields (changes requested):
I'll need to do another read, but it LGTM other than the handful of comments I left. Thanks a lot for the MR!

src/actions/languages/python/motion.ts
J-Fields: This should fix the build error - I recently enabled `strictPropertyInitialization`.

abid-mujtaba: Done.

src/actions/languages/python/motion.ts
abid-mujtaba: Good catch. Removed.

src/actions/languages/python/motion.ts
abid-mujtaba: The logic at the call-site uses `null` to denote an alternate (failed) return state. I considered `undefined` but I thought `null` represents this idea better since `undefined` is more implicit.

I am open to change this.

src/actions/languages/python/motion.ts
abid-mujtaba: Build fails without this.

src/actions/languages/python/motion.ts
J-Fields: Yeah, since the type system in typescript makes the usual use of `undefined` in javascript (checking whether a property exists on an object) useless, they're generally equivalent.
For that reason, I like to keep it simple and pretend `null` doesn't exist (the typescript team [does the same](https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines#null-and-undefined), though it isn't an official recommendation, just their preference).

src/actions/languages/python/motion.ts
J-Fields: Ah - making the type explicit while not asserting it should work as well, without sacrificing type safety:

`type: Type = 'function';`

src/actions/languages/python/motion.ts
abid-mujtaba: Don't know how I missed these. Replaced.

src/actions/languages/python/motion.ts
abid-mujtaba: Good call. Done.

abid-mujtaba:
@J-Fields please re-review.

src/actions/languages/python/motion.ts
abid-mujtaba: Do you know why the type has to be made explicit even though it has been defined in the parent class?

src/actions/languages/python/motion.ts
abid-mujtaba: Done

src/actions/languages/python/motion.ts
abid-mujtaba: Done

J-Fields: No I'm not exactly sure why it's necessary, I've just run into similar things before

src/actions/base.ts
```diff
@@ -83,7 +83,9 @@ export abstract class BaseAction {
     }
 
     const keys2D = BaseAction.is2DArray(this.keys) ? this.keys : [this.keys];
-    const keysSlice = keys2D.map((x) => x.slice(0, keysPressed.length));
+    const keysSlice = keys2D.map((x: string[] | readonly string[]) =>
```
J-Fields, line 86: Why is this necessary?

src/actions/languages/python/motion.ts
```diff
@@ -0,0 +1,232 @@
...
+      yield document.lineAt(index).text;
+    }
+  }
+
+  /*
+   * Calculate the indentation of a line of text.
+   * Lines consisting entirely of whitespace of "starting" with a comment are defined
+   * to have an indentation of "null".
+   */
+  static _indentation(line: string): number | null {
```
J-Fields, line 54: `undefined` -> `null` here and in `_parseLine`. Sorry to nitpick about this 🙂

src/actions/languages/python/motion.ts
J-Fields: This should be `type: Type = 'function';`, as it's more type-safe

src/actions/base.ts
abid-mujtaba: The build was failing (typescript compilation stage) without this change.

src/actions/languages/python/motion.ts
J-Fields: Oh sorry, is this fixed? GitHub's UI is confusing me. The same improvement should be made to `direction` and `edge`, if it's not already

abid-mujtaba: Done

src/actions/languages/python/motion.ts
abid-mujtaba: Done.

abid-mujtaba:
@J-Fields please re-review.

src/actions/base.ts
J-Fields: Weird. I just checked and it built fine, so reverting this line to keep the change contained

src/actions/languages/python/motion.ts
```diff
@@ -0,0 +1,232 @@
...
+  static _parseStructure(lines: LineInfo[]): StructureElement[] {
+    const last = lines.length;
+    const structure: StructureElement[] = [];
+
+    for (let index = 0; index < last; index++) {
+      const info = lines[index];
+      const text = info.text;
+
+      if (text.match(PythonDocument.reDefOrClass)) {
+        const type = text.match(PythonDocument.reDef) ? 'function' : 'class';
```
J-Fields, line 92: I think this distinction can be made by looking at the match group instead of running another regex, but probably not a big deal

src/actions/languages/python/motion.ts
```diff
@@ -0,0 +1,232 @@
...
+    // Return the first match if any exist
+    if (elements.length) {
+      // If direction === 'next' return the first element
+      // otherwise return the last element
+      const index = direction === 'next' ? 0 : elements.length - 1;
+      const element = elements[index];
+      const pos = element[edge];
+
+      // execAction MUST return a fully realized Position object created using new
+      return new Position(pos.line, pos.character);
```
J-Fields, line 157: I think `return pos;` is fine, actually - `Position`s are immutable

src/actions/languages/python/motion.ts
```diff
@@ -0,0 +1,232 @@
...
+  abstract direction: Direction;
+  abstract edge: Edge;
+
+  public async execAction(position: Position, vimState: VimState): Promise<Position> {
+    const document = vimState.document;
+    switch (document.languageId) {
+      case 'python':
+        return (
+          new PythonDocument(document).find(this.type, this.direction, this.edge, position) ||
+          position
```
J-Fields, line 193: I think `|| position` should be replaced by `?? failedMovement(vimState)`. This triggers macros to stop and avoids doing anything if paired with an operator

src/actions/languages/python/motion.ts
```diff
@@ -0,0 +1,232 @@
...
+  static moveClassBoundary(
+    document: TextDocument,
+    position: Position,
+    forward: boolean,
+    start: boolean
+  ): Position {
+    const direction = forward ? 'next' : 'prev';
+    const edge = start ? 'start' : 'end';
+
+    return new PythonDocument(document).find('class', direction, edge, position) || position;
```
J-Fields, line 173: `?? failedMovement(vimState)` again, I think

src/actions/languages/python/motion.ts
```diff
@@ -0,0 +1,232 @@
...
+    const document = vimState.document;
+    switch (document.languageId) {
+      case 'python':
+        return (
+          new PythonDocument(document).find(this.type, this.direction, this.edge, position) ||
+          position
+        );
+
+      default:
+        return position;
```
J-Fields, line 197: If `document.languageId !== 'python'`, that's definitely a bug, right - since the caller is responsible for checking that? If that's the case, I'd rather throw an exception

src/actions/languages/python/motion.ts
abid-mujtaba: Note to self:

``` javascript
    const match = text.match(PythonDocument.reDefOrClass);
    if (match) {
        const type = match[1];
        ...
```

and delete the definition of `PythonDocument.reDef`.

src/actions/languages/python/motion.ts
abid-mujtaba: That is the part I am most uncertain about. Currently I am using [this code](https://github.com/abid-mujtaba/Vim/blob/b0f2b4d866a69dfa3134d92f7573ce95602322cd/src/actions/languages/python/motion.ts#L202-L208) to connect key-strokes to actions. The registration is not specifying a language so a user could, for example, press `]m` in a file of any language (not just Python).

What is the proper way to ensure that these key-stroke actions are only valid in a Python file?

J-Fields: Oh, sorry, you're right - it's just `moveClassBoundary` that's called from outside. The right way to do this is to implement `doesActionApply`

src/actions/languages/python/motion.ts
abid-mujtaba: Done

src/actions/languages/python/motion.ts
abid-mujtaba: Done

src/actions/languages/python/motion.ts
abid-mujtaba: Done

src/actions/languages/python/motion.ts
abid-mujtaba: Done

src/actions/languages/python/motion.ts
abid-mujtaba: Done

abid-mujtaba:
@J-Fields please re-review.

J-Fields (approved):
Seems reasonable to me. Thanks again for the PR and for quickly responding to feedback!

abid-mujtaba:
Thanks @J-Fields, the feedback was great, learned a lot.

max-sixty:
Without wanting to pollute notifications, quick note to add this is very helpful, thank you @abid-mujtaba !

PR #6394, opened 2021-04-01
Feature/vimrc source support

### LINKED ISSUE #5542: Support `source`
And `:source`, obviously. Some people keep keybindings separate from their main .vimrc, and it'd be nice to accommodate this.

Demiguise:

Adds support for the `source` keyword in `.vimrc`, loading the referenced files and parsing their configurations.

Fixes #5542

Not sure whether the loadConfig should live in the vimrc file or somewhere else.
The same 33 tests fail in this branch as they do on master.

### DIFF
--- a/src/configuration/vimrc.ts
+++ b/src/configuration/vimrc.ts
@@ -39,12 +39,15 @@ export class VimrcImpl {
   private static async loadConfig(configPath: string, config: IConfiguration) {
     // Add the new remappings
     try {
-      console.log(`Reading VimRC file from "${configPath}"`);
       const vscodeCommands = await vscode.commands.getCommands();
       const lines = (await fs.readFileAsync(configPath, 'utf8')).split(/\r?\n/);
       for (const line of lines) {
         const source = this.buildSource(line);
         if (source) {
+          if (!(await fs.existsAsync(source))) {
+            console.warn(`Unable to find "${source}" file for configuration`);
+            continue;
+          }
           VimrcImpl.loadConfig(source, config);
           continue;
         }
--- a/src/configuration/vimrc.ts
+++ b/src/configuration/vimrc.ts
@@ -7,9 +7,11 @@ import { IConfiguration, IVimrcKeyRemapping } from './iconfiguration';
 import { vimrcKeyRemappingBuilder } from './vimrcKeyRemappingBuilder';
 import { window } from 'vscode';
 import { configuration } from './configuration';
+import { Logger } from '../util/logger';
 
 export class VimrcImpl {
   private _vimrcPath: string;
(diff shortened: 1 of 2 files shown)


src/configuration/vimrc.ts
```diff
@@ -18,6 +18,60 @@ export class VimrcImpl {
...
+      const vscodeCommands = await vscode.commands.getCommands();
+      const lines = (await fs.readFileAsync(configPath, 'utf8')).split(/\r?\n/);
+      for (const line of lines) {
+        const source = this.buildSource(line);
+        if (source) {
+          if (!(await fs.existsAsync(source))) {
+            console.warn(`Unable to find "${source}" file for configuration.`);
+            continue;
+          }
+          console.log(`Loading "${source}" file for configuration.`);
```
J-Fields, line 50: Create a `Logger` for this class and use `this.logger.debug` here (and `this.logger.warn` above)

src/configuration/vimrc.ts
```diff
@@ -18,6 +18,60 @@ export class VimrcImpl {
     return this._vimrcPath;
   }
 
+  private static readonly SOURCE_REG_REX = /^(source)\s+(.+)/i;
+
+  private static buildSource(line: string) {
+    if (line.trimLeft().startsWith('"')) {
```
J-Fields, line 24: Instead of having this same check in every `build*` function, let's just put it at the start of the loop over lines in `loadConfig`

src/configuration/vimrc.ts
```diff
@@ -18,6 +18,60 @@ export class VimrcImpl {
...
+      const lines = (await fs.readFileAsync(configPath, 'utf8')).split(/\r?\n/);
+      for (const line of lines) {
+        const source = this.buildSource(line);
+        if (source) {
+          if (!(await fs.existsAsync(source))) {
+            console.warn(`Unable to find "${source}" file for configuration.`);
+            continue;
+          }
+          console.log(`Loading "${source}" file for configuration.`);
+          VimrcImpl.loadConfig(config, source);
```
J-Fields, line 51: Is it a problem that we don't `await` this? I think if two sourced .vimrcs have overlapping remaps, it would make the one that is effective non-deterministic, though probably that's not a big deal.

J-Fields (changes requested):
Thanks! Looks good to me aside from a the few comments I made

src/configuration/vimrc.ts
Demiguise: I think `await`ing on this is probably a good idea, don't want it to be non-deterministic!

Demiguise:
Fixed an extra issue in there with `<2` :)

J-Fields (approved):
Thanks a lot!

Issue #5663, opened 2020-11-16
Use `vimState.editor` whenever possible


J-Fields:
`vscode.window.activeTextEditor` can change when we're not looking, which I think is the cause of annoying issues such as the `getLineLength` out of bounds bug. By using `vimState.editor` in its place, we eliminate a lot of race conditions.

sql-koala:
I just found that substitute.ts also uses alot of deprecated methods of Texteditor. I can work on that - if you will merge it :>

sql-koala:
Ok, we _can_ use
```
      vimState.recordedState.transformer.addTransformation({
        type: 'replaceText',
```

But we would need to convert ~22 old tests:
```
  test('Replace with `g` flag', async () => {
    await modeHandler.handleMultipleKeyEvents(['i', 'a', 'b', 'a', '<Esc>']);
    await commandLine.Run('%s/a/d/g', modeHandler.vimState);
    assertEqualLines(['dbd']);
  });
```
to the newTest-method...

sql-koala:
@J-Fields are you interested in using
```
      vimState.recordedState.transformer.addTransformation({
        type: 'replaceText',
```
for replace?
It requires some grunt work (adapting tests, which i can do) but I would need help with 1-2 test for interactive `:s`

J-Fields:
> for replace?

Are you referring to `TextEditor.replace`? Or `:s` specifically?

In general, yes - at this point I'm slowly working toward all document changes being done through transformations (at which point the uniform interface will make further refactoring and optimization easier). The issue, IIRC, is that transformations currently sync the cursor positions from `editor.selections` (overwriting any change you tried to do via `vimState.cursorStopPosition = ...`), which sucks as a default behavior. You can generally get around this by using putting an `ExactCharacter` `PositionDiff` on the transformation, but this is also janky. I've been thinking about making a `syncCursors` transformation which would do that, but haven't gotten around to trying out that sort of refactor yet.

*Anyway*, would love to have some help on this stuff - let me know what you've got in mind (or if you've got a WIP branch, feel free to open a PR for review). I'm more likely to accept small, targeted improvements as I've been thinking about all this stuff for quite a while and need to do some experimentation before settling on how this'll all work.

PR #6402, opened 2021-04-02
Substitute, use transformations

### LINKED ISSUE #4915: Substitution doesn't operate on whole block if newlines are inserted
**Describe the bug**
If I use `s/.../... \n .../` on a highlighted blocks, only half of the lines are changed. (suspicion - it iterates on original line range, without accounting for additions)

**To Reproduce**

1. Create a file with:
```
a: b
c: d
e: f
g: h
```
2. Highlight all lines (`ggVG`)
3. Run replacement (`:s/: /\n- /`)

**Expected behavior**
```
a
- b
c
- d
e
- f
g
- h
```
instead of current:
```
a
- b
c
- d
e: f
g: h
```

- Extension (VsCodeVim) version: 1.14.5
- VSCode version: 1.45.1
- OS: macos mojave

sql-koala:
this refactors substitute `:s` and replaces deprecated methods to access the document.

fixes #4915
relates to #5663 

the diff for tests is quiet large and ugly, because many tests need to be adapted. I provide the commit for tests first, with a CI run,
https://github.com/VSCodeVim/Vim/runs/2254892205 
 so you can see that only a few tests are changed a bit. For some tests, we need to adapt expected cursor position to be the BOL, which is also what gvim does.


sql-koala:
@J-Fields can you look at the one failing test?  `1) : (command) register can be used as a macro` 
I have no clue why this does not work or how this could be connected to the change.

sql-koala:
well, gnaarf :| ... two things
1. when using interactive / confirm `:s` with a transformation, the document is not updated right away, only after last confirmation.
2. executing the : register `@:` does not work after using transformation

what to do?

J-Fields:
Thanks a bunch @sql-koala! I just pushed a change which should fix that test - the issue was that macros are replayed via a `'macro'` transformation, and transformations created in that process weren't responded to. My fix seems like a band-aid on larger architectural issues, but it should do the trick for now.

J-Fields:
> when using interactive / confirm `:s` with a transformation, the document is not updated right away, only after last confirmation.

I don't care enough about this to prevent merging - I've created #6427 to track it

sql-koala:
yay, merged. thanks for the help @J-Fields

PR #6409, opened 2021-04-03
Bugfix substitute with backslash(fixes #2710)

### LINKED ISSUE #2710: Substitute backslash inconsistency

**What happened**:
When using substitute, the backslash character is not handled correctly. Typing in `\\` will print out "\\" instead of only one "\"

This wasn't really a problem until I ran into a case where I have to separate in "\right". If I use `:s/[find pattern]/\right/gc` I end up with the return command being used. The output is therefore "ight" on the next line.

If I use `:s/[find pattern]/\\right/gc`, it will simply output "\\right" which isn't what I wanted either.

`:s/[find pattern]/\\right/gc` to replace "\right" into the corresponding matches.
 
Note: the `:` is simply there for effect. I'm not entering `:` into the command box that pops up.

**Environment**:

- VSCode version: 1.23.1
- OS version: Windows 10

sekai013:
substitute string to backslashes properly without neovim integration.

#2710 

This is my first PR to OSS. I'm going to follow the contributing guide, if there is any inconvenience, please tell me what to do. thx!

### DIFF
--- a/src/cmd_line/subparsers/substitute.ts
+++ b/src/cmd_line/subparsers/substitute.ts
@@ -32,6 +32,9 @@ function parsePattern(pattern: string, scanner: Scanner, delimiter: string): [st
             case 't':
               pattern += '\t';
               break;
+            case '\\':
+              pattern += '\\';
+              break;
             default:
               pattern += '\\';
               pattern += currentChar;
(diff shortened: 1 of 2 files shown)


J-Fields:
Thanks!

PR #6433, opened 2021-04-06
replace with register, gr use transformer

### LINKED ISSUE #5115: Replace with multiple cursors on a line replaces in the wrong location
**Describe the bug**

**To Reproduce**

1.  Copy a word to the clipboard `yaw`
2.  Create multiple cursors on the same line ~`Ctrl-d`~ `Cmd-d`
3.  Run replace `griw`
4.  See the word under the second cursor isn't replaced correctly

**Expected behavior**
Replace under both cursors as though they were run independently

**Screenshots**
This follows the exact keystrokes above: https://youtu.be/569BSfF-3zY
(I got stuck on QuickTime / formats / uploading to GH, hope this is OK)

- Extension (VsCodeVim) version: 1.16.0
- VSCode version: 1.47.3
- OS: MacOS

**Additional context**

sql-koala:
- This brings logic to "gr", that we recently did elsewhere, handling the case of yanking + replacing with multi cursor correctly (like #6065 I think)
- uses transformations, therefore handles multiple cursors on the same line

fixes #5115
relates #5663 

I took the liberty to remove (not adapt) the code that deals with "post cursor placement". Leaving the cursor as is, is - in my judgement - as good (I think of gr as something like gU, g~). If there are workflows that need the cursor at end of replace text, i can look into it.
I have dogfooded this for a few months...


sql-koala:
Hmm, that one test passes locally here. Also, the result is correct if i create a buffer and type the commands manually.

test/plugins/replaceWithRegister.test.ts
```diff
@@ -18,91 +18,84 @@ suite('replaceWithRegister plugin', () => {
...
+    end: ['first |first'],
   });
 
   newTest({
     title: 'Replaces within {} over multiple lines',
     start: ['{', '  first', '  s|econd', '  third', '}'],
     keysPressed: `${YankInnerWord}${ReplaceOperator}i}`,
-    end: ['{', '|second', '}'],
-  });
-
```
sql-koala, line 79: this removes a redundant test.

test/plugins/replaceWithRegister.test.ts
```diff
@@ -18,91 +18,84 @@ suite('replaceWithRegister plugin', () => {
...
-  newTest({
-    title: 'Replaces a multiline register within {} over multiple lines',
-    start: ['{', '  first', '  s|econd', '  third', '}'],
-    keysPressed: `yj${ReplaceOperator}i}`,
-    end: ['{', '  |second', '  third', '}'],
```
sql-koala, line 84: this removes a redundant test.

test/plugins/replaceWithRegister.test.ts
```diff
@@ -18,91 +18,84 @@ suite('replaceWithRegister plugin', () => {
...
   });
 
   newTest({
     title: 'Yanking inside {} then replacing inside {} in a noop, besides the cursor movement',
-    start: ['{', '  first', '  s|econd', '  third', '}'],
+    start: ['{', '  first', '  |second', '  third', '}'],
     keysPressed: `yi}${ReplaceOperator}i}`,
-    end: ['{', '  |first', '  second', '  third', '}'],
+    end: ['{', '  first', '  |second', '  third', '}'],
   });
```
sql-koala, line 92: Hmm, that one test passes locally here. Also, the result is correct if i create a buffer and type the commands manually.

test/plugins/replaceWithRegister.test.ts
```diff
@@ -18,91 +18,84 @@ suite('replaceWithRegister plugin', () => {
...
   newTest({
     title: 'Replaces a multiline register within {} over multiple lines',
     start: ['{', '  first', '  s|econd', '  third', '}'],
     keysPressed: `yj${ReplaceOperator}i}`,
-    end: ['{', '  |second', '  third', '}'],
+    end: ['{', '  second', '  |third', '}'],
   });
 
   newTest({
     title: 'Yanking inside {} then replacing inside {} in a noop, besides the cursor movement',
```
sql-koala, line 88: Hmm, that one test passes locally here. Also, the result is correct if i create a buffer and type the commands manually.

test/plugins/replaceWithRegister.test.ts
```diff
@@ -18,91 +18,84 @@ suite('replaceWithRegister plugin', () => {
...
-    start: ['{', '  first', '  s|econd', '  third', '}'],
+    start: ['{', '  first', '  |second', '  third', '}'],
```
sql-koala, line 89: Hmm, that one test passes locally here. Also, the result is correct if i create a buffer and type the commands manually.

src/actions/plugins/replaceWithRegister.ts
```diff
@@ -24,46 +23,25 @@ export class ReplaceOperator extends BaseOperator {
...
+        text: replaceWith,
+        range: replaceRange,
+      });
     }
-
-    const replaceWith = register.text as string;
-    await TextEditor.replace(vimState.editor, range, replaceWith);
     await vimState.setCurrentMode(Mode.Normal);
-    updateCursorPosition(vimState, range, replaceWith);
+    return;
```
J-Fields, line 45: this `return` isn't needed

J-Fields:
> I took the liberty to remove (not adapt) the code that deals with "post cursor placement". Leaving the cursor as is, is - in my judgement - as good (I think of gr as something like gU, g~). If there are workflows that need the cursor at end of replace text, i can look into it.

Yeah I'm inclined to keep it with the cursor at the end of the replace text because that's what people are used to and that's how the original Vim extension behaves. I agree it's basically arbitrary - I guess the justification is that it works basically like `v<motion>p`. Don't worry about that for now, though - setting the cursor position after a transformation will be much easier after a refactor I'm working on.

J-Fields:
I implemented this in a similar way via 8750e3754b8e2925d75b3c83a9e2634be2cfd6ef

Issue #5081, opened 2020-07-26
Cursor position after surround in visual line mode is wrong


J-Fields:
Split off from #4869

PR #6452, opened 2021-04-08
Surround Rewrite

### LINKED ISSUE #1818: Surround mode doesn't work with multiple cursors
Reported by /u/atomen on Reddit.

### LINKED ISSUE #2893: surround emulation fails with underscore "_"
The _surround_ emulation fails to remove underscores.

**To Reproduce**
1. move cursor under a word
1. type `ysiw_` to add an underscore around the word
1. type `ds_` to remove the underscore just added

**Expected behavior**
I expected the underscore to be removed, but it is not.

**Environment**

 - Extension (VsCodeVim) version: 0.15.7
 - VSCode version: 1.25.1
 - OS: w10p -1803 - 17134.81

**Additional context**
The adding and removing other characters works.

### LINKED ISSUE #3386: Surround doesn't keep history steps properly
When doing surround in multiple lines, the "undo" doesn't keep history right, it undoes multiple commands behind. 

Example: 
do `yss"` in multiple lines, then `u`

asdf

May be related to 
https://github.com/VSCodeVim/Vim/issues/3355

### LINKED ISSUE #5211: When I selected multiple lines in a file, and use vim-surround, it will show just one line
When I selected muntiple lines in a file, and use vim-surround, it will show me that I just selected the original one line, but actually it indeed selected the lines that I want to.

You could reproduce the problem in this way:
(* At first, make sure your vim-surround options are enabled.)
1. open any file
2. press "V" (shift + v) to select current line
3. press "j" or "k" to move cursor (to select more lines)
4. press "S" (shift + s) , and the problem will arise.
5. when vim-surround is finished, you will find that the brackets have been added correctly.

### LINKED ISSUE #6564: Surround emulation gets positions wrong when tabs are involved
**Describe the bug**

When you do `ysw"` on a word with a tab after it, the closing quote ends up after the tab. The `w` motion should not include spaces after the word. Thus, `cw` correctly does not delete the tab in nvim or vscodevim, and vim-surround in nvim puts the quote before the tab as expected with `ysw"`.

**To Reproduce**

```
 » echo -e 'a\tb\n' > bug.txt
 » code bug.txt
```

Put your cursor at the start of the first line in normal mode, type `ysw"`.

You'll get:

```
"a	"b
```

**Expected behavior**

```
"a"	b
```

- Extension (VsCodeVim) version: 1.20.1

VSCode info:

Version: 1.55.2
Commit: 3c4e3df9e89829dce27b7b5c24508306b151f30d
Date: 2021-04-13T09:37:02.931Z
Electron: 11.3.0
Chrome: 87.0.4280.141
Node.js: 12.18.3
V8: 8.7.220.31-electron.0
OS: Linux x64 5.11.16-arch1-1

sql-koala:
Here is something cool: a complete rewrite of the surround plugin.

Features:
- decoupled from yank operator (possible with new remapper)
- "mini state machine" on SurroundInputMode drastically reduced
- integrates much better with modehandler, so that
  - requires no special logic for repeat
  - supports multiple cursors (with no overhead)
- all exec methods fit on less than one screen; made of 3-4 easy to understand steps because
- handling of edge cases and ugly details are extracted and reused in a few methods
- there is exactly one let (vs const) in the whole code

fixes #1818
fixes #5211 
fixes #2893
fixes #6564 
should fix #3386

This is ready to merge from my side. 
...
And i had to disable the tests with tags, because this uses an inputBox now, to read the tag name.

Looking forward to your feedback @J-Fields ;-)


sql-koala:
One remark one visual surround: we have to balance/choose between #5081 vs #5211.

setting this on Visual Operators
`cursorStop = start`
reduces the selection to one line.

Therefore I prefer
`cursorStop = end`

test/plugins/surround.test.ts
```diff
@@ -59,6 +77,7 @@ suite('surround plugin', () => {
     end: ['first |( line ).test'],
   });
 
+  /*
```
J-Fields, line 80: Maybe we can stub `vscode.window.showInputBox` to make these tests possible?

src/state/recordedState.ts
```diff
@@ -184,10 +184,6 @@ export class RecordedState {
 
   public hasRunOperator = false;
 
-  public hasRunSurround = false;
```
J-Fields, line 187: Don't have time yet for a thorough review, but I love seeing jank like this being ripped out!

test/plugins/surround.test.ts
sql-koala: I have adapted the existing tests with a stub.

sql-koala:
I have added two functionality at other places.
- to make it ship-able: a way to provide default loaded mappings for plugins
- to make it testable: support stub in tests. tag tests helped, I fixed a small bug and a not covered feature.
 
This builds and passes locally.

J-Fields:
Sweet! Can you fix the merge conflicts when you get a chance?

sql-koala:
@J-Fields I had to wrestle quiet a bit with git, but that should to it.

J-Fields:
> One remark one visual surround: we have to balance/choose between #5081 vs #5211.
> 
> setting this on Visual Operators
> `cursorStop = start`
> reduces the selection to one line.
> 
> Therefore I prefer
> `cursorStop = end`

That's fine, at least for now. I think we can eventually hack it so they're both as expected, but don't worry about that for this PR.

sql-koala:
Hi @J-Fields, can we get this reviewed and merged?
I just added a few more tests in the last commits, but it is ready from my side.

Because it's a complete rewrite, even small automated changes like
https://github.com/VSCodeVim/Vim/commit/3f6ba739445d469f43324b2757096f3c15c43a5a#diff-9102ca6b9529b6f1b3372b511a9687cd3a0b3cb0229a2cbc4b4c01577cced66d
can result in weird merge conflicts ... which i than need to manage.

test/plugins/surround.test.ts
```diff
@@ -32,6 +33,20 @@ suite('surround plugin', () => {
     end: ['first |(line) test'],
   });
 
+  newTest({
```
J-Fields, line 36: `ysw` doesn't work in some cases. For instance, put the cursor at the start of `newTest` here and do `ysw"`

J-Fields: Seems like this happens if there's a newline or punctuation before whitespace

sql-koala: See https://github.com/VSCodeVim/Vim/pull/6452/commits/20629ff7fd36893454179ce27d737b278bfbd7c6 with whole function as context.
Apparently, we are expected to correct w/aw motions, which is a bit ugly.
I thought about/would prefer adding maps: ysw -> yse but that would not work with counts: ys4w.

This seems to handle all known cases correct.

src/actions/plugins/pluginDefaultMappings.ts
```diff
@@ -0,0 +1,38 @@
...
+  private static defaultMappings: Array<{
+    mode: string;
+    configSwitch: string;
+    mapping: IKeyRemapping;
+  }> = [
+    // default maps for surround
+    {
+      mode: 'normalModeKeyBindingsNonRecursive',
+      configSwitch: 'surround',
+      mapping: { before: ['y', 's'], after: ['<plugys>'] },
```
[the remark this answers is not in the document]
J-Fields: Leaving this as-is for now, planning to come back to it at some point (but will likely forget, if I'm being honest)

J-Fields:
Thanks a ton for this, @sql-koala! To be honest I just skimmed the code, as this is an area I'm very unfamiliar with, but it seems reasonable, works better than before from what I can tell, and leaks fewer implementation details into ModeHandler and such - good enough for me.

Issue #4463, opened 2020-01-02
Can not map to `python.sortImports` in vimrc


ChrisLee63:
**Describe the bug**
`nnoremap <leader>ci python.sortImports` does not work in vimrc

**Environment (please complete the following information):**

- Extension (VsCodeVim) version: latest
- VSCode version: latest
- OS: windows 10

520Chris:
@J-Fields 
I am sorry to disturb you.
Is this a bug or just by design?

J-Fields:
A bug. Does it work when you make that remap with our normal VSCode settings?

520Chris:
```
"vim.normalModeKeyBindingsNonRecursive": [
        {
            "before": [
                "<leader>",
                "d"
            ],
            "commands": [
                "python.sortImports",
            ]
        },
    ],
```
also does not work.

But it can work when I map `<leader>d` to `editor.action.formatDocument`.

fihyer:
I also facing the same problem on mapping keys with `python.sortImports` command

```json
    "vim.normalModeKeyBindingsNonRecursive": [
        {
            "before": ["<leader>", "l", "s"],
            "commands": [
                "python.sortImports",
            ]
        },
    ],
    "vim.visualModeKeyBindingsNonRecursive": [
        {
            "before": ["<leader>", "l", "m"],
            "commands": [
                "python.refactorExtractMethod",
            ]
        },
        {
            "before": ["<leader>", "l", "v"],
            "commands": [
                "python.refactorExtractVariable",
            ]
        }
    ],
```

the first mapping not work, but the following two `python.refactor*` works.

I've tried assgin keyboard shortcuts to the `python.sortImports` command in "keybindings.json", 
```json
[
    {
        "key": "ctrl+l ctrl+s",
        "command": "python.sortImports"
    }
]
```
the shortcuts works fine, then I try remapping the "before" keys to "after" keys (the shortcuts in "keybindings.json")
```json
    "vim.normalModeKeyBindingsNonRecursive": [
        {
            "before": ["<leader>", "l", "s"],
            "after": ["<C-l>", "<C-s>"]
        },
    ],
```
still not working

Any suggestions?

PR #6694, opened 2021-06-04
Implement :vsc to run VSCode commands

meluskyc:
Add a `:vsc` command to run VSCode commands, including extension commands. 

Examples: 

```
:vsc editor.action.organizeImports
:vsc gitlens.toggleFileBlame
```

As a bonus, we can now map extension commands in vimrc. 

```
noremap <leader>glf :vsc gitlens.showFileHistoryView<CR>
```

#4463, #4733, probably others. While PR #4750 is another attempt at solving these issues, adding a new command is more in-line with approaches by other Vim emulators. e.g. VSVim uses `:vsc` and Ideavim uses `:action`. 

### DIFF
--- a/src/cmd_line/commands/vscode.ts
+++ b/src/cmd_line/commands/vscode.ts
@@ -3,7 +3,7 @@ import { VimState } from '../../state/vimState';
 import { CommandBase } from '../node';
 import { Scanner } from '../scanner';
 
-export class VscCommand extends CommandBase {
+export class VsCodeCommand extends CommandBase {
   private command?: string;
 
   constructor(command?: string) {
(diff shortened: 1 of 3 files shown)


J-Fields:
Thanks for the PR! We do already handle mapping built-in commands via .vimrc with special case logic, but I agree - this seems like the correct approach.

src/cmd_line/subparser.ts
```diff
@@ -348,6 +349,11 @@ export const commandParsers = {
     parser: fileCmd.parseEditNewFileInNewVerticalWindowCommandArgs,
   },
 
+  vsc: {
```
J-Fields, line 352: I think a full name of `vscode` makes sense

meluskyc: @J-Fields Updated. Thanks!

src/cmd_line/commands/vscode.ts
```diff
@@ -0,0 +1,27 @@
...
+  }
+
+  public static parse(args: string): VsCodeCommand {
+    const scanner = new Scanner(args);
+    scanner.skipWhiteSpace();
+    return new VsCodeCommand(scanner.isAtEof ? undefined : scanner.nextWord());
+  }
+
+  private async vsc() {
+    await vscode.commands.executeCommand(this.command ?? '');
```
J-Fields, line 21: What does `executeCommand('')` do? I'd think `:vsc` with no argument should just throw `E471`

J-Fields:
Thanks again @meluskyc!

Issue #6825, opened 2021-07-04
:sort u merges two duplicates


karizma:
**Describe the bug**
I found some cases where when you use sort u, it sort of merges two duplicates

**To Reproduce**

1.  Open a file
2.  Input the following content: 
```
there
do
there
```
3.  Use `:sort u`

Output:
```
do
thereere
```
**Expected behavior**
```
do
there
```

**Environment**

- Extension (VsCodeVim) version: 1.21.4
- VSCode version: 1.57.1
- OS: Windows 10

---

Doing a bit more testing, it seems like it works with any 3 lines as long as the first and last are the same, and the middle is 'do':

```
candy
do
candy

=>

candy
dondy
```
```
hello
do
hello

=>

do
hellollo
```

teachermugabo:
This is a fascinating error. I had no idea Vim had this kind of functionality. Looking forward to seeing how this gets resolved and to using it!

karizma:
For the most part you can already use sort, it's pretty stable. Vscode itself also has the sort functionality in case you didn't know (although it's way more limited than vim).

karizma:
Not sure if you wanted to get ping, but it has been solved at #6835. @teachermugabo

teachermugabo:
Thank you!

PR #6835, opened 2021-07-06
Fix sort unique bug

karizma:
This PR Fixes a bug where when you used `:sort u` on certain text it would cause it to not replace everything.
An example:

```
there
do
there
```

Would turn into:

```
do
thereere
```

#6825

If you are interested in what caused the bug:

In the [file that handles the command](https://github.com/VSCodeVim/Vim/blob/master/src/cmd_line/commands/sort.ts), the program modifies originalLines when the unique argument is specified:

```js
    if (this.arguments.unique) {
      const seen = new Set<string>();
      const uniqueLines: string[] = [];
      for (const line of originalLines) {
        const adjustedLine = this.arguments.ignoreCase ? line.toLowerCase() : line;
        if (!seen.has(adjustedLine)) {
          seen.add(adjustedLine);
          uniqueLines.push(line);
        }
      }
      originalLines = uniqueLines;
    }
```

The problem with this is when the program specifies the range of text to replace:

```js
 vimState.recordedState.transformer.addTransformation({
      type: 'replaceText',
      // this line in specific
      range: new vscode.Range(startLine, 0, endLine, lastLineLength),
      text: sortedContent,
      diff: PositionDiff.exactPosition(
        new vscode.Position(startLine, sortedLines[0].match(/\S/)?.index ?? 0)
      ),
 });
 ```
 The first 3 parameters are correct, but the fourth one is wrong. lastLineLength is defined by:
 
 ```js
 const lastLineLength = originalLines[originalLines.length - 1].length;
 ```
 
But this happens **after** the transformation on originalLines (line 46-57).  So lastLineLength is not actually getting the original text's last line length, it's getting the transformed originalLines (uniqueLines).

Here's a quick walkthrough if the text is `there\ndo\nthere`:

```
originalLines = ['there', 'do', 'there']

unique argument is provided so we get the unique lines and set it

originalLines is now ['there', 'do']

lastLineLength = 'do'.length = 2

// => ERROR: It should be 'there'.length which is 5.
```

In this case the range's end line character is only 2, so on the third line of 'there', it only replaces the starting two letters, and leaves 'ere' alone. So we get the result of:

```
do 
thereere
```

### DIFF
--- a/src/cmd_line/commands/sort.ts
+++ b/src/cmd_line/commands/sort.ts
@@ -43,6 +43,8 @@ export class SortCommand extends node.CommandBase {
       originalLines.push(vimState.document.lineAt(currentLine).text);
     }
 
+    const lastLineLength = originalLines[originalLines.length - 1].length;
+
     if (this.arguments.unique) {
       const seen = new Set<string>();
       const uniqueLines: string[] = [];
@@ -55,8 +57,6 @@ export class SortCommand extends node.CommandBase {
       }
       originalLines = uniqueLines;
     }
-
-    const lastLineLength = originalLines[originalLines.length - 1].length;
 
     const sortedLines = this.arguments.ignoreCase
       ? originalLines.sort((a: string, b: string) => a.localeCompare(b))


J-Fields:
Nice catch, thanks! Will merge pending a clean run of the test suite

J-Fields (approved):

PR #6852, opened 2021-07-07
Add debuggingForeground to colorCustomizations

lmlorca:
If you chose to customize your modes under `vim.statusBarColor`, and since `statusBar.debuggingBackground` is registered to colorCustomizations so it uses your current mode background, using debugger can render the statusBar unreadable. I'm simply registering the missing `statusBar.debuggingForeground`, so the statusBar uses both your mode's background and foreground, not just the background.


J-Fields (approved):
Thanks!

Issue #6520, opened 2021-04-21
Search for \\c...


edemaine:
**Describe the bug**
It seems impossible to search for `\char` using `/\\char`.  Instead, it ends up searching for `har`.  Searching for `\c` via `/\\c` works fine.  It seems that `\\ch` gets contracted to `h` somehow.

**To Reproduce**

1.  Open any file containing `\char` as well as false matches like `not hard`.
2.  Type `/\\char`
3.  Observe that `har` matches in both instances.

**Expected behavior**
I expect only `\char` to match when searching for `\\char`.

**Environment (please complete the following information):**

- Extension (VsCodeVim) version: 1.20.0
- VSCode version: 1.55.2
- OS: Windows 10

**Workaround:**
`/\\\bchar` successfully searches for `\char`.

J-Fields:
Related: #2710

zeroxia:
Recently after an update, I found I can't do this ex command to replace all `\` to `/` for a Windows path:

`:s/\\/\//g`

And I have to type commands like this:

`:s/\\\\/\//g`

Yes, you need to repeat backslash four times.

Is this intended?


Screencast to demo:

![Demo recording](https://i.imgur.com/NrtzMUP.gif)

edemaine:
@zeroxia I think that's a different issue, so I opened it as #6890.

PR #6891, opened 2021-07-19
Fix escaping in :s substitutions

### LINKED ISSUE #6890: Backslashes must be duplicated in :s substitution
_Originally posted by @zeroxia in https://github.com/VSCodeVim/Vim/issues/6520#issuecomment-882043850_:

Recently after an update, I found I can't do this ex command to replace all `\` to `/` for a Windows path:

`:s/\\/\//g`

And I have to type commands like this:

`:s/\\\\/\//g`

Yes, you need to repeat backslash four times.

Is this intended?

Error message reported:

```
CommandLine: Error executing cmd=s/\\/\//g. err=SyntaxError: Invalid regular expression: /\/: \ at end of pattern.
```

Screencast to demo:

Demo recording

----

@edemaine analysis:
This seems to be directly caused by #6409 which removed a layer of backslashes in both the pattern and the replacement, when it was presumably only supposed to change the latter.  I'm planning to take a look at this.

edemaine:
Fix `\` escaping in `:s` substitution commands. The pattern and replacement need to be processed differently, because `\`s need to be preserved in the pattern for passing to `RegExp`, whereas they need to actually be processed in the replacement.

* `parsePattern` generally leaves backslashes alone, passing them on  to the regular expression parser.
* New `parseReplace` interprets backslashes.  New definitions for `\b`  (backspace), `\&` (same as `$&`), and `\0` through `\9` (same as  `$1` through `$9`).
* Rewrite these parsers in imperative style to avoid any stack overflow.

Fixes #6890

I'd like to change how search (`/`) works as well, for #6520. But I'm finding `src/actions/commands/search.ts` hard to follow. If you could point me in the right direction, I can try to fix that similarly.

### DIFF
--- a/src/cmd_line/subparsers/substitute.ts
+++ b/src/cmd_line/subparsers/substitute.ts
@@ -8,49 +8,70 @@ function isValidDelimiter(char: string): boolean {
   return !!/^[^\w\s\\|"]{1}$/g.exec(char);
 }
 
-function parsePattern(pattern: string, scanner: Scanner, delimiter: string): [string, boolean] {
-  if (scanner.isAtEof) {
-    return [pattern, false];
-  } else {
+function parsePattern(scanner: Scanner, delimiter: string): [string, boolean] {
+  let pattern = '';
+  while (!scanner.isAtEof) {
     let currentChar = scanner.next();
 
     if (currentChar === delimiter) {
-      // TODO skip delimiter
-      return [pattern, true];
+      return [pattern, true]; // found second delimiter
     } else if (currentChar === '\\') {
       if (!scanner.isAtEof) {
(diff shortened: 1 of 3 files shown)


J-Fields:
LGTM, thanks!

J-Fields:
> I'd like to change how search (`/`) works as well, for #6520. But I'm finding `src/actions/commands/search.ts` hard to follow. If you could point me in the right direction, I can try to fix that similarly.

I think the change would be in `searchState.ts`, particularly `set searchString()`. Haven't read through this code in a while, so could be wrong, though - let me know if you aren't able to figure it out quickly and I'll dig further to help out.

PR #6940, opened 2021-08-01
Added TreeSitter-like text objects for classes and functions 

w-cantin:

This PR adds 2 new types of text objects for functions (aF) and classes (ac) using VSCode's Abstract Syntax Tree. Those text objects select the whole class or function definition. It makes yanking or deleting functions and classes a lot easier. This is strongly inspired by this neovim plugin https://github.com/nvim-treesitter/nvim-treesitter-textobjects#built-in-textobjects

The parsed symbols generated by VSCode are language independent so the text objects work for any  different languages using completely different syntax (for example python vs c). The only requirement is a working LSP (this should not be an issue for most language supported by VSCode) . One way to check if the LSP is working properly is to check if the breadcrumbs are showing when a file is open: https://code.visualstudio.com/docs/editor/editingevolved#_breadcrumbs

This PR also provides solid foundations to add more complex features based on parsed code in the future like jumping to the next function "]m" for any language or switching 2 arguments positions for example. I did not want to add too many features at the same time but it should be quite straight forward.  

Unfortunately VSCode does not provide a language independent way to select inner blocks like inner function or inner class so it is only possible to select the entire class or function definition or select one of its children. 

I am still unsure how to handle errors in vim and I was wondering if I had to create my own error code in error.ts and if so what number should I use. I am going to look more into it soon.

Also I was wondering about the keybindings for those. There might be better ones available. I wanted to use "af" instead of "aF" but I noticed that it was already used by SelectAnExpandingBlock text object. I know its a big no no usually to change keybindings like that but considering that `editor.action.smartSelect.expand` already does the same thing out of the box maybe it could be changed.


src/textEditor.ts
```diff
@@ -266,6 +267,36 @@ export class TextEditor {
...
+  public static async getSymbols(editor: vscode.TextEditor) {
+    const uri = editor.document.uri;
+    const symbols = await vscode.commands.executeCommand<vscode.DocumentSymbol[]>(
+      'vscode.executeDocumentSymbolProvider',
+      uri
+    );
+
+    if (symbols === undefined) {
+      // Not quite sure how errors are handled yet.
+      // Should I define my own error in error.ts and if so, is there a logic to
```
J-Fields, line 293: Good question. All the `ErrorCode`s currently defined are taken straight from Vim. `:help errors` is useful, but I don't think it's exhaustive. There's probably a conscious organization of the error codes, but I'm not sure what that is. At the very least, I'd like a comment where this custom error code is defined making it clear that it's custom. Maybe it should also be `E1000` to reserve `E1` to `E999` for errors we inherit from Vim?

src/textEditor.ts
```diff
@@ -266,6 +267,36 @@ export class TextEditor {
...
+   *
+   * Throws an error if no symbols were found.
+   *
+   * @param editor, the active editor containing the document that is parsed.
+   * @returns the list of symbols found for th
+   */
+  public static async getSymbols(editor: vscode.TextEditor) {
+    const uri = editor.document.uri;
+    const symbols = await vscode.commands.executeCommand<vscode.DocumentSymbol[]>(
+      'vscode.executeDocumentSymbolProvider',
```
J-Fields, line 287: Well shit - I didn't know this existed! Very exciting that we'll be able to leverage LSP like this.

src/textobject/astSymbols.ts
```diff
@@ -0,0 +1,171 @@
...
+      }
+    }
+    return new SymbolNode([], [], undefined, undefined);
+  }
+
+  /**
+   * Returns a list containing all the ancestors of a symbol. We go from
+   * left to right in the list where index 0 is the child symbol and last index
+   * is the root symbol.
+   *
```
J-Fields, line 88: Nit: extra blank line at end of this comment and the one below it (line 107)

src/textobject/textobject.ts
```diff
@@ -971,3 +972,56 @@ export class SelectAroundArgument extends SelectArgument {
...
+ */
+abstract class SelectASymbol extends TextObject {
+  // The whitelist is the filter used to only select the Symbol type that we want
+  abstract whitelist: Set<SymbolKind>;
+
+  public override async execAction(position: Position, vimState: VimState): Promise<IMovement> {
+    let start: Position = vimState.cursorStartPosition;
+    let stop: Position = vimState.cursorStopPosition;
+
+    const symbols = await TextEditor.getSymbols(vimState.editor);
```
J-Fields, line 989: Do you think it would be worth it to cache these symbols for a given `TextDocument.version`?

J-Fields (changes requested):
Thanks for the PR, this would be really cool to have! Left a few comments after skimming the code; I'll do a more thorough review and play around with it when I get the chance.

w-cantin:
Hi,

So I was fixing up the PR and I decided to do a lot more tests using more languages to make sure that everything was ok. I used Python and JS/TS for my initial tests and they would work fine for the most part. But using other languages (c# or java) I soon realized that Microsoft's implementation of their own AST navigation is incredibly lackluster and do not work half of the time. 

The DocumentSymbolProvider API provided by VSCode's LSPs and their own functions like [breadcrumbs](https://code.visualstudio.com/docs/editor/editingevolved#_breadcrumbs) and especially (go to symbols)[https://code.visualstudio.com/docs/editor/editingevolved#_go-to-symbol] do not really work well for other languages. Once the breadcrumbs are shown the symbol navigation works great but it is not as universal as it seemed.

It is unfortunate because it was quite exciting and seemed like a good fit for AST navigation but I do not think it is worth pursuing. I am gonna look into TreeSitter directly instead (they provide a JS api) see if I can make this work.

w-cantin:
Ok so I was playing with the symbols provided by VSCode's typescript and javascript LSP and I found a couple limitation. 

- Arrow functions are not recognized as functions by vscode (even if put inside classes as methods)

- Getters and setters are also not considered functions

That means that the Function textobjects does not properly works for Javascript/typescript and there is no real way around it. Other languages though works fine. Let me know if it is a dealbreaker.

fernandovmacedo:
That would be a super nice feature.

PR #6943, opened 2021-08-01
Doc remap

### LINKED ISSUE #6241: Document the use of vim.remap
I ran across issues trying to remap Ctrl+Shift+K and other key combinations several times over the past few months and gave up each time until today when I ran across berknam's response in #5917.

I tried several combinations in my settings.json along the lines of this:
```json
    {
      "before": ["<C-K>"], // <C-S-k> // <M-k> // <M-K> -- None of these worked
      "after": ["<leader>", "<leader>", "2", "s"]
    },
```

This works however and seems very powerful:
```json
  {
    "key": "ctrl+shift+k",
    "command": "vim.remap",
        "when": "inputFocus && vim.mode == 'Normal'",
        "args": {
            "after": ["<leader>","<leader>", "2", "s"],
        }
  },
```
Sample oddball keyboard shortcut configuration  using vim.remap should be documented in the repo's README  instead of the following misleading statement:

> VSCodeVim explicitly instructs VS Code which key events we care about through the package.json. If the key you are trying to remap is a key in which vim/vscodevim generally does not handle, then it's most likely that this extension does not receive those key events from VS Code. With logging level adjusted to 'debug', as you press keys, you should see output similar to:

w-cantin:

This PR adds better documentation for remapping complex key combinations not supported natively by VSCodeVim. 

This PR directly fixes #6241

### DIFF
--- a/README.md
+++ b/README.md
@@ -368,7 +368,34 @@ Custom remappings are defined on a per-mode basis.
     debug: ModeHandler: handling key=<C-a>.
     ```
 
-    As you press the key that you are trying to remap, do you see it outputted here? If not, it means we don't subscribe to those key events.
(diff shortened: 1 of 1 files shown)


w-cantin:
VSCode messed up the formatting for the README so I will fix it and repush the new changes.

PR #6961, opened 2021-08-06
fix #6960

### LINKED ISSUE #6960: Undo doesn't work in Notebook Cell
Originally reported in https://github.com/microsoft/vscode/issues/130142. Undo doesn't work in notebook.

rebornix:
Notebook cell document is never dirty, as a workaround we filter them out of the local change.

### DIFF
--- a/extensionBase.ts
+++ b/extensionBase.ts
@@ -175,7 +175,12 @@ export async function activate(context: vscode.ExtensionContext, handleLocal: bo
 
     if (handleLocal) {
       setTimeout(() => {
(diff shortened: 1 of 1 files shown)


J-Fields:
This works within a single cell, but it gets confused when you go between cells (edit cell 1, go to cell 2, undo). Does VSCode provide IDs for notebook cells via a context value or something?

rebornix:
@J-Fields every cell's document uri is unique which can be used as cell id. Can you provide suggestions how I can hook that up?

J-Fields:
Superseded by 4585ef2723cceb672922729ada46eebb43f262c7

Issue #6965, opened 2021-08-08
Allow for appending to [a-z] registers


DevinLeamy:
**Is your feature request related to a problem? Please describe.**
- Current I cannot append to `[a-z]` registers 
- vim provides: "`q{0-9a-zA-Z"}` Record typed characters into register {0-9a-zA-Z"} `(uppercase to append)`"

**Describe the solution you'd like**
- Ideally, I could append to my `[a-z]` registers, like in vim

**Describe alternatives you've considered**
- I saw the de-facto vim implementation of this feature on the roadmap. I did not consider alternative solutions.

**Additional notes**
- I've implemented the desired changes on a fork. If this issue is given the OK, I'll submit a PR.

PR #6971, opened 2021-08-10
Append to [a-z] registers

DevinLeamy:
- Implements appending to `[a-z]` registers via `q[A-Z]`
- It's really nice to have, is part of [neo]vim, and was missing

- https://github.com/VSCodeVim/Vim/issues/6965 

- Passes `npx gulp test`


src/actions/commands/actions.ts
```diff
@@ -341,7 +341,11 @@ export class CommandQuitRecordMacro extends BaseCommand {
 
     const existingMacro = (await Register.get(macro.registerName))?.text;
     if (existingMacro instanceof RecordedState) {
-      existingMacro.actionsRun = existingMacro.actionsRun.concat(macro.actionsRun);
+      if (/^[A-Z]+$/.test(macro.registerKey)) {
```
J-Fields, line 344: Make `Register.isValidUppercaseRegister()` public and call it instead of this

J-Fields (changes requested):
LGTM other than one small change. Thanks!

J-Fields (approved):

Issue #6995, opened 2021-08-13
`:delete` and `:yank` should support `{count}` argument


J-Fields:
For example, `:y4` should yank 4 lines

DevinLeamy:
@J-Fields 
Currently, the first (and only) argument passed into both the `DeleteRangeCommand` and the `YankCommand` are register names - if an argument exists at all. So when `:y4` is called, then `{ register: '4' }`. To accommodate a count argument, can we assume that the first argument passed in will always be a count. As in:
```ts
/**
   * Note that to specify a register, a [cnt] for :d[elete][cnt]
   * must be specified.
   * Ex: :d4 i -> specifies register i
   *     :d  i -> does not specify a register
   */
  const scanner = new Scanner(args);
  return new node.DeleteRangeCommand({
    linesToRemove: +scanner.nextWord() ?? 1,
    register: scanner.isAtEof ? undefined : scanner.nextWord(),
  });
```

J-Fields:
I'm away from my PC at the moment so I can't check, but I thought the first argument was the register, unless it's a number, in which case it's used as the count.
How does Vim interpret `:y 4 x` and `:y 4 7`?

DevinLeamy:
_"I thought the first argument was the register, unless it's a number, in which case it's used as the count."_  - Just tested, this is correct.

`:y 4 x` and `:y 4 7` don't work. 

```
:[range]y[ank] [x]	Yank [range] lines [into register x]. Yanking to the
			"* or "+ registers is possible only when the
			|+clipboard| feature is included.

:[range]y[ank] [x] {count}
			Yank {count} lines, starting with last line number
			in [range] (default: current line |cmdline-ranges|),
			[into register x].

```

PR #7007, opened 2021-08-18
Adds count argument to `:yank` and `:delete` commands

DevinLeamy:
- Adds [cnt] to `:y[ank] [register] [cnt]`
- Adds [cnt] to `:d[elete] [register] [cnt]`

- https://github.com/VSCodeVim/Vim/issues/6995

- The first argument is a register, unless it's a number, in which case it's treated as the count
- Passes tests


src/cmd_line/commands/yank.ts
```diff
@@ -34,7 +61,12 @@ export class YankCommand extends CommandBase {
...
-    await this.yank(vimState, vimState.cursorStartPosition, vimState.cursorStopPosition);
+    const linesToYank = this.arguments.linesToYank;
+    const startPosition = vimState.cursorStartPosition;
+    const endPosition = linesToYank
+      ? startPosition.getDown(linesToYank - 1).getLineEnd()
+      : vimState.cursorStopPosition;
+    await this.yank(vimState, startPosition, endPosition);
   }
 
   override async executeWithRange(vimState: VimState, range: LineRange): Promise<void> {
```
J-Fields, line 72: Seems like `[count]` is used if it's given, even if there's a range as well

J-Fields (changes requested):
LGTM, other than one change requested

src/cmd_line/commands/yank.ts
DevinLeamy: I tested this with `:[%, $, number]yank [reg] [cnt]` and `executeWithRange(...)` is called instead of `execute(...)`, in other words, I haven't encountered this behaviour. Is there a particular command where you saw this happen?

J-Fields: Sorry, I mean Vim uses `[count]` if it's given, even with a range. So with VisualLine highlighting of lines 2 and 3 here:
```
one
two
three
four
five
six
```
`:y3` should copy
```
four
five
six
```

DevinLeamy: Added support for `[range]` with `[cnt]` argument to `:delete` and `:yank`, and added a test suite for `:yank`. The build and tests work locally, not sure why Prettier is failing in the Github Action with:
```
Amending the current commit...
You are not currently on a branch.
Please specify which branch you want to merge with.
See git-pull(1) for details.

    git pull <remote> <branch>
```
Can I manually rerun the action?

J-Fields: Thanks!

Yeah, it sees that the formatting isn't right and tries to fix that for you, but I think the action is misconfigured or something. In the future, `gulp forceprettier` should fix it for you.

J-Fields (approved):
Thanks!

Issue #1767, opened 2017-05-27
Emulating `targets.vim`


wagk:
* Click *thumbs-up* 👍 on this issue if you want it!
* Click *confused* 😕 on this issue if not having it makes VSCodeVim unusable. 

The VSCodeVim team prioritizes issues based on reaction count. 

--------

**Is this a BUG REPORT or FEATURE REQUEST?** (choose one): FEATURE REQUEST

Hello,

Is it possible to emulate [`targets.vim`](https://github.com/wellle/targets.vim)? 

To elaborate `targets.vim` is a plugin that defines new text objects;

- Pair text objects (`[`, `]`, `(`, `)`, `{`, `}`, `<`, `>`, tags)
  - Vim already does this but the plugin makes it accept a count amongst other things
- Quote text objects (`'`, `"`, `` ` ``)
- Separator text objects (any separator-delimited list, supported separators being `, . ; : + - = ~ _ * # / | \ & $`)
- Argument text objects (function arguments, etc)

---

Anecdotally `targets.vim` matches `surround.vim` for sheer utility for my day to day tasks, being able to manipulate arguments as text objects (specifically) smooths over a lot things for me.

Chillee:
This is actually a duplicate of https://github.com/VSCodeVim/Vim/issues/601.

It's most likely that support for this will likely come through more neovim integration, imo.

J-Fields:
Duplicate of #601

Issue #601, opened 2016-08-12
Request for plugin support: targets.vim


extr0py:
One plugin that is very useful for my VIM flow is this one:
https://github.com/wellle/targets.vim

It builds on the idea of the ```ci"``` command - where it jumps forward to find the nearest quotation, and changes inside, for other characters, like (), {}, [], etc.

It also supports argument text objects - which are handy for deleting and changing in function arguments. So ```cia``` would jump forward to an argument and change. 

These motions and text objects are nice because they mean you can strike from a distance.

In terms of integration, it seems like it depends on whether the plugin story for this is implemented:
https://github.com/VSCodeVim/Vim/issues/590

If that happens, we might want to look at making the text objects extensible to support these scenarios.
Otherwise, we could potentially bake it in as a setting.

Note that ```ci"``` is already supported in VSCodeVim (and VIM out of the box), so potentially the work at least for those jump commands might not be too bad.

johnfn:
Anyone who wants this feature can feel free to get started on it. It would simply be too much work for me to port over every Vim plugin, but if you (or anyone else) wants to open up a WIP PR, or ask me questions on Slack, I'm happy to help you along the way. :)

Austin-Scott:
Inspired by the recent addition of argument text objects #4653 I'm currently taking a shot at implementing the `in` ("inner next" such as `cin(`), `an` ("around next" such as `can(`), `il` ("inner last" such as `cil(`), and `al` ("around last" such as `cal(`) commands from this plugin.

J-Fields:
@Austin-Scott Sweet! Feel free to reach out if you need any guidance

berknam:
> 
> 
> Inspired by the recent addition of argument text objects #4653 I'm currently taking a shot at implementing the `in` ("inner next" such as `cin(`), `an` ("around next" such as `can(`), `il` ("inner last" such as `cil(`), and `al` ("around last" such as `cal(`) commands from this plugin.

I've been looking into implementing the `:normal` command which would allow to create this type of text objects with operator-pending mode remaps (after PR #4735) which would allow to create remaps like `onoremap in( :<C-u>normal! f(vi(<CR>`  and `onoremap il( :<C-u>normal! F)vi(<CR>` 

It might take a while still.

larrywal:
Any news or progress here? - agree this would be super useful.  In particular, the seeking behavior where I don't have to be within brackets or braces to make something like ci( work.

Austin-Scott:
[...]

PR #7025, opened 2021-08-24
Smart quotes

elazarcoh:

It implements better quotes movement behavior.
* handles quotes match correctly
  ```
  a "b" |c "d"  -> di"  ->  a "b" c "|"   ; instead of   a "b"|"d"
  ```
* added [n]ext and [l]ast quotes (by skipping current)
  ```
  a |"b" c "d"  -> din"  ->  a "b" c "|"
  ```
* added break-through lines support for next/last quotes (toggleable via config)
  ```
  a |"b"
  foo
  c "d"
    
    -> din"  ->  
    
  a "b"
  foo
  c "|"
  ```
  
this behavior can be enabled via `vim.smartQuotes.enable`

partially: #601, #1767


package.json
```diff
@@ -722,6 +722,21 @@
           "markdownDescription": "`#editor.lineNumbers#` is determined by the active Vim mode, absolute when in insert mode, relative otherwise.",
           "default": false
         },
+        "vim.smartQuotes.enable": {
+          "type": "boolean",
+          "markdownDescription": "Enable smart quotes movements (similar to targets.vim)",
```
J-Fields, line 727: Maybe make `targets.vim` a link to the original plugin? ([this anchor](https://github.com/wellle/targets.vim#quote-text-objects) seems like the equivalent)

package.json
```diff
@@ -722,6 +722,21 @@
...
+          "default": true
+        },
+        "vim.smartQuotes.breakThroughLines": {
+          "type": "boolean",
+          "markdownDescription": "Whether to break through lines when using [n]ext/[l]ast motion",
+          "default": true
+        },
+        "vim.smartQuotes.aIncludesSurrouningSpaces": {
+          "type": "boolean",
+          "markdownDescription": "Whether to use default vim behaviour when using `a` (e.g. da') which include surrounding spaces, or not, as for other text objects.",
```
J-Fields, line 737: Capitalize `vim` and add backticks around `da'`

package.json
```diff
@@ -722,6 +722,21 @@
...
           "default": false
         },
+        "vim.smartQuotes.enable": {
+          "type": "boolean",
+          "markdownDescription": "Enable smart quotes movements (similar to targets.vim)",
+          "default": true
+        },
+        "vim.smartQuotes.breakThroughLines": {
+          "type": "boolean",
+          "markdownDescription": "Whether to break through lines when using [n]ext/[l]ast motion",
```
J-Fields, line 732: `markdownDescription` should always end with a period

package.json
```diff
@@ -722,6 +722,21 @@
...
+          "default": true
+        },
+        "vim.smartQuotes.aIncludesSurrouningSpaces": {
```
J-Fields, line 735: typo: `Surrouning`

src/configuration/iconfiguration.ts
```diff
@@ -79,6 +79,21 @@ export interface ICamelCaseMotionConfiguration {
   enable: boolean;
 }
 
+export interface ISmartQuotesConfiguration {
```
J-Fields, line 82: Same changes as requested in `package.json`

test/plugins/smartQuotes.test.ts
```diff
@@ -0,0 +1,993 @@
...
+      end: [
+        'aaa "bbb" c', //
+        'd "e" f', //
+        'g "|" k',
+      ],
+    });
+    newTest({
+      title: 'next: should go next line - 2',
+      start: [
+        'aaa "bbb" c', //
```
J-Fields, line 720: What's with all the empty comments?

test/plugins/smartQuotes.test.ts
```diff
@@ -0,0 +1,993 @@
+import { Configuration } from '../testConfiguration';
+import { newTest } from '../testSimplifier';
+import { cleanUpWorkspace, setupWorkspace } from '../testUtils';
+
+suite('smartQuotes plugin', () => {
```
J-Fields, line 5: Love to see thorough tests like this; thanks! 🙂

src/actions/plugins/targets/smartQuotesMatcher.ts
```diff
@@ -0,0 +1,336 @@
...
+  second: Dir;
+  includeCurrent: boolean;
+};
+type QuotesAction = {
+  search: SearchAction | undefined;
+  skipToLeft: number; // for last quotes, how many quotes need to skip while searching
+  skipToRight: number; // for next quotes, how many quotes need to skip while searching
+};
+
+const quoteDirs: Record<string, QuotesAction> = {
```
J-Fields, line 23: Can you add a short comment explaining what the key is?

src/actions/plugins/targets/smartQuotesMatcher.ts
```diff
@@ -0,0 +1,336 @@
...
+  },
+  '000': {
+    //  |ab
+    search: undefined,
+    skipToLeft: 0,
+    skipToRight: 0,
+  },
+};
+
+export class SmartQuoteMatcher {
```
J-Fields, line 135: Can you add a comment to this class describing how it differs from the usual `i"`?

J-Fields (changes requested):
Still need to do a deeper reading of the core implementation, but it all seems very reasonable. Left a handful of surface-level comments for now, thanks for the PR!

package.json
elazarcoh:
```suggestion
          "markdownDescription": "Enable smart quotes movements (similar to [targets.vim](https://github.com/wellle/targets.vim#quote-text-objects))",
```

package.json
elazarcoh:
```suggestion
          "markdownDescription": "Whether to use default Vim behavior when using `a` (e.g. `da'`) which include surrounding spaces, or not, as for other text objects.",
```

package.json
elazarcoh:
```suggestion
          "markdownDescription": "Whether to break through lines when using [n]ext/[l]ast motion.",
```

package.json
elazarcoh:
```suggestion
        "vim.smartQuotes.aIncludesSurroundingSpaces": {
```

test/plugins/smartQuotes.test.ts
elazarcoh: Prettify automatically joins the lines, and I think it's make it hard to read the code otherwise.

We can do this instead, if you prefer
```suggestion
        'aaa "bbb" c', // linebreak
```

test/plugins/smartQuotes.test.ts
elazarcoh: Tried to cover everything.
Though did I miss a few cases and added them later, using the same code I where I found the bug 🙃

src/actions/plugins/targets/smartQuotesMatcher.ts
elazarcoh:
```suggestion
/**
 * This mapping is used to give a way to identify which action we need to take when operating on a line.
 * The keys here are, in some sense, the number of quotes in the line, in the format of `lcr`, where:
 * `l` means left of the cursor, `c` whether the cursor is on a quote, and `r` is right of the cursor.
 * 
 * It is based on ideas used in `targets.vim`. For each line & cursor position, we count the number of quotes
 * left (#L) and right (#R) of the cursor. Using those numbers and whether the cursor it on a quote, we know 
 * what action to make.
 * 
 * For each entry we have an example of a line & position.
 */
const quoteDirs: Record<string, QuotesAction> = {
```

src/actions/plugins/targets/smartQuotesMatcher.ts
elazarcoh:
```suggestion
/**
 * An extended of `QuoteMatcher` (though not inheritance because it's too differ from the original).
 *
 * This class is intended to handle quotes-object in a "smarter" way than the default one. e.g., when
 * you have a line ` "a"| "b" ` and the cursor in between the quotes-objects, instead of treating the
 * space between as a quote-object, we will take "b" to be our object of interest.
 * So, in this case, `di"` would result in ` "a" "|" `, instead of ` "a""b" `.
 */
export class SmartQuoteMatcher {
```
Does that make sense?

elazarcoh:
Thanks for the review 😄 

I have another thing I want to discuss, regarding the naming/structuring of the configurations.

As I said, I already implemented another part of `targets.vim`, which handles next/last objects, e.g.
```
"|  [abc] "  ->  din[   ->  "  [|]"
```

I was thinking of whether we want to have all of it under the same plugin name: `targets`.
Then the configurations will be something like:
```ts
interface IConfiguration {
    // ...
    targets: {
       enable: boolean
       smartQuotes: {
          enable: boolean  // default to true and will apply only if targets is enabled
       }
       // other stuff related to targets
   }
}
```

I consider it important to make a decision, because we changing it later is more painful. 

I don't know if `targets.vim` will be implemented in full, at least by me. I'm in favor of having a some kind of understanding which functionality are more useful than other and operate by it (need to figure out how to do it).

I did started playing with the idea of custom text objects (means defined by the user in the settings), e.g. https://github.com/kana/vim-textobj-user, which save us from implementing some of the maybe less used `targets.vim` functionality, e.g. `di-` ([separator](https://github.com/wellle/targets.vim#in-separator)).
I made a POC, but it's not ready for a PR. For this we would need to have a another discussion, if you think you can give it a time (for discussion and review, I mean. The implementation is on me).

src/configuration/iconfiguration.ts
```diff
@@ -79,6 +79,21 @@ export interface ICamelCaseMotionConfiguration {
...
+   */
+  enable: boolean;
+  /**
+   * Whether to break through lines when using [n]ext/[l]ast motion
+   */
+  breakThroughLines: boolean;
+  /**
+   * Whether to use default vim behaviour when using `a` (e.g. da') which include surrounding spaces, or not, as for other text objects.
+   */
+  aIncludesSurrouningSpaces: boolean;
```
elazarcoh, line 94:
```suggestion
  aIncludesSurroundingSpaces: boolean;
```

src/configuration/configuration.ts
```diff
@@ -264,6 +265,12 @@ class Configuration implements IConfiguration {
   easymotionKeys = 'hklyuiopnm,qwertzxcvbasdgjf;';
   easymotionJumpToAnywhereRegex = '\\b[A-Za-z0-9]|[A-Za-z0-9]\\b|_.|#.|[a-z][A-Z]';
 
+  smartQuotes: ISmartQuotesConfiguration = {
+    enable: false,
+    breakThroughLines: false,
+    aIncludesSurrouningSpaces: true,
```
elazarcoh, line 271:
```suggestion
    aIncludesSurroundingSpaces: true,
```

test/testConfiguration.ts
```diff
@@ -45,6 +46,11 @@ export class Configuration implements IConfiguration {
   easymotionMarkerFontWeight = 'bold';
   easymotionMarkerMargin = 0; // Deprecated! No longer needed!
   easymotionKeys = 'hklyuiopnm,qwertzxcvbasdgjf;';
+  smartQuotes: ISmartQuotesConfiguration = {
+    enable: false,
+    breakThroughLines: false,
+    aIncludesSurrouningSpaces: true,
```
elazarcoh, line 52:
```suggestion
    aIncludesSurroundingSpaces: true,
```

src/configuration/iconfiguration.ts
elazarcoh: changes were committed

src/actions/motion.ts
```diff
@@ -2000,41 +2002,86 @@ export abstract class MoveQuoteMatch extends BaseMovement {
...
+      }
+      let { start, stop, lineText } = res;
+
+      if (!this.includeQuotes) {
+        // Don't include the quotes
+        start = start.translate({ characterDelta: 1 });
+        stop = stop.translate({ characterDelta: -1 });
+      } else if (
+        this.adjustForTrailingWhitespace &&
+        configuration.smartQuotes.aIncludesSurrouningSpaces
```
elazarcoh, line 2020:
```suggestion
        configuration.smartQuotes.aIncludesSurroundingSpaces
```

src/actions/plugins/targets/smartQuotesMatcher.ts
J-Fields: Yup! I think `extended` and `differ` are typos though

J-Fields:
> I was thinking of whether we want to have all of it under the same plugin name: `targets`.
> Then the configurations will be something like:
> ...

Looks good, though wouldn't `targets.enable` be redundant?

elazarcoh:
> > I was thinking of whether we want to have all of it under the same plugin name: `targets`.
> > Then the configurations will be something like:
> > ...
> 
> Looks good, though wouldn't `targets.enable` be redundant?

I thought about it. I was thinking, since this plugin would have different features, we probably want to be able to disable each of them separately. the reason I suggested to have also a top level enablement so it would me more convenient for plugin users to enable disable all of them at once.

If after reading this you still think so, it's totally fine by me to remove it.

src/actions/plugins/targets/smartQuotesMatcher.ts
elazarcoh: updated. feel free to suggest an alternative description, though.

J-Fields:
> I thought about it. I was thinking, since this plugin would have different features, we probably want to be able to disable each of them separately. the reason I suggested to have also a top level enablement so it would me more convenient for plugin users to enable disable all of them at once.

Yeah, my only concern is the exact semantics. I guess the top level would default to `false`, and the specific flags would default to `undefined`, where if they're defined they override the top level?

package.json
```diff
@@ -722,6 +722,26 @@
...
           "default": false
         },
+        "vim.targets.enable": {
+          "type": "boolean",
+          "markdownDescription": "Enable [targets.vim](https://github.com/wellle/targets.vim#quote-text-objects) plugin (not fully implmeneted yet).",
+          "default": false
+        },
+        "vim.targets.smartQuotes.enable": {
+          "type": "boolean",
+          "markdownDescription": "Enable smart quotes movements",
```
J-Fields, line 732: This should end in a period and should probably include [this link](https://github.com/wellle/targets.vim#quote-text-objects)

package.json
```diff
@@ -722,6 +722,26 @@
...
+          "default": true
+        },
+        "vim.targets.smartQuotes.breakThroughLines": {
+          "type": "boolean",
+          "markdownDescription": "Whether to break through lines when using [n]ext/[l]ast motion.",
```
J-Fields, line 737: What does this mean by `[n]ext/[l]ast motion`?

elazarcoh: It described in targets.vim.
we have new motions: din", which stands for "delete in next double-quotes" and dil" which stands for "delete in last double-quotes".

package.json
elazarcoh:
```suggestion
          "markdownDescription": "Enable the smart quotes movements from [targets.vim](https://github.com/wellle/targets.vim#quote-text-objects).",
```

elazarcoh:
Added support for `iq` `aq` `inq` etc. motions. see https://github.com/wellle/targets.vim#any-quote.
Current implementation doesn't work well for nested quotes (but with the reqular motions, `i"` it works ok as before).
Added simple tests for it as well.

package.json
J-Fields: Can you link to that in the description as well?

elazarcoh:
```suggestion
          "markdownDescription": "Whether to break through lines when using [n]ext/[l]ast motion, see [targets.vim#next-and-last-quote](https://github.com/wellle/targets.vim#next-and-last-quote).",
```

elazarcoh:
@J-Fields do you think I should remove the `diq` etc. from the current PR? It is pretty handy, I think. On the other hand, you can say it's not a complete solution because of the quotes nesting issue I mentioned.
I can make another PR for it, instead.

elazarcoh:
@J-Fields with the release of pre-release, do you think we can push it faster?

J-Fields (approved):
LGTM. Thanks for the clean implementation and thorough test suite!

PR #7092, opened 2021-09-17
#6553: capture file mode and restore it after force write

### LINKED ISSUE #6553: Running :w! on a read-only file results in messed up permissions.
When you try to use :w! on a read-only file, it will mess up the file permissions.

**To Reproduce**

1.  Open a new folder in VSCode
2.  Create a new file
3.  Do a chmod -w on said file.
4.  Edit the file in VSCode. Run :w! on the file. File permissions are now "--w--wx-wT".

**Expected behavior**
File content should be written on disk, and permissions shouldn't change.

**Environment**

- Extension (VsCodeVim) version: v1.20.1
- VSCode version: 1.55.2
- OS: macOS 10.15.7

joecrop:

This PR corrects an issue with the force-write (:w!) feature. It solves two issues:
- In proper vim fashion, force-writing a file should not change the permissions of the file permanently.
- When making a file writable (now temporarily), the chmod value should be octal, but was decimal.

fixes #6553 


J-Fields (approved):
Nice catch, thanks!

PR #7122, opened 2021-10-02
Fix bang command with ranges

tagniam:
Fixes the bang `!` command when used with a range, which is currently broken since any bang commands with ranges operate on all lines since the range isn't passed in properly.

Also improved some tests which should've detected this regression, but didn't because they executed on all lines.

#3136


J-Fields:
Thank you!

Issue #7123, opened 2021-10-02
Iskeyword doesn't work for multiple languages


shinichy:
**Describe the bug**
`vim.Iskeyword` doesn't seem to change when switching to a different language file. For example, when there are two different language files are open (e.g. Clojure and YAML) and I reload a window with the Clojure file focused, `w` command skips `-` in YAML files (`-` is not included in `wordSeparators` of Clojure type). `w` command should stop at `-` in YAML files because it's included in the default `wordSeparatos`.

```
    "[clojure]": {
        "editor.wordSeparators": "\t ()\"':,;~@#$%^&{}[]`"
    }
```

**To Reproduce**

1.  Install [Calva](https://marketplace.visualstudio.com/items?itemName=betterthantomorrow.calva), which is a Clojure extension (it will set `editor.wordSeparators` above by default)
2.  Click `New File` and populate `a-b` as the content
3.  Change Language to Clojure
2.  Click `New File` and populate `a-b` as the content
3.  Change Language to YAML
1.  Open the Clojure file created above
2.  Reload Window
3.  Open the YAML file created above
4.  Type `w` when the cursor is at `a`
5.  The cursor stops at `b` (`-` is skipped)

**Expected behavior**
The cursor stops at `-`

**Environment (please complete the following information):**

- Extension (VsCodeVim) version: 1.21.8
- VSCode version: 1.60.2
- OS: MacOS 11.5.2

PR #7126, opened 2021-10-04
iskeyword is evaluated when a command is called (#7123)

shinichy:

`configuration.iskeyword` is evaluated only once when the extension is initialized. This means that `iskeyword` for a specific language is not respected when it's different from the language used for initialization. This PR solves the issue by evaluating it every time the word movement command is called.

#7123


J-Fields:
Thank you!

PR #7215, opened 2021-11-01
search operator `\%V`

### LINKED ISSUE #4936: `\%V` not recognized in substitutions within visual selections.
**Describe the bug**
In vim, if you want to do a substitution only on text within a visual selection, you need to use the `\%V` atom. See [vimwiki](https://vim.fandom.com/wiki/Search_and_replace_in_a_visual_selection#Substituting_in_a_visual_selection) on the subject.

However, in VSCode Vim, the same technique appears to just result in "E486: Pattern Not Found" instead of doing the substitution.

**To Reproduce**

1. Create a file with the text `"-Hello-There-"`
2. Visually select the `o-T` in the middle of Hello and There.
3. Run the command: `'<,'>s/\%V-//`
4. See "E486: Pattern Not Found"

**Expected behavior**
In Vim, the resulting text leftover after the above command is `"-HelloThere-"`. I expect the same from VSCode Vim.

- Issue Type: `Bug`
- Extension Name: `vim`
- Extension Version: `1.14.5`
- OS Version: `Windows_NT x64 10.0.19041`
- VSCode version: `1.46.0-insider`

elazarcoh:
it implements `\%V` in search.
substitution will come in later (maybe in this PR). **Edit:** substitution works out-of-the-box.

fixes #4936 

### DIFF
--- a/src/vimscript/pattern.ts
+++ b/src/vimscript/pattern.ts
@@ -125,6 +125,7 @@ export class Pattern {
           vimState.document.positionAt(searchOffset + match.index + match[0].length)
         );
         if (
+          !this.inSelection &&
           lineRange &&
           (matchRange.start.line < lineRange.start || matchRange.end.line > lineRange.end)
         ) {
--- a/src/vimscript/pattern.ts
+++ b/src/vimscript/pattern.ts
@@ -1,6 +1,6 @@
 import { escapeRegExp } from 'lodash';
 import { alt, any, lazy, noneOf, oneOf, Parser, seq, string } from 'parsimmon';
-import { Position, Range, TextDocument } from 'vscode';
+import { Position, Range, Selection, TextDocument } from 'vscode';
 import { configuration } from '../configuration/configuration';
 import { VimState } from '../state/vimState';
 import { LineRange } from './lineRange';
@@ -40,6 +40,7 @@ export class Pattern {
   public readonly direction: SearchDirection;
   public readonly regex: RegExp;
   public readonly ignorecase: boolean | undefined;
+  public readonly inselection: boolean | undefined;
 
   private static readonly MAX_SEARCH_RANGES = 1000;
   private static readonly SPECIAL_CHARS_REGEX = /[\-\[\]{}()*+?.,\\\^$|#\s]/g;
@@ -83,8 +84,16 @@ export class Pattern {
       fromPosition = args.fromPosition;
     }
 
-    const haystack = vimState.document.getText();
-    const startOffset = vimState.document.offsetAt(fromPosition);
+    const visualSearch = this.inselection && vimState.lastVisualSelection;
+    const haystack = vimState.document.getText(
+      visualSearch
+        ? new Selection(vimState.lastVisualSelection!.start, vimState.lastVisualSelection!.end)
+        : undefined
+    );
+    const searchOffset = visualSearch
+      ? vimState.document.offsetAt(vimState.lastVisualSelection!.start)
+      : 0;
+    const startOffset = vimState.document.offsetAt(fromPosition) - searchOffset;
     this.regex.lastIndex = startOffset;
 
     const matchRanges = {
@@ -102,8 +111,8 @@ export class Pattern {
         }
 
         const matchRange = new Range(
-          vimState.document.positionAt(match.index),
-          vimState.document.positionAt(match.index + match[0].length)
+          vimState.document.positionAt(searchOffset + match.index),
+          vimState.document.positionAt(searchOffset + match.index + match[0].length)
         );
         if (
           lineRange &&
@@ -176,6 +185,7 @@ export class Pattern {
       : '?';
     // TODO: Some escaped characters need special treatment
     return alt(
+      string('\\%V').map((_) => ({ inselection: true })),
       string('\\')
         .then(any.fallback(undefined))
         .map((escaped) => {
@@ -200,12 +210,15 @@ export class Pattern {
       .map((atoms) => {
         let patternString = '';
         let caseOverride: boolean | undefined;
+        let inselection: boolean | undefined;
         for (const atom of atoms) {
           if (typeof atom === 'string') {
             patternString += atom;
           } else {
             if (atom.ignorecase) {
               caseOverride = true;
+            } else if (atom.inselection) {
+              inselection = atom.inselection;
             } else if (caseOverride === undefined) {
               caseOverride = false;
             }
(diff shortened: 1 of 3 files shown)


src/vimscript/pattern.ts
```diff
@@ -40,6 +40,7 @@ export class Pattern {
   public readonly direction: SearchDirection;
   public readonly regex: RegExp;
   public readonly ignorecase: boolean | undefined;
+  public readonly inselection: boolean | undefined;
```
J-Fields, line 43: Make this `inSelection`, please

src/vimscript/pattern.ts
```diff
@@ -83,8 +84,16 @@ export class Pattern {
       fromPosition = args.fromPosition;
     }
 
-    const haystack = vimState.document.getText();
-    const startOffset = vimState.document.offsetAt(fromPosition);
+    const visualSearch = this.inselection && vimState.lastVisualSelection;
+    const haystack = vimState.document.getText(
```
J-Fields, line 88: Can you write this like so:
```typescript
let haystack: string;
let searchOffset: number;
if (this.inSelection && vimState.lastVisualSelection) {
  ...
} else {
  ...
}
```

This should make the type assertions unnecessary.

elazarcoh: 👍 will do
I thought you would prefer the constness

src/vimscript/pattern.ts
```diff
@@ -83,8 +84,16 @@ export class Pattern {
...
+      visualSearch
+        ? new Selection(vimState.lastVisualSelection!.start, vimState.lastVisualSelection!.end)
```
J-Fields, line 90: No need to import `Selection` - you can make this `new Selection(...)`

src/vimscript/pattern.ts
```diff
@@ -200,12 +210,15 @@ export class Pattern {
       .map((atoms) => {
         let patternString = '';
         let caseOverride: boolean | undefined;
+        let inselection: boolean | undefined;
```
J-Fields, line 213: This approach (if there's a `\%V` anywhere, the whole string must be contained in the visual selection) handles the common case, but can you put a `TODO` somewhere noting that it's technically incorrect (as only that zero-length character has to be in the visual selection)?

src/vimscript/pattern.ts
J-Fields: Also, please make it a `boolean` and default to `false` in the constructor

src/vimscript/pattern.ts
elazarcoh: I wasn't aware of that... I guess I can implement it exactly as described in `:help \%V` if you think it worth it.
Otherwise, I'll add a comment.

src/vimscript/pattern.ts
J-Fields: Yeah this is an annoyance I have with typescript (and C++, which I write in my day job) - rust does it right IMO by making blocks expressions... In any case, I like const-ness, but I care more about type safety.

src/vimscript/pattern.ts
J-Fields: Nah I don't think it's worth it, at least not now. At some point I'll probably build proper support for all the weird zero-length atoms Vim supports, but for now (all this `Pattern` code is brand new), let's avoid the complexity and satisfy what's probably the only sane use case of `\%V`.

src/vimscript/pattern.ts
elazarcoh: fixed

src/vimscript/pattern.ts
elazarcoh: fixed

src/vimscript/pattern.ts
elazarcoh: fixed

src/vimscript/pattern.ts
```diff
@@ -83,8 +84,20 @@ export class Pattern {
       fromPosition = args.fromPosition;
     }
 
-    const haystack = vimState.document.getText();
-    const startOffset = vimState.document.offsetAt(fromPosition);
+    let haystack: string;
+    let searchOffset: number;
+    if (this.inSelection && vimState.lastVisualSelection) {
+      // TODO: This is not exactly how Vim implements in-selection search (\V%), see :help \%V for more info.
```
J-Fields, line 90: typo: `\V%`

src/vimscript/pattern.ts
```diff
@@ -83,8 +84,20 @@ export class Pattern {
...
+      haystack = vimState.document.getText(
+        new Range(vimState.lastVisualSelection.start, vimState.lastVisualSelection.end)
+      );
+      searchOffset = vimState.document.offsetAt(vimState.lastVisualSelection.start);
+    } else {
+      haystack = vimState.document.getText();
+      searchOffset = 0;
```
J-Fields, line 97: Something seems a bit buggy when I test it out...

Shouldn't this be `searchOffset = vimState.document.offsetAt(fromPosition)`, and just use that instead of `searchOffset`?

elazarcoh: Hmm, can you describe what seems buggy? maybe I didn't test this use case.

I'll check it again, but the searchOffset is required since the regex match will return the index relative to the selection start, and fromPosition is just where to start searching.

src/vimscript/pattern.ts
elazarcoh:
```suggestion
      // TODO: This is not exactly how Vim implements in-selection search (\%V), see :help \%V for more info.
```

src/vimscript/pattern.ts
J-Fields: It doesn't always include every match within the visual selection; I think it's starting from the wrong position

elazarcoh: I think it fixed now, by 1a96e70.

elazarcoh:
I seems that substitution is working (almost) out-of-the-box! kudos! In the previous version of the code, before your last refactoring, I saw it would be more painful to implement. 
The only issue with it that it replaces all the lines, instead of one by one. I'll work on it. Do you want it in a separate PR, or we can have both in this one?

elazarcoh:
Substitution works as well. I've added basic tests.
Note that I tagged this PR as "fixes" for the relevant issue (previously it wasn't tagged as such)

J-Fields:
> Substitution works as well (out-of-the-box! kudos for the refactoring!)

Nice, good timing for this PR!

J-Fields (approved):
Thanks a lot for the PR and for working through feedback quickly!

Issue #7100, opened 2021-09-21
Vim filter with WSL2 is unavailable


dpsigor:
**Is your feature request related to a problem? Please describe.**
I wish to filter commands available in my default WSL2 shell, instead of filtering through PowerShell.

**Describe the solution you'd like**
One possible solution is to parse `cmd args | cmd2 args2` into `wsl cmd args | wsl cmd2 args2`. In this case, no configs like bashrc are loaded.
Another possible solution, a better one, is to use the integrated terminal.

**Additional context**

Vim:
![image](https://user-images.githubusercontent.com/71943303/134237763-fff2b1e5-fa65-43a6-9469-2094e8e17919.png)
![image](https://user-images.githubusercontent.com/71943303/134237782-7b39c635-3895-43c5-98e1-0119c34ade5f.png)

VSCodeVim:
![image](https://user-images.githubusercontent.com/71943303/134237823-500e5c92-53dc-4b2e-8a02-0616802291da.png)
![image](https://user-images.githubusercontent.com/71943303/134237852-b021b2b8-4cd8-41cb-8ae7-42aa1dd91a98.png)
As in, sh is not a recognized PowerShell command.

Filter commands were implemented with #5042.

tagniam:
Hi @dpsigor, thanks for bringing this up. I'm testing some code that would involve allowing custom shells for filter commands as a VSCodeVim setting. 

You'll then simply just need to add `"vim.shell": "<path to WSL2 shell>"` to your `settings.json` and your `!` commands will use the WSL2 shell. I'll try and make a PR soon.

> Another possible solution, a better one, is to use the integrated terminal.

I investigated this initially, but unfortunately VSCode doesn't offer a public API to get the output from an integrated terminal command, so redirecting output from an integrated terminal into the buffer is not possible without being very hacky. 

They do have a proposed API `vscode.window.onDidWriteTerminalData()`, but turns out published extensions on the VSCode Marketplace [cannot use such proposed APIs](https://code.visualstudio.com/api/advanced-topics/using-proposed-api#sharing-extensions-using-the-proposed-api).

PR #7252, opened 2021-11-12
Refactor `externalCommand.ts` to not use temporary files

tagniam:
Refactors `externalCommand.ts` to use event emitters instead of temporary files to read the command output into the Vim buffer. This is much more cross-platform friendly. Creating temporary files and file paths are OS-dependent, which has made it difficult to build a `vim.shell` setting for #7100... so instead, this PR lets `exec`'s event emitters handle reading stdout/stderr for us. 

No change in behavior as this is an under-the-hood change, and existing tests in `bang.test.ts` and `filter.test.ts` pass.

N/A


J-Fields:
Thank you!

tagniam:
Thanks for the speedy review! PR for #7100 coming up :)

Issue #5858, opened 2020-12-09
Support `:map <silent>`


stevenguh:
The mode display on status bar is overridden by command executed by Vim when a command and the mode display is only returned switch a user switch mode.

For example: if the follow command is executed, it will show the "vspacecode.space" on the status bar, and the current mode will be hidden until the next mode switch.
```json
  "vim.normalModeKeyBindingsNonRecursive": [
    {
      "before": ["<space>"],
      "commands": [ "vspacecode.space"]
    }
]
```
![Screen Shot 2020-11-29 at 4 32 32 AM](https://user-images.githubusercontent.com/1498727/100541919-f8768000-31fb-11eb-9a32-0db836d19dac.jpg)

**Describe the solution you'd like**
The status bar message will timeout and return back to current mode display automatically.

**Describe alternatives you've considered**
Here are some alternatives that I thought of:
1. A setting to disable the display of the commanded by vim.
2. Not display command executed
4. Have a command a third-library can call to reset to mode display
3. A flag in the remapping setting of Vim like 
```json
  "vim.normalModeKeyBindingsNonRecursive": [
    {
      "before": ["<space>"],
      "commands": [{
                    "command": "vspacecode.space",
                    "args": "m",
                    "displayOnStatusBar": false
                }]
    }
]
```
(This seems very verbose if each command has `displayOnStatusBar`)

**Additional context**
This issue was filed in the extension I maintain which relies on VSCode Vim  https://github.com/VSpaceCode/VSpaceCode/issues/137.

I believe this is where vim set the message
https://github.com/VSCodeVim/Vim/blob/119cf137d6bdc968a4f0e150de28dfc1caaba22a/src/configuration/remapper.ts#L527, and I can help to get a PR if we agree on how we should handle it.

J-Fields:
My preference for a first step would be Vim's behavior - no timeout, but as soon as the view scrolls, the command line is cleared. I'm pretty sure it used to work like that in VSCodeVim (at least for error messages), but it's apparently regressed (or maybe I'm misremembering).

The proper way to support this is via [`:map <silent>`](http://vimdoc.sourceforge.net/htmldoc/map.html#:map-%3Csilent%3E). That's probably not too difficult, but it'll require some thought.

Let me know if these will satisfy your use case.

stevenguh:
> My preference for a first step would be Vim's behavior - no timeout, but as soon as the view scrolls, the command line is cleared.

[...]

Issue #5851, opened 2020-12-08
`showmodename` in config should be called `showmode`


J-Fields:
Abbreviated: `smd`

berknam:
Also `timeout` should be called `timeoutlen`, abbreviated to `tm`

Issue #4483, opened 2020-01-09
Migrate from tslint to eslint


J-Fields:
`tslint` is deprecated in favor of `eslint`. We should switch over.

See [this blog post](https://medium.com/palantir/tslint-in-2019-1a144c2317a9) and [this issue](https://github.com/palantir/tslint/issues/4534).
Edit: also [this](https://code.visualstudio.com/api/advanced-topics/tslint-eslint-migration)

PR #7253, opened 2021-11-12
Add silent option to key remappings

mly32:

Adds a silent option to `IKeyRemapping`. 

This is related to [https://github.com/VSCodeVim/Vim/issues/5858](https://github.com/VSCodeVim/Vim/issues/5858). This change uses keybindings defined in `settings.json` rather than the remaps talked about in the issue.

I am still new to the code base so I had a hard time understanding the testing files. I'm assuming tests for this PR will go into either `remapper.test.ts` or `remaps.test.ts`, but I'm not sure how to write the tests. I'm assuming I'd need to compare step results? I'd like some help in getting one test written.

Currently, I'm manually running the extension and verifying the behavior with these settings:

```jsonc
  "vim.insertModeKeyBindingsNonRecursive": [
    // status bar text should go from "-- INSERT --" to "-- NORMAL --"
    // if silent is false, will go to "leaveSnippet"
    { "before": ["<Esc>"], "commands": ["leaveSnippet"], "after": ["<Esc>"], "silent": true }
  ],
    // both should go from "-- NORMAL --" to "lineBreakInsert"
    { "before": ["K"], "commands": ["lineBreakInsert"] },
    { "before": ["J"], "commands": ["lineBreakInsert"], "silent": false }
  ],
```

Todo:

After testing behavior is confirmed, the README needs to be updated. Are there other places that also need to be updated?

Notes:

When I tried using VSCode's debugging to run `Build, Run Extension`, I got the error message `Could not find the task 'gulp: build-dev'`. Is this expected? To go around this, I ran `yarn build`, then switched the debugging task to just `Run Extension`.

The `.vscode/extensions.json` file still recommends TSLint but TSLint is deprecated. Is this project going to migrate to ESLint?

In addition to `silent`, I also wanted to update the current boolean setting `showmodename` to some kind of enum which supports options like `always, normal, never`. Where `normal` and `never` behave like the current `showmodename` and `always` would change `StatusBar.setText` to prepend the text of `statusBarText(vimState)` to the result, essentially always showing the vim mode. 

The problem is that this change would break people's configs so I'd like some thoughts.


mly32:
Any updates on this?

mly32:
@J-Fields, I know you're busy, but are there any updates on this?

J-Fields:
> I am still new to the code base so I had a hard time understanding the testing files

Meh, don't worry about a test case for now (if you could just put a TODO in one of the files you mentioned, that'd be awesome)

> After testing behavior is confirmed, the README needs to be updated. Are there other places that also need to be updated?

Not that I can think of off the top of my head

> When I tried using VSCode's debugging to run Build, Run Extension, I got the error message Could not find the task 'gulp: build-dev'. Is this expected? To go around this, I ran yarn build, then switched the debugging task to just Run Extension.

Not expected (I always do `gulp build-dev` then `Run Extension`, so I hadn't noticed)

> The .vscode/extensions.json file still recommends TSLint but TSLint is deprecated. Is this project going to migrate to ESLint?

#4483

> In addition to silent, I also wanted to update the current boolean setting showmodename to some kind of enum which supports options like always, normal, never. Where normal and never behave like the current showmodename and always would change StatusBar.setText to prepend the text of statusBarText(vimState) to the result, essentially always showing the vim mode.

👍🏻 

> The problem is that this change would break people's configs

We should be able to use a shim (see `overlapSetting`). The setting needs to be migrated anyway (#5851).

J-Fields (changes requested):
Just a TODO for a test to be written in the future and something in the README and this should be good to go. Thanks!

mly32:
Thanks for the feedback. The README has been updated.

My school semester just started so I don't think I can implement `show-mode` anytime soon.

Is the task already marked anywhere as a todo? or should I create an issue and mark it as an enhancement?

mly32:
@J-Fields, are there other changes that need to be added?

J-Fields:
> Is the task already marked anywhere as a todo? or should I create an issue and mark it as an enhancement?

If you could make an issue for it, that would be great

J-Fields (approved):
Thank you!

PR #7255, opened 2021-11-13
Add `vim.shell` setting for custom `!` shell

tagniam:
Adds a `vim.shell` setting to set a custom shell to use for `!` ex commands.

Includes a test suite for Unix-based systems (assuming most are packaged with `sh` and `bash`) but also tested manually on Windows with command prompt and Git Bash.

#7100


README.md
```diff
@@ -404,23 +404,24 @@ Configuration settings that have been copied from vim. Vim settings are loaded i
...
+| vim.maxmapdepth  | Maximum number of times a mapping is done without resulting in a character to be used. This normally catches endless mappings, like ":map x y" with ":map y x". It still does not catch ":map g wg", because the 'w' is used before the next mapping is done. | Number  | 1000                                                           |
+| vim.report       | Threshold for reporting number of lines changed.                                                                                                                                                                                                              | Number  | 2                                                              |
+| vim.shell        | Path to the shell to use for `!` and `:!` commands.                                                                                                                                                                                                           | String  | `/bin/sh` on Unix, `%COMSPEC%` environment variable on Windows |
```
tagniam, line 418: diff looks large here, but just added a row for `vim.shell` and sorted the table.

src/util/externalCommand.ts
```diff
@@ -80,6 +84,8 @@ class ExternalCommand {
   public async run(command: string, stdin: string = ''): Promise<string> {
     command = this.expandCommand(command);
     this.previousExternalCommand = command;
+    // combines stdout and stderr (compatible for all platforms)
+    command += ' 2>&1';
```
J-Fields, line 88: You removed this in your previous PR - is it necessary?

tagniam: Yeah it's necessary, I previously removed it because I thought the child process pipes would maintain the correct ordering of stdout/stderr, but they don't - so it's best to redirect both to stdout to get the right order.

J-Fields:
Thank you!

Issue #4017, opened 2019-08-29
Ctrl-P and Ctrl-N can‘t work in the latest version


s97712:
**Describe the bug**
Ctrl-P and Ctrl-N can‘t work in the latest version

**To Reproduce**

1.  Go to '...'
2.  Click on '....'
3.  Scroll down to '....'
4.  See error

**Expected behavior**

**Screenshots**
If remapping-related, please attach log output: https://github.com/VSCodeVim/Vim#debugging-remappings.

**Environment (please complete the following information):**

- Extension (VsCodeVim) version: Vim 1.10.0
- VSCode version: 1.36.1
- OS:

**Additional context**

qingtao:
the key Ctrl+n and Ctrl+p in path: C:\Users\tom\.vscode\extensions\vscodevim.vim-1.10.0

```json
{
    "key": "ctrl+n",
    "command": "extension.vim_ctrl+n",
    "when": "(editorTextFocus || vim.mode == 'SearchInProgressMode' || vim.mode == 'CommandlineInProgress') && vim.active && vim.use<C-n> && !inDebugRepl"
}````

```json
{
    "key": "ctrl+p",
    "command": "extension.vim_ctrl+p",
    "when": "(suggestWidgetVisible || vim.mode == 'SearchInProgressMode' || vim.mode == 'CommandlineInProgress') && vim.active && vim.use<C-p> && !inDebugRepl"
}
```

then look them in "Keyboard Shorcuts":
```json
{
  "key": "ctrl+n",
  "command": "extension.vim_ctrl+n",
  "when": "(editorTextFocus || vim.mode == 'SearchInProgressMode' || vim.active && vim.use<C-n> && !inDebugRepl && vim.mode == ''CommandlineInProgress')'"
}
```
```json
{
  "key": "ctrl+p",
  "command": "extension.vim_ctrl+p",
  "when": "(suggestWidgetVisible || vim.mode == 'SearchInProgressMode' || vim.active && vim.use<C-p> && !inDebugRepl && vim.mode == ''CommandlineInProgress')'"
}
```

J-Fields:
Duplicate of #4011

nelsyeung:
I just want to say that while Ctrl-N works from #4011, Ctrl-P still doesn't work, so this issue is valid.

ITenthusiasm:
[...]

PR #7261, opened 2021-11-17
Change "when" for ctrl+p key bind in package.json

grosssoftware:
Fixes when the ctrl+p key bind is used to enable those who want to remap it. Changing from suggestWidgetAvailable to editorTextFocus allows the key bind to work in either mode. With this change, ctrl+p will work the same as ctrl+n which was fixed some time ago.

None


grosssoftware:
This has been open for almost a year. Any chance this can be merged?

J-Fields:
Sorry this slipped through the cracks; thanks!

tamird:
This broke the ability to open files with ctrl-p? Seems reminiscent of https://github.com/VSCodeVim/Vim/issues/4017.

bradengc-github:
> This broke the ability to open files with ctrl-p? Seems reminiscent of #4017.

I was able to work around this by adding an entry for <C-p> in `vim.handleKeys' in my user settings. e.g.:

image

kendallroth:
I had to do the same thing with `<C-n>` to allow it to open new files, instead of simply moving cursor down a line 🤷

HengHuH:
> > This broke the ability to open files with ctrl-p? Seems reminiscent of #4017.
> 
> I was able to work around this by adding an entry for in `vim.handleKeys' in my user settings. e.g.:
> 
> image

It is make vim action 'navigation previous in pop-up menu' disable.

Issue #7274, opened 2021-11-23
Typing speed is very slow


KahnDepot:
The vim plug-in will slow down the performance of vscode, which is very serious in insert mode.

As shown in the figure below, this is the delay when I repeat the keystrokes after opening the vim plug-in. When my mouse starts to move, my hands have left the keyboard, but it still hasn't finished typing:

![vscode_delay](https://user-images.githubusercontent.com/81478335/142991070-52ec7062-434a-48ee-8d78-41404ecc9b40.gif)

If I disable the vim plugin, the speed increase of vscode will be obvious:

![vscode_delay2](https://user-images.githubusercontent.com/81478335/142991450-9e6770a6-bf10-4c24-b063-8b3afd14ef12.gif)

Is there any good solution to avoid this situation? this It greatly affected my programming speed！

marcwimmer:
I have the same issue.

gruvw:
same problem here

mateusdcc:
I hope this gets fixed, loosing my overall productivity because of that

MurtadhaInit:
This is the first thing I've noticed immediately after installing the plugin.
What's bizarre is that it seems to randomly stop for no reason mid navigation. E.g. say you're holding the h key in normal mode or simply just pressing it not holding it: sometimes it just won't register for some reason. But 80% of the times it does. It's a bizarre behaviour.

trungdvu:
Same issue here. It works fine on my old mac

imbrn:
Same problem here

smiglo:
Same here
_quite a lot of time passed since 2021 :(_

Issue #7270, opened 2021-11-20
search in explorer and then use j/k/h/k/n/N to move


XiaoXiaoLui:
**Is your feature request related to a problem? Please describe.**
setting workbench.list.keyboardNavigation=highlight
After using / to search in file explore, I can only use arrow keys to move cursor which is awkward for a vim user.

**Describe the solution you'd like**
using / to search in file explore and press enter to confirm the highlighted matching text, after that I can use n/N/j/k/h/l to move like normal vim.

**Describe alternatives you've considered**
Other ways I can jump to each matches conviniencely.

Issue #7262, opened 2021-11-18
v1.21.10 not work


wangduanduan:
**Describe the bug**

after install v1.21.10, nothing happen, all key bind not work。so i change to v1.11.3， all vim key bind work

**To Reproduce**

1.  install latest v1.21.10
2.  click some file
3.  nothing work

**Expected behavior**
all key bind work

**Screenshots**
If remapping-related, please attach log output: https://github.com/VSCodeVim/Vim#debugging-remappings.

**Environment (please complete the following information):**

Version: 1.61.2
Commit: 6cba118ac49a1b88332f312a8f67186f7f3c1643
Date: 2021-10-19T14:58:13.605Z
Electron: 13.5.1
Chrome: 91.0.4472.164
Node.js: 14.16.0
V8: 9.1.269.39-electron.0
OS: Linux x64 5.10.60-amd64-desktop

Issue #7245, opened 2021-11-10
Unexpected behavior with `vito`


spigelli:
### **Describe the bug**
Unexpected behavior with `vito`

### **To Reproduce**

Cursor location marked with [ ]
Selection marked with ` `
Newline marked with ↵

> \<div>↵
> \<span>text\</span>↵
> \</di[v]>↵

Press `vit`
*Note this selection isn't right either, see #6555*

> \<div`>↵`
>   `<span>text\</span>↵`
> []\</div>↵

Press `o`

> `<div>↵`
> []\<span>text\</span>↵
> \</di[v]>↵

### **Expected behavior**

> \<div>↵
> \<span>text\</span>↵
> \</di[v]>↵

Press `vit`

> \<div>`↵`
>   `<span>text\</span>[↵]`
> \</div>↵

Press `o`

> \<div>`[↵]`
> `<span>text\</span>↵`
> \</di[v]>↵

### **Environment**

- Extension (VsCodeVim) version: 1.21.5
- VSCode version: 1.54.3
- OS: MacOS and Ubuntu

PR #7284, opened 2021-11-25
Add a structured bug report template

max-sixty:

I see there are 1100 issues on this repo, and that @J-Fields does much of the triaging as well as the development. Would a more structured issue log template increase the quality of issues? I see a mixed bag out there — e.g. https://github.com/VSCodeVim/Vim/issues/7274, https://github.com/VSCodeVim/Vim/issues/7270, https://github.com/VSCodeVim/Vim/issues/7262, https://github.com/VSCodeVim/Vim/issues/7223 would probably be more helpful to both the reporter and the project if they added more detail.

A few ideas contained here:
- Use the new structured yaml format for issues
- Ask people to fill out a "machine-readable" testcase where possible. 
  - This is probably a minority of issues but some recent ones where this would help are https://github.com/VSCodeVim/Vim/issues/7214, https://github.com/VSCodeVim/Vim/issues/7245, https://github.com/VSCodeVim/Vim/issues/7159, https://github.com/VSCodeVim/Vim/issues/7213, https://github.com/VSCodeVim/Vim/issues/7204, https://github.com/VSCodeVim/Vim/issues/7168, https://github.com/VSCodeVim/Vim/issues/7095) . 
  - As well as removing ambiguity, it's also a testcase that development can start from — a way to shift burden from maintainers & contributors to the original reporter.
  - Another option would be to split this into a different type of bug report
- Ask people to confirm the bug isn't a duplicate

We don't have to take all of these at once if some of them aren't good ideas.

[Rendered version](https://github.com/max-sixty/Vim/issues/new?assignees=&labels=&template=bug_report.yaml)


max-sixty:
@J-Fields let me know if something like this would be of interest — no problem if not!

Issue #4844, opened 2020-05-13
Fix tests that fail locally on Windows


J-Fields:
I've been ignoring these for too long

Issue #4079, opened 2019-09-16
Test suite fails in various scenarios on Windows


caleywoods:
**Describe the bug**
When trying to contribute to VSCode VIM it's advised to write tests and make sure that they pass before submitting your PR. To execute the test suite a contributor can either run `npm test` to test locally or `gulp test` to build a docker container and execute the test suite within a linux environment.

The issue happens when a potential contributor is using Windows as the host OS and manages to do the following (or vice-versa):

* Run `npm test` for the first time, this will likely download a stable Windows build of VSCode into `.vscode-test`.

* Attempt to run `gulp test` to execute the suite in the Linux container, this will result in an `ENONT` error since the download script sees a `.vscode-test/vscode-n.nn.n` directory and assumes VSCode is downloaded and good to go.

**To Reproduce**

1.) On a Windows host, run cmd prompt as Administrator (required to run test suite locally)
2.) Clone the VSCode VIM repo and `cd` into it
3.) Execute `gulp build` and then `npm test`
4.) Attempt to run `gulp test`

**Expected behavior**
`vscode-test` should make an effort to make sure that a stable version of VSCode that matches `process.platform` is downloaded. This prevents any foul-up during executing the test suite. 

**Environment (please complete the following information):**

- OS: Windows 10 x64

**Additional context**
I think this makes sense to create as an issue on the upstream `vscode-test` repo where a fix would make the most sense. I am just opening this here because I think VSCode VIM can implement a stop-gap in the meantime by checking the error code and giving the end-user some feedback on steps to take to remedy the issue.

I have tested a "fix" locally by checking `data.code === ENONT` within `cmd.on('error', ()=>{});` and then console.logging out a helpful message with a yellow background. The message tells the user to try deleting the `.vscode-test` directory and re-running the command they just tried, this helps in both cases whether the user-tested with docker first (and downloaded a Linux version of VSCode) or whether they tried `npm test` first  (and downloaded VSCode for Windows).

caleywoods:
Current Workaround for `ENONT` when running the test suite: Delete `.vscode-test` directory and re-run whichever command you were attempting to run.

PR #7293, opened 2021-12-01
Silence failing tests on Windows and add Windows build step

tagniam:
While working on the bang/filter command feature, I've noticed that quite a few tests are failing only on Windows. This PR proposes we temporarily silence those tests on Windows and document them with TODO comments, which I hope should be helpful towards fixing #4844 (and #4079).

Additionally, I propose we add a Windows build step to our Github Actions workflow to encourage cross-platform compatibility for future PRs. However, this may affect this repo's Github Actions minutes limit - [jobs that run on Windows consume minutes at twice the rate of Linux-based jobs](https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions#:~:text=Jobs%20that%20run%20on%20Windows%20and,10%2C000%20minutes%20included%20in%20your%20account.). This multiplier means that this would probably cause the minutes consumption to triple. I'm not sure what the minutes limit is for VSCodeVim nor if we currently run near that limit, so if this would cause problems, I can omit this change. Lemme know your thoughts!

N/A, but progresses towards fixing #4844 and #4079.


test/jumpTracker.test.ts
```diff
@@ -358,19 +369,22 @@ suite('Record and navigate jumps', () => {
...
-      newJumpTest({
+      // TODO: this fails on Windows
+      newJumpTestSkipOnWindows({
         start: ['|start', 'a1', 'a2', 'a3', 'a4', 'a5', 'end'],
         keysPressed: '/^\nnnnkdd',
         end: ['start', 'a1', 'a2', '|a4', 'a5', 'end'],
         jumps: ['start', 'a1', 'a2', 'a4'],
       });
-      newJumpTest({
+      // TODO: this fails on Windows
```
J-Fields, line 386: Can you make these `// TODO(#4844): this fails on Windows` so it's associated with the right ticket?

J-Fields:
Thanks for the PR! I'm definitely supportive of this idea. We should do something similar for the web version as well at some point (maybe Mac as well, but it seems that uses 10x the minutes).

Any idea why this last step won't finish? 
image

> I'm not sure what the minutes limit is for VSCodeVim nor if we currently run near that limit

I think the free limit is 2000 minutes/month - I don't think we're anywhere near that. If this causes issues, we can always change to run on all platforms only for releases, or something like that.

test/jumpTracker.test.ts
tagniam: Done!

tagniam:
> I think the free limit is 2000 minutes/month - I don't think we're anywhere near that. If this causes issues, we can always change to run on all platforms only for releases, or something like that.

Sounds good!

> Any idea why this last step won't finish?

Not sure, the other two steps are fine :thinking: I've tried closing/reopening the PR, force pushing and even changing the base branch. Going to try and open another PR to see if this is a fluke.

tagniam:
Sorry for the delay! From what I can tell, the last `build` step is forever waiting because it comes from the existing enforced status check rule - it is applying to this branch as well, but since the rule was updated and no longer exists on this branch, it's waiting forever.

You might need to merge this in first and then modify the required status checks in this repo's `Settings > Branches > Require status checks to pass before merging` to these new `build / build (ubuntu-latest)` and `build / build (windows-latest)` checks. Related: https://circleci.com/docs/2.0/enable-checks/#troubleshooting-github-checks-waiting-for-status-in-github

---
One thing I'd like to bring up is that it seems running tests on Windows is a bit flakier, sometimes the Windows build fails while the ubuntu build succeeds, in which a re-run usually fixes it. If that is a concern we can shelve the Windows build for now, or perhaps make it a non-required check for PRs. Thoughts?

J-Fields:
> Sorry for the delay!

No trouble whatsoever, I haven't been active on this project at all recently.

> You might need to merge this in first and then modify the required status checks

Sounds good, I'll give that a shot

> One thing I'd like to bring up is that it seems running tests on Windows is a bit flakier, sometimes the Windows build fails while the ubuntu build succeeds, in which a re-run usually fixes it. If that is a concern we can shelve the Windows build for now, or perhaps make it a non-required check for PRs. Thoughts?

Yeah good point. Non-required is probably the way to go for now

J-Fields:
Seems to have worked, thanks very much!

Issue #7359, opened 2021-12-31
Vimrc file won't reload after saving file with vimrcPath's home environment variable


waynewaynetsai:
**Describe the bug**

Vimrc file won't reload after saving file with vimrcPath's home environment variable

**To Reproduce**

1.  Set `$HOME/_vimrc` to vimrc paths 

(or any path with $HOME variable, Ex. `$HOME/testdir/_vimrc`)

![](https://i.imgur.com/ZKOM9Kf.png)

2.  Add a _vimrc file under home directory

3.  Write any keybinding configs, save file

4.  `Sourced new .vimrc` message won't display

**Expected behavior**

Should display `Sourced new .vimrc` message then reload `_vimrc` file

![](https://i.imgur.com/lOuVgB5.png)

**Environment (please complete the following information):**

- Vim  Extension:  v1.21.10
- MacOS Mojave: 10.14.6/ win10

Issue #7358, opened 2021-12-31
vim.editVimrc command failed with vimrcPath's  home environment variable


waynewaynetsai:
**Describe the bug**

`Vim: Edit .vimrc` command failed with vimrcPath's  home environment variable

**To Reproduce**

1.  Set `$HOME/_vimrc` to vimrc path

(or any path with $HOME variable, Ex. `$HOME/testdir/_vimrc`)

![](https://i.imgur.com/ZKOM9Kf.png)

2.  Add _vimrc file under home directory

3.  Execute `Vim: Edit .vimrc` command 

4.  See error messages at output channel `Log(Window)`

![](https://i.imgur.com/kYwWJ6v.png)

**Expected behavior**

Should open _vimrc file under home directory directly

**Environment (please complete the following information):**

- Vim  Extension:  v1.21.10
- MacOS Mojave: 10.14.6/ win10

PR #7360, opened 2021-12-31
Fix .vimrc file's issues with vimrcPath's home environment variable

waynewaynetsai:

Fix .vimrc file's issues with vimrcPath's home environment variable

- #7359 
- #7358

Reference: 

- https://github.com/VSCodeVim/Vim/blob/master/src/configuration/iconfiguration.ts#L343


J-Fields:
Thank you!

PR #7376, opened 2022-01-05
Disable other extensions while running tests for avoiding unexpected side effect

waynewaynetsai:

- Disable other extensions while running tests for avoiding unexpected side effect at local environment

- Reference: [Disabling other extensions while debugging](https://code.visualstudio.com/api/working-with-extensions/testing-extension#disabling-other-extensions-while-debugging)


J-Fields:
Did this cause an issue for you?

Issue #5789, opened 2020-11-30
Update README to include key repeat instructions for Mac Exploration builds


edwinhollen:
Hi, I'm using Mac VS Code Exploration builds temporarily, and I discovered that the set of "enable key-repeating" commands found in the README only accommodates `com.microsoft.VSCode` (stable), `com.microsoft.VSCodeInsiders` (Insider), and `com.visualstudio.code.oss` (Codium), and does not currently include the necessary command to enable this helpful behavior for Exploration users. 

**Expected behavior**
The instructions should include this command to enable key repeat for VS Code Exploration users:

`defaults write com.microsoft.VSCodeExploration ApplePressAndHoldEnabled -bool false`

(This is the CFBundleIdentifier found in the `Visual Studio Code - Exploration.app/Contents/Info.plist` file) 🙂

AMMAR-62:
I'll take up this issue ... if it's open

PR #7408, opened 2022-01-20
added enable key-repeating doc for Codium Exploration Users 

AMMAR-62:
Documentation for exploration users of VS Codium for apple users.
#5789


J-Fields:
Thank you!

Issue #6656, opened 2021-05-23
Show previews for `:s` replacements


The-Compiler:
**Is your feature request related to a problem? Please describe.**
I'm coming from Spacemacs/Doom Emacs (after switching there from vim), and I really miss the preview it shows on `:s/.../...` commands while typing the replacement. Especially for regexes, this can be very useful.

**Describe the solution you'd like**
Something like this (implemented by evil-mode in emacs):

![image](https://user-images.githubusercontent.com/625793/119263108-0111a700-bbde-11eb-8cc9-0674b695a142.png)

**Describe alternatives you've considered**
Doing the replacement, checking manually if it worked or not, undoing and doing it again, etc.

**Additional context**
Looks like [vim-over](https://github.com/osyo-manga/vim-over) is a plugin for vim doing something similar.

J-Fields:
Related: #4934

J-Fields:
nvim has this built-in, see `:help inccommand`

Issue #6655, opened 2021-05-23
ex-commandline renders icon codes


The-Compiler:
**Describe the bug**
When e.g. typing `:s/$(rocket)`, that shows up as `:s/🚀`

**Expected behavior**
The text would be escaped in some way so that VS Code [doesn't interpret it as icon](https://code.visualstudio.com/api/references/icons-in-labels). I'm not sure if that's possible, though...

**Screenshots**
![image](https://user-images.githubusercontent.com/625793/119262654-40d78f00-bbdc-11eb-93d5-96badfaa2a10.png)

**Environment (please complete the following information):**

- Extension Version: `1.20.3`
- OS Version: `Linux x64 5.12.3-arch1-1`
- VS Code version: `1.56.2`

J-Fields:
@The-Compiler Could you make an issue in the VSCode repo asking if there's a way to do this? I'm not aware of one.

The-Compiler:
Okay, so it wasn't just me who couldn't figure out how to do this... Done: https://github.com/microsoft/vscode/issues/124703

The-Compiler:
Ah, looks like doing e.g. `\$(rocket)` works. A quick attempt at fixing this:

```ts
diff --git i/src/statusBar.ts w/src/statusBar.ts
index 47962738..42379355 100644
--- i/src/statusBar.ts
+++ w/src/statusBar.ts
@@ -149,6 +149,10 @@ class StatusBarImpl implements vscode.Disposable {
 
 export const StatusBar = new StatusBarImpl();
 
+function escapeIcons(text: string) {
+  return text.replace(/\$\([a-zA-Z0-9~-]+\)/g, '\\$&');
+}
+
 export function statusBarText(vimState: VimState) {
   const cursorChar =
     vimState.recordedState.actionKeys[vimState.recordedState.actionKeys.length - 1] === '<C-r>'
@@ -184,12 +188,12 @@ export function statusBarText(vimState: VimState) {
       const leadingChar =
         globalState.searchState.searchDirection === SearchDirection.Forward ? '/' : '?';
 
-      const searchWithCursor = globalState.searchState.searchString.split('');
+      const searchWithCursor = escapeIcons(globalState.searchState.searchString).split('');
       searchWithCursor.splice(vimState.statusBarCursorCharacterPos, 0, cursorChar);
 
       return `${leadingChar}${searchWithCursor.join('')}`;
     case Mode.CommandlineInProgress:
[...]

PR #7416, opened 2022-01-23
Implement `inccommand`

adrsm108:
This PR introduces the option `inccommand`, which enables a live preview of `:s` commands as they are typed. It also makes some improvements to interactive substitution with the `c` flag. 

The feature is configured by setting `vim.inccommand` to `'append'` or `'replace'` (or `''` to disable).
When set to `'append'`, substitution previews are inserted after each highlighted match, a behavior found in some popular vim emulators like spacemacs.
When set to `'replace'`, the substitution previews replace each match, which is consistent with the default behavior of neovim.

The config keys `vim.substitutionColor` and `vim.substitutionTextColor` are provided to control the appearance of preview text. 

#6656
Also addresses #6655

Probably the biggest change from the original neovim behavior is how multi-line substitutions are handled. Since decorations can't change the apparent number of lines in the editor, concealed multi-line matches still have line breaks that won't be present once the substitution is performed. Multi-line replacements also have to be displayed inline; I currently do this by replacing line breaks with the character `⏎` (U+23CE, "RETURN SYMBOL"), but it might be more sensible to use standard escape sequences.


README.md
```diff
@@ -131,6 +131,8 @@ These settings are specific to VSCodeVim.
...
 | vim.searchMatchTextColor         | Foreground color of current search match                                                                                                                                                                                                                                                                                                                                                                                            | String  | None                                      |
+| vim.substitutionColor            | Background color of substituted text when `vim.inccommand` is enabled                                                                                                                                                                                                                                                                                                                                                               | String  | "#50f01080"                               |
```
J-Fields, line 134:
```suggestion
| vim.substitutionColor            | Background color of substitution text when `vim.inccommand` is enabled                                                                                                                                                                                                                                                                                                                                                               | String  | "#50f01080"                               |
```

Maybe? This is the text that is doing the substituting, not being substituted, right?

README.md
```diff
@@ -423,6 +425,7 @@ Configuration settings that have been copied from vim. Vim settings are loaded i
...
 | vim.ignorecase   | Ignore case in search patterns                                                                                                                                                                                                                                | Boolean | true                                                           |
 | vim.incsearch    | Show the next match while entering a search                                                                                                                                                                                                                   | Boolean | true                                                           |
+| vim.inccommand   | Show the results of a `substitute` command while typing.                                                                                                                                                                                                      | String  | `replace`                                                      |
```
J-Fields, line 428:
```suggestion
| vim.inccommand   | Show the results of the `:substitute` command while typing                                                                                                                                                                                                      | String  | `replace`                                                      |
```

adrsm108:
I'm not super familiar with github actions, but is this check working as intended? The only changes were to the README, and things seem to be going wrong when it's reformatted.

J-Fields (approved):
Thanks a lot!

adrsm108:
My pleasure!

Swoorup:
I have been using this plugin for quite sometime, and didn't realise this feature had landed until I saw the preview when using the replace command. Just came to say thank you. 🙏 😊 🤩

PR #7418, opened 2022-01-23
add operation for generating a console log of targeted text object

osandell:
Hi!

When coding I find myself using the snippet `clo` from [vscode-react-javascript-snippets](https://github.com/dsznajder/vscode-react-javascript-snippets) all the time. It pastes `console.log("<cursor>", <cursor>)` and thereby let you log a specific object (or array, variable etc...) while in the same time log the name of that object.

Having the double cursor helps speeding up the process since you typically will type simply the object name. However, autocomplete won't work since we're typing a string. This is a major hassle since a lot of times I want to type something like `myObject1.myObject2.myPropertyWithLongName`.

I therefore forked vscodevim and made a new operator that I call ”logObject”. It's mapped to `'<leader>l’` and lets you target a text object and then console logs that object on a new line below, in a similar fashion as the snippet mentioned above.

For example, I can place the cursor within the brackets of this line:

`headerImage = isImageEmpty(page.headerImage) ? null : page.headerImage;`

and then type this motion:

`'<leader>', 'l', 'i', '('`

It will then create a new line below with the following:

`console.log("page.headerImage", page.headerImage);`

The default template of the log message is for JavaScript, but this can be modified by setting `vim.logObjectTemplate` to a message of choice your.

There is no specific issue that it fixes.

This fork is made primarily because it speeds up the workflow for me personally. But I think other people might benefit from it as well, especially since it's possible to customize the log message. Please let me know if you think it would suit the project. I understand if you might want to keep to a minimal amount of well established plugins though.

### DIFF
--- a/package.json
+++ b/package.json
@@ -615,6 +615,16 @@
           "description": "Ignore case in search patterns.",
           "default": true
         },
+        "vim.logObject": {
+          "type": "boolean",
+          "markdownDescription": "Generate console log of targeted text object.",
+          "default": false
+        },
+        "vim.logObjectTemplate": {
+          "type": "string",
+          "markdownDescription": "The template for the console message. {object} will be replaced by the text object in the generated output. Default template is in JavaScript format. Please replace it if using another language.",
+          "default": "console.log(\"{object}\", {object});"
+        },
         "vim.smartcase": {
           "type": "boolean",
           "markdownDescription": "Override the `ignorecase` option if the search pattern contains upper case characters.",
--- a/src/actions/include-all.ts
+++ b/src/actions/include-all.ts
@@ -2,7 +2,6 @@ import './base';
 import './operator';
 import './motion';
 import '../textobject/textobject';
-
 // commands
 import './commands/insert';
 import './commands/replace';
@@ -11,11 +10,11 @@ import './commands/commandLine';
 import './commands/search';
 import './commands/put';
 import './commands/digraphs';
-
 // plugin
 import './plugins/camelCaseMotion';
 import './plugins/easymotion/easymotion.cmd';
 import './plugins/easymotion/registerMoveActions';
+import './plugins/logObject';
 import './plugins/sneak';
 import './plugins/replaceWithRegister';
 import './plugins/surround';
--- a/src/actions/include-plugins.ts
+++ b/src/actions/include-plugins.ts
@@ -2,6 +2,7 @@
 import './plugins/camelCaseMotion';
 import './plugins/easymotion/easymotion.cmd';
 import './plugins/easymotion/registerMoveActions';
+import './plugins/logObject';
 import './plugins/sneak';
 import './plugins/replaceWithRegister';
 import './plugins/surround';
--- a/src/actions/plugins/logObject.ts
+++ b/src/actions/plugins/logObject.ts
@@ -1,0 +1,54 @@
+import * as vscode from 'vscode';
+
+import { configuration } from '../../configuration/configuration';
+import { BaseOperator } from '../operator';
+import { Mode } from '../../mode/mode';
+import { Position } from 'vscode';
+import { RegisterAction } from '../base';
+import { TextEditor } from './../../textEditor';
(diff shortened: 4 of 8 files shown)


elazarcoh:
I would love to have something like that, but a bit more generic. 

I might consider moving it into a separate extension command, and just use VSCodeVim keybinding to execute those commands.
what I'd like to have is a way to use it with multiple snippets. e.g.

`<leader>l` will do `print(|)`, and all the goodies with text objects.

`<leader>m` will do `|.mkdir(exist_ok=True)`

Basically, I want to be able to take any snippet and use the vim text objects, the the PR does.

I'm not sure if my suggestion is already possible with VSCodeVim (maybe an extension that takes a selection and insert it into a snippet already exists. worth searching. In that case, we probably can use the existing functionality of VSCodeVim to get the same effect).

osandell:
@elazarcoh 

Thanks for the input. Could you please elaborate a bit, I'm not quite sure I follow what functionality you are after?

> \<leader\>l will do print(|), and all the goodies with text objects.

When you say "all the goodies with text objects", do you mean that you want to be able to put your cursor over let's say the word `cat` and do `'<leader>', 'l', 'i', 'w'`, and that would produce `print(cat)`?

In that case that functionality is already implemented in this PR. You would just go to settings.json and put `vim.logObjectTemplate = print({object})`

And as for other types of snippets... For me personally I find that most snippets work pretty good in VS Code. Like for example `rfc` will create a React Functional Component and let me enter a name for it. I don't see the need for vim in these kind of circumstances. But I'm not sure I'm following your second example with mkdir. Do you mean that `'<leader>m'` would just type out `|.mkdir(exist_ok=True)` and put your cursor in front of it.. or do you mean that you should be able to target a specific text object?

elazarcoh:
Yes, that what I meant about "goodies".

So, for the other snippet I mantioned, I would like to have the same functionality, mainly the "goodies" part. Just stand on a directory variable, and `<leader>miw` and it'll add the snippet I mentioned with the variable name already in place.

osandell:
@elazarcoh Got you. Yes that's a good point.. I'll look into that!

J-Fields:
Very sorry, but I'll decline this PR because I'd like to limit the plugins we explicitly support to de-facto "standard" ones which are repeatedly requested.

With a bit better vimscript support (which is planned, but I'm not sure when/if I'll have time to do), this should be doable without any explicit support. With some clever remaps you may even be able to do it now. It seems very cool and useful though! I may have to add something like this to my neovim config...

Thank you for the PR!

elazarcoh:
@osandell 
found out it's already possible!
```json
        {
            "before": [
                "<leader>",
                "l"
            ],
            "after": [
                "y",
                "<Esc>",
                "o",
            ],
            "commands": [
                {
                    "command": "editor.action.insertSnippet",
                    "args": {
                        "snippet": "console.log('$CLIPBOARD');"
                    }
                }
            ]
        }
    ],
```

osandell:
> Very sorry, but I'll decline this PR because I'd like to limit the plugins we explicitly support to de-facto "standard" ones which are repeatedly requested.
> 
> With a bit better vimscript support (which is planned, but I'm not sure when/if I'll have time to do), this should be doable without any explicit support. With some clever remaps you may even be able to do it now. It seems very cool and useful though! I may have to add something like this to my neovim config...
> 
> Thank you for the PR!

Alright, no problem. Makes sense to keep some restraints :-)

osandell:
> @osandell found out it's already possible!
> 
> ```json
> "vim.visualModeKeyBindingsNonRecursive": [
>         {
>             "before": [
>                 "<leader>",
>                 "l"
>             ],
>             "after": [
>                 "y",
>                 "<Esc>",
>                 "o",
>             ],
>             "commands": [
>                 {
>                     "command": "editor.action.insertSnippet",
>                     "args": {
>                         "snippet": "console.log('$CLIPBOARD');"
>                     }
>                 }
>             ]
>         }
>     ],
> ```

@elazarcoh Oh, that's a really cool find... I'll use that instead then :-D

PR #7450, opened 2022-02-03
Fixed a bug where insertLineAbove would leave extra whitespace.

half-potato:
After some more testing by myself, I noticed a case where VSCode would leave extra whitespace. I have fixed this bug and added a unit test for this case.


src/actions/commands/actions.ts
```diff
@@ -1536,28 +1536,34 @@ export class CommandInsertNewLineAbove extends BaseCommand {
     }
 
     vimState.cursors = getCursorsAfterSync(vimState.editor);
-    const indentAmt = Math.max(charPos - vimState.cursors[0].start.character, 0);
+    const endPos = vimState.cursors[0].start.character;
+    const indentAmt = charPos - endPos;
+    console.log(charPos);
+    console.log(indentAmt);
```
J-Fields, line 1542:
```suggestion
```

J-Fields (approved):
Thank you!

PR #7467, opened 2022-02-08
Support `:breaka[dd]`, `:breakd[el]`, and `:breakl[ist]`

### LINKED ISSUE #7186: Support `:breaka[dd]`, `:breakd[el]`, and `:breakl[ist]`
These allow you to interact with breakpoints

elazarcoh:
It adds support for interacting with breakpoints.

fixes #7186


src/cmd_line/commands/breakpoints.ts
```diff
@@ -0,0 +1,271 @@
...
+    });
+    return vscode.window.showQuickPick(lines).then((selected) => {
+      if (selected) {
+        const [idWithHashtag] = selected.split('\t');
+        const id = parseInt(idWithHashtag.replace('#', ''), 10) - 1;
+        const breakpoint = breakpoints[id] as AnyBreakpoint;
+        if (breakpoint && breakpoint.location !== undefined) {
+          return vscode.window.showTextDocument(breakpoint.location.uri).then(
+            () => {
+              vimState.cursorStartPosition = breakpoint.location.range.start;
```
elazarcoh, line 200: @J-Fields couldn't get it to work while in the same file... Jumping to another file seems to be working.

src/cmd_line/commands/breakpoints.ts
```diff
@@ -0,0 +1,271 @@
...
+}
+
+/**
+ * Add Breakpoint Command
+ */
+type AddBreakpointExpr = { expr: string };
+type AddBreakpointFunction = { function: string };
+type AddBreakpointFile = { line: number; file: string };
+type AddBreakpointHere = 'here';
+type AddBreakpoint =
```
J-Fields, line 22: I'd prefer a discriminated union here (with a `type` string enum). I think
```typescript
if (this.addBreakpoint.type === 'file') {...}
```
is cleaner than custom type guards in cases like this

src/cmd_line/commands/breakpoints.ts
```diff
@@ -0,0 +1,271 @@
...
+
+/**
+ * Delete Breakpoint Command
+ */
+type DelBreakpointById = { id: number };
+type DelAllBreakpoints = 'all';
+type DelBreakpointFunction = { function: string };
+type DelBreakpointFile = { line: number; file: string };
+type DelBreakpointHere = 'here';
+type DelBreakpoint =
```
J-Fields, line 89: Discriminated union, as above

src/cmd_line/commands/breakpoints.ts
```diff
@@ -0,0 +1,271 @@
...
+}
+
+/**
+ * List Breakpoints Command
+ */
+class ListBreakpointsCommand extends ExCommand {
+  public override isRepeatableWithDot: boolean = false;
+
+  async execute(vimState: VimState): Promise<void> {
+    type AnyBreakpoint = vscode.Breakpoint & vscode.SourceBreakpoint & vscode.FunctionBreakpoint;
```
J-Fields, line 175: Isn't this type equivalent to `vscode.Breakpoint`?

src/cmd_line/commands/breakpoints.ts
```diff
@@ -0,0 +1,271 @@
...
+      } = b as AnyBreakpoint;
+      let entry = '';
+      entry += `#${i + 1}\t`;
+      entry += enabled ? '$(circle-filled)\t' : '$(circle-outline)\t';
+      entry += condition ? '$(debug-breakpoint-conditional)\t' : '\t';
+      entry += location ? `${path.basename(location.uri.fsPath)}:${location.range.start.line + 1}` : '';
+      entry += functionName ? `$(debug-breakpoint-function)${functionName}` : '';
+      return entry;
+    });
+    return vscode.window.showQuickPick(lines).then((selected) => {
```
J-Fields, line 192: Is there a reason we don't want to use `await` here?

src/cmd_line/commands/breakpoints.ts
```diff
@@ -0,0 +1,271 @@
...
+        const id = parseInt(idWithHashtag.replace('#', ''), 10) - 1;
+        const breakpoint = breakpoints[id] as AnyBreakpoint;
+        if (breakpoint && breakpoint.location !== undefined) {
+          return vscode.window.showTextDocument(breakpoint.location.uri).then(
+            () => {
+              vimState.cursorStartPosition = breakpoint.location.range.start;
+            },
+          );
+        } else {
+          return;
```
J-Fields, line 204: I think both of these `else` blocks are unnecessary?

src/cmd_line/commands/breakpoints.ts
J-Fields: Hmm... try setting `vimState.cursorStopPosition`?

J-Fields (changes requested):
Thank you! Left a handful of mostly minor comments - I'll need to do another read and some manual testing, but this is looking good.

src/cmd_line/commands/breakpoints.ts
elazarcoh: I've returned the promise, so it's essentially the same. But I'll do it, as it seems you prefer it this way. 👍

src/cmd_line/commands/breakpoints.ts
elazarcoh: Works!

elazarcoh:
Hi,
I resolved all of the issues, I think.

As a general workflow, who should click "Resolve conversation"?

src/cmd_line/commands/breakpoints.ts
elazarcoh: not exactly. but I made it clearer now

J-Fields:
> As a general workflow, who should click "Resolve conversation"?

You can resolve them most of the time - if there's something you're still unsure about (or think I might be unsure about, like if you disagree with a change I suggested), leave it unresolved

PR #7471, opened 2022-02-09
Tab completion of file names is case insensitive on Windows

### LINKED ISSUE #7160: Tab completion of file names should be case insensitive on Windows

elazarcoh:
Tab completion of file names should be case insensitive on Windows

fixes #7160 

### DIFF
--- a/src/actions/commands/commandLine.ts
+++ b/src/actions/commands/commandLine.ts
@@ -115,9 +115,14 @@ class CommandLineTab extends CommandLineAction {
         isRemote,
         shouldAddDotItems
       );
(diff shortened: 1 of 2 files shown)


elazarcoh:
Hmm.. my test is failing for some reason. Maybe the workaround of setting the `process.platform` isn't working?
Both tests passes on my machine. I tried to reproduce with WSL but couldn't get the tests to even start on it.

Should I simply run each test (linux/windows) only if the platform match. instead of hacking it like I did?

J-Fields:
Thanks!

PR #7483, opened 2022-02-12
Fix the remaining dependency cycles.

### LINKED ISSUE #6177: multiple circular dependencies between TypeScript files in the VSCodeVim code base

inspirer:
What this PR does / why we need it:
This PR removes all the remaining dependency cycles and adds circular-dependency-plugin to prevent backslide.

Which issue(s) this PR fixes
#6177

### DIFF
--- a/extensionBase.ts
+++ b/extensionBase.ts
@@ -17,6 +17,7 @@ import { taskQueue } from './src/taskQueue';
 import { Register } from './src/register/register';
 import { SpecialKeys } from './src/util/specialKeys';
 import { HistoryTracker } from './src/history/historyTracker';
+import { exCommandParser } from './src/vimscript/exCommandParser';
 
 let extensionContext: vscode.ExtensionContext;
 let previousActiveEditorUri: vscode.Uri | undefined;
@@ -97,6 +98,8 @@ async function loadConfiguration() {
  * The extension's entry point
  */
 export async function activate(context: vscode.ExtensionContext, handleLocal: boolean = true) {
+  ExCommandLine.parser = exCommandParser;
+
   // before we do anything else, we need to load the configuration
   await loadConfiguration();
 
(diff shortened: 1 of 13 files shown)


kpengboy:
Hi @J-Fields, I wonder if you've had any chance to take a look at this?

inspirer:
Hi @J-Fields, I updated this PR to the latest head and tried to come up with the minimum change to make it work. I'm pretty sure that these solutions might not be perfect and I'm happy to redo them in a better way. Let me know what you think.

This change will help us build VSCodeVim with the closure compiler, and will remove the need to maintain and update a local fork with these patches.

J-Fields (approved):
Awesome, thanks @inspirer!

> I'm pretty sure that these solutions might not be perfect and I'm happy to redo them in a better way. Let me know what you think.

They seem reasonable to me. I'd rather get the plugin in place and refactor later, especially the number of times I've broken these PRs already 🙃

Issue #6957, opened 2021-08-05
Reverse searching missing some matches


babramczyk:
**Describe the bug**
In certain scenarios, a reverse search does not match everything.

Especially when it doesn’t match the search when I’ve only typed a character or two, it can be really jarring, as it will momentarily scroll way out of view and I’ll lose context. The video shows what I mean better.

**To Reproduce**
Steps aren’t necessarily super well known, but I’ve ran into this before (and have noticed due to the jarring scroll jumping), but finally managed to get a video of it.

As you can see in the video, I do a reverse search `?t`. It skips over about 200 lines, plenty of which have `t`s. Those matches are skipped over, and do not show in the gutter.

Then, when I expand the search to `?type`, it matches a `t` that was there all along, just a few lines up from the cursor.

I do it again, to demonstrate the jarring scroll effect this has

(at the end is Vim in the VS Code terminal on the right, showing expected behavior).

https://streamable.com/qr3fyh

**Environment (please complete the following information):**

- Extension (VsCodeVim) version:
- VSCode version: v1.21.5
- OS: macOS 11.5

PR #7495, opened 2022-02-14
Use timeout for search

elazarcoh:
Using timeout is more correct then limiting the search results.

#6957

I think 1 second should be enough. We can finetune this value for better experience.
We can use both timeout and max number of ranges, but as a conjuction:
```ts
if (timeout && before.length + after.lenth > MAX_RANGES) {
    break;
}
```
This will keep the current behavior, while allowing more searches to be found.

### DIFF
--- a/src/vimscript/pattern.ts
+++ b/src/vimscript/pattern.ts
@@ -44,6 +44,7 @@ export class Pattern {
   public readonly closed: boolean; // was an ending delimiter present?
 
   private static readonly MAX_SEARCH_RANGES = 1000;
+  private static readonly MAX_SEARCH_TIME = 1000;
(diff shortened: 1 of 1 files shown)


src/vimscript/pattern.ts
```diff
@@ -115,6 +116,11 @@ export class Pattern {
     while (true) {
       const match = this.regex.exec(haystack);
 
+      let searchTimeout = false;
+      setTimeout(() => {
+        searchTimeout = true;
```
J-Fields, line 121: Aha! I tried doing this once but was thinking about it wrong (I don't have much experience with js/node) and got into a classic XY problem googling "how to cancel a promise nodejs", which turns out to be hard or impossible (I forget). Turns out to be so simple 🙃

J-Fields:
> I think 1 second should be enough
I thought I've read this is configurable, but I can't find anything about it in `:help` or google. 1 second seems very reasonable.

PR #7503, opened 2022-02-15
Using text-registers for macros

elazarcoh:
This commit is a POC for using plain text registers as Vim macro.
Example:
1. Create a new document, and insert "hello".  `ihello<Esc>`
    ```plaintext
    hello
    ```
2. Create an empty macro at register a.  `qaq`.  This is currently required to get register a to be executable.
3. Insert the requested text macro.  `oviwc123\<Esc\>A456\<Esc\><Esc>"ayiW`.  Note: I've escaped `\<` to note that I want the character '<', and not the '\<Esc\>' key. (point 4 will make this clearer, I think.)
4. See that register a contains the required text.  `:reg a`
5. Delete this line. `dd`
6. Execute register a. `@a`
7. You should get:
    ```plaintext
    123456
    ```


test/macro.test.ts
```diff
@@ -178,4 +179,39 @@ suite('Record and execute a macro', () => {
...
+
+      ['dd', ['|two', 'three']],
+
+      ['yyp', ['one', '|one', 'two', 'three']],
+
+      ['VGJ', ['one two| three']],
+
+      ['gUU' + 'j.' + 'j.', ['|ONE', 'TWO', 'THREE']],
+
+      [':2d', ['|one', 'three']],
```
elazarcoh, line 201: This one should include an `<Enter>` at the end, this is the behavior in Vim also.  Try `:2d` vs `:2d\n`. Only the latter works.

I've tried with `<Enter>` and with `\n` but couldn't get it to work (yet). It's interesting, since in the remap config I have something like this that works. I'll have to see why it works there and not here.
```json
        {
            "after": [
                ":",
                "s",
                "/",
                "\\",
                "\\",
                "/",
                "\\",
                "/",
                "/",
                "g",
                "<Enter>"
            ],
            "before": [
                "<leader>",
                "s",
                "/"
            ]
        },
```

elazarcoh:
Ok. All tests should pass now (passes on my machine).

What else we need to fully support it? Some more tests, probably. I'll add some, hopefully soon.

elazarcoh:
Okay. so I fixed some bugs in the tests. Now all of them passing.

elazarcoh:
I'll add some more, hopefully soon.
Do you think anything else is needed?

elazarcoh:
For some reason I cannot make it work with `<leader>`. 
@J-Fields Do you know why?

J-Fields:
Nope, I'll try to take a look sometime soon

elazarcoh:
`<leader>` as a string is not handled by the `handleKeyEvent`. It has to be converted to the actual leader key. I added support for it + tests.

J-Fields:
Hey @elazarcoh what's the status of this - is it ready to merge?

elazarcoh:
It should be, yes.
I couldn't think of any more things that need testing to make sure it doesn't break anything else. If you have an idea for what else should be tested, I can implement it.

src/transformations/execute.ts
```diff
@@ -161,38 +162,64 @@ export async function executeTransformations(
...
+        if (!recordedMacro) {
           return;
-        }
 
-        vimState.isReplayingMacro = true;
+        } else if (typeof recordedMacro === 'string') {
+          // A string was set to the register. We need to execute the characters as if they were typed (in normal mode).
+          const keystrokes = keystrokesExpressionParser.parse(recordedMacro);
+          if (!keystrokes.status) {
+            return;
```
J-Fields, line 172: Should this be communicated to the user somehow?

src/transformations/execute.ts
```diff
@@ -161,38 +162,64 @@ export async function executeTransformations(
...
+            vimState.lastMovementFailed = false;
+            return;
+          }
 
-        vimState.recordedState = new RecordedState();
-        if (transformation.register === ':') {
-          await new ExCommandLine(recordedMacro.commandString, vimState.currentMode).run(vimState);
-        } else if (transformation.replay === 'contentChange') {
-          await modeHandler.runMacro(recordedMacro);
         } else {
```
J-Fields, line 193: Any reason you see that we can't (eventually) drop this `else` entirely and always store registers as text, rather than `RecordedState`s?

elazarcoh: The only reason I can think of is that using `RecordedState` might have better runtime performance.

src/transformations/execute.ts
elazarcoh: See updated

src/transformations/execute.ts
J-Fields: Yeah, we'll need to measure that

J-Fields (approved):

PR #7506, opened 2022-02-16
`cur[sor]`

elazarcoh:
Add cursors by search pattern.
Can add cursor at a specific location using special indicator. Currently I'm using `[$]{0}`.
We should discuss a better indicator. It has to be zero-length regex pattern, when searching. We can add to the search parser another special indicator, e.g. `<Cursor>` and replace it with a zero-length regex such that `[$]{0}` (or something more robust rather than `$`).

Feature:
Consider this example: (cursor indicated by `|`)
```python
def foo():
|def bar(): 
def bizbaz():
def moo():
```
1. `:cur def`    (all matches)
    ```python
    |def foo():
    |def bar():
    |def bizbaz():
    |def moo():
   ```
2. `:cur 2 def`   (limit number of cursors)
    ```python
    def foo():
    |def bar():
    |def bizbaz():
    def moo():
   ```
3. `:cur de[$]{0}f`   (at specified location)
    ```python
    de|f foo():
    de|f bar():
    de|f bizbaz():
    de|f moo():
   ```
4. `:cur [$]{0}de[$]{0}f`      (multiple cursor)
    ```python
    |de|f foo():
    |de|f bar():
    |de|f bizbaz():
    |de|f moo():
   ```
5. `:cur def \w+[$]{0}`     (using regex)
    ```python
    def foo|():
    def bar|():
    def bizbaz|():
    def moo|():
   ```
5. `vj :'<,'>cur def`     (in selection range)
    ```python
    def foo():
    |def bar():
    |def bizbaz():
    def moo():
   ```

### DIFF
--- a/src/cmd_line/commands/cursor.ts
+++ b/src/cmd_line/commands/cursor.ts
@@ -1,24 +1,24 @@
-import { optWhitespace, Parser, seq, whitespace } from "parsimmon";
-import { Position } from "vscode";
-import { Cursor } from "../../common/motion/cursor";
-import { VimState } from "../../state/vimState";
-import { ExCommand } from "../../vimscript/exCommand";
-import { numberParser } from "../../vimscript/parserUtils";
-import { Pattern, PatternMatch, SearchDirection } from "../../vimscript/pattern"
-
+import { optWhitespace, Parser, seq, whitespace } from 'parsimmon';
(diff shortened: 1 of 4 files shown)


elazarcoh:
Added new feature: cursors from current word/selection, when not giving a pattern to the command.
examples is PR description.

PR #7527, opened 2022-02-22
Publish VSCodeVim to OpenVSX

### LINKED ISSUE #7183: publish to Open VSX
Hi 👋 We are running Web version of VS Code in Gitpod: www.gitpod.io. VS Code Vim is very popular extension among our users and we would like to ensure that they can use latest versions on time: https://github.com/gitpod-io/gitpod/issues/6220#issuecomment-947288571

Right now it is published by OpenVSX service account which updates versions from time to time: https://github.com/open-vsx/publish-extensions We would like to ask you to take an ownership of the vscodevim namespace and publish directly 🙏 

Docs are here: https://github.com/eclipse/openvsx/wiki/Publishing-Extensions Technically it should be straightforward. You are basically have to get a token and call ovsx publish with it, for instance how RedHat is publishing the java extension: https://github.com/redhat-developer/vscode-java/blob/ff51dbffc30dd23190335c929e9f34ccdf511554/Jenkinsfile#L73-L82

For the background:
Only MS products like VS Code may use MS VS Code extensions Marketplace. So we have to use Open VSX instead: https://open-vsx.org/ It is a vendor neutral alternative.

filiptronicek:
Hey there VS Code Vim team 👋!

This PR adds automatic publishing to the OpenVSX registry and therefore fixes #7183.

There are some more steps to take:

1. Add a repo secret for GitHub Actions called `OPEN_VSX_TOKEN` with a PAT obtained via following [this guide](https://github.com/eclipse/openvsx/wiki/Publishing-Extensions) (steps 1-4)
2. After merging, try publishing for the first time 🚀
3. (Optional) claim the namespace by following [this doc](https://github.com/eclipse/openvsx/wiki/Namespace-Access#how-to-claim-a-namespace) so that the extension is verified on the registry.

cc @J-Fields

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -62,3 +62,10 @@ jobs:
         run: yarn run vsce publish --yarn
         env:
           VSCE_PAT: ${{ secrets.VSCE_PAT }}
+          
+      - name: Publish to Open VSX Registry
+        uses: HaaLeo/publish-vscode-extension@v1
+        id: publishToOpenVSX
+        with:
+          pat: ${{ secrets.OPEN_VSX_TOKEN }}
+


J-Fields:
Thanks for helping me get this set up! I've made the request to claim the namespace, after that's accepted I'll publish a new version.

Issue #6046, opened 2021-01-26
Surrogate pairs (emojis, etc.) are not handled well in insert mode


athos:
**Description**

When I'm typing in a text containing [surrogate pairs](https://en.wikipedia.org/wiki/Universal_Character_Set_characters#Surrogates) (like emoji, relatively complicated kanji, etc.) using my [input method](https://en.wikipedia.org/wiki/Input_method) in the insert mode, the text gets garbled as shown in the GIF below:

<img src="https://user-images.githubusercontent.com/27441/105862486-29740500-6033-11eb-96af-184a846a3795.gif" width="382">

After digging into the codebase, I found that [this line of code](https://github.com/VSCodeVim/Vim/blob/7f8da746873e51070a211f6fa22496b0ff399e60/src/actions/commands/insert.ts#L306) does not seem to handle surrogate pairs well:

```javascript
    const char = this.keysPressed[this.keysPressed.length - 1];
```

This can be confirmed as follows:

```javascript
> var text = '🤔'
> text.length
2
> text[text.length-1]
'�'
> var text2 = '𩸽'
> text2.length
2
> text2[text2.length-1]
'�'
```

To fix this behavior, I think we need to do something like what VSCode does in [getPrevCodePoint](https://github.com/microsoft/vscode/blob/578e2dc4ee7f2f508be6910c99af3f9ab866dab0/src/vs/base/common/strings.ts#L468-L477). If we had it, we should be able to handle text containing surrogate pairs well like the following:

```javascript
> String.fromCodePoint(getPrevCodePoint(text, text.length))
'🤔'
```

**To Reproduce**
Typing in a text containing surrogate pairs in the insert mode.

**Expected behavior**
Expected characters are successfully entered in the editor.

**Environment:**
- Extension (VsCodeVim) version: 1.18.5
- VSCode version: 1.52.1
- OS: macOS Big Sur (Darwin x64 20.2.0)

xiekeyi98:
I have the same issue.

Would you mind taking a look @J-Fields   , thanks.

J-Fields:
Can you confirm this is still an issue? This implementation was changed a bit in 1.19.0

athos:
Thank you for letting me know! 🤝

I've checked with 1.19.2, but I still see the same issue, unfortunately.

midorizemi:
Hello. I still have same issue too.
[...]

Issue #6894, opened 2021-07-20
Emojis are treated as two characters when moving cursor


fwcd:
**Describe the bug**

Moving the cursor over an emoji character, e.g. 👍, causes it to 'swallow' the second press, presumably because it treats the emoji as two characters. This occurs both in insert and in edit mode, not however in VSCode's default editor with the extension disabled.

**To Reproduce**

1. Type some text with emoji
2. Move the cursor over it with h/l or the arrow keys

**Expected behavior**

The cursor/extension should treat the emoji as a single character.

**Environment:**

- Extension (VsCodeVim) version: 1.12.5
- VSCode version: 1.58.2
- OS: Ubuntu 20.04

**Additional context**
I think the issue is that the extension internally deals in terms of UTF-16 (?) code units rather than code points (or ideally grapheme clusters), causing the cursor to move "incorrectly" when a character is encoded with multiple code units. The same issue probably also occurs for other characters that aren't encoded as a single code unit.

Related issue, that complicates working with emojis currently: https://github.com/VSCodeVim/Vim/issues/6046

J-Fields:
Duplicate of #6046

Issue #2722, opened 2018-06-11
Windows 10 Emojis require two characters for navigation


liamdawson:
Issue Type: <b>Bug</b>

**What happened**:
I typed the 🏢 emoji using the Windows 10 emoji picker, and it changed it to `��` in-editor when I closed the picker.

**What did you expect to happen**:
I expected it to retain the 🏢 emoji.

**How to reproduce it (as minimally and precisely as possible)**:

1. Install this extension
2. Create a new file
3. Press `win`+`.` to open the emoji picker
4. Insert the 🏢 emoji
5. Press `ctrl`, `esc`, or any key that closes the emoji picker (or press the close button on the emoji picker)

Disabling the `vim` extension prevented the issue. Copying and pasting the emoji from Notepad also prevents the issue.

Extension version: 0.12.0
VS Code version: Code 1.24.0 (6a6e02cef0f2122ee1469765b704faf5d0e0d859, 2018-06-06T17:35:40.560Z)
OS version: Windows_NT x64 10.0.17133

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|AMD Ryzen 5 1600X Six-Core Processor            (12 x 3593)|
|GPU Status|2d_canvas: enabled<br>flash_3d: enabled<br>flash_stage3d: enabled<br>flash_stage3d_baseline: enabled<br>gpu_compositing: enabled<br>multiple_raster_threads: enabled_on<br>native_gpu_memory_buffers: disabled_software<br>rasterization: disabled_software<br>video_decode: enabled<br>video_encode: enabled<br>vpx_decode: enabled<br>webgl: enabled<br>webgl2: enabled|
|Memory (System)|15.96GB (1.43GB free)|
|Process Argv|C:\Program Files\Microsoft VS Code\Code.exe [project path]|
|Screen Reader|no|
|VM|0%|

</details>

liamdawson:
I found a separate issue in that the extension also treats `🏢` as two characters for navigation purposes (e.g. I have to press `ll` or `2l` to move to the next character after the `🏢` character. Worth raising as a separate issue?

jpoon:
Workaround for the first issue, is to be in `insert` mode while using the emoji picker.

J-Fields:
Related: #2722

J-Fields:
Duplicate of #6046

PR #7562, opened 2022-03-16
Consider surrogate pair for moving(h/l) and yank

smallkirby:

This PR is related to #6046 .

This fix adds a check if the character the cursor is moving to  is a part of surrogate pairs. If the character is a part of surrogate pairs, it additionally moves a cursor.

Before this PR, yank and moving for surrogate pairs look like below (the first character is `𩸽`, and the second is `🐕`):

https://user-images.githubusercontent.com/47111091/158623636-d5a42600-f001-4f7b-9ffd-023f1beb8775.mp4

After this PR, it looks like below:

https://user-images.githubusercontent.com/47111091/158623701-ad5c1bd8-9e8d-4087-9f65-a6ae96ec78ab.mp4

Related to #6046, but I don't know this fixes it or not.

- I'm sorry for dirty commits which just fixes typo. If it is bad, I will rebase them, or open new PR.

### Environment of PoC movie above
- VSCode version: 1.64.2
- OS: Linux x64 5.16.9-custom (Ubuntu Jammy Jellyfish (devel branch))


J-Fields:
Thanks for the PR!

> I'm sorry for dirty commits which just fixes typo. If it is bad, I will rebase them, or open new PR.

No trouble at all; except for very large PRs, I tend to ignore individual commits and just squash them at the end.

Would you mind adding a few test cases, please?

smallkirby:
Thanks for your quick reply. I added a test suite which tests moving and yank of surrogate-pairs.

smallkirby:
Let me summarize a little bit more in detail.

- Surrogate pairs are pairs of two bytes, which consist of a single Unicode character, such as Emoji and Kanji. Surrogate pairs consist of `high surrogate` and `low surrogate`. In VSCodeVim, these pairs are just treated as two separate characetrs when moving cursor or yanking characters.
  - Therefore as #6894 and #2722, you need to type `ll` to move left across surrogate pairs.
  - Yanking surrogate pairs by `vy` just copies a single byte of the pair. It is pasted to as invalid character.

- In this PR, I managed that cursor position must always be at `high surrogate`, not `low surrogate` for navigation.
  - If the cursor is located at `low surrogate` byte of surrogate pairs, move left or right additionaly depending on navigation direction.

- Same for yank. If the start position of yank is located at `low surrogate`, move start position to left. If the end position of yank is located at `high surrogate`, move end position to right.
  - In short, range of selection should not laid across `high surrogate` and `low surrogate`.

- The largest impact for users would be that surrogate pairs are treated as single character for navigation. It might need discussion though vanilla vim (I mean vim-tiny or something) treats surrogate pairs in the same way as far as I know.

- As for #6046, I don't know whether this PR fixes it or not, because I can't reproduce it. I can copy `𩸽` outside VSCodeVim, and paste it to VSCode without any modification.

J-Fields (approved):
LGTM. Thank you!

PR #7681, opened 2022-05-07
Fix a bug that marks not deleted properly

chewcw:
Before this fix we cannot delete marks using this command :delm A-Za-z, etc.


J-Fields:
Thanks!

Issue #5272, opened 2020-09-28
Add "clever-f.vim" functionality to VSCodeVim


macintacos:
I like `f`/`F`/`t`/`T` for quickly navigating to the next occurrence of a letter, and think that easymotion is a bit too heavy-handed in accomplishing this. [`clever-f.vim`](https://github.com/rhysd/clever-f.vim) is a plugin that makes "go to next occurrence of character you just searched with `f`/`F`/`t`/`T`" the same character you started the motion with instead of using `;`/`,`, allowing you to free those up for other things (like remapping for `<leader>` or whatever). The addition of visual queues (similar to easymotion) that help you quickly navigate to the appropriate match that you intended makes this plugin really nice in practice when in Vim, and I wish I had it in VSCode.

**Describe the solution you'd like**
I'd like `clever-f.vim`, but in this extension.

**Describe alternatives you've considered**
Tried figuring out how to implement it, but I'm pretty new to extension authoring. Don't know how to accomplish this.

**Additional context**
Full details of how the plugin works is [in their README](https://github.com/rhysd/clever-f.vim/blob/master/README.md).

macintacos:
Just realized that they have a handy little demo here to see what it's like: http://rhysd.github.io/clever-f.vim/

JW9506:
doesn't `/` + `[stuff you want to search]` + `n / N`  do exactly what it does?

macintacos:
Sure, but you could say the same thing about easymotion, which is far more involved.

JW9506:
> Sure, but you could say the same thing about easymotion, which is far more involved, but is implemented in this extension anyway.

To me, `easymotion` helps me go straight to the target without having to cycle through the list, which is inherently different. This `clever-f.vim` is just another version of `/ search` but disguise itself using `t/f` prefix. `t/f` is great in itself because it locks the scope in just the line you are in.

macintacos:
It has inherent additional utility like freeing up `;` and `,` for other mappings, as well as just being a more intuitive (IMO) implementation of those motions overall (with the addition of visual cues to help navigate). I find that plugin to be really helpful when navigating around in Vim, and think it'd be a useful addition.

If you just want to use `/` to "do the same thing", that's fine, nobody's forcing you to use something else. Just like easymotion is something you can enable/disable, I'm not suggesting that clever-f be the "default" functionality, but rather something you can enable if you'd like it. It's just something that I would like to be added.

JW9506:
[...]

PR #7696, opened 2022-05-12
💗Implementation of clevef-F plugin!!!!

Mr-peipei:

This plugin is so amazing!!!!🎊
clever-F is a more advanced version of the f and F command!.
This plugin was created with reference to the following vim plugin.
[rhysd/clever-f](https://github.com/rhysd/clever-f.vim)

But I didnt implenet t and T command.
Because I felt it was better to implement it step by step.

clever-F

And I add a clever-F test!
All test is clear.

If you hava any question, please add a comment in this PR.

implement #5272


Mr-peipei:
Why does the test fail on windows

src/actions/plugins/cleverF.ts
```diff
@@ -0,0 +1,253 @@
...
+
+  public override async execActionWithCount(
+    position: Position,
+    vimState: VimState,
+    count: number
+  ): Promise<Position | IMovement> {
+    count ||= 1;
+    // First if is repeat "f" movement. Ex) "f","a","f" called
+    // Second else if is repeat "F" movement.Ex) "F","a","f" called
+    // Third else is first "f" command. Ex) "f","a" called
```
tani, line 40: Excuse me, your English is weird. I guess what you want to say like this.

The first if-clause states that the current operator is `f` and also the last operator is `f`.
Next else-if-clause states that the current operator is `f`-command and the last operator is `F`.
The final else-clause states that the current operator is the initial `f`.


Is this true?

Mr-peipei: thank you for add comment!
Sorry for my word is not corrected
that is true. 

I will fix my comment!

Mr-peipei: fixed

tani: Thank you. I confirmed that my concern is solved. Thank you for the great extra plugin :)

elazarcoh:
Thanks for the implementation! I pulled it into my custom VSCodeVim build I'm using for everyday work. 
I have a little issue with it. It seems to break on some special characters, e.g. `f[`, where without the plugin it worked as expected.

src/actions/plugins/cleverF.ts
```diff
@@ -0,0 +1,253 @@
...
+  constructor() {
+    this.decorations = [];
+  }
+  public updateDecorations(position: Position, editor: vscode.TextEditor, character: string) {
+    this.decorations = [];
+    const sourceCode = editor.document.getText();
+    if (character === '') {
+      return;
+    }
+    const regex = '(' + character + ')';
```
elazarcoh, line 233: Need to escape special characters

Mr-peipei: thank you for using this PR plugin!!!
understood, I will try to fix it.

Mr-peipei: I fixed it.
How about it ?

elazarcoh: Seems to be working well. Thank you!

elazarcoh:
Hi, I noticed operators are failing now (`dfa` for example).
I think disabling the clever-f on operator pending might be the correct solution for it.

Thanks again!

d1y:
Amazing! Looks great, looking forward to merging it into the main branch 🔥

jensenojs:
+1

Mr-peipei:
I thought this PR will not be reviewed for a while,  so I stopped to repair this PR.
Is there still demand this PR?

Itachi1998:
THANK YOU SO MUCH

Issue #6286, opened 2021-03-16
Support <C-g> motions in insert mode


J-Fields:
See https://vimhelp.org/insert.txt.html#ins-special-special

For instance, `<C-g>j` moves down and puts your cursor on the column you started insert mode in

PR #7740, opened 2022-06-05
Add <C-g> motions in insert mode

chewcw:
Add `<C-g>` motions in insert mode
Peek 2022-06-05 12-56

#6286

### DIFF
--- a/src/actions/commands/insert.ts
+++ b/src/actions/commands/insert.ts
@@ -530,3 +530,29 @@ class CreateUndoPoint extends BaseCommand {
     vimState.historyTracker.finishCurrentStep();
   }
 }
+
+@RegisterAction
+export class CommandMoveCursorOneLineDownInInsertMode extends BaseCommand {
+  modes = [Mode.Insert];
+  keys = [
+    ['<C-g>', 'j'],
+    ['<C-g>', '<down>'],
+  ];
+
+  public override async exec(position: Position, vimState: VimState): Promise<void> {
+    vimState.cursorStopPosition = position.getDown(1);
+  }
+}
+
+@RegisterAction
+export class CommandMoveCursorOneLineUpInInsertMode extends BaseCommand {
+  modes = [Mode.Insert];
+  keys = [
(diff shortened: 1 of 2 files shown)


src/actions/commands/insert.ts
```diff
@@ -530,3 +530,29 @@ class CreateUndoPoint extends BaseCommand {
     vimState.historyTracker.finishCurrentStep();
   }
 }
+
+@RegisterAction
+export class CommandMoveCursorOneLineDownInInsertMode extends BaseCommand {
+  modes = [Mode.Insert];
+  keys = [
+    ['<C-g>', 'j'],
```
J-Fields, line 538: `<C-g>`, `<C-j>` as well (and the same below)

src/actions/commands/insert.ts
```diff
@@ -530,3 +530,29 @@ class CreateUndoPoint extends BaseCommand {
...
+    ['<C-g>', '<down>'],
+  ];
+
+  public override async exec(position: Position, vimState: VimState): Promise<void> {
+    vimState.cursorStopPosition = position.getDown(1);
```
J-Fields, line 543: This doesn't seem to match Vim's behavior, which is to go one line up (or down), at the column Insert mode began in

test/mode/modeInsert.test.ts
```diff
@@ -625,4 +625,38 @@ suite('Mode Insert', () => {
       endMode: Mode.Insert,
     });
   });
+
+  suite('<C-g>', () => {
```
J-Fields, line 629: Going along with behavior change I mentioned above, I'd want to see tests with insertions, deletions, and arrow key movements before the `<C-g>` movement

J-Fields (changes requested):
See comments; it's a good start but these motions are more complicated than the current implementation.

I think a proper implementation would require adding an Insert mode variant to `ModeData` storing the starting position, and making sure that gets set properly.

J-Fields:
Closing for now due to inactivity; feel free to ping me if comments are addressed

PR #7745, opened 2022-06-08
feat: support Chinese phonetic in easymotion

JMwill:
1. Can use english character to search Chinese phonetic character with easymotion plugin
2. Useful to Chinese language users
3. Not effect the user who doesn't need it.

None


JMwill:
The project author does not need this feature, close it

Issue #4993, opened 2020-07-06
Configuring vscode vim by language (i.e. textwidth for COMMIT_MESSAGE)


raggi:
**Is your feature request related to a problem? Please describe.**

I use vscode much as I used vim, wherein it is $EDITOR and I use it to edit commit messages. The standard commit message format is:

* Subject line is no more than 50 characters
* Body is no wider than 72 characters

I have been able to adjust visual indicators for commit messages, using vscode settings as follows:

```json
    "[git-commit]": {
        "editor.rulers": [
            50, 72
        ],
        "editor.wordWrapColumn": 72,
    },
```

I would like to be able to format the commit body text using `gq` from visual mode, however it does not respect that word wrap column. I have to manually set `textwidth` using a vim command each time when I edit a commit message. I would like to remove the need to set the textwidth, however, I do not seem to be able to modify vim plugin settings inside of a language declaration block such as this.

**Describe the solution you'd like**

I would like to be able to do something like:

```json
    "[git-commit]": {
        "editor.rulers": [
            50, 72
        ],
        "editor.wordWrapColumn": 72,
        "vim.textwidth": 72,
    }
```

As a general feature. As of today this reports: `This setting does not support per-language configuration.`

I would also like to not have to do this - that textwidth would be able to be globally configured such that it stays in sync with `editor.wordWrapColumn`.

**Describe alternatives you've considered**

* I type the workaround myself all the time.
* I have tried a number of other ways to format this text easily.
* I have considered violating conventional commit message style.

raggi:
Friendly bump.

If I was to take a look at implementing this feature:
 - Would the feature be acceptable?
 - What blockers/challenges should I expect?
 - Are there any pointers for things I should read early on?

red8888:
adding `set textwidth=88` in my  `~/.vscodevimrc` seemed to also have no effect

PR #7747, opened 2022-06-08
Make vim.textwidth configurable per-language

BlakeWilliams:

Currently the vim.textwidth setting is global which means that all
languages must share the same value. This works most of the time but
for some cases like git commits it can be useful to have a custom value
to adhere to specific guidelines.

This updates setting configuration to be "language-overridable" which
allows VSCode to autocomplete the value in the settings page for
languages. This allows us to update the `textwidth` property to a getter
where VSCode will automatically return the correct value.

Fixes https://github.com/VSCodeVim/Vim/issues/4993

I don't typically work on VSCode extensions so if there's a better way to test this I'd appreciate any feedback on how to improve it.

### DIFF
--- a/src/configuration/configuration.ts
+++ b/src/configuration/configuration.ts
@@ -219,8 +219,6 @@ class Configuration implements IConfiguration {
 
   overrideCopy = true;
 
-  textwidth = 80;
-
   hlsearch = false;
 
   ignorecase = true;
@@ -471,6 +469,16 @@ class Configuration implements IConfiguration {
   operatorPendingModeKeyBindingsMap: Map<string, IKeyRemapping> = new Map();
   visualModeKeyBindingsMap: Map<string, IKeyRemapping> = new Map();
   commandLineModeKeyBindingsMap: Map<string, IKeyRemapping> = new Map();
+
+  get textwidth(): number {
+    const textwidth = this.getConfiguration('vim').get('textwidth', 80);
+
+    if (typeof textwidth !== 'number') {
+      return 80;
+    }
+
+    return textwidth;
+  }
 
   private static unproxify(obj: object): object {
     const result = {};
--- a/test/configuration/configuration.test.ts
+++ b/test/configuration/configuration.test.ts
@@ -1,9 +1,11 @@
 import * as assert from 'assert';
 import * as srcConfiguration from '../../src/configuration/configuration';
 import * as testConfiguration from '../testConfiguration';
-import { cleanUpWorkspace, setupWorkspace } from './../testUtils';
+import * as vscode from 'vscode';
+import { cleanUpWorkspace, setupWorkspace, createRandomFile } from './../testUtils';
 import { Mode } from '../../src/mode/mode';
 import { newTest } from '../testSimplifier';
+import { utils } from 'mocha';
 
 suite('Configuration', () => {
   const configuration = new testConfiguration.Configuration();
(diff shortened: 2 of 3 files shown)


test/configuration/configuration.test.ts
```diff
@@ -37,6 +39,27 @@ suite('Configuration', () => {
     assert.deepStrictEqual(normalizedKeybinds[0].after, ['o', '<Esc>', 'k']);
   });
 
+  test.only('textwidth is configurable per-language', async () => {
```
J-Fields, line 42:
```suggestion
  test('textwidth is configurable per-language', async () => {
```

test/configuration/configuration.test.ts
```diff
@@ -1,9 +1,11 @@
 import * as assert from 'assert';
 import * as srcConfiguration from '../../src/configuration/configuration';
 import * as testConfiguration from '../testConfiguration';
-import { cleanUpWorkspace, setupWorkspace } from './../testUtils';
+import * as vscode from 'vscode';
+import { cleanUpWorkspace, setupWorkspace, createRandomFile } from './../testUtils';
 import { Mode } from '../../src/mode/mode';
 import { newTest } from '../testSimplifier';
+import { utils } from 'mocha';
```
J-Fields, line 8: Unnecessary import, I think

test/configuration/configuration.test.ts
```diff
@@ -37,6 +39,27 @@ suite('Configuration', () => {
     assert.deepStrictEqual(normalizedKeybinds[0].after, ['o', '<Esc>', 'k']);
   });
 
+  test.only('textwidth is configurable per-language', async () => {
+    const globalVimConfig = vscode.workspace.getConfiguration('vim');
+    const jsVimConfig = vscode.workspace.getConfiguration('vim', { languageId: 'javascript' });
+
+    try {
+      assert.equal(jsVimConfig.get('textwidth'), 80);
```
J-Fields, line 47: Use `assert.strictEqual` or replace the import with

```typescript
import { strict as assert } from 'assert';
```

test/configuration/configuration.test.ts
BlakeWilliams: Ah whoops, didn't mean to let that make it into the commit. 😅

test/configuration/configuration.test.ts
BlakeWilliams: Good catch, fixed in [1bf1d5c](https://github.com/VSCodeVim/Vim/pull/7747/commits/1bf1d5ce27eaeffbdf7f706f823fc2593506ced4)

test/configuration/configuration.test.ts
J-Fields: I've probably put the same on `master` a dozen times by now 🙃

test/configuration/configuration.test.ts
BlakeWilliams: Updated in [Use strictEqual](https://github.com/VSCodeVim/Vim/pull/7747/commits/cb07ddcc1c56f8c7a4f3dc06ed35c2c92d7ab83a)

test/configuration/configuration.test.ts
```diff
@@ -1,7 +1,8 @@
 import * as assert from 'assert';
 import * as srcConfiguration from '../../src/configuration/configuration';
 import * as testConfiguration from '../testConfiguration';
-import { cleanUpWorkspace, setupWorkspace } from './../testUtils';
+import * as vscode from 'vscode';
+import { cleanUpWorkspace, setupWorkspace, createRandomFile } from './../testUtils';
```
BlakeWilliams, line 5:
```suggestion
import { cleanUpWorkspace, setupWorkspace } from './../testUtils';
```

I was using it at one point, but `createRandomFile` is no longer needed for the test.

src/configuration/configuration.ts
```diff
@@ -472,6 +470,16 @@ class Configuration implements IConfiguration {
   visualModeKeyBindingsMap: Map<string, IKeyRemapping> = new Map();
   commandLineModeKeyBindingsMap: Map<string, IKeyRemapping> = new Map();
 
+  get textwidth(): number {
```
J-Fields, line 473: Is the getter necessary? These properties *should* get updated when the setting changes

BlakeWilliams: I just tested this out locally, it looks like that is true but the property is cached based on the most recently updated value, at least based on my limited testing.

e.g. If you update Ruby to have a textwidth of 20 but the global value (default or configured value) is 80, non-Ruby files will start wrapping at 20 since the value is cached and not dynamically fetched. By dynamically fetching the value the VSCode API seems to be returning the correct settings for the language based on the active file. 

These API's are all new to me so I'm not 100% certain if this is the best way to accomplish this but it seems like it's working as intended.

J-Fields: Ah, got it, thanks. The whole config system probably needs a deeper refactor to allow easily changing `scope` for other settings, but this looks like a step in the right direction.

PR #7765, opened 2022-06-18
Fix incorrect error message for :q in dirty file

### LINKED ISSUE #7544: Error message is incorrect
**Describe the bug**
`:q!` is a command to finish editing without saving. However, the error message says to use this command to override it, which is a mistake for `:wq`.

**To Reproduce**

1.  Enter `:q` in file with unsaved changes.

**Expected behavior**
This message need to change to `No write since change (add w to overwrite)`

**Screenshots**
スクリーンショット 2022-03-09 17 43 33

- Extension (VsCodeVim) version: latest
- VSCode version: latest
- OS: macos Montery

itta611:
The error message for `:q` in dirty file was incorrect.

Fix #7544

### DIFF
--- a/src/error.ts
+++ b/src/error.ts
@@ -47,7 +47,7 @@ export const ErrorMessage: IErrorMessage = {
   33: 'No previous substitute regular expression',
   34: 'No previous command',
   35: 'No previous regular expression',
-  37: 'No write since last change (add ! to override)',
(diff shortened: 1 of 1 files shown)


J-Fields:
The existing message seems to be correct according to https://vimhelp.org/message.txt.html (search for `E37`)

PR #7807, opened 2022-06-30
fix 'db', 'dB', 'c' + backward/left motion at beginning of file

### LINKED ISSUE #6407: `d B` at the begin of file should not delete any thing.
`B` WORDS backword is an exclusive motion, which is described in Vim motion.txt. So when you at the begin (first row, first column) of some file, type `d B` should not delete any thing.
Peek 2021-04-03 13-58
Peek 2021-04-03 13-59

monjara:

Fixed that when cursor is at the beginning of document and in Normal mode
* 'db', 'dB' don't delete anything and stay in Normal mode.
* 'c' + backward motion don't delete anything and stay in Normal mode.
* 'c' + left motion don't delete anything and in Insert mode.

fixes #6407

### DIFF
--- a/src/actions/operator.ts
+++ b/src/actions/operator.ts
@@ -104,6 +104,14 @@ export class DeleteOperator extends BaseOperator {
   public modes = [Mode.Normal, Mode.Visual, Mode.VisualLine];
 
   public async run(vimState: VimState, start: Position, end: Position): Promise<void> {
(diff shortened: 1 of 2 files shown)


J-Fields:
I think this is a bit too special-case for me; IIRC this behavior is really a byproduct of `:help exclusive`. Appreciate the PR though!

PR #7811, opened 2022-07-01
Fix `set tw=N`

### LINKED ISSUE #7809: `set tw=#` throws an error
**Describe the bug**
I used to use the `set tw=#` command on the fly for changing the text width. However, now when I try and run this command I get the following error:

```
CommandLine: Error executing cmd=set textwidth=100.
err=TypeError: Cannot set property textwidth of #<f> which has only a getter.
```

**To Reproduce**

1. Open a file
2. Type `:set tw=120`

**Expected behavior**
After using the `set` command, `textwidth` should be set to whatever number I gave and I should therefore be able to use `gq` to wrap text to that width.

- VSCode version: 1.68
- OS: Windows 10

adamjhawley:
- Fixes `set tw=N` command
- Add test for `set tw=N`

Since the v1.23.0 release, setting 'textwidth' on the fly with `set tw=100` has
been broken and would cause the following error:

```
CommandLine: Error executing cmd=set textwidth=100.
err=TypeError: Cannot set property textwidth of #<f> which has only a getter.
```

This PR provides a fix and also adds a test so we can tell if the
command breaks again.

Fixes #7809

### DIFF
--- a/src/configuration/configuration.ts
+++ b/src/configuration/configuration.ts
@@ -470,14 +470,28 @@ class Configuration implements IConfiguration {
   visualModeKeyBindingsMap: Map<string, IKeyRemapping> = new Map();
   commandLineModeKeyBindingsMap: Map<string, IKeyRemapping> = new Map();
 
+  _textwidth = 80;
+
   get textwidth(): number {
-    const textwidth = this.getConfiguration('vim').get('textwidth', 80);
+    const textwidth = this.getConfiguration('vim').get('textwidth', this._textwidth);
 
     if (typeof textwidth !== 'number') {
-      return 80;
+      return this._textwidth;
     }
 
     return textwidth;
+  }
+
+  set textwidth(textwidth: number) {
+    // If in a workspace, update the workspace configuration. Otherwise update
+    // the global configuration.
+    if (vscode.workspace.name) {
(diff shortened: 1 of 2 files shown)


src/configuration/configuration.ts
```diff
@@ -470,16 +472,6 @@ class Configuration implements IConfiguration {
   visualModeKeyBindingsMap: Map<string, IKeyRemapping> = new Map();
   commandLineModeKeyBindingsMap: Map<string, IKeyRemapping> = new Map();
 
-  get textwidth(): number {
```
J-Fields, line 473: This is admittedly a bit janky (config system is in need of an overhaul), but it is necessary to let `textwidth` be configured per-language (see #7747).

I think we should instead add `set textwidth(number)`

J-Fields (changes requested):

adamjhawley:
(Sorry, didn't mean to close this PR temporarily.)

I've come up with something for `set textwidth(number)` but I'm not sure it is the right approach so very open to alternative suggestions! This PR is now more of an RFC.

tamird:
It'd be nice if this worked! Can this be rebased and landed?

edgimar:
@J-Fields Would you be willing to take a look at this one?  Having this will be a pretty big impact on usability.  Thanks!

Issue #7357, opened 2021-12-30
E23: No alternate file when using CTRL+^ to jump between files


schneems:
**Describe the bug**

Previously I was able to use CTRL+^ to jump between files, but something changed recently and now I'm unable to use that feature. I get an error instead. Here's a video. I open one file via the finder (CMD+P) then I try to jump back using CTRL+^ and  I get an error "E23: No alternate file". 

https://imgur.com/a/Oj4Lpsi

**To Reproduce**

1.  Open a file
2. Open another file
2.  Press CTRL+^
4.  See error

**Expected behavior**

vscode should jump back to the previous file

**Screenshots**

<img width="1712" alt="image" src="https://user-images.githubusercontent.com/59744/147773171-0418b743-d1d6-4c8d-9af6-c32cdbc27f8c.png">

I checked my logs after setting them to debug, but got nothing with the key press CTRL+^ I may have done it incorrectly

**Environment (please complete the following information):**

Extension version: 1.21.10
Vscode version: 1.63.2
OS: Mac Big Sur 11.6.2 (20G314)

J-Fields:
Thanks for the bug report - this was fixed recently on `master` and will be in the next release

schneems:
> Thanks for the bug report - this was fixed recently on master and will be in the next release

OMG thank you! This has been driving me up the wall. It's hard to know if a problem is on my end or an extension. Thanks for the the update and thanks for all your work ❤️

schneems:
Hey sorry to be a bother. When do you think y'all can cut a release?

schneems:
I'm on the latest release and it's not working.

Error when using CTRL+^:


<img width="1904" alt="image" src="https://user-images.githubusercontent.com/59744/155562976-c0fa9184-6e5f-4a70-a83e-f8f7af0508d2.png">


Vim version:

<img width="866" alt="image" src="https://user-images.githubusercontent.com/59744/155562837-9e42b7a0-7889-4c42-bb9a-15db61969f5c.png">

Vscode version:

<img width="866" alt="image" src="https://user-images.githubusercontent.com/59744/155562881-1ca31c2c-4907-4c20-bbc0-0a237563a80a.png">

schneems:
I tried this again today and IT WORKS!!!!!!!! Thank you! thank you! thank you! ❤️💜💚

PR #7862, opened 2022-07-22
Fix the jump list (#7825)

### LINKED ISSUE #7825: <c-o> jump back to wrong location after 'gd'
**Describe the bug**
<c-o> jump back to wrong location after 'gd'
version 1.22.2 is good,  version 1.23.0 and 1.23.1 not work.

**To Reproduce**
use `gd` jump from file-A to file-B ,jump to file-C, jump to file-D. 
then use `c-o` jump back , 
at version 1.22.2 , jump back from D to C to B to A. 
at version 1.23.0 , jump back from D to C to D to C. 

**Screenshots**
good case:
goodCase

bad case:
badCase

- VSCode version:  1.69.0  92d25e35d9bf1a6b16f7d0758f25d48ace11e5b9
- OS: Darwin x64 20.6.0

**Additional context**

pitkali:
Fix for #7357 not only started setting # register but also added
additional call to file jump handling that broke the jump list. So,
still set # register if activeTextEditor is undefined, but skip the
extra task.

Fixes #7825.

Fixes a jump list regression (#7825) caused by #7357.

#7825 

Since I've done zero VsCode/Typescript/JS development before today, I can't tell you why this was the problem, only that the fix seems to work.

### DIFF
--- a/extensionBase.ts
+++ b/extensionBase.ts
@@ -244,6 +244,9 @@ export async function activate(context: vscode.ExtensionContext, handleLocal: bo
         Register.setReadonlyRegister('%', relativePath);
       }
 
+      if (activeTextEditor === undefined) {
+        return;
+      }
       taskQueue.enqueueTask(async () => {
         const mh = await getAndUpdateModeHandler(true);
         if (mh) {


max-sixty:
Thanks @pitkali ! Will be great to have this back

catskul:
@J-Fields anything still needed before this one can land?

J-Fields (approved):
Thanks a lot! Seems this was caused by b64799c2b5045e6c9319a13e2accc6feb7660b97

PR #7866, opened 2022-07-24
Preserve case when inc/decrement (`<C-a>`/`<C-x>`) num

smallkirby:

Current impl of `<C-a>` and `<C-x>` lowercases hex string even if original string is capital.

This PR preserves lowercase/uppercase in below strategy:

1. If the string is hex string, find the last alphabet in the word, check if it is lowercase/uppercase.
2. If the hex string doesn't have alphabet in it, think it as lowercase.
3. If the string is not hex, think it as lowercase.
4. Finally, when doing `toString()`, transform whole hex string to lowercase or uppercase. 

### DIFF
--- a/src/common/number/numericString.ts
+++ b/src/common/number/numericString.ts
@@ -35,6 +35,7 @@ export class NumericString {
   suffix: string;
   // If a negative sign should be manually added when converting to string.
   negative: boolean;
+  isCapital: boolean;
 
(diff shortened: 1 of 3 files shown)


smallkirby:
I'm so sorry if I should've created an issue first :bow:

J-Fields (approved):
Neat, thanks!

Issue #7055, opened 2021-09-03
Support `set`


samedamci:
**Describe the bug**
`:set` command works fine but if you have `set` statements in your `.vimrc` file these options are not loaded.

**To Reproduce**

1. Add for example `set relativenumber` in your `.vimrc`
2. Load config to VSCode
3. See that feature not working
4. Run `:set relativenumber` command
5. See the efects

**Expected behavior**
Support `set` statement in `.vimrc`

**Environment (please complete the following information):**

- Extension (VsCodeVim) version: v1.21.7
- VSCode version: 1.58.0
- OS: Linux

diningPhilosopher64:
As a temporary work around , we can use the editor level setting : `editor.cursorSurroundingLines` to mimick the behaviour of scrolloff in vimrc.

PR #7893, opened 2022-08-09
Support `:set` in `.vimrc`

elazarcoh:
Add support for `set` in vimrc. 

fixes: #7055

To make use of the already existing `SetCommand`, I needed to allow passing a `null` as the `VimState`, which might not be the best solution to tackle this. 
The `SetCommand` not really needs the `vimState` to handle the vimrc usecase, I think.
An alternative approach is to parse the vimrc, and capture all the operations that needs `vimState` into a list, which we'll execute later when we have a `vimState`.
Something like:
```ts
class Vimrc {
  public load(vimrcPath) {
     const lines = ...
     for (const line in lines) {
       const action: ((_: VimState) => void) | undefined = parseLine(line);
       if (action) this.actions.push(action)
    }
  }
 public onStart(vimState) {
    for (const action in this.actions) {
      action(vimState);
    }
  } 
}

```

### DIFF
--- a/src/cmd_line/commands/set.ts
+++ b/src/cmd_line/commands/set.ts
@@ -13,7 +13,7 @@ type SetOperation =
       option: string | undefined;
     }
   | {
-      // :se[t] {option}?
+     // :se[t] {option}?
       type: 'show';
       option: string;
     }
@@ -153,7 +153,8 @@ export class SetCommand extends ExCommand {
     this.operation = operation;
   }
 
-  async execute(vimState: VimState): Promise<void> {
+  // We allow vimState to be `null` so we can use this command for vimrc loading (where vimState is not available).
+  async execute(vimState: VimState | null): Promise<void> {
     if (this.operation.option === undefined) {
       // TODO: Show all options that differ from their default value
       return;
@@ -179,13 +180,15 @@ export class SetCommand extends ExCommand {
           if (type === 'boolean') {
             configuration[option] = true;
           } else {
-            this.showOption(vimState, option, currentValue);
+            if(vimState)
+              this.showOption(vimState, option, currentValue);
           }
         }
         break;
       }
       case 'show': {
-        this.showOption(vimState, option, currentValue);
+        if(vimState)
+          this.showOption(vimState, option, currentValue);
         break;
       }
       case 'unset': {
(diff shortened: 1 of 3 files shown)


src/cmd_line/commands/set.ts
```diff
@@ -153,7 +153,8 @@ export class SetCommand extends ExCommand {
     this.operation = operation;
   }
 
-  async execute(vimState: VimState): Promise<void> {
+  // We allow vimState to be `null` so we can use this command for vimrc loading (where vimState is not available).
+  async execute(vimState: VimState | null): Promise<void> {
```
J-Fields, line 157: This codebase prefers `undefined` to `null` when given the option

src/cmd_line/commands/set.ts
```diff
@@ -179,13 +180,15 @@ export class SetCommand extends ExCommand {
           if (type === 'boolean') {
             configuration[option] = true;
           } else {
-            this.showOption(vimState, option, currentValue);
+            if(vimState)
```
J-Fields, line 183: Looks like `prettier` didn't run

src/configuration/vimrc.ts
```diff
@@ -68,6 +69,11 @@ export class VimrcImpl {
           VimrcImpl.clearRemapsFromConfig(config, clearRemap);
           continue;
         }
+        const setOptionAction = vimrcSetOptionBuilder.buildSetAction(line);
```
J-Fields, line 72: In the long run we'd want this to be more general (as many ex-commands can be run from `.vimrc`), but this makes sense in the shorter term

elazarcoh: Totally agree about that.

src/cmd_line/commands/set.ts
elazarcoh: done.

src/cmd_line/commands/set.ts
elazarcoh: Done. Though prettier changed more lines now for some reason.

J-Fields: This should be handled by our CI but it's been broken for a while now

mefengl:
when this is merged, there should also have a big update in README, to show which option now can be configured in .vimrc

Issue #4865, opened 2020-05-15
Folding visual selected lines (`zf`)


alac1984:
**Is your feature request related to a problem? Please describe.**
In Vim we can visual select some lines and create a custom fold with `zf` command. In VSCodeVim it doesn't work.

**Describe the solution you'd like**
I want to visual select a group of lines and create a fold with `zf` command, and after that open or close it with `za`. These folds need to be stored in some place, to obviously be reloaded everytime we open the code. If it could include a description of what is folded (a snippet of the code folded or even a custom description) it would be very appreciated.

J-Fields:
VSCode doesn't allow arbitrary folding; foldable regions are defined either by the language server or by indentation (depending on config).

jonboh:
It seems that this is now possible in VSCode 1.70.0 with "Create Manual Folding Ranges from Selection" and "Remove Manual Folding Ranges"
https://code.visualstudio.com/updates/v1_70#_fold-selection

PR #7903, opened 2022-08-14
Folding visual selected lines (`zf`)

elazarcoh:
Add folds using the new vscode commands "editor.createFoldingRangeFromSelection" and "editor.removeManualFoldingRanges".

Everything seems to work fine except for the cursor position in some cases. For some reason setting `vimState.cursors` does not work.

fixes: #4865

Testing is in WIP.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@
 - Yanking block-wise now pads shorter lines with spaces ([@J-Fields](https://github.com/J-Fields)).
 - `<C-]>` now goes to definition, not declaration ([@J-Fields](https://github.com/J-Fields)).
 
+### Added
+
+- Support for `zf` and `zd` commands, for folding and unfolding custom code blocks ([@elazarcoh](https://github.com/elazarcoh)).
+
 ## [v1.23.2](https://github.com/vscodevim/vim/tree/v1.23.2) (2022-08-01)
 
 ### Fixed
--- a/src/actions/commands/fold.ts
+++ b/src/actions/commands/fold.ts
@@ -87,9 +87,9 @@ class AddFold extends BaseOperator {
     const previousSelections = vimState.lastVisualSelection;  // keep in case of Normal mode
     vimState.editor.selection = new vscode.Selection(start, end);
     await vscode.commands.executeCommand(this.commandName);
-    await new CommandCloseFold().exec(start, vimState);
     vimState.lastVisualSelection = previousSelections;
     vimState.cursors = [new Cursor(start, start)];
+    await vimState.setCurrentMode(Mode.Normal);  // Vim behavior
   }
 }
 
--- a/src/actions/commands/fold.ts
+++ b/src/actions/commands/fold.ts
@@ -1,7 +1,10 @@
-import { Position } from "vscode";
-import { Mode } from "../../mode/mode";
-import { VimState } from "../../state/vimState";
-import { BaseCommand, RegisterAction } from "../base";
+import * as vscode from 'vscode';
+import { Position } from 'vscode';
+import { Cursor } from '../../common/motion/cursor';
+import { Mode } from '../../mode/mode';
+import { VimState } from '../../state/vimState';
+import { BaseCommand, RegisterAction } from '../base';
+import { BaseOperator } from '../operator';
 
 type FoldDirection = 'up' | 'down' | undefined;
 abstract class CommandFold extends BaseCommand {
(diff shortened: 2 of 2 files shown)


src/actions/commands/fold.ts
```diff
@@ -72,3 +75,31 @@ class CommandOpenAllFoldsRecursively extends CommandFold {
...
+abstract class FoldOperator extends BaseOperator {
+  public modes = [Mode.Normal];
+  abstract commandName: string;
+
+  public async run(vimState: VimState, start: Position, end: Position): Promise<void> {
+    const previousSelections = vimState.editor.selections;
+    vimState.editor.selection = new vscode.Selection(start, end);
+    await vscode.commands.executeCommand(this.commandName);
+    await new CommandCloseFold().exec(start, vimState);
+    vimState.editor.selections = previousSelections;
```
J-Fields, line 88: Why set `vimState.editor.selections` then set `vimState.cursors`?

The latter should generally be the source of truth and `ModeHandler` will bring the former into line with it after the action's been run. Setting `vimState.editor.selections` directly in an action is sometimes necessary (for stupid reasons) but should be avoided because it can trigger `TextEditorSelectionChangeEvent` callbacks, which can cause issues.

elazarcoh: So, the previous selections are relevant in `'<'>` commands, so I have to save and restore them.
The cursor position is the position of where the cursor should be set after the folding, matching Vim's behavior.

J-Fields: `'<'>` relies on `vimState.lastVisualSelection` - can't we just make sure that's preserved?

elazarcoh: Possibly. I'll look into it 👍.

elazarcoh: Okay. Investigating it also led me to find few bugs. Seems to be corrected now. I'll see if I can add some tests (is testing folds is even possible?).

J-Fields: > is testing folds is even possible?

Probably not easily - I wouldn't worry about it too much for now

src/actions/commands/fold.ts
```diff
@@ -72,3 +75,42 @@ class CommandOpenAllFoldsRecursively extends CommandFold {
...
+@RegisterAction
+class AddFold extends BaseOperator {
+  override modes = [Mode.Normal, Mode.Visual];
+  keys = ['z', 'f'];
+
+  readonly commandName = 'editor.createFoldingRangeFromSelection';
+
+  public async run(vimState: VimState, start: Position, end: Position): Promise<void> {
+    const previousSelections = vimState.lastVisualSelection;  // keep in case of Normal mode
+    vimState.editor.selection = new vscode.Selection(start, end);
```
J-Fields, line 88: Why's this necessary?

src/actions/commands/fold.ts
```diff
@@ -72,3 +75,42 @@ class CommandOpenAllFoldsRecursively extends CommandFold {
...
+  readonly commandName = 'editor.removeManualFoldingRanges';
+
+  override async exec(position: Position, vimState: VimState): Promise<void> {
+
+    const selectionForFold = vimState.currentMode === Mode.Visual
+      ? vimState.editor.selection
+      : new vscode.Selection(position, position);
+
+    const previousSelections = vimState.lastVisualSelection;  // keep in case of Normal mode
+    vimState.editor.selection = selectionForFold;
```
J-Fields, line 109: Ditto - shouldn't the editor's selection be correct already?

J-Fields:
Mind adding a bullet point describing the change to `CHANGELOG.md`?

elazarcoh:
> Mind adding a bullet point describing the change to `CHANGELOG.md`?

Will have to wait for tomorrow, unfortunately.
But will do.

src/actions/commands/fold.ts
elazarcoh: do you mean the `vimState.editor.selection =`? If so, the command `editor.createFoldingRangeFromSelection` does just that: it create a fold from the selection. I haven't found any other way of doing that. I don't think there is other API for that. 
https://github.com/microsoft/vscode/blob/3edcbec59a15e6bec9538d37b9ac47702d90a3a4/src/vs/editor/contrib/folding/browser/folding.ts#L1081

src/actions/commands/fold.ts
elazarcoh: This is needed for the normal mode case.

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # Change Log
 
+- Support for `zf` and `zd` commands, for folding and unfolding custom code blocks ([@elazarcoh](https://github.com/elazarcoh)).
```
J-Fields, line 3: needs a rebase - this can go under a new `### Added` section

src/actions/commands/fold.ts
J-Fields: Ah, I missed that this was an operator - makes sense.

src/actions/commands/fold.ts
J-Fields: Why's that? Shouldn't `vimState.editor.selection` be `(position, position)` already, in Normal mode?

CHANGELOG.md
elazarcoh: Hmm the rebase messed it up... I'll see what I can do.

elazarcoh: Done.

src/actions/commands/fold.ts
elazarcoh: I'll check it.

elazarcoh: You're correct. Updated.

J-Fields (approved):
Thanks! Will merge pending green CI

tlkahn:
Super cool! I've been waiting for this for eternity! 🥳

PR #7918, opened 2022-08-21
Exclude source map from extension package

kidonng:

I noticed the source maps are taking a lot of space in the extension:

```shellsession
$ du -sh ~/.vscode/extensions/vscodevim.vim-1.23.2/out/* | sort -h | tail -n4
988K	/Users/kid/.vscode/extensions/vscodevim.vim-1.23.2/out/extension.js
2.3M	/Users/kid/.vscode/extensions/vscodevim.vim-1.23.2/out/extensionWeb.js.map
4.0M	/Users/kid/.vscode/extensions/vscodevim.vim-1.23.2/out/extension.js.map
8.0M	total
```

This pull request excluded them from the extension package, which goes down from 2.02MB to just 515.02KB!


J-Fields:
I *think* we need these for getting nice stack traces on errors

kidonng:
On the renderer process side, there is no extension but VSCode itself shown in the DevTools, and even then:

- It's only renderer code (of course)
- The source maps are loaded from a remote: `https://ticino.blob.core.windows.net`

On the main process side, Node requires explicitly [enabling source map](https://nodejs.org/dist/latest-v16.x/docs/api/cli.html#cli_enable_source_maps) and [VSCode does not do this currently](https://github.com//microsoft/vscode/issues/145473). Situation is probably the same on VSCode Web.

So basically, shipping source maps _now_ is useless.

J-Fields:
Currently we're using https://www.npmjs.com/package/source-map-support to put useful stack traces in auto-generated issues (see https://github.com/kidonng/Vim/blob/346c3df3b09d2c79f3a3af0a4ad292c2b3997f56/src/taskQueue.ts#L24) - do you know of a way to do this without shipping source maps?

J-Fields (approved):
Decided I'll just rip out the code that needs these source maps - automated error reporting is better done via VS Code's telemetry API if we're gonna do it. Thanks for the MR!

Issue #7687, opened 2022-05-09
Behavior of `:tabn[ext] {count}` is wrong


km-1021:
Thank you for providing this extension.
I use this everyday.

**Describe the bug**
When ":tabn[ext] 3" is executed, it does not go to the third tab from the top, but to the third tab from the current tab.
When ":tabn 3" is executed in the original Vim, it moves to the third tab from the top.
The same problem exists when executing ":tabn 1", ":tabn 2", and ... .

**To Reproduce**

1.  Open many tabs on VS Code.
2.  Enter ":tabn 3" when current tab page is 2nd tab.
3.  Current tab is changed to 5th tab.

**Expected behavior**
I expect current tab is changed to 3rd tab.

**Environment (please complete the following information):**

- Extension (VsCodeVim) version: Ver 1.22.2
- VSCode version: 1.67.0
- OS: Windows11

**Additional context**
The command "gt", which has the same behavior as ":tabn", worked correctly in ver 1.22.2. 
":tabn" in ver 0.14.0 is also working correctly.

elazarcoh:
This can (probably) be implemented using `vscode.window.tabGroups` introduced in vscode 1.67.0. 
I've tried to start look at it, but there seems to be lots of breaking changes in the API (I mostly encountered `readonly` stuff they added). I'll try to gather a PR updating the vscode version first. @J-Fields, would you prefer to update to the latest vscode (1.71.0), or to the minimal supported version (1.67.0)?

J-Fields:
Minimal supported version, generally

PR #7945, opened 2022-09-06
fixes for vscode 1.67.0

elazarcoh:
This is needed to support some requested features, e.g. #7687 


src/actions/motion.ts
```diff
@@ -68,16 +68,18 @@ abstract class MoveByScreenLine extends BaseMovement {
         // for some reason vscode doesn't update them. But doing it this way does
         // update vscode's selections.
         const selections = vimState.editor.selections;
```
J-Fields, line 70: Can we just collapse these 3 lines to `vimState.editor.selections = vimState.editor.selections.map(...)`?

elazarcoh: Sure.

J-Fields (approved):

PR #7977, opened 2022-09-20
Fix insert mode for surrogate pairs

garzj:

When inserting surrogate pairs, for example emojis (`Win` + `.` / `Ctrl` + `Cmd` + ` `), it would show two "unknown glyph" characters.

Fixes https://github.com/VSCodeVim/Vim/issues/6046


src/mode/modeData.ts
```diff
@@ -24,6 +24,11 @@ export type ModeData =
       /** The first line number that was visible when SearchInProgressMode began */
       firstVisibleLineBeforeSearch: number;
     }
+  | {
+      mode: Mode.Insert;
+      /** The high surrogate of an incomplete pair */
+      highSurrogate: string | null;
```
J-Fields, line 30: This codebase prefers `undefined` to `null` when possible

src/actions/commands/insert.ts
```diff
@@ -236,9 +237,35 @@ export class CommandInsertInInsertMode extends BaseCommand {
   public override async exec(position: Position, vimState: VimState): Promise<void> {
     const char = this.keysPressed[this.keysPressed.length - 1];
 
+    let text = char;
+
+    if (char.length === 1) {
+      const prevHighSurrogate =
+        vimState.modeData.mode === Mode.Insert ? vimState.modeData.highSurrogate : null;
```
J-Fields, line 244: `undefined` here as well

J-Fields (changes requested):
LGTM other than trivial issue of `undefined` vs. `null`. Thanks for the PR!

J-Fields (approved):

Issue #4840, opened 2020-05-13
Support sidescrolloff option


ronakg:
**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you'd like**
An option that mimics the behavior of `scrolloff` and `sidescrolloff` settings in vim.

**Describe alternatives you've considered**
No alternatives

**Additional context**
None

J-Fields:
`editor.cursorSurroundingLines` is equivalent to `scrolloff`. Last I checked horizontal scrolling was not part of the VSCode API so `sidescrolloff` is not currently possible.

ronakg:
Thanks. Didn't know about `editor.cursorSurroundingLines`. I can live without `sidescrolloff` for now.

PR #8014, opened 2022-10-07
Bind vim `scrolloff` with VS Code `editor.cursorSurroundingLines`;

LinHeLurking:

This PR implements VIM `scrolloff` setting by binding it with VSCode `editor.cursorSurroundingLines`.
With this PR, you can execute command `:set scrolloff=5` or `:set so=5` to justify scroll offset.

There are many issues mentioning `scrolloff` (such as #7055 , #4840 and #1427). Current walk-around is manually set `editor.cursorSurroundingLines`.

This PR is intendedly work with PR #7893, with which supports `:set` command in `.vimrc`. After both of these PRs are merged, one can write `set so=5` in `.vimrc` file just like in normal VIM.


J-Fields (approved):
Thanks!

j127:
Has anyone gotten this working? I enabled `.vscodevimrc` and added `set scrolloff=0`, but there's still padding when I use `zt`. The suggestions [here](https://github.com/VSCodeVim/Vim/issues/1427#issuecomment-2241444292) don't work.

PR #8022, opened 2022-10-11
Add goto reference command

blennster:

Adds better vscode integration with goto references command using `gr`. Inspired by [helix](https://github.com/helix-editor/helix).
None yet

This has a bug same as `gd` when the multiple references are displayed with jump points. This is because the `executeCommand`call finishes before the picker is closed and there does not seem to be a way of knowing when a selection is made.


J-Fields:
Thanks for the MR! Hate to decline it, but I'm reluctant to add mappings which aren't Vim-native. You can add the mapping yourself directly in VS Code or in your vimrc with `nnoremap gr editor.action.goToReferences` or similar

PR #8045, opened 2022-10-22
`va(` searches forward

### LINKED ISSUE #8024: `va(` not working
First of all, thanks for the awesome plugin 👍🏻

The issue is that in my case `va(` not working like in vim or nvim. For example: `myFunctionName(arg1,arg2)` and if positioned at `F` in normal mode, by pressing `va(` one would expect to be selected: `(arg1,arg2)`, but nothing happens. How can I fix this? Thank you very much 🙏🏽

smallkirby:

In current impl, `va(` searches only backward for an opening character.

```txt
// If positioned as `te|st`, `vi(d` would delete `test`.
// But if positioned as `|function`, `vi(d` would delete only `f`
function (test) {
  hoge
}
```

As far as I checked, Vim 8.2 searches forward for an opening char if it can't find it backward. So I fixed to behave the same.

After this PR, `va(`-family would work as below:

```txt
// If positioned as `te|st`, `vi(d` would delete `test`.
// And if positioned as `|function`, `vi(d` would also delete `test`
function (test) {
  hoge
}
```

- Fixes #8024 

- I didn't touch behavior of indent block and tag block, because vanilla Vim behaves as same already.


src/actions/motion.ts
```diff
@@ -1860,7 +1860,20 @@ export abstract class MoveInsideCharacter extends ExpandingSelection {
...
     let openPos = PairMatcher.nextPairedChar(selStart, closingChar, vimState, true);
     if (openPos === undefined) {
-      return failedMovement(vimState);
+      // If opening character not found, search forwards
+      let lineNum = selStart.line;
+      while (true) {
+        if (lineNum >= vimState.document.lineCount) {
+          break;
+        }
+        const lineText = vimState.document.lineAt(lineNum).text.split('');
```
J-Fields, line 1869: Isn't the `split()` unnecessary?

smallkirby: You're right, fixed.

smallkirby:
Thanks for your reveiw.

- I checked the comment and fixed.
- I fixed a bug in my commit.

J-Fields:
Thanks!

PR #8046, opened 2022-10-22
Explicitly assign new active editor to `vimState` before jumping to global marks

### LINKED ISSUE #7975: Cursor position incorrect after jumping to a global mark
**Describe the bug**
When I navigate to an uppercase (global) mark in another file, vscode opens the file but does not set the cursor position correctly. (The cursor stays in its previous position.) Navigating to the mark a second time fixes the issue.

**To Reproduce**

1.  Set an uppercase mark in a file (with `mA` for example)
2.  Move the cursor to another line in the file
3.  Open a different file
4.  Navigate to the mark (with `'A` for example)

**Expected behavior**
The cursor should move to the marked line on the first invocation.

**Screenshots**

https://user-images.githubusercontent.com/702318/191124071-4d0f3848-405e-49f4-a77b-2f6f7fec4173.mov

- Extension (VsCodeVim) version: 1.23.2
- VSCode version: 1.71.2
- OS: macOS 12.6

smallkirby:

When jumping to global marker across files, it awaits `showTextDocument()` switches active editor. However, even after the await, `vimState` sometimes has old editor (as far as I tested, it happens only when first `jump to global marker` across files for that name).
So this patch explicitly assigns new Editor to `vimState` before jumping to global marker.

- Fixes #7975 

### DIFF
--- a/src/actions/motion.ts
+++ b/src/actions/motion.ts
@@ -615,6 +615,12 @@ class MarkMovementBOL extends BaseMovement {
       await ensureEditorIsActive(mark.document);
     }
 
+    const editor = vscode.window.activeTextEditor;
+    if (!editor) {
+      throw VimError.fromCode(ErrorCode.MarkNotSet);
+    }
+    vimState.editor = editor;
+
     return TextEditor.getFirstNonWhitespaceCharOnLine(vimState.document, mark.position.line);
   }
 }


src/actions/motion.ts
```diff
@@ -615,6 +615,12 @@ class MarkMovementBOL extends BaseMovement {
       await ensureEditorIsActive(mark.document);
     }
 
+    const editor = vscode.window.activeTextEditor;
+    if (!editor) {
+      throw VimError.fromCode(ErrorCode.MarkNotSet);
+    }
+    vimState.editor = editor;
```
J-Fields, line 622: Unfortunately setting `VimState.editor` directly is misguided. It's not `readonly`, but it should be.

src/actions/motion.ts
```diff
@@ -615,6 +615,12 @@ class MarkMovementBOL extends BaseMovement {
...
+
     return TextEditor.getFirstNonWhitespaceCharOnLine(vimState.document, mark.position.line);
```
J-Fields, line 624: Is it sufficient to just use `activeTextEditor.document` here?

J-Fields (changes requested):
see comments

src/actions/motion.ts
smallkirby: Unfortunately, if L622(assignment to `vimState`) is removed, it doesn't work.
It would be because `vimState` remains to be old Editor when `Position` is returned.

Do you have any idea?

J-Fields: Yeah, I figured as much. I'll have to think about it some more - our current architecture is just very clumsy for actions which relate to multiple editors, since `VimState` is tightly coupled with a `TextEditor`, and an `Action` is run with respect to a given `VimState`.

smallkirby:
It seems I can’t solve the problem in [the comment](https://github.com/VSCodeVim/Vim/pull/8046#discussion_r1007457603) without bunch of breaking changes.

Closing the issue for now. cc: @J-Fields

Issue #7695, opened 2022-05-12
Split some configuration to support multiple OS in the same time


hukacode:
Hello,

**Is your feature request related to a problem? Please describe.**
I'm using multiple OS (Windows and Linux) and I found there is only one setting for NeoVim.
`"vim.neovimPath"`

**Describe the solution you'd like**
I would like to have separate config for Windows, Linux, Mac, for example:
```
"vim.neovimPath.Windows"
"vim.neovimPath.Linux"
"vim.neovimPath.Mac"
```

J-Fields:
What's the use case? If sharing config between computers with VSCode's settings sync, this shouldn't be a problem because `vim.neovimPath` has `machine` scope (and if sharing via a dotfile repo or whatever, just don't include the setting in the shared config)

hukacode:
Hi @J-Fields
Thank you for your response.

The use case is I use `scoop` to install `neovim`, the path will be, for example, `C:\Users\huka\scoop\apps\neovim`. But on `brew` (MacOS, Linux), the path of neovim is `/opt/homebrew/bin/nvim`. Because I use 2 PCs at the same time (one for work, one for personal), then I will get an error message.
This issue is happen to `im-select` too.

Best regards,
Huka

soimy:
Hope for any updates after 3 years

PR #8051, opened 2022-10-24
Add scope for vim.autoSwitchInputMethod

Quanuanc:
- The im-select path of each machine is different, so it should not be synced.

- #7695 


J-Fields (approved):
Thanks!

PR #8067, opened 2022-10-28
Run showOpenDialog for empty `:e` command

### LINKED ISSUE #8039: `:e` command opens new file instead of open dialog
**Describe the bug**
I want to open an existing file and use vscode's open dialog. Looking at the code, it should be possible with `:e` without arguments, but currently `:e` opens a new untitled file.

**To Reproduce**
Execute `:e` command in normal mode.

**Expected behavior**
VSCode's open dialog is shown (same as File -> Open File...)

- Extension (VsCodeVim) version: v1.24.1
- VSCode version: 1.72.2
- OS: Win10

**Additional context**
This behavior seems to be already implemented:
https://github.com/VSCodeVim/Vim/blob/5f312cb5ef86eb4c8ffd6840b3442f471eeba3a9/src/cmd_line/commands/file.ts#L172
but the code is currently unreachable because omitting the file argument to the command results in `args.file` being `undefined`. This also means that `args` is identical for `:e` and `:enew` commands, and currently the `:enew` behavior is executed first.

I would propose to set the default for the `fileNameParser` to `''` to fix this here:
https://github.com/VSCodeVim/Vim/blob/5f312cb5ef86eb4c8ffd6840b3442f471eeba3a9/src/cmd_line/commands/file.ts#L87

If you agree, I could create a PR for this.

Trying `:e .` to open the current folder results in an error that I am opening a directory, where a sensible behavior would be to show the file open dialog at the specified folder, but maybe that is a different feature request.

hborchardt:
I want to be able to open the "Open File" dialog by executing the `:e` command.
The code for this already exists, but was unreachable because the default value for `file` while parsing was `undefined`.
Set the default value to `''`, so that arguments for `:e` are distinguishable from the `:enew` command.

Fixes #8039

❤️

### DIFF
--- a/src/cmd_line/commands/file.ts
+++ b/src/cmd_line/commands/file.ts
@@ -26,7 +26,7 @@ export type IFileCommandArguments =
       bang: boolean;
       opt: FileOpt;
       cmd?: FileCmd;
-      file?: string;
+      file: string;
       createFileIfNotExists?: boolean;
     }
   | {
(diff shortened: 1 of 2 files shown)


J-Fields:
I understand why you'd want this, but `:e` in Vim re-loads the current file, which is behavior some people depend on, and I try to keep compatibility when possible. Thanks for the PR!

PR #8070, opened 2022-10-31
special treatment for {

### LINKED ISSUE #2982: `da{` misbehavior
When using delete a block, with braces "da{" and the cursor is not on the opening brace, but a tab or space before, it will delete 

To reproduce:
- Get some code that includes curly braces.
- Go to the line of an opening one
- Go a few spaces back
- Execute "da{" in normal mode

Expected (Expected this, even if the cursor was few spaces off like the next gif):

vimexpected

Actual:
vimactual

elazarcoh:
fixes special case for hadling `{`, when there are no characters before it in the same line.

fixes #2982


J-Fields:
Thanks!

Issue #7870, opened 2022-07-27
Text is not aligned when pasting a range yanked in blockwise.


ryota-abe:
**Describe the bug**
Vim 9.0:
In blockwise visual mode, you can select a non-rectangular range (e.g., by pressing $).
Yank it and paste it in the middle of another line, and spaces will be added so that the inserted text is a rectangular shape.
![vim](https://user-images.githubusercontent.com/15438757/181258169-f3e104c5-e50c-48f7-870f-1168a2be5cab.gif)

VSCodeVim:
pasted text is not aligned.
![vscode](https://user-images.githubusercontent.com/15438757/181258185-8169edc7-ab50-40c4-b3aa-4dd907e96187.gif)

**To Reproduce**

1.  Press `CTRL-v` and select some lines with unequal line lengths. e.g. 
```
a
bb
ccc
```
2. Press `$` to select to the end of the lines.
3. Press `y`
4. Move to the middle of another text. e.g.
```
123 # caret is on "2"
456
789
```
5. Press `p`, will result in
```
12a3
45bb6
78ccc9
```

**Expected behavior**
```
12a  3
45bb 6
78ccc9
```

**Screenshots**
see above

**Environment (please complete the following information):**

- Extension (VsCodeVim) version: v1.23.1
- VSCode version: 1.69.1
- OS: Windows 10

**Additional context**
N/A

J-Fields:
Neat - looks like this was changed in Vim 9 and neovim, and sure seems like the right behavior to me

PR #8080, opened 2022-11-05
Fix cursor position after visual block gp and gP operations

burnsdy:
Cursor position may be incorrect after `gp` and `gP` operations. This PR correctly adjusts the cursor position and fixes one test case that was incorrect.

Related to the changes for #7870


J-Fields:
Thanks!

PR #8118, opened 2022-11-26
When searching for a vimrc path, first look for ~/.vscodevimrc

dominicpalmer:
Most Vim emulators first look for a vimrc file in `~` whose name is related to the target editor in some way. This is to prevent a standard `.vimrc` config file being overloaded by multiple editors on the same machine, when it shouldn't be. For example, JetBrains products first look for `.ideavimrc`, Visual Studio first looks for `.vsvimrc`, and both configuration files can contain settings that are designed to target only the respective editor. Examples...

Snippet from an `.ideavim` configuration file:
```
noremap <leader>n :action NewFile<CR>
```

Snippet from a `.vsvimrc` file:
```
noremap <leader>f :vsc Edit.FindInFiles<CR>
```

If either of these lines were placed in a general-purpose `.vimrc`, and each editor first looked for `.vimrc`, then the configuration would fail to load in each case.

VSCodeVim does allow the user to specify a path to a config file to use, but this is overhead not present in other emulators. A user of multiple emulators across multiple editors might intuitively expect a VSCodeVim specific `.vscodevimrc` to be parsed first, with the standard `.vimrc` afterwards as fallback.

This change adds an initial search for a `.vscodevimrc`, before continuing the existing searches as they are now.

N/A

N/A


J-Fields:
Nice idea, thanks!

PR #8140, opened 2022-12-05
fix: correct typo on vim.targets.enable desc

RoRep1ay:
Fixes a typo on `vim.targets.enable` description


J-Fields:
Thanks!

PR #8151, opened 2022-12-07
Fix typo in remapping error message

AsterisMono:
Key binding configurations accept a key-value pair with the key of "commands", but the error message asks for a "command" key, which is quite confusing.

None


J-Fields (approved):
Thanks!

PR #8178, opened 2022-12-26
fix #8083 - support the line range with no start line

### LINKED ISSUE #8083: Commands with relative range do not work
**Describe the bug**

At the moment, you always have to specify both start and finish range for a command to be applied.

For example, this works:

```
:25,$d
```

but if I am on line 25 and write:

```
:,$d
```

this does not work.

Same goes for commands like:

```
:1,d
```

that do not work.

**Expected behavior**

If start or end position is missing, it should use the current line.

VSCodeVim:  v1.24.3

Version: 1.73.0
Commit: 8fa188b2b301d36553cbc9ce1b0a146ccb93351f
Date: 2022-11-01T15:44:09.336Z
Electron: 19.0.17
Chromium: 102.0.5005.167
Node.js: 16.14.2
V8: 10.2.154.15-electron.0
OS: Linux x64 5.19.0-21-generic
Sandboxed: No

lazygyu:

This PR makes the line range with an empty start number like the one below;

```
:,3d     // this deletes from the current line to the 3rd line
```

#8083 

### DIFF
--- a/src/vimscript/lineRange.ts
+++ b/src/vimscript/lineRange.ts
@@ -277,19 +277,30 @@ export class LineRange {
     this.separator = separator;
   }
 
-  public static parser: Parser<LineRange> = seq(
-    Address.parser.fallback<Address>(new Address({ type: 'current_line' }, 0)).skip(optWhitespace),
+  public static parser: Parser<LineRange> = alt(
     seq(
+      // with the start line
+      Address.parser.skip(optWhitespace),
+      seq(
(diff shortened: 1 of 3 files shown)


J-Fields:
Mind adding a test case or two to `lineRangeParse.test.ts`?

lazygyu:
I've changed the implementation and added some tests.

J-Fields:
Thank you!

PR #8193, opened 2023-01-02
Fixes #8192: Command args are passed as lists

### LINKED ISSUE #8192: Remapping to a command passes args as a list
When a keybinding is mapped to a VS Code command, the arguments are not passed one by one but instead as a list. When no argument is specified, an empty list is passed.

**To Reproduce**
Define keybindings such as
```json
"vim.normalModeKeyBindings": [{
  "before": ["<leader>", "a"],
  "commands": "workbench.action.quickOpen"
},
{
  "before": ["<leader>", "b"],
  "commands": [{
    "command": "foo",
    "args": ["bar", "baz"]
  }]
}]
```
and trigger the bindings.

**Expected behavior**
The command is called with the given position arguments. I.e., no arguments in the first case and two arguments `"foo"` and `"bar"` in the second case.

**Actual behavior**
The command is called with a single list argument. An empty list `[]` in the first case and `["foo", "bar"]` in the second.

elmar-peise:
Fixes passing arguments to commands.

Fixes #8192


J-Fields (approved):
Thanks!

Issue #8144, opened 2022-12-06
Support leap.nvim


kabouzeid:
Similar but advanced version of sneak. https://github.com/ggandor/leap.nvim

GustafXTT:
Agree, it is quite graceful

cuixiaorui:
hi  @kabouzeid @GustafXTT 
I have implemented leap in my vim branch
![vim-leap](https://user-images.githubusercontent.com/12064746/210035757-42ee1fe0-c585-4938-b079-fbb8d5eedaff.gif)
But so far I haven't finished them all 
So there is no mention of pr
But it's already working. I've been using it for a few days now.
If you want to experience it, you can install my [vim branch](https://github.com/cuixiaorui/vim)
Simply execute the following command
```shell
1. yarn package
2. code --install-extension vim-1.24.3.vsix --force
```
then open leap switch in settings.json
```json
  "vim.leap": true
```

kabouzeid:
This works quite well already, thank you! I have noticed a few issues, but you are probably already aware of them. Let me know if I should write them up for you.

cuixiaorui:
haha @kabouzeid 
Thank you for your approval 
Just implemented some new features can be repackaged to experience
Wait for your feedback, let's do it together!

kabouzeid:
Just tried your updated version, and it works better already.

One thing that I noticed is that you show the marker at the first column, while leap.nvim shows the marker at the third column. For example when I type `sta` I get  `ta<marker>rget` with leap.nvim and `<marker>target` with your implementation.

cuixiaorui:
Haha, you found it. There are actually three differences in my implementation, one is not jumping directly to the first marker, and one is the way to select multiple markers. Another one is the different position of the marker point display

## Instead of jumping to the first marker point
In leap.nvim, the default is to jump to the first marker and then go to normal mode, then if you press the letter w, then there may be two cases, one case is that w is executed as a command, jumping to the first part of the next letter, and the other case is that w will be the name of the marker and pressing w will jump to the position of the corresponding marker. This makes pressing w not a stable behavior, and the user needs to check if w is a marker point first, which is difficult. I was often plagued by this problem when I was using leap.nvim
Of course, there is another way to solve the above problem, that is, after jumping to the first marker point, I press a non-marker point command to cancel the other marker points, such as pressing ESC or pressing Space. I need to press an extra button to cancel the other markers.
[...]

PR #8197, opened 2023-01-06
Feat implementation leap plugin

cuixiaorui:
## **What this PR does / why we need it**:
hi @J-Fields 

I have implemented a leap-like functionality plugin in vscodevim

leap is currently a very popular mobile plugin in the neovim community, and it is really efficient. I believe there are many vscodevim users waiting for its arrival #8144 

### Functional Features
- s c c -> search down
- S c c -> Search up
- s c -> Show cue marker points
- s \<enter\> -> Execute the previous search
- s c \<space\> -> Search c + \\t

You can see how it behaves in vscodevim

show-vscode-leap

## Compare the functional differences between leap.nvim

Because the implementation of vscodevim is different from neovim. So in some functions it is not the same as leap.nvim

- No automatic jumping to the first marker point
- Selection of multiple markers
- No support for using operators together

### Does not automatically jump to the first marker point

When matching multiple tag points, it does not automatically jump to the first tag point position.

In leap.nvim you can recognize that when you press a character that is the tag of a marker point then it will automatically jump to the marker point, otherwise it will execute the character as if it was a command in normal mode, but unfortunately this is not possible in vscodevim.

Another problem is that this feature can be confusing. Sometimes it jumps automatically and sometimes it doesn't (for example when there are multiple jump points, or you set the corresponding configuration). So I simply killed this feature in my implementation.

Instead of jumping to the first match point by default, the implementation now requires you to press an s. This s can be modified by the configuration

Of course, if only one marker is matched, then it will automatically jump over

### Multiple markers are selected in the following way

In leap.nvim, if there are too many matching tokens, then the selection method is to select the corresponding group by /.The problem with this method is that if three groups are matched, then needs to be pressed twice to select them. (Of course, this is a very extreme case, so you will not encounter it)

In my implementation, I refer to easymotion's jump implementation, which starts with one letter and uses two letters if it goes beyond that, so the advantage is that no matter how many matches are made, the two letters will directly locate the marker point.Of course, another reason is that this implementation is unified with easymotion. Because some mobile scenes use easymotion. So it would be difficult to have two sets of selections. (I believe many users of vscodeVim also use easymotion, so this behavior may be more preferable for them, but I'm not sure... haha... what do you think?)

vim-leap-01

### Use with operators is not supported

Unfortunately, this feature is also not supported, again because of the underlying implementation of vscodevim. Currently there is no way to do this

But this feature scenario is not used very often

## Can I map other keys if I don't want to use s S?

You can refer to the following configuration

Replace s with f
```
  {

      "before":[
        "f"
      ],

      "after":[
        "s"
      ]
    },

    {

      "before":[
        "s"
      ],

      "after":[
        "c",
        "l"
      ]
    },
]
```

## How should I experience this feature before this pr merge?

You can install the experience locally based on my code branch

From time to time, I will merge new code from vscodevim into my branch until this pr is merged.

Here's how to do it: 

1. Download [my code branch](https://github.com/cuixiaorui/VSCodeVim/tree/leap)
2. yarn   
3. yarn package
4. code --install-extension vim-1.24.3.vsix --force

### DIFF
--- a/src/actions/plugins/leap/LeapAction.ts
+++ b/src/actions/plugins/leap/LeapAction.ts
@@ -1,0 +1,139 @@
+import { Mode } from '../../../mode/mode';
+import { BaseCommand, RegisterAction } from '../../base';
+import { Position } from 'vscode';
+import { VimState } from '../../../state/vimState';
+import { configuration } from '../../../configuration/configuration';
+import { LeapSearchDirection, createLeap } from './leap';
+import { getMatches, generateMarkerRegex, generatePrepareRegex } from './match';
+import { StatusBar } from '../../../statusBar';
+import { Marker } from './Marker';
+import { VimError, ErrorCode } from '../../../error';
+
+@RegisterAction
+export class LeapPrepareAction extends BaseCommand {
+  modes = [Mode.Normal];
+  keys = [
+    ['s', '<character>'],
+    ['S', '<character>'],
+  ];
+
+  public override doesActionApply(vimState: VimState, keysPressed: string[]) {
+    return super.doesActionApply(vimState, keysPressed) && configuration.leap;
+  }
+
+  public override async exec(cursorPosition: Position, vimState: VimState): Promise<void> {
+    if (!configuration.leap) return;
+
+    if (this.keysPressed[1] === '\n') {
+      this.execRepeatLastSearch(vimState);
+    } else {
+      await this.execPrepare(cursorPosition, vimState);
+    }
+  }
+
+  private async execPrepare(cursorPosition: Position, vimState: VimState) {
+    const direction = this.getDirection();
+    const firstSearchString = this.keysPressed[1];
+
+    const leap = createLeap(vimState, direction, firstSearchString);
+    vimState.leap = leap;
+    vimState.leap.previousMode = vimState.currentMode;
+
+    const matches = getMatches(
+      generatePrepareRegex(firstSearchString),
+      direction,
+      cursorPosition,
+      vimState.document
+    );
+
+    vimState.leap.createMarkers(matches);
+    vimState.leap.showMarkers();
+    await vimState.setCurrentMode(Mode.LeapPrepareMode);
+  }
+
+  private execRepeatLastSearch(vimState: VimState) {
+    if (vimState.leap?.leapAction) {
+      vimState.leap.isRepeatLastSearch = true;
+      vimState.leap.direction = this.getDirection();
+      vimState.leap.leapAction.fire();
+    } else {
+      StatusBar.displayError(vimState, VimError.fromCode(ErrorCode.LeapNoPreviousSearch));
+    }
+  }
+
+  private getDirection() {
+    return this.keysPressed[0] === 's' ? LeapSearchDirection.Backward : LeapSearchDirection.Forward;
+  }
+}
+
+@RegisterAction
+export class LeapAction extends BaseCommand {
+  modes = [Mode.LeapPrepareMode];
+  keys = ['<character>'];
+  override isJump = true;
+  private vimState!: VimState;
+  private searchString: string = '';
+  public override async exec(cursorPosition: Position, vimState: VimState): Promise<void> {
+    if (!configuration.leap) return;
+    this.vimState = vimState;
+    this.searchString = vimState.leap.firstSearchString + this.keysPressed[0];
+    const markers: Marker[] = this.getMarkers(cursorPosition);
+
+    if (markers.length === 0) {
+      await this.handleNoFoundMarkers();
+      return;
+    }
+
+    // When the leapAction is executed, it needs to be logged
+    // This is to repeat the last search command
+    // As long as it is recorded, it means that the search was successfully executed once.
+    // As long as the search has been executed successfully, it will be ok when we execute "repeat last search".
+    vimState.leap.leapAction = this;
+
+    if (markers.length === 1) {
+      await this.handleOneMarkers(markers[0]);
+      return;
+    }
+
+    await this.handleMultipleMarkers();
+  }
+  private async handleMultipleMarkers() {
+    this.vimState.leap.keepMarkersBySearchString(this.searchString);
+    await this.vimState.setCurrentMode(Mode.LeapMode);
+  }
+
+  private async handleOneMarkers(marker: Marker) {
+    this.vimState.cursorStopPosition = marker.matchPosition;
+    this.vimState.leap.cleanupMarkers();
+    await this.vimState.setCurrentMode(this.vimState.leap.previousMode);
+  }
+
+  private async handleNoFoundMarkers() {
+    StatusBar.displayError(
+      this.vimState,
+      VimError.fromCode(ErrorCode.LeapNoFoundSearchString, this.searchString)
+    );
+    this.vimState.leap.cleanupMarkers();
+    await this.vimState.setCurrentMode(this.vimState.leap.previousMode);
+  }
+
+  private getMarkers(cursorPosition: Position) {
+    if (this.vimState.leap.isRepeatLastSearch) {
+      const matches = getMatches(
+        generateMarkerRegex(this.searchString),
+        this.vimState.leap.direction!,
+        cursorPosition,
+        this.vimState.document
+      );
+      this.vimState.leap.createMarkers(matches);
+      this.vimState.leap.showMarkers();
+      return this.vimState.leap.markers;
+    } else {
+      return this.vimState.leap.findMarkersBySearchString(this.searchString);
+    }
+  }
+
+  public fire() {
+    this.exec(this.vimState.cursorStopPosition, this.vimState);
+  }
+}
--- a/src/actions/plugins/leap/leap.ts
+++ b/src/actions/plugins/leap/leap.ts
@@ -1,0 +1,129 @@
+import { Mode } from '../../../mode/mode';
+import { Match } from './match';
+import { Marker, createMarker, generateMarkerNames } from './Marker';
+import { VimState } from '../../../state/vimState';
+import { LeapAction } from './LeapAction';
+
+export enum LeapSearchDirection {
+  Forward = -1,
+  Backward = 1,
+}
+
+export class Leap {
+  markers: Marker[] = [];
+  vimState: VimState;
+  previousMode!: Mode;
+  isRepeatLastSearch: boolean = false;
+  direction?: LeapSearchDirection;
+  leapAction?: LeapAction;
+  firstSearchString?: string;
+
+  constructor(vimState: VimState) {
+    this.vimState = vimState;
+  }
+
+  public createMarkers(matches: Match[]) {
+    this.markers = matches.map((match) => {
+      return createMarker(match, this.vimState.editor);
+    });
+
+    this.setMarkersName();
+
+    return this.markers;
+  }
+
+  private setMarkersName() {
+    const map: { [key: string]: Marker[] } = {};
+
+    this.markers.forEach((marker) => {
+      const key = marker.searchString;
+      if (!map[key]) {
+        map[key] = [];
+      }
+
+      map[key].push(marker);
+    });
+
+    Object.keys(map).forEach((key) => {
+      const group = map[key];
+      const markerNames = generateMarkerNames(group.length);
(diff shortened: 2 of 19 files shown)


kabouzeid:
The settings need to be contributed to the `package.json` such that eg `vim.leap=true` can be set from the GUI.

cuixiaorui:
> The settings need to be contributed to the `package.json` such that eg `vim.leap=true` can be set from the GUI.

thx your tips,  it's fixed now.

src/actions/plugins/leap/LeapAction.ts
```diff
@@ -0,0 +1,139 @@
...
+    return this.keysPressed[0] === 's' ? LeapSearchDirection.Backward : LeapSearchDirection.Forward;
+  }
+}
+
+@RegisterAction
+export class LeapAction extends BaseCommand {
+  modes = [Mode.LeapPrepareMode];
+  keys = ['<character>'];
+  override isJump = true;
+  private vimState!: VimState;
```
elazarcoh, line 74: can you get rid of the `!`?

2 suggestions for how to do it:
1. Create a handler class with the vimState as constructor parameter, and use this class.
2. Pass the vimState from function to another as parameter.

elazarcoh (changes requested):
Hi, I would love to see this PR get merge. Already built you version into my custom VSCodeVim build I'm using for my everyday work.
In the meantime, I made a review.
My main complaint is that you use `!` too freely for my taste 🙃

src/actions/plugins/leap/match.ts
```diff
@@ -0,0 +1,105 @@
...
+    const searchChars = rawSearchString.split('').map(escapeString);
+
+    const firstChar = searchChars[0];
+    const secondChar = searchChars[1];
+    let pattern = '';
+    if (secondChar === ' ') {
+      pattern = firstChar + '\\s' + '|' + firstChar + '$';
+    } else {
+      pattern = firstChar + secondChar;
+    }
```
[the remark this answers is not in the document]
cuixiaorui: haha I forget to refactor this code, Thank you for reminding me

src/actions/plugins/leap/LeapAction.ts
cuixiaorui: Haha thank you for the advice
As you said, this code-style problem

kabouzeid:
in visual mode `x` and `X` don't jump to the right character. and `s` and `S` are not there at all.

cuixiaorui:
@kabouzeid  I just found this detail

The x in visualization mode does not contain the match, while the s in visualization mode contains both characters of the match
But now the problem is that S is used by the surround plugin.
How can I solve this key conflict problem? Do you have any good ideas?

cuixiaorui:
fixed the bug that in visual mode x and X don't jump to the right character

> in visual mode `x` and `X` don't jump to the right character. and `s` and `S` are not there at all.

Leiyi548:
@cuixiaorui leap.nvim can be set to support two-way search, do you plan to support it later?

cuixiaorui:
> @cuixiaorui leap.nvim can be set to support two-way search, do you plan to support it later?

Is there a description document for this function? I'll look into it.

Leiyi548:
@cuixiaorui 
image
more information please see https://github.com/ggandor/leap.nvim#calling-leap-with-custom-arguments

cuixiaorui:
> @cuixiaorui image more information please see https://github.com/ggandor/leap.nvim#calling-leap-with-custom-arguments

Haha
Interesting feature 
Sometimes I really don't like to use `S` because I have to press one more `shift`
I will implement this feature when I have time in a few days
I will update the documentation and then

Leiyi548:
> Yes, sometimes I have to think in my head whether to press shift or not, and it bothers me.

cuixiaorui:
@Leiyi548  
vim_leap_search

I finally finished implementing this function yesterday
You can use it to see
If you have problems, just give me feedback!

set `vim.leapBidirectionalSearch": true` enable the feature

Shaker-Pelcro:
Would love to see this merged.

package.json
```diff
@@ -1190,6 +1215,7 @@
     "ts-loader": "9.4.2",
     "tslint": "6.1.3",
     "typescript": "4.9.5",
+    "vitest": "^0.29.2",
```
J-Fields, line 1218: I think this is unintended?

src/actions/plugins/leap/LeapAction.ts
J-Fields: Yeah I'd prefer `vimState` is passed explicitly to each method

src/state/vimState.ts
```diff
@@ -61,6 +62,8 @@ export class VimState implements vscode.Disposable {
 
   public easyMotion: IEasyMotion;
 
+  public leap!: Leap;
```
J-Fields, line 65: Can we pull this out into `ModeData`?

src/actions/plugins/leap/leap.ts
```diff
@@ -0,0 +1,238 @@
...
+
+    this.setMarkersName();
+
+    return this.markers;
+  }
+
+  private setMarkersName() {
+    const map: { [key: string]: Marker[] } = {};
+
+    this.markers.forEach((marker) => {
```
J-Fields, line 41: Let's use a `Map` here, rather than a simple object - I think it's cleaner than relying on `Object.keys` for instance

src/actions/plugins/leap/leap.ts
```diff
@@ -0,0 +1,238 @@
...
+      return createMarker(match, this.vimState.editor);
+    });
+
+    this.setMarkersName();
+
+    return this.markers;
+  }
+
+  private setMarkersName() {
+    const map: { [key: string]: Marker[] } = {};
```
J-Fields, line 39: Let's use a `Map` here, rather than a simple object - I think it's cleaner than relying on `Object.keys` for instance

test/plugins/leap.test.ts
```diff
@@ -0,0 +1,144 @@
+import { Configuration } from '../testConfiguration';
+import { newTest } from '../testSimplifier';
+import { cleanUpWorkspace, setupWorkspace } from '../testUtils';
+
+function leapCommand(key: 's' | 'S', searchStrings: string[], jumpKeys: string[] = []) {
+  return [key, ...searchStrings, ...jumpKeys].join('');
+}
+
+suite('leap plugin', () => {
```
J-Fields, line 9: I'd like to see some more test cases - across lines for instance

cuixiaorui: Okay, I'll add a more comprehensive test case.

Leiyi548:
@cuixiaorui Can you add the option to uncheck visual mode x leap search, because most of the time I checked x for delete

cuixiaorui:
Hi @J-Fields @elazarcoh
thank you very much for taking the time to review Leap's code. 
I have just fixed all the issues mentioned above. In addition, I have also updated the following features:

1. Added the functionality of bidirectional search.
3. Supported marker background color and font color configuration.
4. Refactored the way to get Leap (removed it from vimState).

Please feel free to contact me if you have any questions. 
I really like vscodeVim and promote it in many places. 
I sincerely hope that vscodeVim will become better and better.

MaqingOnes:
I have a suggestion, make the first match always be "j",  or have a setting where user can choose what character they want to jumo to  the first match point. (you know they did similar stuff in neovim ,in neovim they support command ,  but that's whole another thing though)
eg: use  "Sse" here , "j" will get me to the "use" of this line since in this line we have the first match. am I making myself clear?

MaqingOnes:
> I have a suggestion, make the first match always be "j", or have a setting where user can choose what character they want to jumo to the first match point. (you know they did similar stuff in neovim ,in neovim they support command , but that's whole another thing though) eg: use "Sse" here , "j" will get me to the "use" of this line since in this line we have the first match. am I making myself clear?


You know what maybe make the first match the **last command character**
if the command = "sse" make the firt match jumping character is e, if command = "sid" make the first match jumping character is d. dynamic.

cuixiaorui:
> I have a suggestion, make the first match always be "j", or have a setting where user can choose what character they want to jumo to the first match point. (you know they did similar stuff in neovim ,in neovim they support command , but that's whole another thing though) eg: use "Sse" here , "j" will get me to the "use" of this line since in this line we have the first match. am I making myself clear?

You can set vim.leap.labels to modify the default value of "sklyuiopnm,qwertzxcvbahdgjf;", where "s" is the first jumping point.

wroyca:
Thank you for providing this plugin! While testing your branch, I encountered an issue where attempting to transition to an uppercase character does not work as expected:

[Screencast from 2023-05-13 10-51-01.webm](https://github.com/VSCodeVim/Vim/assets/91024200/3d9821a0-731e-4a57-8259-5ba627bfdf13)

Eitheages:
This PR wasn't merged yet?

archilkarchava:
@cuixiaorui, thanks a lot for the plugin implementation.
I found a small issue. When multiple cursors are present and the leap plugin is invoked the extension gets stuck in Leap mode and it's not possible to move the cursor until the file is closed and reopened. The issue is reproducible with multiple cursors both in normal and visual modes.

https://github.com/VSCodeVim/Vim/assets/16840190/722190aa-f10a-4a64-87c2-69abe870d4a4

archilkarchava:
@cuixiaorui, the leap action is dot-repeatable, which is not desirable. Here's how we can exclude it from being marked as dot-repeatable https://github.com/archilkarchava/Vim/commit/12da1ca1d4a0e19283056808f329d2c598bec3b7.

archilkarchava:
Also, I've discovered that the <s-Enter> to execute the previous search doesn't work with the `vim.leap.bidirectionalSearch` enabled.

dykwok:
@J-Fields is this PR okay to merge?

clamydo:
Love it!

For further reference and inspiration, if required, there is yet another alternative https://github.com/folke/flash.nvim, that also features remote actions, which I find rather ingenious.

Remote actions work like that:

1. Cursor is at position A
2. Start remote action: for example for yanking remotely press `yr` and then leap to a position B
3. Do a motion, e.g. `iw` or another leap
4. Cursor jumps back to A

cuixiaorui:
> Love it!
> 
> For further reference and inspiration, if required, there is yet another alternative https://github.com/folke/flash.nvim, that also features remote actions, which I find rather ingenious.
> 
> Remote actions work like that:
> 
> 1. Cursor is at position A
> 2. Start remote action: for example for yanking remotely press `yr` and then leap to a position B
> 3. Do a motion, e.g. `iw` or another leap
> 4. Cursor jumps back to A

Soon I will implement flash-like functionality in vscodeVim! 
Progress will be synchronized to this [issue](https://github.com/cuixiaorui/VSCodeVim/issues/6)

archilkarchava:
@cuixiaorui, thanks for your kind words. I've been using this plugin ever since and it's been a major improvement to my workflow.
I found another small issue: all matches get a label. This includes cases where a search string has a single match.
These labels only confuse the user since the cursor will jump to the match automatically without pressing a label key.

We can fix the issue by not assigning a label in case a search string only has a single corresponding match: https://github.com/archilkarchava/Vim/commit/318fd31049589a211a5a0615a559c7a6467666d7.

### Before the fix

https://github.com/VSCodeVim/Vim/assets/16840190/f4ca8cfa-0c45-4e59-9796-5cd3ea64704e

### After the fix

https://github.com/VSCodeVim/Vim/assets/16840190/4b5f7f70-553c-41d5-b974-93eb9c94444a

cooljser:
Why this PR not be merged? This is a realy used plugin for vim user.

Edgar-Avila:
I'd love to see the PR merged, leap is a great plugin.

Shaker-Hamdi:
Any updates on this PR? It's been so long, and it's a much-needed plugin.

wires5210:
+1 to wanting to see this merged, leap is by far my favorite nvim plugin

darianmorat:
Waiting for this PR merged we need leap I'm vscodevim... For now I'll use the edited extension that I forked

dz0039:
Can we get this merged? This extension could really benefit from something more than easymotion.

clovu:
@J-Fields Hi, has there been any progress on this PR?

PR #8238, opened 2023-02-08
readme: key repeat for VS Codium

asher-gh:
The previously suggested command didn't work for me. However, this [SO answer](https://stackoverflow.com/a/73897433/14616597) worked.

Tested with **VS Codium 1.75.0** (installed with HomeBrew) on **MacOS 13.2**.


J-Fields:
Thanks!

Issue #1876, opened 2017-06-27
Save As with `:w <filename>`


ghost:
* Click *thumbs-up* 👍 on this issue if you want it!
* Click *confused* 😕 on this issue if not having it makes VSCodeVim unusable. 

The VSCodeVim team prioritizes issues based on reaction count. 

--------

**FEATURE REQUEST** 

The normal `:w` only works with files that have already been saved. If I open a new undefined file and want it to save as in the currently active folder `:w filename.txt` does nothing. It would be nice if this would just follow the usual save as procedure.

skosch:
Not to dismiss this idea at all, but in the meantime, for those looking to save files under an explicit name there already is `:sav <filename>` as well.

con-ssc:
Ah, very good. That would already suffice for me.

skosch:
**RETRACTION – I eat my words!** :disappointed_relieved:   `:sav` does *not* work either (although I promise it does in vim!) ... the only thing that seems to do it for new files is to type `:w` without a filename, hit enter, and use the "Save File" dialog that pops up.

This really would be nice to have!

con-ssc:
Oh okay. Well that I knew already was the reason for the issue :P

skosch:
... or rather, the real bug is that `:sav <filename>` does indeed save the file, but the tab's path isn't set to the filename, so subsequent `:w`'s don't do anything.

asyks:
Any update on this? Not having `:sav <filename>` support is somewhat crippling.

PR #8270, opened 2023-03-01
Added option to "Save As" with `:w <filename>`

JLMSC:
Added a feature to the :w[rite] command, allowing the option to "Save As" instead of just saving the current file.

Fixes the issue #1876


J-Fields (approved):
LGTM. Thanks!

PR #8281, opened 2023-03-04
Fix issue with shift + mouse selection from bottom to top

lqqyt2423:
When the cursor is at the end of a line, and I use the shift key and mouse to select an area above it, I encounter a problem where the selection cannot be made. The reason for this is that the selection range in the code is covering the wrong direction. This commit fixes this issue.

### DIFF
--- a/src/mode/modeHandler.ts
+++ b/src/mode/modeHandler.ts
@@ -371,7 +371,7 @@ export class ModeHandler implements vscode.Disposable, IModeHandler {
         // If we prevented from clicking past eol but it is part of this selection, include the last char
         if (this.lastClickWasPastEol) {
           const newStart = new Position(selection.anchor.line, selection.anchor.character + 1);
-          this.vimState.editor.selection = new vscode.Selection(newStart, selection.end);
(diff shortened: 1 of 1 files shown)


J-Fields:
Can you please describe in a bit more detail the bug this fixes?

lqqyt2423:
First, click the mouse to move the cursor to the end of the line, and then hold down the shift key while clicking the mouse to select the other lines.

When using the selection method above, selecting from top to bottom works correctly.

However, when selecting from bottom to top, there is a calculation error in the code that prevents successful selection.

Here is the code:

```js
const newStart = new Position(selection.anchor.line, selection.anchor.character + 1);
this.vimState.editor.selection = new vscode.Selection(newStart, selection.end);
```

When selecting from bottom to top, `selection.anchor` and `selection.end` are the same, causing the selection to fail. To fix this, replace `selection.anchor` with `selection.active`.

Based on the code, `selection.anchor` should appear in pairs with `selection.active`.

J-Fields (approved):
Thanks for elaborating. LGTM!

PR #8323, opened 2023-03-20
add :m[ove]

zhanyi22333:
Add "move" excommand.
For excommand like : a,bmc  (a,b,c are line numbers), it works as follow:
1. Copy text in range [a,b].
2. Paste text below c, in range[c+1,c+1+(b-a)].
3. delete text in  range [a,b].

Futhermore, I just ban the situation that range [a,b] and range[c+1,c+1+(b-a)], which will lead to wrong performance and has no practical use. I also ban the situation that move [a,b] to [a,b], in which case the "move" excommand just do useless work.

I'm just a newbie in PR.Let me know if there is something not in convention.


src/cmd_line/commands/move.ts
J-Fields, line 1: Can you write a few test cases for `:move` please?

J-Fields (changes requested):

src/cmd_line/commands/move.ts
zhanyi22333: Ok.I'll work on this.
By the way,ls there any doc for writing test or a good sample test script?

J-Fields: `sort.test.ts` is a decent example

zhanyi22333:
I've add a test script for m[ove].During the process, I compare its behavior with orignal vim and find some misunderstanding. So I change the workflow and also change the ban situation.
Now, the performannce is finetuned like this:
For excommand like : a,bmc (a,b,c are line numbers), it works as follow:

1. Copy text in range [a,b].
2. Find target position  by c and decide cursor position.
3. Delete text in range [a,b].
4. Paste text to target positon and move cursor to cursor position.

Ban the situtations that  move a range into itself and the situations that move to adress just right abow or below the range.

All test scripts are passed on my pc. If there are some problems with the added test script, let me know. Thanks.

zhanyi22333:
There are two tests failed when build on windows-lastest. I've re-test on my pc and can not reproduce them. And the failed tests seem no relationship with m[ove]. I feel confused. Should I do something for the failed tests?

J-Fields:
> Should I do something for the failed tests?

Nah don't worry about it; a few tests are just flaky

J-Fields (approved):
Thanks!

Miigon:
Thank god this is implemented!

PR #8328, opened 2023-03-23
fixes #7022

### LINKED ISSUE #7022: `s` is not working properly in visual mode
**Describe the bug**
`s` is not working properly, the vim editor mode does not enter `Insert` mode immediately after pressing `s`.

`v1.21.5` wors fine.

**To Reproduce**

1.  In `Visual` mode pressing `s`
2.  Continue to input `s`, a character after the cursor will be deleted
3.  This seems to be because edit mode did not enter insert mode

listenerri:
related oprations/actions: 
- s
- c

issue: #7022

Fix ChangeOperators not enter InsertMode when autoSwitchInputMethod is enabled.

#7022

I found that there are other codes that use the `vimState.setCurrentMode` / `vimState.setModeData` interface, and don't write the `await` keyword, but I don't know whether it needs to be modified, such as:

https://github.com/VSCodeVim/Vim/blob/2d703280762663e8d8567e603f1e178f8ead0453/src/actions/commands/actions.ts#L2349
https://github.com/VSCodeVim/Vim/blob/2d703280762663e8d8567e603f1e178f8ead0453/src/actions/commands/insert.ts#L246
https://github.com/VSCodeVim/Vim/blob/2d703280762663e8d8567e603f1e178f8ead0453/src/actions/commands/insert.ts#L258

### DIFF
--- a/src/actions/operator.ts
+++ b/src/actions/operator.ts
@@ -675,7 +675,7 @@ export class ChangeOperator extends BaseOperator {
       vimState.recordedState.transformer.delete(deleteRange);
     }
 
-    vimState.setCurrentMode(Mode.Insert);
+    await vimState.setCurrentMode(Mode.Insert);
   }
 }
 


J-Fields:
Nice catch, thanks!

> I found that there are other codes that use the `vimState.setCurrentMode` / `vimState.setModeData` interface, and don't write the `await` keyword, but I don't know whether it needs to be modified

These other instances may be less likely to cause issues, but I do think they should be `await`ed as well. Want to add those to this PR too?

listenerri:
> These other instances may be less likely to cause issues, but I do think they should be `await`ed as well. Want to add those to this PR too?

I modified those previously mentioned places, and did a little testing of the changes locally, and everything seems to be ok :)

PR #8077, opened 2022-11-03
Make `(`/`[`/`{`/`T`/`F` surrounds on visual line indent wrapped lines

mogelbrod:
**What this PR does / why we need it**:

The surround plugin implementation is mostly great, but one thing I miss is a way to automatically indent the surrounded text in visual line mode. To my knowledge this wasn't supported in the original vim plugin either, but I'd argue that the current behaviour isn't exactly intuitive:

```
one
tw|o
three
```
`VS{` results in:
```
one
{ 
two
 }
three
```

This PR changes the visual line mode behaviour of the surround keys that usually wrap the content with a space (`(`/`[`/`{`/`T`/`F`) to instead indent the surrounded line(s). The indentation for the delimeters are determined from the first line, with surrounded lines being indented one more level.

Surrounds with `)`/`]`/`}`/`t`/`f` behave as before, as well as regular visual mode surrounds.

I also added <kbd>Enter</kbd> to the list of supported surround keys, which will wrap the selection in a multiline curly brace block (even outside visual line mode).

<details><summary><h3>Additional examples</h3></summary>

```
one
tw|o
three
```
`VS{` results in:
```
one
{
  two
}
three
```

```
<a class="nav" href="/">Home</a>
|<a class="nav" href="/contact">Contact</a>
```
`VkSTnav<Enter>` results in:
```
<nav>
  <a class="nav" href="/">Home</a>
  <a class="nav" href="/contact">Contact</a>
</nav>
```

```
one,
|two,
```
`VkSFconsole.log<Enter>` results in:
```
console.log(
  one,
  two,
)
```

```
one tw|o three
```
`ysiw<Enter>` results in:
```
one {
  two
} three
```

</details>

**Which issue(s) this PR fixes**

- Desired indentation in https://github.com/VSCodeVim/Vim/issues/7103

**Special notes for your reviewer**:

Two tests are failing - I'm struggling to get the cursor position to be where expected, would love to hear some ideas on how to solve it. I've seen that cursor position after surround actions have been problematic in general going by the number of issues referencing it.

PR #8336, opened 2023-03-26
Enable sourcemaps in dev builds

mogelbrod:
Webpack sourcemaps were disabled in f02ed71ed848d8aca2e1fef77d8d2490a3b9e1e7, which inadvertedly also stopped vscode from picking up sourcemaps when debugging the extension. This resulted in vscode breakpoints not working at all.

Couldn't find any reported issues regarding this - I encountered it while working on https://github.com/VSCodeVim/Vim/pull/8077.


J-Fields:
Thanks!

PR #8352, opened 2023-04-04
Clear search highlight on normal hjkl movement

### LINKED ISSUE #5528: Option to disable search highlight after cursor is moved
This is a small change, but it'd be nice if you could clear the search highlight after moving the cursor (i.e., hitting any other key other than `n`/`N`). I couldn't figure out a good way to do this via keymappings, so if anyone has ideas I'm all ears. Things I've tried:

- Map various movement keys to the same key + `:noh` (this doesn't perform very well and can sometimes give unexpected results, and isn't needed like 80% of the time so it feels way too hacky/inelegant)
- Map `ESC` to `ESC` + `:noh` - this is where I'm at now, but again, it doesn't feel very elegant

Thanks!

shlomi-aknin:
Clears search highlight on normal mode hjkl movement
fixes #5528 

### DIFF
--- a/src/actions/motion.ts
+++ b/src/actions/motion.ts
@@ -291,6 +291,10 @@ class MoveDown extends BaseMovement {
   override preservesDesiredColumn = true;
 
   public override async execAction(position: Position, vimState: VimState): Promise<Position> {
+    if (vimState.currentMode === Mode.Normal) {
+      globalState.hl = false;
+    }
+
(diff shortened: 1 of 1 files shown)


J-Fields:
Sorry, but this shouldn't be the default behavior and I'm reluctant to add an setting for it specifically since I think it's fairly niche

Issue #8390, opened 2023-04-23
Update sentence definition to allow for closing quotes/parens/squarebrackets


matthew-elgart:
Text to reproduce behavior:

`"First sentence." Second sentence. (Third sentence.) [Fourth sentence.] 'Fifth sentence.'`

In vim, `is`, `as`, and `(` / `)` treat these as 5 separate sentences. But in vscodevim, these are treated as only 2, because punctuation `.!?` followed by any non-whitespace character does not count as a sentence end.

From poking around in the vim source, I believe that the allowed characters after punctuation are `"`, `'`, `)`, and `]` ([link](https://github.com/vim/vim/blob/887748742deae3d6de7aa0fdbb042afe1ccf5e7a/src/textobject.c#L114)).

So the proposed fix would be to update [sentenceEndRegex](https://github.com/VSCodeVim/Vim/blob/master/src/textobject/sentence.ts#L6) to something like `/[\.!\?]["')\]]*?([ \n\t]+|$)/g`.

I'm not sure of the preferred process for changes, but if this update is desired I'd be happy to make the change!

PR #8400, opened 2023-04-29
update sentence definition to allow for closing quotes/parens/squarebrackets

matthew-elgart:

This change updates the definition of a sentence to be more in line with vim's definition.

Today, a sentence end is a punctuation character (`.`, `?`, `!`) followed by whitespace/line end - which means that something like `(potential sentence end.)` doesn't count as a sentence end (even though it does in vim).

With this change, we allow a sentence end to include any amount of closing: 1. single quotes; 2. double quotes; 3. parentheses; 4. square brackets.

#8390 

none

### DIFF
--- a/test/motion.test.ts
+++ b/test/motion.test.ts
@@ -539,7 +539,6 @@ suite('sentence motion', () => {
       motion = motion.getSentenceBegin({ forward: true });
       assert.strictEqual(motion.line, 11);
       assert.strictEqual(motion.character, 96);
-
     });
   });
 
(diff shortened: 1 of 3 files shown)


test/motion.test.ts
```diff
@@ -515,11 +517,30 @@ suite('sentence motion', () => {
       assert.strictEqual(motion.character, 0);
     });
 
-    test('next sentence when paragraph contains a line of whilte spaces', () => {
+    test('next sentence when paragraph contains a line of white spaces', () => {
```
matthew-elgart, line 520: There were a few errors in test descriptions that I fixed while I was here.

J-Fields (approved):
Looks good to me. Thanks a lot!

Issue #7606, opened 2022-04-05
Add "extensions.experimental.affinity" to doc


haoadoreorange:
In latest [Insiders Build](https://code.visualstudio.com/insiders/) it's possible to run vim extension in dedicated host process, which is supposed to fix some problems where vim is stall due to other running extensions. Maybe we should add this to the doc ?

See the [announcement](https://github.com/microsoft/vscode/issues/65876#issuecomment-1078827200).

lawrence-laz:
This improved responsiveness by a lot on my machine.

mezdelex:
It was night and day for me, definitely a must have in your settings.json if you're using VSCodeVim

joeytwiddle:
May 2023, this feature is stable for me too. I also recommend adding it to the docs.

It could be hugely beneficial for users (especially those with slower machines).

PR #8424, opened 2023-05-15
Document the affinity setting for improved performance

joeytwiddle:

Adds some docs about the affinity setting, which can enhance the performance of VSCodeVim, and could be beneficial to many users.

The setting has been praised [here](https://github.com/VSCodeVim/Vim/issues/7606) and [here](https://github.com/microsoft/vscode/issues/75627#issuecomment-1078827311).

#7606

I wanted to display this near the top of the README, because I think many people may want to consider this setting. But I put more detailed documentation lower down, in the F.A.Q.


TimFelixBeyer:
I think you should change `"vscode.vim": 1` to `"vscodevim.vim": 1`, as the first option doesn't work (at least on my machine).

joeytwiddle:
OK, I think you are right. Although I don't understand how I have been using the wrong setting for so long! :laughing:

I have force pushed the fix.

J-Fields (approved):
Thanks!

Issue #8200, opened 2023-01-08
Cursor jumps to the beginning of the file after switching from another one


rivaldi8:
**Describe the bug**
Sometimes, after switching from one file to another (through tabs), pressing a key makes the cursor jump to the beginning of the file.

**To Reproduce**
I haven't found yet what's triggering the issue. However, once it happens, it can be reproduced like this:

1.  One a file editor, move away from the beginning of the file (row 0, column 0, if you are there), either with the keyboard or with the mouse. For example, press 'j' to move down one line.
2.  Switch to another file tab with the keyboard or mouse.
3.  Switch back to the previous file.
4.  Pressing any key (Esc included) but modifiers (Ctrl, Alt...) jumps to the beginning.
5. Closing and reopening the file fixes it.

**Expected behavior**
The cursor should remain in the same position.

**Environment (please complete the following information):**

- Extension (VsCodeVim) version: 1.24.3
- VSCode version: 1.74.2
- OS: Linux

**Additional context**
This issue appeared after the upgrade to VS Code 1.74, so it may be related to #8157, although a file affected by the former might not be by the latter.

Vanaldito:
I have the same problem and I found a way to reproduce it.

In a file with enough lines (in my case 80 are enough). If you go to the last line and start going up by holding k to the first line. Then, when changing to another file and returning to the original, the problem will happen.

Sometimes it is necessary to repeat the process again to cause the bug.

@rivaldi8, could you confirm if the same thing happens to you?

rivaldi8:
@Vanaldito I confirm I can reproduce following your steps. Although it doesn't happens always, it does frequently enough.

It seem something related with the scrolling and hitting the bottom of the file. Sometimes, just hitting the bottom triggers the issue. I also found it doesn't always jump to the first line. Sometimes it goes to the second or third instead.

Strangely in my other computer, with similar settings and extensions, I can't seem to reproduce any of these bugs. I have to investigate if there's some setting that's causing it.

By the way, I just checked with VS Code 1.74.3 and the bug is still there.

nviennot:
Same. It's awful.

pixeldin:
Any ideas or progress? looking forward to solving:)

basepi:
I'm was newly hitting this as of today. I think I updated to 1.75 yesterday, but I'm not sure from which version. For me it was happening 100% of the time when switching tabs between files.

[...]

PR #8426, opened 2023-05-16
Dont let the pesky (0,0) jump after the go to definition action.

magniff:
"Fixing" an extremely annoying bug.

Addressing the Annoying Cursor Bug Affecting Go to Definition Actions.
You can find a ton of already existing chats, mentioning that specific bug, e.g. https://github.com/VSCodeVim/Vim/issues/8200

This is a hack that seems to be working, but, we need a better solution for sure.

**Details:**
The bug is hard to reproduce on purpose, it seem to just appear after 5-10 minutes of intense using of the plugin. At some point, after an yet another jump action (usually after the jump to definition), when you jump back by pressing `ctrl+o` [here](https://github.com/VSCodeVim/Vim/blob/master/src/mode/modeHandler.ts#L137) `vscode` returns a single `([0, 0] | [0, 0])` selection, which causes the mess. Have no idea why, maybe the bug of the `vscode` itself?


magniff:
@Chillee or someone else, review please. The bug in question is making the user experience unbearable. Thanks!)

FrankFang:
Why not merge the pr?

Gredelston:
cc @J-Fields, can you please review this PR? This fixes a bug that makes the Vim extension quite unpleasant to use.

src/mode/modeHandler.ts
```diff
@@ -135,6 +136,13 @@ export class ModeHandler implements vscode.Disposable, IModeHandler {
   public syncCursors() {
     // TODO: getCursorsAfterSync() is basically this, but stupider
     const { selections } = this.vimState.editor;
+    if (
+      selections.length == 1 &&
```
J-Fields, line 140: Needs to be `===`

src/mode/modeHandler.ts
```diff
@@ -135,6 +136,13 @@ export class ModeHandler implements vscode.Disposable, IModeHandler {
...
+      selections[0].isEqual(new Range(new Position(0, 0), new Position(0, 0)))
+    ) {
+      return;
```
J-Fields, line 143: Let's put a `TODO` here referencing this MR

J-Fields (approved):
Thanks!

PR #8460, opened 2023-06-07
b[uffer] parser and tests

sudo97:
I tried :b2, saw PRs are welcome, so here we are

**Special notes**:
I love constructive criticism, I just want to have this command, so please let me know if anything needs to change.

I tried couple commands, I think it doesn't break anything, but maintainers know better


src/cmd_line/commands/tab.ts
```diff
@@ -142,7 +145,7 @@ export class TabCommand extends ExCommand {
         if (this.arguments.count !== undefined && this.arguments.count >= 0) {
           await vscode.commands.executeCommand(
             'workbench.action.openEditorAtIndex',
-            this.arguments.count
+            this.arguments.count - 1
```
J-Fields, line 148: Looks like `{count}gt` uses this same code path, so I think we need to adjust the offset [here](https://github.com/VSCodeVim/Vim/blob/1fc416d0513144676d0484ce985eb4b6a07ccfc6/src/actions/commands/actions.ts#L1319) to avoid breaking it

J-Fields (approved):
Thanks!

sudo97:
Thanks for helping me to wrap this up

Issue #2747, opened 2018-06-19
[FEATURE REQUEST] Add an option to support `:set selection=exclusive`


tiansh:
**Is this a BUG REPORT or FEATURE REQUEST?** (choose one): FEATURE REQUEST

Add an option to support `:set selection=exclusive`

With `:set selection=exclusive`, starting with an empty file, press

```
iHello world!␛^vwd
```

`world!` should be kept on the screen instead of `orld!`.

RichardBronosky:
Wow! That's a really good vim setting I didn't know about. With `:set selection=exclusive` you get the same result from `^vwd` that you would get from `^dw`. And that is what a logical person would expect. I love it!

sfinktah:
It's a far more intuitive setting.  Otherwise when you visually select something, it deletes an extra character at the end.

When can we have it for vim?  It's in Visual Studio's vsvim and JetBrains vim extension.  Vim becomes almost unusable for me without it.

AndyP3r3z:
Any update on this? I can't live without my beloved exclusive selection 🫠

sfinktah:
I can't believe this one is still outstanding.  Such a small thing, yet it this is exactly the sort of tiny behavioural discrepancy that becomes intolerable once your fingers have learned it.  I wonder if one of these blessed AIs can make it.

okay, well... apparently it can.  

https://github.com/VSCodeVim/Vim/pull/8464 <-- PULL REQUEST... Like and subscribe or contact your local senator or whatever works for you.

PR #8464, opened 2023-06-08
set selection=exclusive

sfinktah:
In answer to feature request https://github.com/VSCodeVim/Vim/issues/2747

**What this PR does / why we need it**: Because visual selection in vim is _retarded_ without `set selection=exclusive`.

~**Special notes for your reviewer**: This PR only partially fixes the issue.  It correctly alters the start of the selection as per traditional vim in `selection=exclusive` mode and *appears* to alter the end of the selection also, but that appearance is deceptive.  Can someone with more experience offer an opinion?~

~TL;DR:
`stop = stop.getLeft();` change the visual appearance of the selection, but not the actual selection.~

~postnote:  this has all been implemented as `"vim.selectionExclusive": true`, rather than the true vim selection=inclusive|exclusive|old`.  Would be happy to rectify that, if/when above issue is resolved.~

### DIFF
--- a/package.json
+++ b/package.json
@@ -1171,6 +1171,16 @@
         "vim.langmap": {
           "type": "string",
           "description": "Language map for alternate keyboard layouts. When you are typing text in Insert (or Replace, etc.) mode, the characters are inserted derectly. Otherwise, they are translated based on the provided map."
+        },
+        "vim.selection": {
+          "type": "string",
+          "enum": [
+            "old",
+            "inclusive",
+            "exclusive"
+          ],
+          "description": "Controls whether the final character of a Visual mode selection is included in an operation.",
+          "default": "inclusive"
         }
       }
     },
--- a/src/configuration/configuration.ts
+++ b/src/configuration/configuration.ts
@@ -53,6 +53,7 @@ export const optionAliases: ReadonlyMap<string, string> = new Map<string, string
   ['scr', 'scroll'],
   ['so', 'scrolloff'],
   ['scs', 'smartcase'],
+  ['sel', 'selection'],
   ['smd', 'showmode'],
   ['sol', 'startofline'],
   ['to', 'timeout'],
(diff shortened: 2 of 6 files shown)


sfinktah:
## Updated

The earlier version of this PR was incomplete. Internally, operators still received the original inclusive range and duly modified the supposedly excluded character.

This version corrects those problems.

* `ModeHandler.updateView()` renders character-wise Visual selections as half-open when `selection=exclusive`.
* `ModeHandler.executeOperator()` adjusts the internal range before handing it to operators, so the text highlighted and the text modified are finally describing the same reality.
* Forward and backward selections now behave consistently, an achievement that should not have required special mention.
* A single-character Visual selection remains inclusive, matching Vim’s refusal to permit an empty Visual selection.
* The provisional `vim.selectionExclusive` boolean has been discarded in favour of the actual string option:

  ```json
  "vim.selection": "exclusive"
  ```
* The option is also configurable through Vim’s Ex syntax:

  ```vim
  :set selection=exclusive
  :set sel=exclusive
  ```

Supported values are `old`, `inclusive`, and `exclusive`. `inclusive` remains the default.

### Remaining caveat

`selection=old` is accepted, but currently behaves identically to `selection=inclusive`. 

Vim distinguishes the two mainly in cases where the cursor may move beyond the end of a line. This PR does not attempt to reproduce those edge cases. Its purpose is to resolve issue #2747 and make `selection=exclusive` function correctly, not to recreate every historical eccentricity of Vim’s selection semantics in one sitting.

## Testing

Regression coverage includes:

* Rendering an exclusive selection without allowing it to collapse into emptiness.
* Forward character-wise deletion.
* Backward character-wise deletion.
* Vim’s single-character inclusive exception.
* Changing the option via `:set selection=exclusive`.

Validation completed without incident:

* Focused regression tests: **5 passing**
* Complete Visual-mode test family: **300 passing, 2 pre-existing pending**
* TypeScript compilation: passing
* ESLint: passing
* Prettier: passing

PR #8505, opened 2023-07-19
Fix tab keybinding from blocking Find menu tab-navigation

### LINKED ISSUE #7000: Tabbing through VSCode's built-in find and replace menu also runs whatever command is mapped to `tab` in the editor
**Describe the bug**
After remapping `tab` and `shift+tab` in vscodevim's normal mode, opening VSCode's built-in find and replace functionality (ordinarily mapped to `ctrl+h`) and tabbing from the 'find' text box to the 'replace' text box will also run whatever Vim command that `tab` is mapped to into the editor that called find/replace.

**To Reproduce**
As an example, I had the following remaps in my `settings.json`: 

```
  ...
  "vim.leader": " ",
    // Find and replace
    { "before": ["<leader>", "r"], "commands": ["editor.action.startFindReplaceAction"] }
    // Handle tab key indentation in VSCode as terminals treat ctrl-i and tab the same
    { "before": ["<tab>"], "commands": [">", ">"] },
    { "before": ["<S-tab>"], "after": ["<", "<"] }
   ...
```

With this in place, in normal mode I can press the `tab` key to indent immediately. But when I then open up the find and replace window, search a 'find' term and hit tab (at this point focus still on the find and replace window), the find/replace window does tab through to the 'replace' text box as expected, but a side effect is that the actual tab command (in my case, indentation via `>>`) is also called in the editor. Consistently tabbing through the find and replace menu will then keep indenting the line in the editor that I was last on.

This also appears to happen when calling find and replace from visual mode, if `tab` is remapped there too:

```
    // Line shifting is better handled by VSCode
    { "before": ["J"], "commands": ["editor.action.moveLinesDownAction"] },
    { "before": ["K"], "commands": ["editor.action.moveLinesUpAction"] },
    // Visual mode indentation
    { "before": ["<tab>"], "after": [">", "g", "v"] },
    { "before": ["<S-tab>"], "after": ["<", "g", "v"] },
    { "before": ["u"], "after": ["<Esc>", "u"] },
    // Find and replace
    { "before": ["<leader>", "r"], "commands": ["editor.action.startFindReplaceAction"] },
  ],
```

In visual mode with the above settings, the visual selection is indented when tabbing through the find and replace menu.

**Expected behavior**
When focus is on the search/find and replace windows build into VSCode, hitting tab should cycle through the window's options without also running the command that tab is mapped to.

Note that this bug is not present for other built-in VSCode windows such as the quick open menu, the command palette or the tasks menu.

**Environment**
- Extension (VsCodeVim) version: v1.21.6
- VSCode version: 1.59.0, x64
- OS: Kubuntu Linux

**Other notes**
It may or may not be relevant that there's a VSCode keybinding `when:` condition that exists to validate focus on the find menu - `findInputFocussed`.

devrelm:

See https://github.com/VSCodeVim/Vim/issues/7000#issuecomment-1642352941

In short, the extension's `tab` and `tab+shift` keybindings are checking for `editorFocus` instead of `editorTextFocus`.

`editorFocus` is [defined as](https://code.visualstudio.com/api/references/when-clause-contexts#available-context-keys):

> An editor has focus, either the text _**or a widget**_.

_emphasis mine_

Because `editorFocus` is `true` when the user tabs in the editor-level Find menu (or at the very least it's `true` after tabbing once to focus the "Match Case" button) the user cannot tab past "Match Case" to "Match Whole Word", "Use Regular Expression", or beyond.

This PR updates the keybindings' `where` clause to use `editorTextFocus` in place of `editorFocus`.

fixes #7000


dominicpalmer:
Nice. @J-Fields, please can this be merged?

J-Fields:
Thanks a lot!

PR #8510, opened 2023-07-21
Do not revert to a single cursor when escape is pressed after a block selection

### LINKED ISSUE #7008: Exiting Insert mode loses multiple cursors when Insert mode was entered from Visual mode via <Shift + i> or <Shift + a>
**Describe the bug**
When selecting multiple occurrences of a word or selection via `Ctrl + D`, then entering Insert mode with `c`, then entering Normal mode via `Escape`, the multiple cursors are preserved (great!). However if Insert mode was entered via `Shift +i` or `Shift + a`, then the additional cursors are lost. In both cases Insert mode was entered directly from Visual mode but afterwards they behave differently. I believe this is a discrepancy that shouldn't exist and makes me sad :(

The best workaround I can see is to enter Normal mode (from Visual mode) via `Ctrl + [` before entering Insert mode. That way the cursors are preserved upon exiting Insert mode.

**To Reproduce**

1. Select multiple occurrences of a word or selection via `Ctrl + D`
2. Enter Insert mode via `Shift + i` or `Shift + a`
3. Type some text (optional)
4. Exit Insert mode via `Escape`
5. Only the first cursor remains

**Expected behavior**
Exiting Insert mode into Normal mode should retain all cursors. Then pressing `Escape` once more in Normal mode should clear the additional cursors.

- Extension (VsCodeVim) version: 1.21.6
- VSCode version: 1.59.0
- OS: Windows 11 Pro

### LINKED ISSUE #8147: Option to control <ESC> behavior in Multiple Cursor mode
No

When in multiple cursor mode, and I hit escape, I'd _really_ like the ability to _stay_ in multiple cursor mode and move around with multiple cursors while in normal mode, and then perform other operations. 

Atom

**Additional context**
No

lord:
Previously, if you created a multicursor block selection, pressed `I`, and then pressed escape, all selections except the first would be deleted. This change updates the escape behavior to preserve these selections.

I've also added tests that confirm this behavior is fixed.

I'll admit I don't perfectly understand the purpose of `isFakeMultiCursor`, but I'm absolutely willing to put time into changes or improvements as suggested by a reviewer.

Fixes #8147
Fixes #7008

n/a

### DIFF
--- a/src/actions/commands/insert.ts
+++ b/src/actions/commands/insert.ts
@@ -109,7 +109,6 @@ export class CommandEscInsertMode extends BaseCommand {
     }
 
     if (vimState.isFakeMultiCursor) {
-      vimState.cursors = [vimState.cursors[0]];
       vimState.isFakeMultiCursor = false;
     }
   }
--- a/test/multicursor.test.ts
+++ b/test/multicursor.test.ts
@@ -166,4 +166,18 @@ suite('Multicursor selections', () => {
     keysPressed: 'gbgb<Esc>Vjjj<Esc><Esc>gg afd',
     end: ['| is a test', '1', '2', 'this is another test', '1', '2', '3', '4', '5'],
   });
+
+  newTest({
+    title: 'Multiple cursors are preserved after inserting before a block selection and then pressing escape once',
+    start: ['|testing', 'testing'],
+    keysPressed: 'gbgbI<Esc>ifoo',
+    end: ['foo|testing', 'footesting'],
+  });
+
+  newTest({
+    title: 'Multiple cursors are truncated to one cursor if you press escape twice',
+    start: ['|testing', 'testing'],
+    keysPressed: 'gbgbI<Esc><Esc>ifoo',
+    end: ['foo|testing', 'testing'],
+  });
 });


J-Fields:
> I'll admit I don't perfectly understand the purpose of `isFakeMultiCursor`, but I'm absolutely willing to put time into changes or improvements as suggested by a reviewer.

I think the context is this: vanilla Vim doesn't support multicursor, but some actions had effects that sort of mimicked it (like insertion from VisualBlock mode, as this PR relates to). Changing these to be real multi-cursor interactions is much nicer, of course, but this breaks Vim compatibility slightly, as one `<Esc>` won't take you from multi-cursor Insert to single-cursor Normal. `isFakeMultiCursor` tracks these situations so one `<Esc>` will *also* collapse the cursors.

I do wonder if all this isn't nitpicky bullshit that we can just do away with.

lord:
I could add a configuration option that defaults to getting rid of `isFakeMultiCursor`. That way users who expect and depend on the old behavior would instantly see it break, and, if they complain, we can mention this configuration option as a way to get back to the old behavior. If they don't complain, then it's unlikely folks are dependent on this old behavior, and we can get rid of it?

musjj:
Any progress with this PR? Adding a configuration seems like a good idea to me.

musjj:
Also, another behavior I'm noticing: when you deleting in a block selection (e.g. `<C-v>jjjdf`) also reverts to a single cursor. Can we also add an option for this too?

Cojasm:
Will this be fixed/updated?

J-Fields:
Sorry - I don't think this is the right approach. I've done quite a bit of work on multicursor recently and have a branch locally that fixes this issue and similar ones. Hopefully will be able to merge soon.

PR #8533, opened 2023-08-10
Fix #7159 bad handling of auto-surrounding selections

### LINKED ISSUE #7159: Surrounding selection with quotes
- Issue Type: `Bug`
- Extension Name: `vim`
- Extension Version: `1.21.8`
- OS Version: `Windows_NT x64 10.0.19043`
- VS Code version: `1.61.1`

This used to work before but now is acting broken.

Select some keyword, or for example when using clg snippet (it creates console.log(object) and selects word 'object'.
Press ' or SHIFT + '
Results: The selected word gets surrounded with quotes but the problem is that the selection is lost and the insert cursor is placed between penultimate and the last letter which is weird. Continuing to type creates a result like 'objecnewly typed textt'

Expected behavior: The selected word gets surrounded with quotes and remains selected. Continuing to type replaces the selection with newly typed characters.

VSCode: https://github.com/microsoft/vscode/issues/133995#issue-1009674353

Elliot-Roberts:
By default in VSCode with VSCodeVim disabled, selecting text and typing an opening bracket or quote will surround the selection, while *keeping the selection*, allowing for further non-surrounding keystrokes to overwrite the selected text.

Currently, with VSCodeVim enabled, in insert mode the functionality is still partially present, but broken.  
The text gets surrounded, but the selection is lost, and the cursor is placed one character short of the right end.

This is a pain when filling in snippet holes with strings.
I want to be able to just type my new text in, but when VSCodeVim is enabled, if I am inserting a string or bracketed expression, I have to delete the placeholder text first to prevent it from interfering.  
e.g. if the selected placeholder is `name` and I type in `"Jeff`, I get `"namJeffe"` when I wanted `"Jeff"`.

This PR fixes the behavior in insert mode to match the vim-disabled behavior.

Fixes #7159 

I did not get super familiar with the codebase. I hope you can pardon the tacked-on quality of the additions. They at least do not break existing tests, and I'm of course open to suggestions for restructuring.

Thanks :)

### DIFF
--- a/src/mode/modeHandler.ts
+++ b/src/mode/modeHandler.ts
@@ -1272,6 +1272,10 @@ export class ModeHandler implements vscode.Disposable, IModeHandler {
             }
             break;
 
+          case Mode.Insert:
+            // Don't collapse existing selections in insert mode
+            selections.push(new vscode.Selection(start, stop));
+            break;
           default:
             // Note that this collapses the selection onto one position
             selections.push(new vscode.Selection(stop, stop));
--- a/src/transformations/execute.ts
+++ b/src/transformations/execute.ts
@@ -241,13 +241,19 @@ export async function executeTransformations(
     }
   }
 
-  const selections = vimState.editor.selections.map((sel) => {
-    let range = Cursor.FromVSCodeSelection(sel);
-    if (range.start.isBefore(range.stop)) {
-      range = range.withNewStop(range.stop.getLeftThroughLineBreaks(true));
-    }
-    return new vscode.Selection(range.start, range.stop);
-  });
+  let selections;
+  if (vimState.currentMode === Mode.Insert) {
+    // Insert mode selections do not need to be modified
+    selections = vimState.editor.selections;
+  } else {
+    selections = vimState.editor.selections.map((sel) => {
+      let range = Cursor.FromVSCodeSelection(sel);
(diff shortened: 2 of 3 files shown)


szechp:
is there a way to improve this to add surroundings for visual selections? right now i have to go to insert mode, shift select and then press " to surround a word.

Elliot-Roberts:
VSCodeVim [includes emulation of the `surround.vim` vim plugin](https://github.com/VSCodeVim/Vim#vim-surround), which is a popular solution for surrounding in vim. Does that work for you? 
The auto-surrounding behavior in insert mode comes from VSCode natively and operates on "VSCode-style" selections.
Visual mode selections are a vim concept, and so have a more vim-style solution.

szechp:
I'm coming from astronvim, which has leader+s+a to add surrounding (and a
bunch of others to replace etc) so at first I wanted a way to rebind vscode
vom to that, which is not working BC surround does net let me rebind. I
have my problems with surround, I don't want y to be a leader, I don't want
S to be remapped, for instance. So I was hoping for the easy solution
namely using visual selection and " to do this. But I figured this runs in
a lot of other issues since brackets are remapped somewhere else, so I will
just get used to one of the new ways that currently work 🤷‍♂️ but thanks
for the quick reply

Issue #2217, opened 2017-12-13
[Feature Request] add in a langmap setting


ghost:
* Click *thumbs-up* 👍 on this issue if you want it!
* Click *confused* 😕 on this issue if not having it makes VSCodeVim unusable. 

The VSCodeVim team prioritizes issues based on reaction count. 

--------

**This a FEATURE REQUEST**:

**Environment**:

- **VSCode Version**: 1.19.1
- **VsCodeVim Version**: v0.10.6
- **OS**: Linux

**What the feature is:**
Langmaps are a feature which allows you to do a string of simplistic mappings with ease on a single line. For example, as a Dvorak user i have several keys I like to remap in order to improve my ability to navigate and to greatly improve comfort while using vim or vim emulators. This could be achieved with normal mappings of each key (although VSCode Vim doesn't do this properly - see #2233 and #2234) or using a single langmap.

**important differences between langmaps and maps**
Langmaps are inflexible and only allow mapping one key to one other key, many times.
Maps allow mapping a key or key combination to another key, key combination, or action. They allow you to do much more complex things and are more flexible.

**Why i feel like this is important:**
I use a Dvorak keyboard, and Vim was made for use on a QWERTY keyboard. This can make some default commands very very awkward for me, including using w and especially j and k. Langmaps can be very very useful to those who use an alternate keyboard layout, whether it means restoring the ability to go off of muscle memory by using a langmap to recreate a QWERTY layout in normal mode, or moving awkward keys to better locations. 
I personally feel langmaps do a better job of this specific use case than maps, which, *especially in VSCode Vim*, bloat the settings file and can draw attention away from or overshadow the more complex and important mappings that may be present and need tweaking, or just to freshen up on what one or more mappings you created but rarely use actually do.

Adding in a langmap at this point could shrink my current settings file to half its current size.

For reference: in Vim it's set with the syntax  \<desiredKey>\<originalKey> OR \<desired keys>;\<original keys> and separated by commas. Such as
`set langmap=tj,nk,sl,.>;wW,\\,';nN`
Although I'd expect only a single backslash would be needed for the escape on the comma.

ohjames:
> For example, mapping '.' to 'w' means that '.' will act as the command 'w', but only on it's own. Typing 'd.' will do nothing since '.' is not a movement.

[...]

PR #8541, opened 2023-08-15
Added Langmap Feature

Opisek:
This pull request implements the langmap feature of Vim. It's an essential accessibility feature for people who use an alternate keyboard layout like Dvorak, Colemak, or a layout specific to a language like Greek or Russian.

It works by specifying the map under `vscode.langmap` in Code's JSON settings.
Example langmap for Greek:
```
ΑA,ΒB,ΨC,ΔD,ΕE,ΦF,ΓG,ΗH,ΙI,ΞJ,ΚK,ΛL,ΜM,ΝN,ΟO,ΠP,QQ,ΡR,ΣS,ΤT,ΘU,ΩV,WW,ΧX,ΥY,ΖZ,αa,βb,ψc,δd,εe,φf,γg,ηh,ιi,ξj,κk,λl,μm,νn,οo,πp,qq,ρr,σs,τt,θu,ωv,ςw,χx,υy,ζz
```
Example langmap for Dvorak:
```
'q,\\,w,.e,pr,yt,fy,gu,ci,ro,lp,/[,=],aa,os,ed,uf,ig,dh,hj,tk,nl,s\\;,-',\\;z,qx,jc,kv,xb,bn,mm,w\\,,v.,z/,[-,]=,\"Q,<W,>E,PR,YT,FY,GU,CI,RO,LP,?{,+},AA,OS,ED,UF,IG,DH,HJ,TK,NL,S:,_\",:Z,QX,JC,KV,XB,BN,MM,W<,V>,Z?
```
You may find the exact format under `:help langmap` in Vim.

Keystrokes are then remapped based on the passed langmap, unless they're to be interpreted as text. Most frequently, one would use this feature to at least remap the movement keys to be at the physical positions of the "hjkl" keys on the keyboard, whilst still being able to type with one's chosen layout.

[Feature Request #2217](https://github.com/VSCodeVim/Vim/issues/2217)

**Special notes for your reviewer**
I have made sure to keep the changes to working code minimal. Additionally, no remapping is called if the passed langmap is an empty string (default value). That way, no unexpected behaviour or performance penalty should be introduced, unless you actively override the default value in your settings.
Furthermore, I'm willing to write unit tests for this feature in the future.


ryan-mooore:
It would be nice to also support the from:to syntax specified in `help langmap`:

>  ```
> 2.  A list of "from" characters, a semi-colon and a list of "to"
>             characters.  Example: "abc;ABC"
>         Example: "aA,fgh;FGH,cCdDeE"```

Also being able to pull the setting from `.vimrc` if applicable would be very convenient!

Thanks for making this PR, been wanting this feature for a while! Hopefully it gets approved soon..

Opisek:
I believe it should already work with .vimrc but let me test it when I'm back home in a few days.

ryan-mooore:
Hmm, my colemak from:to langmap doesn't seem to work here, but works fine in vim:
`"vim.langmap": "qwfpgjluy\\;arstdhneiozxcvbkmQWFPGJLUY:ARSTDHNEIOZXCVBKM;qwertyuiopasdfghjkl\\;zxcvbnmQWERTYUIOPASDFGHJKL:ZXCVBNM",`

If I convert it to the other format however, it works fine:
```"vim.langmap": "qq,ww,fe,pr,gt,jy,lu,ui,yo,\\;p,aa,rs,sd,tf,dg,hh,nj,ek,il,o\\;,zz,xx,cc,vv,bb,kn,mm,QQ,WW,FE,PR,GT,JY,LU,UI,YO,:P,AA,RS,SD,TF,DG,HH,NJ,EK,IL,O:,ZZ,XX,CC,VV,BB,KN,MM",```

Opisek:
Will check the bug, thanks for reporting!

Opisek:
I've identified and fixed the issue @ryan-mooore

ryan-mooore:
All works beautifully for me now 👍

Opisek:
Is there any hope on getting this merged? @J-Fields

J-Fields (changes requested):
Mind adding a few test cases?

Opisek:
> Mind adding a few test cases?

Done.

One annoyance I'm not happy with yet is the need for [14eba0bd](https://github.com/VSCodeVim/Vim/pull/8541/commits/14eba0bd07f7d672585c92bd8ee894b1ceec346a).

Basically, Vim handles registers separetly from other `<character>`-like directives when it comes to remapping:
- When for example searching for a character, it must not be langmapped: `θθ` becomes `fθ` and not `ff`.
- When using registers, the register must be langmapped, otherwise you'd be trying to use an invalid register.

As is right now, there is no way to distinguish mappings like `f<character>` from `@<character>`, so I have to manually go to all the places that registers are used and do the mapping logic there.

If instead something like `@<register>` or `@<r-character>` could be introduced, then the central langmap logic could just handle it.

(Side note: Why does recording a macro use `q<alpha>`, but playing `@<character>`? What's the difference?)

J-Fields:
>If instead something like `@<register>` or `@<r-character>` could be introduced, then the central langmap logic could just handle it all.

Yeah adding `<register>` sounds good to me.

>(Side note: Why does recording a macro use `q<alpha>`, but playing `@<character>`? What's the difference?)

I'm not sure, only thing that really comes to mind is it may be a hacky way to ensure `q/` and `q?` have the right mappings.

Opisek:
@J-Fields I changed the code accordingly. I ended up having to introduce `<register>` and `<macro>` separately. This is due to some edge cases with macro recording, which used `<alpha>` and `<number>` instead of `<character>`, as previously mentioned. Using just <register> caused some issues with obscure tests concerning remapping that I could not decipher.

With the current setup, all tests are passing, and the branch is up-to-date with upstream.

J-Fields (approved):
Looks good to me. Thanks a lot!

PR #8554, opened 2023-08-31
fixes #8095 now can scroll with vimState.recordedState.count

### LINKED ISSUE #8095: Ctrl+{f, d, u, b} not using numeric prefix to repeat
**Describe the bug**
Full page or half-page up/down commands are not utilizing the numeric prefix entered before their invocations with ctrl+{f, d, u, b}

In order to retain all vs-code ctrl+* functionality (eg. ctrl+f) each command ctrl+[a-z] was bound to 'ctrl+* v' for vim and 'ctrl+* ctrl+*' for the vscode original command.

I prefer this because I can not only keep both versions, it's not difficult to get used to, and it opens up my available keybind options to 26^2 since they are all now mapped to a 2-input keybind.

ctrl+{i, o} does accept the numeric prefix, so some vim-functions are not broken and I have confirmed this is not the issue by testing page-down with 'ctrl+f'

The only reason I mention it is because 'ctrl+f v' is more cumbersome than 'ctrl+f', especially to invoke repeatedly, it would be nice if the numeric prefix were utilized by those commands.

**To Reproduce**

1.  type `4<C-f>`
2. page only moves down length of 1 page instead of 4

**Expected behavior**
page should move down number of full/half pages equal to the numeric prefix

- Extension (VsCodeVim) version: 1.24.3
- VSCode version: 1.73.1
- OS: Windows 10 64bit

rpuhalovich:

Super simple PR, there was a simple mistake when using `vimState.actionCount` to multiply the amount of scroll actions to do.

Fixes: #8095

I've noticed that there aren't any tests that go through this specific functionality. Let me know if you'd like one written up.

### DIFF
--- a/src/actions/commands/scroll.ts
+++ b/src/actions/commands/scroll.ts
@@ -85,10 +85,13 @@ abstract class CommandScrollAndMoveCursor extends BaseCommand {
     if (visibleRanges.length === 0) {
       return;
     }
+
     const smoothScrolling = configuration
       .getConfiguration('editor')
       .get<boolean>('smoothScrolling', false);
-    const moveLines = (vimState.actionCount || 1) * this.getNumLines(visibleRanges);
+
+    const timesToRepeat = vimState.recordedState.count || 1;
+    const moveLines = timesToRepeat * this.getNumLines(visibleRanges);
 
     let scrollLines = moveLines;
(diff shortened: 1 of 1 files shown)


J-Fields (approved):
Nice catch, thanks!

PR #8566, opened 2023-09-08
fix: avoid escaping the exact word search twice

Tal500:
Apparently, on exact word search, the word is getting regexp-escaped twice - once in https://github.com/VSCodeVim/Vim/blob/ee4888ab3b18f11baf60e6ee44db84f2ba52c6f6/src/actions/commands/search.ts#L39-L41 and again in https://github.com/VSCodeVim/Vim/blob/ee4888ab3b18f11baf60e6ee44db84f2ba52c6f6/src/actions/commands/search.ts#L102-L103, so this PR just drop the former.

The impact for the user happens when the user configures for VSCode "what is a word". For example, my VSCode configuration on the R language is:
```json
"[r]": {
    // This makes sure that the "." character doesn't separate symbol.
    "editor.wordSeparators": "`~!@#$%^&*()-=+[{]}\\|;:'\",<>/?",
}
```
It's hard to notice, but the difference is that `.` isn't appear there, but appear in the following. The default value for this option is:
```json
"editor.wordSeparators": "`~!@#$%^&*()-=+[{]}\\|;:'\",.<>/?"
```

I do like that VSCodeVim treat my word correctly according to VSCode prefs, but I think that VSCodeVim should treat this word correctly on search (probably it was just a mistake in the code).

### DIFF
--- a/src/actions/commands/search.ts
+++ b/src/actions/commands/search.ts
@@ -36,9 +36,6 @@ async function searchCurrentWord(
       isExact = false;
     }
 
-    if (isExact) {
-      currentWord = _.escapeRegExp(currentWord);
-    }
     // If the search is going left then use `getWordLeft()` on position to start
     // at the beginning of the word. This ensures that any matches happen
     // outside of the currently selected word.


J-Fields (changes requested):
Could you please write a test that fails before this change and passes after?

Tal500:
> Could you please write a test that fails before this change and passes after?

Added tests

J-Fields:
Thanks! But these tests seem to succeed before your change as well - mind adding one that tests the change? You can adjust the config for a test by passing `config` to `newTest`.

Tal500:
> Thanks! But these tests seem to succeed before your change as well - mind adding one that tests the change? You can adjust the config for a test by passing `config` to `newTest`.

I was debating how to do this so. The problem with the "config" parameter is that it doesn't allow me to change the non-vim-specific-extension configuration of VSCode. Here, what I need is to modify the global VScide config `editor.wordSeparators`, how to then?

J-Fields:
Ah, right, I should really add a way to do that easily. For now just put `vscode.workspace.getConfiguration(...).update(...)` in `setup` and reverse it in `teardown`.

PR #8572, opened 2023-09-10
Add white as the default foreground colour.

chandradeepdey:
Fixes illegible text on the new 'Light Modern' theme.


chandradeepdey:
Requesting review @J-Fields

J-Fields:
I'm not able to repro the problem - would you mind sharing your config?

J-Fields:
Thanks!

PR #8575, opened 2023-09-10
Persist global marks on disk

### LINKED ISSUE #3841: Marks are reset/removed after some time
**Describe the bug**
I work on a file, set some marks for key areas I jump around between. After a while I notice that my marks are gone, this is happening repeatedly while working with the file.

**To Reproduce**

1.  Go to a file
2.  Set a couple of marks
3.  Continue to use the application, work for some time (maybe an hour? not quite sure)
4.  Try to go to previously set mark, it is now no longer there. Error message `E20: Mark not set`

**Expected behavior**
Got to the previously set mark

**Screenshots**
N/A

- Extension (VsCodeVim) version: 1.8.1
- VSCode version: Version 1.34.0 (1.34.0)
- OS: macOS 10.14.5

EerikSaksi:
Fixes #3841

This PR ensures that marks are persisted between sessions through a new history file. This makes marks much more usable as marks set before the editor was closed/loaded will still work. This persists local and global marks. Also includes duplicate dedection, so when old marks are overriden the history file also removes the old mark from the file.

#3841 

### DIFF
--- a/src/history/historyTracker.ts
+++ b/src/history/historyTracker.ts
@@ -22,6 +22,8 @@ import { Mode } from '../mode/mode';
 import { ErrorCode, VimError } from '../error';
 import { Logger } from '../util/logger';
 import { earlierOf } from '../common/motion/position';
+import { MarkHistory } from './historyFile';
+import { ExtensionContext } from 'vscode';
 
 const diffEngine = new DiffMatchPatch.diff_match_patch();
 diffEngine.Diff_Timeout = 1; // 1 second
@@ -106,7 +108,7 @@ export interface IMark {
 /**
  * An undo's worth of changes; generally corresponds to a single action.
  */
-class HistoryStep {
+export class HistoryStep {
   /**
    * The insertions and deletions that occured in this history step.
    */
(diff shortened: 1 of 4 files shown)


HenryTSZ:
I merged this PR and roughly used it, and found two issues: 

1. When 'vim. showMarksInGutter' is set to true, the initialization plugin prompts: mark. position. getLineBegin is not a function
2. Using ` + marker cannot jump and will result in an invalid button state

jasonwilliams:
Hey @EerikSaksi was this something you were hoping to come back to? Persistent marks would be great to have

EerikSaksi:
I fixed the gutter issues and fixed some linting stuff, I think the `+ thing should also be solved now

jasonwilliams:
Hey @EerikSaksi I can test this, before I do some questions..

Are global marks supposed to be workspace wide? So if I set a global mark (uppercase letter after `m`) then move to another file and call it (`'M`) should it take me to that file? Or are they only file dependent?

EerikSaksi:
Yeah global marks mark a position and a file and will take you to that file, lower case marks are specific to files

jasonwilliams:
I've tested @EerikSaksi's PR and I can confirm it works, I've tried setting a permanent mark and reloading the window and it does take me back to the right place, this doesn't happen on the main branch

src/platform/node/history.ts
```diff
@@ -62,6 +62,11 @@ export class HistoryBase {
     }
   }
 
+  // filter history based on predicate
+  public filter(predicate: (value: string) => boolean) {
```
J-Fields, line 66: `filter` sounds like a pure function to me. `remove` maybe?

src/history/historyTracker.ts
```diff
@@ -571,21 +603,40 @@ export class HistoryTracker {
...
     }
+
+    // add the entry to the history to persist it over sessions
+    if (mark.isUppercaseMark) {
+      void HistoryStep.markHistory.add(JSON.stringify(mark));
+    } else {
+      // local marks do not store the path, but we still need to include it in the
+      // history file so we know which file it belongs to
+      void HistoryStep.markHistory.add(JSON.stringify({ ...mark, document }));
+    }
```
J-Fields, line 639: I think all this should be encapsulated in
```ts
void MarkHistory.addMark(mark: IMark)
```

greg-nagy:
is there any way I can help pushing this PR to be merged? 

I really want to have this feature :)

EerikSaksi:
Sorry been busy keep forgetting I’ll try look into getting this merged this weekend

greg-nagy:
@J-Fields  I think this might be ready for prime time and to be merged

greg0x:
sad panda is sad

altough moved to neovim since then

Issue #8603, opened 2023-09-17
Insert mode keystrokes are sometimes dropped since 1.26.0


TommyX12:
**Describe the bug**
In the latest release version of vscode vim, when I type in insert mode, about 50% of the time one or more of the characters I input will be missing. The problem is worse when there are heavy auto-complete or github copilot completions. Reverting to 1.25.2 solves this issue.

**To Reproduce**
1. Open a file (preferably within an existing project).
2. Go to insert mode, very quickly type "asdfghjk".
3. Notice that some of the characters are missing.

**Expected behavior**
The typed string should always be "asdfghjk".

**Environment:**

- Extension (VsCodeVim) version: 1.26.0
- VSCode version: 1.82.2
- OS: Windows 10

nikitavoloboev:
This is incredibly frustrating. Please fix this.

https://github.com/pqrs-org/Karabiner-Elements/issues/3595

Hitting the issue above, I can't automate keypresses with this extension on.

Is it possible to downgrade to earlier versions?

nikitavoloboev:
@TommyX12 

> Reverting to 1.25.2 solves this issue

How did you do it?

dj-blake:
Same issue. Really frustrating as it drops key presses quite frequently.

Note I don't use co-pilot, so it is not specifically due to interaction with that extension, though it does occur particularly when Intellisense is popping up suggestions.

Downgraded to 1.25.2 as well to avoid the issue.

hawkinsw:
I want to say that I, too, am experiencing this issue. While it *is* frustrating, I completely understand that you are all doing the best that you can to make this incredibly awesome tool for the community and doing it in your free time. Thank you all for the work that you do!

nachomglz:
Please fix this, when you type fast it drops like 30% of the characters, it's really frustrating.

darrenchurchill:
It looks like the issue is from 50bc4fe1e32434cc15098173a5dfc42c7818ac8e.

I'm not familiar with the code at all, but I set up a macro on my keyboard to rapidly output the string `"asdfghjkl asdfghjkl "`, and ran the extension various times during a `git bisect`. After removing the `TaskQueue`, the extension seems capable of losing track of the order keys were pressed in, missing keypresses, and even generating duplicate keypresses.

I hope this is helpful, I attached a couple screencasts of bad and good behavior.

[...]

Issue #8577, opened 2023-09-11
korean typing error


astrataraxia:
**Describe the bug**

**There is a typing error in Korean.** It keeps breaking weirdly when I type in Korean. This error occurred after the vscode update, so I thought it was a vscode problem, but when I deleted the vim and used it, it worked fine. So I don't know exactly which error it is, but I think I need an update from vim.

**To Reproduce**

Just typing Korean
 -> typing : 안녕하세요 
 -> But typing : ㄴ안ㄴ세하세요 

**Environment (please complete the following information):**

- Extension (VsCodeVim) version: 1.26.0
- VSCode version: 1,82.0
- OS: window 11

cake-monotone:
I'm experiencing the same issue on a Mac. Korean characters are being decomposed or not inputted correctly.

I think this is a bug related to the vim-extension version 1.26.0, and not with VSCode itself.

I installed version 1.25.2 and confirmed that the issue doesn't exist there.

SE-June:
I have the same issue on typing in Korean

eduardosilva:
I think this is not a specific error when typing in Koren because I have the same issue typing in English or Portuguese (pt-br)

okamototomoyuki:
Japanese input also frequently becomes incorrect.
Although the Japanese input system involves entering alphabets and converting them to Japanese words with the same pronunciation,
the position of the character replacement often becomes incorrect.

shinypond:
I have the same issue when typing in Korean.

Kimnabin:
Since it may take some time to be fixed, here's a small workaround: use previous version
I deleted the vim extension from VSCode and ran this command in the terminal.

code --install-extension vscodevim.vim@1.25.2

This should do until the issue is resolved.
Happe Coding.

for Mac users

to enable the code command in the terminal, you need to cmd + shift + p and choose Shell Command: Install 'code' command in PATH
also cmd + shift + p and select Extensions: Disable Auto Update for All Extensions to prevent auto update

Or add Vim Extensions ver 1.25.2

![Screenshot 2023-09-16 012845](https://github.com/VSCodeVim/Vim/assets/96113006/c01bfaf3-fa02-4d70-a352-a3c443b2ac06)

ianhenrygim:
@Kimnabin The solution you gave works. Thanks.

NeuroWhAI:
Same here.

thisnome94:
@Kimnabin 감사합니다.

[...]

PR #8622, opened 2023-09-24
Revert TaskQueue removal

nullbus:
The commit 50bc4fe1e32434cc15098173a5dfc42c7818ac8e removed `TaskQueue`. But some issues rised after this commit like: #8577 #8603

I have investigated this issue and foud that there are some nested event invocations. With `TaskQueue`, these events are added to a queue and processed event by event. But as of 1.26.0, some events are invoking other event processing immediately before finishing current event and this behaviour is different from pre 1.26.0 versions.

This commit is just a revert of 50bc4fe1e32434cc15098173a5dfc42c7818ac8e.

#8577 #8603

I think there are two ways to solve this problem:
- Find and fix all nesting issues case by case
- Task queue recovery

For this commit I choosed latter because of its simplicity.

Keep throwing off `queue` dependency and make (simple) self task queue implementation is also a option but I wanted to preserve the previous source state if possible.

### DIFF
--- a/extensionBase.ts
+++ b/extensionBase.ts
@@ -16,6 +16,7 @@ import { globalState } from './src/state/globalState';
 import { Register } from './src/register/register';
 import { SpecialKeys } from './src/util/specialKeys';
 import { exCommandParser } from './src/vimscript/exCommandParser';
+import { taskQueue } from './src/taskQueue';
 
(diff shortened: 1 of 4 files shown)


J-Fields (approved):
Thanks for tracking this down!

PR #8638, opened 2023-10-01
Mark reg[ister] as not repeatable with dot

### LINKED ISSUE #8596: Dot is repeating the last command-line (:reg) command
**Describe the bug**
I want to dot-repeat the change of a word. Once set up it works as expected:
I then use ":reg" to check my registers. Now the "dot" just opens the registers.

**To Reproduce**
File:
```
Foo
Foo
Foo
```

Starting in Normal Mode on first line: ``ciwBar<Esc>``
Second line: ``.`` ✅ 
In between I check the registers with ``:reg``
Third line: ``.`` ❌ (Just opens the registers again)

**Expected behavior**
Do not take command-line commands into the dot command.

- Extension (VsCodeVim) Version: 1.26.0
- VSCode version: Version: 1.82.1 (user setup)
OS: Windows_NT x64 10.0.22621

dannoe:
Fixes #8596

This PR will fix that the reg[ister] command is repeatable with the dot command.

#8596 


J-Fields (approved):
Thanks!

PR #8639, opened 2023-10-01
Prevent non-strings in QuickPick

### LINKED ISSUE #8506: :reg messes with other dropdown options like CMD + T
**Describe the bug**
The Dropdown shown with the `:reg` command messes with other dropdown options like `CMD + T`

**To Reproduce**

1.  Put something in any registers
2.  execute `:reg` (the dropdown appears)
3.  close the dropdown
4.  open the dropdown again with `CMD + T`
5. There's double text/icons.

**Expected behavior**
There shouldn't be overlapping texts.

**Screenshots**
Screenshot 2023-07-20 at 15 53 59

**Environment:**

- Extension (VsCodeVim) version: v1.25
- VSCode version: v1.80.1
- OS: macOS Ventura 13.1

dannoe:
Non-strings in the description property of QuickPickItem will cause VS Code to somehow "cache" entries and display them in other QuickPicks.

Fixes #8506

This PR will add an additional check if the incoming type is a string, so VS Code won't show weird menus after opening the registers.

#8506 

This is just a workaround, for the following problem: When the ``.registers`` file is saved and later imported again, it is just recognized as an javascript object and not as the correct typescript type ``RecordedState``.
The object is then later consumed by the ``getRegisterDisplayValue`` function which won't handle the ``object`` correctly and just return the ``object`` itself. 
The better fix would be to type the incoming JSON to the correct ``RecordedState`` type or to handle the object inside the ``getRegisterDisplayValue`` function. But my typescript experience isn't good enough to properly fix this. (I'm not even sure there's a better way to check for a string type in Typescript than my method in the pull request.)


J-Fields (approved):
Thanks!

PR #8698, opened 2023-10-27
Support to `:sp`/`:vs` open non-existing files.

### LINKED ISSUE #7283: `:vs`/`:sp` can't open non-existing files.
I completely agree with the features mentioned in #2683 about `gf`, which was fixed but appeared again, and #2680 should be implemented.
By the way, I think `:e` works properly now while `:vs` or `:sp` should open non-existing files like `:e` but by aside. 

Environment:
Extension (VsCodeVim) version: `1.21.10`
VSCode version: `1.62.2 x64`
Commit: `3a6960b964327f0e3882ce18fcebd07ed191b316`
OS: `CentOS Linux release 7.8.2003 (Core)`

### LINKED ISSUE #8671: not support ":sp test.md", ":ls", ":pwd" command
- not support the ":sp test.md" to create the new file and split windows
- not support ":ls", ":pwd" commands

JLMSC:
Fixes the use of `:sp`/`:vs` with non-existing files, opening a new split with the specified file and creating it right away (the file will need to be manually saved to apply changes).

Fixes #7283
Partially fixes #8671

### DIFF
--- a/src/cmd_line/commands/file.ts
+++ b/src/cmd_line/commands/file.ts
@@ -68,9 +68,9 @@ function getLegacyArgs(args: IFileCommandArguments): LegacyArgs {
       createFileIfNotExists: true,
     };
   } else if (args.name === 'split') {
-    return { file: args.file, position: FilePosition.NewWindowHorizontalSplit };
+    return { file: args.file, position: FilePosition.NewWindowHorizontalSplit, createFileIfNotExists: true };
   } else if (args.name === 'vsplit') {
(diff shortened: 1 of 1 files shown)


J-Fields:
Thanks!

dsharma29:
Problem:
---------
With this change now when we only type `:vsp` to split the opened file it creates a new file always.  Now  we have to type the entire filename to be able to split it. 

Solution:
----------
This change should check if a file name is not specified it should split the opened file and not create a new one. It should create a new one only when a filename is specified which does not exists.

PR #8705, opened 2023-10-30
feat: gf relative to project root

### LINKED ISSUE #1940: `gf` should use the current project root too

--------

**Is this a BUG REPORT or FEATURE REQUEST?** (choose one): **FEATURE REQUEST**

Currently, the `gf` command only uses the current location of the currently opened file. Sometimes, it's correct/handy to open a path relative to the top of the project directory

**Environment**:

- **VSCode Version**: 1.14.0
- **VsCodeVim Version**:  0.8.4
- **OS**: 
Linux

Foo-x:

Jump relative to project root with `gf`.

Closes #1940

### DIFF
--- a/src/actions/commands/actions.ts
+++ b/src/actions/commands/actions.ts
@@ -1027,7 +1027,7 @@ class CommandOpenFile extends BaseCommand {
     if (fileInfo) {
       const workspaceRootPath = vscode.workspace.getWorkspaceFolder(
         vimState.document.uri,
-      )?.uri.path;
+      )?.uri.fsPath;
       const filePath =
         path.isAbsolute(fileInfo[1]) || !workspaceRootPath
           ? fileInfo[1]
--- a/src/actions/commands/actions.ts
+++ b/src/actions/commands/actions.ts
@@ -1025,9 +1025,8 @@ class CommandOpenFile extends BaseCommand {
 
     const fileInfo = fullFilePath.match(/(.*?(?=:[0-9]+)|.*):?([0-9]*)$/);
     if (fileInfo) {
-      const workspaceRootPath = vscode.workspace.getWorkspaceFolder(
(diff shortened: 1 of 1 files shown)


src/actions/commands/actions.ts
```diff
@@ -1024,7 +1025,13 @@ class CommandOpenFile extends BaseCommand {
 
     const fileInfo = fullFilePath.match(/(.*?(?=:[0-9]+)|.*):?([0-9]*)$/);
     if (fileInfo) {
-      const filePath = fileInfo[1];
+      const workspaceRootPath = vscode.workspace.getWorkspaceFolder(
+        vscode.window.activeTextEditor!.document.uri,
```
J-Fields, line 1029: Use `vimState.document.uri` instead

Foo-x: Thank you for reviewing!
I have fixed.

Foo-x:
I have noticed that `Uri.path` does not work on Windows, especially for network paths, so fixed to use `fsPath`.

J-Fields:
Thanks!

huba71:
I miss the original behaviour! I used it a lot.
We have projects with several git submodules that mention relative file paths.

Anyway the command line vim also behaves the way vscodevim did before this feature.
I think that vscodevim is created originally for users switching from command line vim to vscode (as I did) with similiar behaviour as command line vim.

So I think that this feature - in this form that breaks compatibility with command line vim -  is a not a good idea.

Perhaps it could behave the original way and there can be a fallback to locate file relative to project root if no file found.

Watching the diff (https://github.com/VSCodeVim/Vim/commit/8204b724df7e447f9d8c005290202768f6840ed0) I think it can not be very difficult to keep backward compatibility with vim command line.

Thanks!

PR #8708, opened 2023-11-01
fix: gf to jump with line number

Foo-x:

Fixed `gf` with line number was not working.


J-Fields:
Thank you!

Issue #7364, opened 2021-12-31
Search highlight defaults should include border color, etc.


J-Fields:
My theme uses a border, not a background highlight, to show search results. We should, by default, use `editor.findMatchHighlightBorder` just like we do `editor.findMatchHighlightBackground`.

adrsm108:
I could do this, though I wanted to share my thoughts before proceeding.

Easymotion aside, the plugin currently emulates 4 vim highlight groups, `Search`, `IncSearch`, `Substitute`, and `HighlightedYankRegion`, each configured via separate background and text color keys.
Since they're all implemented as decorations, we can feasibly style much more than just borders: font weight, slant, and even overview ruler color can all be customized (In fact, most of this is already possible with simple css property injections: `"vim.searchMatchTextColor": "#fff; border: 1px solid white; font-weight: bold; font-style: italic;"`).

Rather than add more keys to control these things, I think it might be worth deprecating the old `vim.xxxColor`/`vim.xxxTextColor` keys in favor of a new `vim.xxxHighlight` option, the value of which would be some subset of properties from [DecorationRenderOptions](https://code.visualstudio.com/api/references/vscode-api#DecorationRenderOptions):

```ts
type ConfigurableHighlight = {
  backgroundColor?: string;
  color?: string;
  borderColor?: string;
  fontStyle?: string;
  fontWeight?: string;
  textDecoration?: string;
  overviewRulerColor?: string;
};
```

The upside would be a more consistent interface offering a degree of customization similar to vim's highlight groups.
Of course, the downsides are an API change, and potentially more complexity. Compared to what we have now, however, I think this has the potential to be much more direct, and ultimately less confusing.

PR #8722, opened 2023-11-10
Fix: Add border to search decorations

bryclee:
Add border to search highlights, for themes that do not use background but use border instead to indicate find results

https://github.com/VSCodeVim/Vim/issues/7364

I saw some comments in the issue about refactoring the decoration, but as I don't have much context on the code, I went with the simple change to add the border as-is.
 
Screenshots using "Light High Contrast" theme:

Search (cursor is on line 1, run `/test`, current match is on line 2):
image
Search replace (cursor is on first word "that", running `:s/test/replace/g`):
image
Search append (cursor is on first word "test", running `:s/this/that/g`):
image


J-Fields:
Thank you!

J-Fields (approved):

Issue #8730, opened 2023-11-18
Files are automatically closed when splitting windows


Dawn-K:
In the new version `v1.27.0`. When using `:vs` or `:sp` to split a file, the file will be opened in the split window and then closed immediately. There will be only a blank file in the newly opened window (just like using Crtl+ N same), I moved the version back to `v1.26.2` to solve the problem

- Extension (VsCodeVim) version: v1.27.0`
- VSCode version: Version: 1.84.2 (user setup)
- OS: Windows 10 Pro 22H2

ankrgyl:
Reproduces for me too, on Mac OS X, v.1.27.0:

```
Version: 1.84.2 (Universal)
Commit: 1a5daa3a0231a0fbba4f14db7ec463cf99d7768e
Date: 2023-11-09T10:52:33.687Z (1 wk ago)
Electron: 25.9.2
ElectronBuildId: 24603566
Chromium: 114.0.5735.289
Node.js: 18.15.0
V8: 11.4.183.29-electron.0
OS: Darwin arm64 22.5.0
```

astroshot:
me too

```plaintext
Version: 1.84.2
Commit: 1a5daa3a0231a0fbba4f14db7ec463cf99d7768e
Date: 2023-11-09T10:50:47.800Z
Electron: 25.9.2
ElectronBuildId: 24603566
Chromium: 114.0.5735.289
Node.js: 18.15.0
V8: 11.4.183.29-electron.0
OS: Linux x64 6.1.62-1-MANJARO
```

PR #8733, opened 2023-11-19
fix: only create new file on split when specified

bcho:

#8698 introduced support for creating new file on split. However, the create behavior should only be used when there is a file arg specified for the split command.

| command case | behavior |
|-----------|---------|
| `(v)sp` | open new split with existing open file |
| `(v)sp <file>` | open new split with the giving file path, create if needed |

#8698 fixed the 2nd case, while broke the 1st case.

#8730 


bcho:
I am not sure how to proper add tests for this behavior, will try submit one in the new pr.

bcho:
hi @J-Fields could you please take a look on this? Sorry for the ping, but the bug is quite annoying :(

pimpeters:
Any ETA on when this will be merged?

pimpeters (approved):

J-Fields:
Thanks!

PR #8756, opened 2023-11-30
fixes #8576 Cursor position is broken at v1.26.0

### LINKED ISSUE #8576: Cursor position with `Foldfix` enabled is broken as of `v1.26.0` update
**Describe the bug**
Using "native" keyboard navigation of `CTRL+<arrow-key left/right>` to jump to beginning end of a word is now broken. 

This happens after vim enters a "broken state". If you then jump forward a word using `ctrl+right` or backward using `ctrl+left` then try to navigate using `left` or `right`, the cursor jumps back to its position before using `ctrl+arrow-key`. The console reports `[warning] Ignoring slipped selection: [n, n; n, n]`. Same for using `Home` or `End`. Other navigation could also be impacted but I mostly use the ones I mentioned.

Upon further testing VIM appears to enter the "broken state" consistently when moving the cursor between lines of difference length, but only at the end of the line.

Toggling VIM off then on fixes the behavior until navigating over the last character of a foldable region (even if it is unfolded).

Disabling `Foldfix` resolves the issue as does reverting to `v1.25.2` with `Foldfix` still enabled.

**To Reproduce**

1.  Enable `Foldfix` (`"vim.foldfix": true`)
2.  Move cursor up/down between lines of same length = ok
3. Move cursor up/down between lines of different length but not at the end of the line = ok
4. Move cursor up/down between lines of different length at the end of the line = enters "broken state"
5. Notice console log of `[warning] invalid cursor position. [n:n | n:n]` followed immediately by `[warning] Ignoring slipped selection: [n, n; n, n]`
7. Try navigating using `ctrl+arrow` or `home`/`end`
8. Notice console log of `[warning] Ignoring slipped selection: [n, n; n, n]` every time you do
9. Try navigating using `arrow-key` either left or right
10. Notice the cursor does not behave as expected
11. Restart vim using `toggleVim` then `toggleVim` again
12. Perform the same movement using `ctrl` or `home`/`end` followed by `left`/`right`
13. Notice the behavior is back to normal and no warning logs in console

**Expected behavior**
It's as if the cursor position is not updated properly when in the "broken state" and using `ctrl` or the other ones.

So if my cursor position is at column 3, then I jump to the end of the next word which puts me at column 10, when I hit `right` arrow I expect the cursor to move to col 11 but it instead moves to col 4.

**Screencasts**

---
*Example 1 - Normal behavior then entering "broken state"*

https://github.com/VSCodeVim/Vim/assets/75916477/2d2a984e-7d48-4440-a413-8bb84bc86613

---
*Example 2 - `Invalid cursor position` + `Ignoring slipped selection` when navigating around bracket (?)*

https://github.com/VSCodeVim/Vim/assets/75916477/8ee2a38b-db15-450d-b689-c0ce07423077

---

*Example 3 - Issue is fixed after restarting VIM (I have `alt+v` mapped to do this)*

https://github.com/VSCodeVim/Vim/assets/75916477/521a1754-da68-4c96-9527-51c4e9d7a5b9

---

- Extension (VsCodeVim) version: `1.26.0`
- VSCode version: `1.82.0`
- OS: `Windows`, `Browser (Vscode.dev)`

**Additional context**
I have not tested this behavior with VIM handling of arrow keys disabled.

HenryTSZ:
fixex #8576 

When moving from rows with more columns to rows with fewer columns, this bug is triggered 
because changing the value of the moved column to the value before the move results in invalid coordinates

https://github.com/VSCodeVim/Vim/blob/452136021881765e1fb98f198a4603ceab23b4bd/src/actions/motion.ts#L281-L283
https://github.com/VSCodeVim/Vim/blob/452136021881765e1fb98f198a4603ceab23b4bd/src/actions/motion.ts#L384-L386

#8576 

This code was submitted 5 years ago, and I am not sure about the specific logic and functionality. When I deleted these two pieces of code, the problem fixed by this PR did not reproduce


J-Fields (approved):
> This code was submitted 5 years ago, and I am not sure about the specific logic and functionality.

Yeah I don't understand what's going on here, but if deleting some janky looking code seems to fix an issue, I'm all for it. Thanks!

Issue #8724, opened 2023-11-12
not support ":ls" commands


riyuejiuzhao:
There was an issue mentioned before but this feature was not added, is there any problem?

 If I want to add this feature, what should I pay attention to?

PR #8761, opened 2023-12-02
add ls command

riyuejiuzhao:
Added the ls directive #8724


J-Fields (approved):
Seems reasonable. Thanks!

Issue #8588, opened 2023-09-13
ctrl+p not working


hhy3:
**Describe the bug**

**To Reproduce**

1.  Go to '...'
2.  Click on '....'
3.  Scroll down to '....'
4.  See error

**Expected behavior**

**Screenshots**
If remapping-related, please attach log output: https://github.com/VSCodeVim/Vim#debugging-remappings.

**Environment (please complete the following information):**

- Extension (VsCodeVim) version:
- VSCode version:
- OS:

**Additional context**

SamHasler:
Duplicate of #8574

Issue #8587, opened 2023-09-13
Files not loading with ctrl+p


jfeliu:
Type: <b>Bug</b>

Files search and dropdown are not loaded with ctrl+p when vscodevim extension is enabled.

Extension version: 1.26.0
VS Code version: Code 1.82.0 (8b617bd08fd9e3fc94d14adb8d358b56e3f72314, 2023-09-06T22:07:18.759Z)
OS version: Linux x64 6.2.0-32-generic
Modes:

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|12th Gen Intel(R) Core(TM) i7-1255U (12 x 3903)|
|GPU Status|2d_canvas: enabled<br>canvas_oop_rasterization: disabled_off<br>direct_rendering_display_compositor: disabled_off_ok<br>gpu_compositing: enabled<br>multiple_raster_threads: enabled_on<br>opengl: enabled_on<br>rasterization: enabled<br>raw_draw: disabled_off_ok<br>video_decode: enabled<br>video_encode: disabled_software<br>vulkan: disabled_off<br>webgl: enabled<br>webgl2: enabled<br>webgpu: disabled_off|
|Load (avg)|3, 2, 1|
|Memory (System)|15.06GB (5.27GB free)|
|Process Argv|--unity-launch --crash-reporter-id f802dc3d-0e9c-4186-a4c3-b81595088d9c|
|Screen Reader|no|
|VM|0%|
|DESKTOP_SESSION|ubuntu-xorg|
|XDG_CURRENT_DESKTOP|Unity|
|XDG_SESSION_DESKTOP|ubuntu-xorg|
|XDG_SESSION_TYPE|x11|
</details><details>
<summary>A/B Experiments</summary>

```
vsliv368cf:30146710
vsreu685:30147344
python383:30185418
vspor879:30202332
vspor708:30202333
vspor363:30204092
vswsl492cf:30256860
vslsvsres303:30308271
vserr242:30382549
pythontb:30283811
vsjup518:30340749
pythonptprofiler:30281270
vshan820:30294714
vstes263:30335439
vscod805cf:30301675
binariesv615:30325510
bridge0708:30335490
bridge0723:30353136
vsaa593:30376534
pythonvs932:30410667
py29gd2263:30792226
vsclangdf:30486550
c4g48928:30535728
dsvsc012:30540252
pynewext54:30695312
azure-dev_surveyone:30548225
vscccc:30803845
282f8724:30602487
[...]

Issue #8581, opened 2023-09-11
Conflict with 'CTRL + P'


RiskExe:
'CTRL + P' is an essential code navigation tool in native vscode. I really value being able to seemlessly navigate a codebase through the keyboard, as I'm sure many of us vim users can relate to. 

Suddenly I find myself unable to use this amazing code navigation tool built in to vscode because VsCodeVim is now using it for seemingly no reason. This happened on both my PC and laptop yesterday so this has to have been some update.

Of course this was easy to rectify. Just delete the VscodeVim binding and you're all set. But having the extension's whole userbase all of a sudden have to identify and solve this problem seems ridiculous. To be honest I don't even know what vim's use for the binding is, but whatever it was it can occupy another keyboard binding.

srackhall:
vim.mode == 'SearchInProgressMode'  

I think it may be caused by the failure of this' when 'expression, but unfortunately, I don't know how to solve this problem. Because this shortcut key is still useful to me in search mode, I cannot handle this issue by deleting or modifying the shortcut key. I hope friends who are also facing this situation can provide some help.

Panzer04:
The issue is someone removed "suggestWidgetVisible" from the old keymapping.

Previously the ctrl-p keymapping would apply what seems to be an up action when the editor was suggestion an autocomplete. Now, it's remapping it in all circumstances to up. It needs to be readded to fix the issue (or just remove? I don't really understand why it's present in the first place, but I am no Vim guru)

ShadowManu:
Just an extra note in case anyone wonders how to quickly workaround this in the meantime, edit your `settings.json` file with:
```
{
  "vim.handleKeys": {
    "<C-p>": false
  }
}
```

davdmoo:
I think the same issue has been mentioned in https://github.com/VSCodeVim/Vim/issues/8574

ghost:
As an alternative solution, you could also map double space to the quick open window.

```
    "vim.leader": "<space>",
    "vim.normalModeKeyBindingsNonRecursive": [
        {
            "before": [
                "<leader>",
                "<space>"
            ],
            "commands": [
                "workbench.action.quickOpen"
            ]
        }
    ]
```

Issue #8574, opened 2023-09-10
Default `Go to File...` (CTRL+P) is not working properly


FaisalBudiono:
**Describe the bug**
After updating to the latest version ([v1.26.0](https://github.com/VSCodeVim/Vim/releases/tag/v1.26.0)). The default behavior for `Go to File...` command is not working properly. It goes to previous line instead of showing `Quick Open` panel. Reverting to previous version `v1.25.2` fixed this issue.

**To Reproduce**
1. Open some project on your vscode
1. Update to the latest version ([v1.26.0](https://github.com/VSCodeVim/Vim/releases/tag/v1.26.0)).
1. Open any file
1. Make sure to focus on that file 
1. Press `ctrl+p`

The cursor will move up.

**Expected behavior**
The `Quick Open` panel should be open.

**Environment (please complete the following information):**

- Extension (VsCodeVim) version: v1.26.0
- VSCode version: 1.82.0
- OS: Windows 10 Home with WSL Ubuntu 20.04

ktvdev:
According to the changelog there's now handling for `CTRL+P` so it can be remapped. I fixed the issue by disabling its handling in `settings.json` under `vim.handleKeys` setting.

E.g. 
```json
"vim.handleKeys": {
	"<C-p>": false
},
```

0900304132:
> According to the changelog there's now handling for `CTRL+P` so it can be remapped. I fixed the issue by disabling its handling in `settings.json` under `vim.handleKeys` setting.
> 
> E.g.
> 
> ```json
> "vim.handleKeys": {
> 	"<C-p>": false
> },
> ```

Doing so will disable the ability to select the previous one in the prompt box.

davdmoo:
You can either disable the vim handler just like https://github.com/VSCodeVim/Vim/issues/8574#issuecomment-1712962173 said - or override it in vim.normalModeKeyBindings.

```json
    {
      "before": [
        "<C-p>"
      ],
      "commands": [
        "workbench.action.quickOpen",
      ]
    },
```

jestarray:
I thought I was going nuts when this wasn't working. I noticed it works when your cursor is focused on the terminal and I was like: huh... I coulda sworn it worked while I was in editor before too! Will disable, thanks

FaisalBudiono:
> According to the changelog there's now handling for `CTRL+P` so it can be remapped. I fixed the issue by disabling its handling in `settings.json` under `vim.handleKeys` setting.
> 
[...]

PR #8835, opened 2024-01-18
[Bug Fix] revert back the ctrl-p function for vscode file navigation

wswslzp:
Previous PR #7261 brought in a severe bug, which is complained by issue #8574, #8581, #8587, #8588. 
#7261 change hot-key ctrl-p's original function that allows for easily file navigation. So this PR is aimed to revert the #7261 change. And it does not affect the ctrl-p's vim function that move upward in suggestion. 

#8574, #8581, #8587, #8588. 

@grosssoftware can you help explain why you do this in #7261? IMO, it did not fix anything but broke an important and basic feature of VSCode.

### DIFF
--- a/package.json
+++ b/package.json
@@ -257,7 +257,7 @@
       {
         "key": "ctrl+p",
         "command": "extension.vim_ctrl+p",
-        "when": "editorTextFocus && vim.active && vim.use<C-p> && !inDebugRepl || vim.mode == 'CommandlineInProgress' && vim.active && vim.use<C-p> && !inDebugRepl || vim.mode == 'SearchInProgressMode' && vim.active && vim.use<C-p> && !inDebugRepl"
+        "when": "suggestWidgetVisible && vim.active && vim.use<C-p> && !inDebugRepl || vim.mode == 'CommandlineInProgress' && vim.active && vim.use<C-p> && !inDebugRepl || vim.mode == 'SearchInProgressMode' && vim.active && vim.use<C-p> && !inDebugRepl"
       },
       {
         "key": "ctrl+q",


grosssoftware:
So, back in the day (like pre-2000), it was very common to map ctrl-p and ctrl-n to :bp and :bn in Vim. Some of us still use those mappings. The power of Vim is in its ability to customize key bindings, but before the change in #7261, it wasn't possible to remap ctrl-p correctly.

I wouldn't say that it's broken becuase there are simple workarounds discussed in the issues you mentioned (and maybe those should be the default?), but I agree that masking a useful feature like "Go to File..." (on Windows, that is -- Mac uses cmd-p) isn't ideal or expected. If this reversion is accepted, then we still need a way to override ctrl-p. Suggestions?

tskj:
> I wouldn't say that it's broken becuase there are simple workarounds discussed in the issues you mentioned

I would like to make a strong case that it is in fact broken. While I have personally adopted the simple workaround, this isn't obvious for people new to the plugin. For anyone today checking out this great plugin, be it because they get referred by a friend or simply want to get into vim, they will think the plugin is completely broken, uninstall it and probably write it off permanently in their minds as "unusable" (and we risk losing that user forever). 

Personally I also thought for a long time I had messed something up in my QMK mappings before I realized this was a bug with the plugin, and it therefore took quite a long time before I even discovered there was a simple workaround. I think this applies to most users, and certainly 99% of new users - they will never find this workaround, they'll simply think this is a broken and unmaintained plugin. 

So again I strongly urge that fixing this bug must be the single highest priority on this project.

grosssoftware:
@tskj Sorry, I guess I misspoke: I agree 100% that "Go to File..." is broken. However, I think the change made in #7261 is not broken or a bug, but... it is incomplete. The change is necessary for anyone who hopes to remap ctrl-p. It is the same change that was accepted for ctrl-n.

After looking into this issue more and playing around with some changes on BOTH Windows AND macOS, I think the most sensible approach is to add `"<C-p>": false` to the defaults for `vim.handleKeys` [here](https://github.com/VSCodeVim/Vim/blob/9f467afae084034904f802db56c3fce8b1721eb8/package.json#L816), akin to the workaround suggested by others in the related issues. Ideally, we'd only make this the default for Windows because there is no issue in macOS, but I don't think that's a simple fix. Also, the README should be updated to note that overriding ctrl-p requires setting `"<C-p>": true` in `vim.handleKeys`.

Logically, this makes the most sense because that default value is saying "ctrl-p is very important to VsCode and the plugin shouldn't mess with it".

Thoughts?

tskj:
No worries, I definitely have a very poor understanding of all the moving pieces here. I don't have any particular opinion on that specific feature, except that I absolutely would be in favour of allowing people to remap ctrl-p as they please - that's a very good goal!

I think setting `"<C-p>": false` as a default on windows to fix this bug sounds like a very sensible course of action!

spiffytech:
> Ideally, we'd only make this the default for Windows

Linux too 🙂

tskj:
I can confirm that the solution with `"<C-p>": false` does indeed disable it in code suggestions. I agree that this is suboptimal, but I still think it's worth it (if reverting #7261 or whatever is off the table).

I would consider the current state of affairs an emergency for the project; every day this goes unfixed is quite a lot of new users we'll lose (potentially permanently).

black-pwq:
I strongly recommend adding the following to settings.json (and README.md in this project) instead of disabling use-ctrl-p if this commit is not accepted:
```json
  {
    "before": ["ctrl+p"],
    "commands": ["workbench.action.quickOpen"],
    "silent": true
  },
]
```
Arguments between vscode and vscodevim are unavoidable since both are editors. Personally, I prefer the vscode's way of ctrl+p. If anyone like @grosssoftware want to remap keys which aren't under vim's control, I would recommend them to try to add those in vscode keybindings.json.

wswslzp:
I agree with @tskj. 

I think we all take one principle for granted: **adding** new features is OK but **not breaking anything that work fine before**. 
What exactly the new feature @grosssoftware you want to **ADD** is enabling ctrl-p remapping. Breaking original function "Go to File ..." of ctrl-p of VSC is the **unexpected SIDE EFFECT**. From plugin user point of view, just one day after they updated the VIM plugin the "Go to File ..." function is not work anymore. And they don't know what happen ( which is exactly what I experienced before). They don't even know that now CTRL-P can be remapped implicitly!

So as @grosssoftware you said, #7261 is incomplete. Either you find a nice way to add enabling ctrl-p remapping feature without breaking anything else, or you should at least find some way to **notify** user of this change and provide some way to close feature to keep everything unchanged just like before.

wswslzp:
> I strongly recommend adding the following to settings.json (and README.md in this project) instead of disabling use-ctrl-p if this commit is not accepted:
> 
> ```json
> "vim.normalModeKeyBindings": [
>   {
>     "before": ["ctrl+p"],
>     "commands": ["workbench.action.quickOpen"],
>     "silent": true
>   },
> ]
> ```
> 
> Arguments between vscode and vscodevim are unavoidable since both are editors. Personally, I prefer the vscode's way of ctrl+p. If anyone like @grosssoftware want to remap keys which aren't under vim's control, I would recommend them to try to add those in vscode keybindings.json.

Your workaround is really nice. I haven't update VIM plugin since #7261 checked in until seeing your WAR. Before that, I have seen many WAR in some of the issues I have mentioned in PR description, but none of them actually work fine (ctrl-p not work in suggestion). 

But finding WAR is not the solution and it's not our user's responsibility to find such a WAR. It's @grosssoftware work and he should be the guy who provide fix.

codythompson-phoenix:
I ran into this while setting up a new work computer. This seems to be the most relevant thread/PR out of the several bugs/discussions listed. The last post on the subject seems to be almost exactly 1 year ago and the issue is still present. So I'm here to try to resurrect this thread.

 +1 for the normalMode workaround suggested by [black-pwq](https://github.com/black-pwq) being the default / fresh-install behavior. FWIW it's not just VS Code that uses ctrl-p as the default binding for find-file. Sublime Text does as well, if I'm not mistaken.

MauroPfister:
Does anyone else struggle with the work around that @black-pwq suggested? When I apply it to my settings.json I always get a `E492: Not an editor command: Files` error when pressin `ctrl + p` in the editor :thinking: . Is there something I overlooked?

> I strongly recommend adding the following to settings.json (and README.md in this project) instead of disabling use-ctrl-p if this commit is not accepted:
> 
> ```json
> "vim.normalModeKeyBindings": [
>   {
>     "before": ["ctrl+p"],
>     "commands": ["workbench.action.quickOpen"],
>     "silent": true
>   },
> ]
> ```

aYorky:
@MauroPfister Best I can say is double check for typos? I added this to my `settings.json` literally just now and it worked fine.

aYorky:
The thread I initially came from was opened in 2023 for v26. It is now just over 2 years to the day and ext. v30 and this is still not fixed. Given that the original change by @grosssoftware broke core functionality of vscode, imo this reversion change should have been a no-brainer to merge right away, and figure out how to support legacy workflows after that. Can this PR be updated and reconsidered for merging?

wswslzp:
Hi @aYorky, now I'm not using this VSCode extension anymore. 
**I strongly recommend you use VSCode NeoVim extension instead of this if your OS have installed NeoVim. VSCode NeoVim extension is far better than this and has no such stupid bug.** 
Since they would not accept this change, then we can leave and choose a better one. 
For all other people who have the same complaint and finally get here, leave and abandon this extension. You have another much better choice. 
Link: https://github.com/vscode-neovim/vscode-neovim

aYorky:
Thanks for the information. Ironically, now that you mention it, I'm pretty sure I'd tried the NeoVim extension several years ago when I was trying to learn vim motions, but this happened to be the extension I found this time around. I'll switch over.

PR #8890, opened 2024-02-19
Update ROADMAP.md

JaiminBrahmbhatt:
Marking jumps and marks from low priority to done.

Updating roadmap.md as it is source of truth of this repo on what is supported and what is not, after a certain point instead of adding everything and marking what is not done we started to just add what is supported leaving the room where I found :jumps and :marks being implemented but not being updated in roadmap.sh

N/A

I was not very sure if this should be marked as vscode specific implementation or not but atleast can mark it is as done for future VScodeVimmers you come to repo to find documentation on supported things.


J-Fields (approved):
Thanks!

PR #8955, opened 2024-03-25
Fix .vimrc warning & more

PainOchoco:
- Fix a typo in the warning message that appears when the .vimrc file fails to open.
- Now uses `loadConfiguration` from extension.ts instead of `configuration.load`.
- Now uses `context.globalStorageUri.fsPath` instead of `context.globalStoragePath` which is deprecated.


J-Fields:
Thank you!

Issue #8959, opened 2024-03-28
Add :Explore command


bvoq:
**Describe the solution you'd like**
Instead of running CMD+SHIFT+E I would like to use :Explore, as this is also the way I open files on vim 

**Additional context**
https://vim.fandom.com/wiki/File_explorer

PR #8960, opened 2024-03-28
Add Explore command (#8959)

bvoq:
Explore command is a default vim command that normally opens the file explorer. It is better than using CMD + SHIFT + E.
See: https://vim.fandom.com/wiki/File_explorer

#8959


J-Fields (approved):
Seems reasonable to me. Thanks!

J-Fields (changes requested):
Please run `yarn prettier`

Issue #8913, opened 2024-03-01
gv does not work with mouse selected text


zyd2001:
**Describe the bug**

**To Reproduce**

1. In the normal mode, use mouse to select some text.
2. Exit visual mode with "\<Esc\>" or "y"
3. Press "g", "v"
4. Nothing is selected or text previous selected with keyboard is selected.

**Expected behavior**
Enter the visual mode previous selected text by mouse is selected.

**Environment (please complete the following information):**

- Extension (VsCodeVim) version:
- VSCode version: 1.27.2
- OS:

PR #8965, opened 2024-03-31
Make 'gv' work with selection by mouse (and command)

zyd2001:
This PR makes 'gv' works with selection by mouse and also editor command. In Vim with mouse support, 'gv' can select previous mouse selected text without issue, but it does not work with this extension.

https://github.com/VSCodeVim/Vim/issues/8913

I understand that `handleSelectionChange()` is quite a complicated function. I didn't change anything but added code to assign `lastVisualSelection` accordingly.


J-Fields (approved):
Thanks!

PR #8985, opened 2024-04-09
Set scroll option with <C-u> and <C-d>

### LINKED ISSUE #8984: Set scroll option for <C-u> and <C-d>
**Describe the bug**
For commands `<C-u>` and `<C-d>` the preceding quantifier should not repeat the command that many times but rather set the scroll option and apply it once. I.e. `2<C-u>` should be equal to `:set scroll=2` followed by `<C-u>`.

**To Reproduce**

1.  `2<C-u>`

**Expected behavior**
Scroll two lines up (equal to `:set scroll=2` followed by `<C-u>`).

ontanj:
Set scroll option with `[count]<C-u>` and `[count]<C-d>`.

Rename `CommandMoveHalfPageDown` to `CommandCtrlD` and `CommandMoveHalfPageUp` to `CommandCtrlU`, since HalfPage is only true when `scroll=0`. This also aligns with `CommandCtrlE` and `CommandCtrlY`.

Wrap a docstring with \` since `<` and `>` need escaping in rendered format.

Fixes  #8984

### DIFF
--- a/src/actions/commands/scroll.ts
+++ b/src/actions/commands/scroll.ts
@@ -157,7 +157,7 @@ class CommandMoveFullPageDown extends CommandScrollAndMoveCursor {
 }
 
 @RegisterAction
-class CommandMoveHalfPageDown extends CommandScrollAndMoveCursor {
+class CommandCtrlD extends CommandScrollAndMoveCursor {
   keys = ['<C-d>'];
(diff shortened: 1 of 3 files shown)


J-Fields (approved):
Neat, I don't think I was aware of this behavior. Thanks!

PR #9008, opened 2024-04-28
Add `gB` keybinding for finding the previous match

### LINKED ISSUE #9000: Wrong cursor orientation after adding multi-cursor selection
**Describe the bug**
When using the `editor.action.addSelectionTo{Next,Previous}FindMatch` command, the direction/orientation of the selection is not taken into account.

So for example if you have selected the word `foo`, with the cursor on `f`, when you run `addSelectionToNextFindMatch` you will get a new selection with the cursor at `o` like this:
```bash
>|foo| # original selection

|foo|< # new selection
```

**To Reproduce**
`settings.json`
```json
{
  "vim.visualModeKeyBindings": [
    {
      "before": ["g", "n"],
      "commands": ["editor.action.addSelectionToNextFindMatch"]
    },
    {
      "before": ["g", "N"],
      "commands": ["editor.action.addSelectionToPreviousFindMatch"]
    }
  ]
}
```
**Expected behavior**
I expect the orientation of the cursor to always match the original selection.

Extension version: 1.26.1
VS Code version: Code 1.88.1 (e170252f762678dec6ca2cc69aba1570769a5d39, 2024-04-10T17:34:12.840Z)
OS version: Linux x64 6.6.28

musjj:
Fix: https://github.com/VSCodeVim/Vim/issues/9000

`gb` is a nice binding, so the inverse of it would be useful. 

You can bind `editor.action.addSelectionToPreviousFindMatch` in your config, but the nice thing about `gb` is that it aligns all the selection ends of the cursors.

fixes #9000

### DIFF
--- a/src/actions/commands/actions.ts
+++ b/src/actions/commands/actions.ts
@@ -2498,6 +2498,27 @@ class ActionTriggerHover extends BaseCommand {
   }
 }
 
+@RegisterAction
+export class ActionPreviousFindMatch extends BaseCommand {
+  modes = [Mode.Normal, Mode.Visual];
+  keys = ['g', 'B'];
+  override runsOnceForEveryCursor() {
+    return false;
+  }
+  override runsOnceForEachCountPrefix = true;
+
+  public override async exec(position: Position, vimState: VimState): Promise<void> {
+    await vscode.commands.executeCommand('editor.action.addSelectionToPreviousFindMatch');
+    vimState.cursors = getCursorsAfterSync(vimState.editor);
+
+    // If this is the first cursor, select 1 character less
+    // so that only the word is selected, no extra character
+    vimState.cursors = vimState.cursors.map((x) => x.withNewStop(x.stop.getLeft()));
+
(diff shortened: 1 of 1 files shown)


HenryTSZ:
I merged this PR and tested it locally, and found an issue: When the word are not in current view, It flashed the view to the selected word but jump back to the original one.

https://github.com/VSCodeVim/Vim/assets/55341525/cace293c-4a90-4e10-8b7d-fd780292e9c5

I discovered the problem through debugging

https://github.com/VSCodeVim/Vim/blob/3c33d00fdd37ba21d97a6a8c2daa7257d2722867/src/mode/modeHandler.ts#L1472-L1475

You need to add ActionPreviousFindMatch here as well

like this:

```ts
      const lastActionRun =
        this.vimState.recordedState.actionsRun[this.vimState.recordedState.actionsRun.length - 1];
      const isLastCursorTracked =
        lastActionRun instanceof ActionOverrideCmdD ||
        lastActionRun instanceof ActionPreviousFindMatch;
```

PR #9014, opened 2024-05-02
Don't close editors in other groups with `:only`

kopiczko:

Currently when running `:only` all editors from other groups are closed. This is surprising for a seasoned vim user because in vim all buffers would be kept open. It's especially annoying when there are terminal opened in editor area executing. some long-running tasks.


J-Fields (approved):
Thanks!

PR #9075, opened 2024-06-13
fixes #8975 The literal string "<leader>" is substituted for the leader key when repeating a command-line command

### LINKED ISSUE #8975: The literal string "<leader>" is substituted for the leader key when repeating a command-line command.
**Describe the bug**
The literal string "`<leader>`" is substituted for the leader key when repeating a command-line command using `@:` or `@@`. The result is that the command either errors or does something else than intended. The following will show the bug with two different leader keys. These were also tested on two environments, showing the same symptoms.

**To Reproduce the bug in case with leader `\` (backslash)**

1.  Go to vscode settings and search for "vim leader".
2.  Find the option "What key should `<leader>` map to in remappings?", and set the leader key to `\` (backslash).
3.  Open a new text file and insert the sentence `The	quick	brown	fox	jumps	over	the	lazy	dog.` (using tabs rather than spaces), and save the file.
4.  Try to replace tabs by spaces by doing command `:s/\t/ ` (after the last `/` is a single space).
5.  See the resulting string `The quick	brown	fox	jumps	over	the	lazy	dog.` (It replaced the first tab by a space, as expected).
6.  Try to repeat last command by `@:`.
7.  See thrown error `E486: Pattern not found: <leader>t`.
8.  Type `:` then arrow up to see most recent command, this reveals `:s/<leader>t/ `.

**To Reproduce the bug in case with leader ` ` (one space)**

1.  Go to vscode settings and search for "vim leader".
2.  Find the option "What key should `<leader>` map to in remappings?", and set the leader key to ` ` (one space).
3.  Open a new text file and insert the sentence `The	quick	brown	fox	jumps	over	the	lazy	dog.` (using tabs rather than spaces), and save the file.
4.  Try to replace tabs by spaces by doing command `:s/\t/ ` (after the last `/` is a single space). 
5.  See the resulting string `The quick	brown	fox	jumps	over	the	lazy	dog.` (It replaced the first tab by a space, as expected).
6.  Try to repeat last command by `@:`.
7.  See the resulting string `The quick<leader>brown	fox	jumps	over	the	lazy	dog.` (not the intended result, it replaced character `\t` by the string "`<leader>`").
9. Type `:` then arrow up to see most recent command, this reveals `:s/\t/<leader>`.
10.  Repeately press `@@`
11. See the resulting string `The quick<leader>brown<leader>fox<leader>jumps<leader>over<leader>the<leader>lazy<leader>dog.`.

**Expected behavior (same for both cases)**
Repeating the command with `@:` or `@@` replaces a tab by a space each time it is used.

**Environment 1:**

- Extension (VsCodeVim) version: v1.27.2
- VSCode version: March 2024 (version 1.88)
- OS: Windows 10

**Environment 2:**

- Extension (VsCodeVim) version: v1.27.2
- VSCode version: March 2024 (version 1.88)
- OS: Ubuntu 22.04.4 LTS (Jammy Jellyfish)

HenryTSZ:

fixes #8975 The literal string "<leader>" is substituted for the leader key when repeating a command-line command

#8975 

The commandList here saves the ex command, which typically do not include leader. 

https://github.com/VSCodeVim/Vim/blob/5365305330a69288f38acfd12ae85cba7a593fa6/src/state/recordedState.ts#L58-L59

Therefore, we do not need the following replace

https://github.com/VSCodeVim/Vim/blob/5365305330a69288f38acfd12ae85cba7a593fa6/src/state/recordedState.ts#L62-L64


src/state/recordedState.ts
```diff
@@ -56,7 +56,7 @@ export class RecordedState {
       this.commandList.length > 0
     ) {
       // Used for the registers and macros that only record on commandList
```
J-Fields, line 58: Let's clarify & expand this comment - it's not obvious why we're doing this early return.

J-Fields:
Can you please write a test case for this?

src/state/recordedState.ts
HenryTSZ: The commandList here saves the ex command, which typically do not include leader. 

https://github.com/VSCodeVim/Vim/blob/5365305330a69288f38acfd12ae85cba7a593fa6/src/state/recordedState.ts#L58-L59

Therefore, we do not need the following replace

https://github.com/VSCodeVim/Vim/blob/5365305330a69288f38acfd12ae85cba7a593fa6/src/state/recordedState.ts#L62-L64

HenryTSZ:
> Can you please write a test case for this?

I don't know how to write test cases. Do you have any documentation for reference?

mvandenhoek:
> I don't know how to write test cases. Do you have any documentation for reference?

I think the test will be similar to this one: https://github.com/VSCodeVim/Vim/blob/cfbd076c24e1497b957e7ab106ec80f46dcb60ea/test/macro.test.ts#L25-L30

To test for my issue, you'll need two such testcases. One where leader is ` ` (space) and one where it is `\` (backslash), and those keys then need to be used for a substitution command in `keysPressed`.

My guess is that you'll have to use the configuration to set the leader key before starting the test, see https://github.com/VSCodeVim/Vim/blob/master/src/configuration/configuration.ts

I'm not sure if for the test to simulate those key presses, the `keysPressed` will have to have the actual character e.g. `\`, or you have to write `<leader>`. Perhaps @J-Fields  can shed some light on this?

Either way I would first check that it fails the test before your change, and then confirm that it passes after your change.

J-Fields:
> I'm not sure if for the test to simulate those key presses, the `keysPressed` will have to have the actual character e.g. `\`, or you have to write `<leader>`. Perhaps @J-Fields can shed some light on this?

I _think_ you have to write `<leader>`, but could be misremembering

HenryTSZ:
@mvandenhoek @J-Fields 
Hi, Sorry, I've been a bit busy these past few days
After restoring the code
https://github.com/VSCodeVim/Vim/blob/5365305330a69288f38acfd12ae85cba7a593fa6/src/state/recordedState.ts#L58-L59
and adding the following test code

```js
  suite('`:` (command) register used as macro and command with leader key', () => {
    setup(async () => {
      const configuration = new testConfiguration.Configuration();

      // for testing with <leader>
      configuration.leader = 'o';

      await setupWorkspace(configuration);
    });

    newTest({
      title: 'Repeat :s and command with leader key',
      start: ['|old', 'old', 'old'],
      keysPressed: ':s/old/new\nj@:j@@',
      end: ['new', 'new', '|new'],
    });

    newTest({
      title: 'Repeat :s and command with leader key',
      start: ['|old', 'old', 'old'],
      keysPressed: ':s/old/new\nj@:j@@',
      end: ['new', 'old', '|old'],
      statusBar: 'E486: Pattern not found: <leader>ld',
    });
  });
```

The running result is as follows

```
    `:` (command) register used as macro and command with leader key
      1) Repeat :s and command with leader key
      ✔ Repeat :s and command with leader key (40ms)

  1 failing
  1) Record and execute a macro
       `:` (command) register used as macro and command with leader key
         Repeat :s and command with leader key:

      Document content does not match.
      + expected - actual

       new
      -old
      -old
      +new
      +new
```

it fails the test

mvandenhoek:
> @mvandenhoek @J-Fields Hi, Sorry, I've been a bit busy these past few days
...
> I think this should prove that I solved this bug and did not damage the original functionality

I think that if
- `\n` in the test is the same as pressing enter (I expected you'd have to write `<enter>` or something alone those lines, but I suppose that's just my confusion).
-  The test before the change produces an error: `Pattern not found: <leader>ld`
-  The test after the change produces the expected result of `end: ['new', 'new', '|new']`

Then I think that solves the issue. So if it didn't break anything else I would approve of it.

src/state/recordedState.ts
mvandenhoek: I think J-Fields means your explanation should be in the comment in the code rather than in the comment on this PR.

HenryTSZ:
> \n in the keysPressed of the test simulates pressing enter (I expected you'd have to write <enter> or something alone those lines, but I suppose that's just my confusion).

I think the existing test cases are all used `\n`, so there should be no problem

> I was also trying to run the tests myself, and I could run it locally with sudo yarn test, but it runs the entire test suite and I couldn't find how to just run a single test. Did you manage to figure that out?

newTest -> newTestOnly

src/state/recordedState.ts
HenryTSZ: Oh, I understand. Let me add a comment

src/state/recordedState.ts
```diff
@@ -56,6 +56,8 @@ export class RecordedState {
...
+      // The commandList here saves the ex command, which typically do not include <leader>.
+      // Therefore, we do not need the following replace
```
mvandenhoek, line 60: I think this would be a bit clearer than referring to "the following replace".

```suggestion
      // The commandList here saves the ex command, so we should not regex replace the
      // leader key by the literal string '<leader>', as would have been done for a normal command.
```

mvandenhoek (changes requested):

mvandenhoek (approved):
Looks good to me, but my knowledge of the internal workings of the vscodevim plugin is limited, so there should also be another reviewer.

test/macro.test.ts
```diff
@@ -129,6 +129,24 @@ suite('Record and execute a macro', () => {
...
+
+      await setupWorkspace(configuration);
+    });
+
+    newTest({
+      title: 'Repeat :s and command with leader key',
+      start: ['|old', 'old', 'old'],
+      keysPressed: ':s/old/new\nj@:j@@',
+      end: ['new', 'new', '|new'],
+    });
```
J-Fields, line 147:
```suggestion
    newTest({
      title: 'Repeat :s and command with leader key',
      config: { leader: 'o' },
      start: ['|old', 'old', 'old'],
      keysPressed: ':s/old/new\nj@:j@@',
      end: ['new', 'new', '|new'],
    });
```

J-Fields:
Thank you!

PR #9089, opened 2024-06-24
make `tabedit` work with relative path, fix #4730

### LINKED ISSUE #4730: Tab autocomplete in command does not match path with tabedit
**Describe the bug**
Strangely the tab autocomplete reflects the cwd of the open file, but the filepath it tries to open is the project root. 

**To Reproduce**
with a file tree like:

project
.+ _ _ banana.txt
.+ _ _ inner
....... + - - inner_file.txt
....... + - - other_inner.txt
....... + - - another_inner.txt

If I have `inner_file.txt` open and I type `:tabedit+tab` it will cycle through `inner_file.txt`, `other_inner.txt`, `another_inner.txt` (as expected). if I press `return` when `another_inner.txt` is displayed, I get the following error: 
```
Unable to open 'another_inner.txt': Unable to read file '/Users/myuser/project/another_inner.txt' (Error: Unable to resolve non-existing file '/Users/myuser/project/another_inner.txt').
```
so the path it attempts to open != the path it displays during autocomplete (for tabedit). 

**Expected behavior**
When a file is presented in autocomplete for tabedit it will open in a new tab when I hit enter.

- Extension (VsCodeVim) version: 1.13.1
- VSCode version: 1.44.0
- OS: Catalina 10.15.4

**Additional context**
This works fine as far as I can see with `e:`, but not with `:tabedit`.

iblislin:
make `tabedit` work with relative path (like the `:e <path>` does)

 #4730

### DIFF
--- a/src/cmd_line/commands/tab.ts
+++ b/src/cmd_line/commands/tab.ts
@@ -9,6 +9,7 @@ import {
   FileOpt,
   bangParser,
   fileCmdParser,
+  fileNameParser,
   fileOptParser,
   numberParser,
 } from '../../vimscript/parserUtils';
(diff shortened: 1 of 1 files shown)


J-Fields (approved):
Thank you!

PR #9095, opened 2024-06-25
Remove leading "$" characters in README so terminal commands can be copied and used directly

markjaquith:

GItHub helpfully adds a "copy" button to code snippets. But the prefixing of the commands with `$` means you cannot copy/paste the commands without editing them.

This just removes the `$ ` prefixes from the terminal commands in the README.


J-Fields:
Thanks!

PR #9100, opened 2024-06-28
fixes #8783 add easymotionSearchLines

### LINKED ISSUE #8783: Is there any way to set a limit for EasyMotion?
I really like easymotion, I've made a remap of `leader, leader, 2, f` to `f` (same with F), so it's much better to use that way. The problem is when u have long text / code, and here is not as efficient, cause u can literally have just 1 word in your view, but at the end u need to press the keys to go there, cause easymotion is taking all the words on the whole file. 

For a better use of Insearch, Easymotion & Skeak: 
- / & ? = search whole file 
- f & F = search just in a certain amount of lines (e.g. 50)
- s & S = search nearest word

**Solution**
I think that adding a limit like. `vim.easymotion.linesSearch: 50` that is gonna look just in a range of 50 lines e.g:
f = 50 lines forward, F = 50 lines backward)

HenryTSZ:

fixes #8783 add easymotionSearchLines

#8783 

### DIFF
--- a/README.md
+++ b/README.md
@@ -546,6 +546,7 @@ Based on [vim-easymotion](https://github.com/easymotion/vim-easymotion) and conf
 | vim.easymotionMarkerFontWeight                   | The font weight used for the marker text.                                                                | String  | 'bold'                                             |
 | vim.easymotionKeys                               | The characters used for jump marker name                                                                 | String  | 'hklyuiopnm,qwertzxcvbasdgjf;'                     |
 | vim.easymotionJumpToAnywhereRegex                | Custom regex to match for JumpToAnywhere motion (analogous to `Easymotion_re_anywhere`)                  | String  | `\b[A-Za-z0-9]\|[A-Za-z0-9]\b\|_.\|#.\|[a-z][A-Z]` |
+| vim.easymotionSearchLines                        | Set the number of lines for the easymotion search.                                                       | Number  | 100                                                |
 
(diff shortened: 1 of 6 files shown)


J-Fields:
Does https://github.com/easymotion/vim-easymotion have an equivalent setting?

J-Fields:
And is there any good reason we shouldn't just stop at the screen's edge, rather than make this configurable or hardcoded to 100?

HenryTSZ:
> Does https://github.com/easymotion/vim-easymotion have an equivalent setting?

Vim EasyMotion itself does not directly provide a configuration item for setting the number of rows to be crossed during search

HenryTSZ:
> And is there any good reason we shouldn't just stop at the screen's edge, rather than make this configurable or hardcoded to 100?

Firstly, we usually only see the content within the current screen range

Secondly, if using a vertical display, the screen display range may exceed 100 lines

kaiguang:
> Secondly, if using a vertical display, the screen display range may exceed 100 lines

Besides the large display, if I have a fold with more than 100 lines, I cannot jump over that fold area with EasyMotion. Being able to change the 100 to something larger will be helpful.

Also, an idea solution would be jumping anywhere within the screen range, and I don't really to need to search within the folded lines.

```
1 <cursor_here>
2
3...100 <folded_lines>
101
102
103                   <-- cannot jump here
```

PR #9112, opened 2024-07-03
Fix escaping of multiple forward slashes in visualstar search

### LINKED ISSUE #9111: visualstar fails to properly escape search string
**Describe the bug**
When using the visualstar feature on a selection containing multiple forward slashes, only the first slash is escaped in the regex used for the search.

**To Reproduce**
1. Paste the following text into a file
    ```text
    //a
    //a
    ```
2. Select `//a` in visual mode
3. Press `*` (with `vim.visualstar` enabled in settings)
4. Observe no matches (the regex used is `\//a`)

**Expected behavior**
The regex used for the search should be `\/\/a`

- Extension (VsCodeVim) version: 1.27.3
- VSCode version: 1.90.2
- OS: Windows 10

zaneduffield:
Allows searching for visual selections that contain multiple forward slashes.

Fixes #9111

### DIFF
--- a/src/actions/commands/search.ts
+++ b/src/actions/commands/search.ts
@@ -99,7 +99,7 @@ async function createSearchStateAndMoveToMatch(args: {
     return;
   }
 
-  const escapedNeedle = escapeRegExp(needle).replace('/', '\\/');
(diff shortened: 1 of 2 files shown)


J-Fields (approved):
Nice catch; thank you!

PR #9120, opened 2024-07-06
Fix 7397 dot with numbered registers

### LINKED ISSUE #7397: Repeat command (.) when using numbered registers keeps using the same register instead of incrementing it by one each time
**Describe the bug**
The repeat command (.) used with an explicit numbered register from 1 to 9 should increase the register number by one each time it is called.

From vim's  `:h undo-tree`:
```
                                                redo-register
If you want to get back more than one part of deleted text, you can use a
special feature of the repeat command ".".  It will increase the number of the
register used.  So if you first do '"1P', the following "." will result in a
'"2P'.  Repeating this will result in all numbered registers being inserted.
```

**To Reproduce**

1.  Open an empty file
2.  Press `i` to enter Insert mode
3.  Type `1 2 3`
4. Press Escape to exit Insert mode
4.  Press `gg` to go to the beginning of the file
5. Press `"1daw..`
6. Press `"1p..`

**Expected behavior**
The output shoud be `1 2 3`. The output now is `333`

- Extension (VsCodeVim) version: v1.21.10
- VSCode version: 1.63.2
- OS:  Ubuntu 20.04.2 LTS

SirTomme:

It fixes #7397 and adds three tests for this.

#7397 

Hi there I am happy to make my first contribution to this amazing project!

I implemented the fix for all commands that start with a (numbered) `CommandRegister` action. I could not think of or find a dot repeatable command that has a numbered register in it but does not not start with it. If there is one, please let me know.

BTW: is there a way to filter tests? I do not want to run all tests _while_ I'm trying to fix a small bug 😉

### DIFF
--- a/src/mode/modeHandler.ts
+++ b/src/mode/modeHandler.ts
@@ -31,6 +31,7 @@ import {
   CommandInsertAtCursor,
   CommandNumber,
   CommandQuitRecordMacro,
+  CommandRegister,
   DocumentContentChangeAction,
 } from './../actions/commands/actions';
 import {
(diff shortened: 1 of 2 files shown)


J-Fields:
> BTW: is there a way to filter tests? I do not want to run all tests _while_ I'm trying to fix a small bug 😉

See `newTestOnly`

J-Fields (approved):
Thanks!

PR #9127, opened 2024-07-12
Implement basic separator targets from `targets.vim`

rwblickhan:

This adds basic [separator text objects](https://github.com/wellle/targets.vim/#separator-text-objects) from `targets.vim`. In particular, it supports the basic `i`/`a` behavior for all separators supported by `targets.vim`. This PR notably does _not_ implement:

- [`I`](https://github.com/wellle/targets.vim/#inside-separator) or [`A`](https://github.com/wellle/targets.vim/#around-separator) behavior to select inside-without-whitespace or around-including-trailing
    - These are complicated by visual mode — `I` and `A` immediately fire `ActionGoToInsertVisualLineModeCommand` when in visual / visual line / visual block mode. That said, `i` and `a` seem more broadly useful — for instance, `mini.ai` actually [only implements those two](https://github.com/echasnovski/mini.nvim/blob/main/doc/mini-ai.txt)! 
- [Next/last separators](https://github.com/wellle/targets.vim/tree/master?tab=readme-ov-file#next-and-last-separator)
- Expanding selections
    - Arguably all of these should be subclasses of `ExpandingSelection`, but that feels out-of-scope for this basic PR (and anyway `targets.vim` doesn't support that either as far as I can tell). I welcome someone else to refactor 😉

Implements a bit more of #601.

### DIFF
--- a/package.json
+++ b/package.json
@@ -697,6 +697,11 @@
           "markdownDescription": "Enable last/next movements for bracket objects.",
           "default": true
         },
+        "vim.targets.separatorObjects.enable": {
+          "type": "boolean",
+          "markdownDescription": "Enable separator text objects.",
+          "default": true
+        },
         "vim.targets.smartQuotes.enable": {
           "type": "boolean",
           "markdownDescription": "Enable the smart quotes movements from [targets.vim](https://github.com/wellle/targets.vim#quote-text-objects).",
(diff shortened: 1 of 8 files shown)


max-sixty:
Would love to see this merged — particularly helpful for removing `bar()` in `foo.bar().baz()`...

PR #9140, opened 2024-07-18
fixes #8872 issue with surround

### LINKED ISSUE #8872: [bug]: vim surround using html tags
**Describe the bug**

When I use Vim surround key binding for adding an HTML tag around something, and also add an attribute like `class` or `id`, the closing tag generated also has the same attributes which is not what is expected.

**To Reproduce**

1.  Go to any line.
2.  Select the entire line with `V` (shift + v) in `normal` mode
3.  Add a tag around it with the following key presses `St`
4.  It'll show a text box to add the tag which you want to surround with.
5. Write a tag like `div class="container"`
6. It'll generate a closing tag like `</div class="container">`

**Expected behavior**

When we insert a tag with attributes, the attributes should only be included in the opening tag and excluded in the closing tag.

**Screenshots**

https://github.com/VSCodeVim/Vim/assets/108579856/19c969d9-3a00-427c-afb0-07a2e6e898d0

- Extension (VsCodeVim) version: 1.27.2
- VSCode version: 1.86.0 (Universal)
- OS: MacOS 14.0

Nestastnikos:
The current behavior is clearly a bug, see related issue.

#8872 

One special case that I found, is that if we do the repro steps from the issue (`VStt`) and we input a string ` p` (the first character is space). The resulting closing tag will be `</>`. It might be not be intutiive, but the same behavior occurs also in the original vim-surround implementation, so I intentionally didn't deal with this special case.


J-Fields (approved):
Thanks!

J-Fields:
Tests seem to be failing - mind looking into that @Nestastnikos?

Issue #9019, opened 2024-05-06
Cursor horizontal character column position not restored when vertically navigating lines with 'j' or 'k' "jk", :startofline desiredColumn


MrMattBusby:
**Describe the bug**
When navigating vertically using, e.g. <kbd>j</kbd> and <kbd>k</kbd>, the cursor's horizontal position (column) isn't restored after navigating through a shorter line.

**To Reproduce**
Normal mode:
```
1)  int variable1[*4]; // Cursor starts on "4"
2)  char c[3]*;        // Use "j" twice past this shorter line, cursor is moved to end of this line
3)  int variable2[*6]; // Should end up on "6" (does in VIM on MING and Unix)
3') int varia*ble2[6]; // On VSCodeVim you end up here instead directly below the line 2's ";"
```
the cursor's column position was reset by navigating vertically but should have saved the "intended" column from earlier.

**Expected behavior**
The horizontal cursor position should be restored when navigating between lines of different lengths.

Native VIM behavior. See `:help startofline`: "cusor is kept in the same column (if possible)".

**What I've Tried**
Playing with wrapping like `vim.foldfix:false`, and `set` for `sol` and `compatible` didn't fix.

**Workaround !!**
- ✅ https://github.com/VSCodeVim/Vim/issues/8912#issuecomment-1971575743

**Environment (please complete the following information):**

- Extension (VsCodeVim) version: 1.27
- VSCode version: 1.89
- OS: Windows
- This works **natively** VIM in MING64 and Unix versions 8.1 and 9.0.

**Additional context**
[VIM which "set" setting maintains cursor horizontal character position when navigating up/down lines?](https://stackoverflow.com/questions/78438200/vim-which-set-setting-maintains-cursor-horizontal-character-position-when-navi)

**Related**

> Maybe these are all related and could be merged?

- related https://github.com/VSCodeVim/Vim/issues/8912
- related https://github.com/VSCodeVim/Vim/issues/8840
- related https://github.com/VSCodeVim/Vim/issues/8200
- related https://github.com/VSCodeVim/Vim/issues/8108 
- related https://github.com/VSCodeVim/Vim/pull/5080 (failed)
- related https://github.com/VSCodeVim/Vim/issues/4797
- related https://github.com/VSCodeVim/Vim/issues/4441
- related https://github.com/VSCodeVim/Vim/issues/4428
- related https://github.com/VSCodeVim/Vim/issues/1323

Issue #8912, opened 2024-02-29
j/k don't preserve column


bowernite:
I know this has been posted in the past for gj/gk, but this also seems to be broken whenever moving the cursor up or down.

Basically, in this scenario:

```
abc
a
abc
```

If you have your cursor on `c` and move to the other `abc` line, the cursor ends up on `a` in the other line for any of the following:

1. normal mode `j`/`k`
2. insert mode `up`/`down`
3. visual mode `j`/`k`

If I reload VS Code with extensions disabled and use `up`/`down`, this bug does not exist.

babramczyk:
FWIW, this seems to be a bandaid fix:

```
  "vim.normalModeKeyBindingsNonRecursive": [
    {
      "before": ["j"],
      "commands": ["cursorDown"]
    },
```

babramczyk:
FWIW, this seems to be a bandaid fix:

```
  "vim.normalModeKeyBindingsNonRecursive": [
    {
      "before": ["j"],
      "commands": ["cursorDown"]
    },
```

darianmorat:
This is caused for this option: `"vim.foldfix": true`, once disabled, is gonna work just fine.

p-se:
> FWIW, this seems to be a bandaid fix:
> 
> ```
>   "vim.normalModeKeyBindingsNonRecursive": [
>     {
>       "before": ["j"],
>       "commands": ["cursorDown"]
>     },
> ```

Please note that this seems to work with `"vim.foldfix": true`!

Issue #8840, opened 2024-01-21
j and k don't move into the same column


tinhajj:
**Describe the bug**
If you push j or k and you are moving to a line with a different tab indentation the cursor doesn't stay in the same column even if it could have.

**To Reproduce**
1. Start on a line that has no whitespace in the beginning
2. Use j or k to move to a line that has tab characters in front of it

Also I believe you need to make sure tab display size is greater than 1

**Expected behavior**
The cursor should stay in the same column if its possible

**Screenshots**
[2024-01-21 15-15-02.webm](https://github.com/VSCodeVim/Vim/assets/5882922/9080a453-44a0-4fac-904a-6cd963b06d3d)

**Environment (please complete the following information):**
- Extension (VsCodeVim) version: v1.27.2
- VSCode version: 1.85.1
- OS: Windows, happens on Mac OS too

TheDevMinerTV:
This is probably related to https://github.com/VSCodeVim/Vim/issues/8108

mmviamm:
and it happens when you set the foldfix true in settings
`"vim.foldfix": true`

PR #9184, opened 2024-08-02
[Small Fix] Fixed an issue with the foldfix setting (v0.2)

DragosStefanNecula:
### What this PR does / why we need it:
When enabling the foldfix setting, the cursor stops remembering the column. This fixes that and it is needed because the foldfix implementation is really useful - however it is rendered unusable because the remember column feature is really important as well.

### Which issue(s) this PR fixes
Fixed an issue with the foldfix setting where the cursor would stop remembering the column it was on (and resetting to the smallest line-end it would encounter). This fix unfortunately only works on the j and k keys, and doesn't also fix the same issue with the arrow keys. 

### Special notes for your reviewer:
In the second commit I found a easier way to do it, however I left both in here in case the maintainer might prefer one over the other. If this is not up to the standard of this extension, I am more than happy to discuss how to fix this bug in a way that is. I hope that you will have a wonderful week ahead of you :).


DragosStefanNecula:
related https://github.com/VSCodeVim/Vim/issues/9019
related https://github.com/VSCodeVim/Vim/issues/8912
related https://github.com/VSCodeVim/Vim/issues/8840
related https://github.com/VSCodeVim/Vim/issues/8200
related https://github.com/VSCodeVim/Vim/issues/8108
related https://github.com/VSCodeVim/Vim/issues/4797
related https://github.com/VSCodeVim/Vim/issues/4441
related https://github.com/VSCodeVim/Vim/issues/4428
related https://github.com/VSCodeVim/Vim/issues/1323

J-Fields:
Thanks!

Issue #9214, opened 2024-08-21
mouseSelectionGoesIntoVisualMode from insert mode too


sygi:
**Describe the bug**
Currently, even when `vim.mouseSelectionGoesIntoVisualMode` is on, selecting with a mouse when being in an insert mode doesn't change it (see [here](https://github.com/VSCodeVim/Vim/blob/acd49784aefa5ada88dbc7c382f8925e4bce49b3/src/mode/modeHandler.ts#L417)). This is different than a default behavior in vim.

**To Reproduce**
Enter insert mode with `i`, then select with mouse in any file.

**Expected behavior**
vim should enter `visual` mode.

**Additional context**
As I understand it, the original plan was to implement full-on Select Mode (#5716), however, 4 years after #5842 was opened, I think it's the right time to declare PR bankruptcy there and address the issues in smaller scale.

Note: this issue is particularly annoying when working in jupyter notebooks, where each cell has a separate "mode" and it's difficult to track in which cell when selecting with mouse you'll be in an insert vs virtual modes.

sygi:
It seems that the decision to diverge from the vim behavior was made [here](https://github.com/VSCodeVim/Vim/commit/8e23c88ff43e591919dfca9859c7e8cce0ac3647). Is there any context available? Can we maybe have a setting which controls this?

nguymin4:
@sygi It was so long ago so I cannot say for sure. Nowadays I only use neovim or vim, and the behaviour is if I'm in Insert mode and select some text with my mouse, it actually triggers the terminal selection mode instead of going to neovim/vim visual mode

It could be my setting that disables mouse somewhere as well, but then it also means that behavior similar to `mouseSelectionGoesIntoVisualMode: false`

Side note: I have terrible perfomance issue with VSCodeVim when using Jupyter notebook so I usually disable it ...

sygi:
Hmm, the behavior you describe happens when you disable mouse handling by vim; you'll always be able to achieve that via ~/.vimrc.

I kinda like the vim mode in jupyter notebook, except for this issue ^^
I prepared a quick fix above.

nguymin4:
I just want to say that, please double check with other maintainers about the intention and plan with that setting. Even if it may not affect me anymore, people may notice the changed behavior and ask about it (think about non-jupyter users). And we end up in a complaining loop.

Anyways, I do wish the vim support for notebook become better.

sygi:
Of course, that's the point of this thread / issue :)

Issue #2410, opened 2018-03-02
Using mouse in insert mode causes VIM to think it is in normal mode


kjgoodrick:
**Is this a BUG REPORT or FEATURE REQUEST?** (choose one): BUG REPORT

**What happened**: When in insert mode using the mouse to move the cursor causes the cursor to switch to block and the mode "changes" to normal; however, VIM is actually in insert and as soon as any letter is pressed the mode "switches" to insert and the character(s) are inserted. If the user settings file is open the cursor behaves as expected.

**What did you expect to happen**: Stay in insert mode, cursor remains a line and moves to clicked location. This happens if the user settings file is open.

**How to reproduce it (as minimally and precisely as possible)**: be in insert mode click anywhere else in the document.

This has also been reported by christiangenco [here](https://github.com/VSCodeVim/Vim/issues/1871#issuecomment-369363878) (Includes a video).

**Environment**:
- Extension Name: vim
- Extension Version: 0.11.0
- OS Version: Windows_NT x64 10.0.16299
- VSCode version: 1.20.1

```json 
{
	"activationTimes": {
		"startup": true,
		"codeLoadingTime": 283,
		"activateCallTime": 2,
		"activateResolvedTime": 19,
		"activationEvent": "*"
	}
}
```

christiangenco:
I've been having the same bug and it's driving me absolutely crazy. Here's [a screen recording of it happening to me](https://www.youtube.com/watch?v=9JUkdspipQw&feature=youtu.be).

No joke I feel like I'm having a stroke every time it happens.

Chillee:
I can't seem to reproduce this. Could you confirm that something we did in the last release didn't somehow fix it?

jpoon:
Can't repro on my end with 0.11.1

ivanvoznyakovsky:
this is fixed in last release of the extension but the fixe introduced another bug. now when you in insert mode and change cursor position by clicking status bar is blinking showing normal mode and that going back to insert mode.

christiangenco:
@ivanvoznyakovsky's bug is happening to me too (and was happening before 0.11.2).

Here's [another video of it happening](https://youtu.be/BzAPnMtnOeY). Again, strangely, it's fixed if I open the settings window.

Here are the plugins and versions I have installed:

[...]

Issue #1871, opened 2017-06-25
Selecting with mouse or double clicking on a word to select it should NOT go into visual mode


gotofritz:
Unlike what #290 says, selecting with mouse should NOT go into visual mode.

We are still using VS Code, not Vim - if I use the mouse it means I DON'T want to use Vim commands. If I wanted to only use Vim commands, then I would use Vim.

At least make it configurable so we can toggle the behaviour on / off in the preferences!

Thanks.

Chillee:
Sounds reasonable, but out of curiosity, what are the downsides of going into visual mode?

gotofritz:
Because some of us, those whose vim foo is not yet so strong, prefer to dip in and out of vim.

Note that you can still use the cursor in insert mode, and if you are in insert mode and use LEFT + SHIFT + ALT you select the whole word - remaining in insert mode. Then if you type letters you will replace the selection with what you typed. Basically VS Code behaves as if VSCodeVim was not installed when you use the cursor keys. Vim purists will probably consider it a bug, but for us beginner it's a god send. I suggest the mouse could be treated the same way. If I use the mouse it means I am not doing Vim stuff, basically. If I want visual mode I use the keyboard. 

The problem is that (1) it wasn't expected, but even now that I know (2) it have to frantically click escape - i escape - i escape i until it gets back to insert mode. 

Maybe it's just because I was used to Vintageous in Sublime. VSCodeVim is generally much better, but I think vintageous gets the balance of Vim + non-Vim right. Sometimes it's good to take a step back and just let VS Code be VS Code IMHO (CMD-D is another area where VSCodeVim gets in the way).  A config settings to allow the mouse to be ignored would be great.

Thanks for all the great work, either way.

galenjiang:
I think it is reasonable, this make me can not multi select something to use normal operation.

galenjiang:
Many wired operation cause by it,  selected by mouse, and can not delete normally by backspace and so on.

knownasilya:
Same here, this has changed in the recent release I think. Basically, if you hit 'i' you are in insert mode, now if you go to select text and try to do some non-vim stuff it used to just work. Now it no longer works at all. I have to disable vim mode to use the mouse. This is a pretty big bug.

Chillee:
Alright I hear you guys. I'll make it an option in the settings.

heeton:
@Chillee Thanks, an option is ideal

coryshaw1:
[...]

PR #9215, opened 2024-08-21
Move to visual mode on mouse selection

sygi:
Currently, selecting text with mouse changes mode to Visual, unless we started from the insert mode. Multiple people [complained](https://github.com/VSCodeVim/Vim/issues/2410#issuecomment-574908393) about this departure from learned vim behavior. There is a setting that supposed to control it (#1871), but it doesn't address insert->visual for some reason.

#9214


sygi:
dropping this as it pollutes my space and I stopped using vscode anyway

PR #9226, opened 2024-08-26
Avoid needing two `MoveLeft`s if the current character is a low surrogate

semicube:
It fixes an issue where you'll sometimes need to press 'h' twice to go to the next left char, if your current char is a surrogate pair. 

Happens when the character corresponding to the current cursor position is a low surrogate, so you'll only get to the next left char after you've moved to the high surrogate char (two `MoveLeft`s).

Before:
before-moveleft

After this PR:
after-moveleft


J-Fields:
Mind adding a test case?

semicube:
> Mind adding a test case?

Sure, I've added a simple test case. Would this be enough?

J-Fields (approved):
Thanks!

Issue #7068, opened 2021-09-08
Unexpectedly entered visual mode in snippet


KenMSK:
**Describe the bug**
With a custom snippet, vscode enters visual mode when [tab] is pressed (to use the snippet) 
This blocks further editing until [esc] and [i] is pressed to enter insert mode again.

**To Reproduce**
1. Add the following custom snippet:
( How to use snippets: https://code.visualstudio.com/docs/editor/userdefinedsnippets )
```
{ // file: Buggy_Snippet.code-snippets
  "Snippet Test": {
    "prefix": "SnippetTest",
    "body": ["${1:default}", "${1}"],
    "description": "Snippet that triggers visual mode"
  },
}
```
2. In vscode editor, open a file and type "snippettest" and wait/trigger suggestions. Select the appeared snippet and press [tab]/[enter] to use the snippet.
3. Visual mode is triggered.

**Expected behavior**
Vscode should stay in insert mode. 

**Screenshots**
![image](https://user-images.githubusercontent.com/51538854/132468593-884143c6-43b1-4961-ac20-76cb2200bc7e.png)

**Environment:**

- Extension (VsCodeVim) version: 1.21.7
- VSCode version: 1.60.0
- OS: macOS Big Sur version 11.1

About VSCode:
Version: 1.60.0
Commit: e7d7e9a9348e6a8cc8c03f877d39cb72e5dfb1ff
Date: 2021-09-01T10:54:53.442Z
Electron: 13.1.8
Chrome: 91.0.4472.164
Node.js: 14.16.0
V8: 9.1.269.39-electron.0
OS: Darwin x64 20.2.0

KenMSK:
A temporary work around is to remove the default value. But this bug can be triggered by preceding snippet variables, sometimes it is impossible to fix.

draivin:
Seems related to https://github.com/VSCodeVim/Vim/issues/5240, I'm not sure if it was never fixed properly or if a regression happened.

Calling `editor.insertSnippet` with a non-empty range in the `location` parameter changes the active mode to visual mode. The expected behavior is for it to stay in insert mode.

bselwe:
The same issue happens when using code actions (CMD .) for Flutter such as "Wrap with Padding". Notice how it creates two cursors (one in the insert mode, the other in the visual mode). In this case, the visual mode cursor should be cancelled and VIM should go into the insert mode from Flutter code action.

[...]

Issue #6465, opened 2021-04-11
Multi cursor comes up in Snippets


rickalex21:
I think this might be a duplicate but the problem is still happening, so ... I did get an error #6464 but maybe I entered a wrong key. I later figured out that pressing 'c' makes it work as normal.

**Describe the bug**

Multi cursor comes up, when using this snippet. Pressing 'c' makes it work as normal but don't know if it's still a bug that needs to be fixed.

```json
"Shortcode link": {
  "prefix": "link",
  "body": [
    "[${1:pandoc page}]({{< link \"/code/pandoc\" >}} \"$1\")"
  ],
  "description": "Shortcode link"
},
```

**To Reproduce**

Use the snippet above.

**Expected behavior**

Work normally without multi cursor. This snippet works fine.

```json
"foo": {
  "prefix": "foo",
  "body": [
    "foo is ${1:bar} now $1"
  ],
  "description": "foo"
},

```

**Environment (please complete the following information):**

- Extension (VsCodeVim) version: v1.19.2
- VSCode version:
 Version: 1.55.1
Commit: 08a217c4d27a02a5bcde898fd7981bda5b49391b
Date: 2021-04-09T00:42:12.838Z
Electron: 11.3.0
Chrome: 87.0.4280.141
Node.js: 12.18.3
V8: 8.7.220.31-electron.0
OS: Darwin x64 20.3.0
- OS: Mac os X

Issue #5969, opened 2021-01-06
Failed to enter insert-mode if a snippet starts with a linked placeholder.


fingershi:
**To Reproduce**

See this simple snippet:
```
	"test": {
		"prefix": "test",
		"body": [
			"${1:some_placeholder}, $1" 
		],
	},
```

After typing "test" and expanded, vim enters visual mode instead of insert mode.

The bug occurs only if all 3 conditions below are met:
1. The snippet must start with a placeholder.  "blah blah ${1:some_placeholder} $1" works fine.
2. The placeholder must be linked.    "${1:some_placeholder} blah blah" works fine.
3. It should be a placeholder, not a simple tabstop.   "$1 blah blah $1" works fine.

**Environment (please complete the following information):**

- Extension (VsCodeVim) version: 1.18.5
- VSCode version: 1.52.1
- OS: win10

vdegenne:
I have the exact same bug with an even more simple snippet for testing :
```
${1:foo}${2:bar}
```

In other words the bug is not related to "linked placeholders" as you thought.

**version:** 1.19.0

*note:* Have you find a workaround @fingershi ?

vdegenne:
I think one way to fix that would be to add a command in this file, but I am not sure how things work: 

https://github.com/VSCodeVim/Vim/blob/5a2dc0210df86cff80f1c6be559316d50e40191c/src/actions/commands/insert.ts

Maybe add a command on `<Tab>` in `Mode.Insert` and it should check if the editor is not editing snippet, and if it is it should prevent to switch in `Visual Mode`, is it possible ?

@J-Fields @Chillee @jbaiter @jpoon @xconverge @fatanugraha @cvaldev @shawnaxsom @lyup @ldm0 @kylecarbs

J-Fields:
We're not sure how to handle `<Tab>` in insert mode currently, see #5131

PR #9239, opened 2024-09-03
Fix going into visual mode when filling out snippets with adjacent placeholders etc

### LINKED ISSUE #9236: Vim strangely turns into normal mode for adjacent snippet placeholders
After creating a snippet with several placeholders, my expected behavior is that, vim keeps in the insert mode, and the user inputs some letters in the first placeholder, and then presses the `TAB` key to go to the second to input, until all placeholders are filled. During this process, vim is always in the insert mode unless the user presses the `ESC` key. This holds for most cases so I think this should be the expected behavior.

However, if two placeholders in a snippet are directly adjacent, i.e. there's no other characters between them, when the user finishes filling the first placeholder, and presses `TAB` to go to the second, vim will unexpectedly go to the visual mode, selecting the second placeholder. If the second placeholder doesn't have another placeholder nested in it, pressing a `c` will deal with the case; otherwise, more key presses are necessary.

The attached MP4 video file shows the expected and unexpected behaviors mentioned above.

https://github.com/user-attachments/assets/8eaabd62-d968-4800-8a62-eae698107bfd

semicube:

Fixes going into visual mode on certain placeholders when you fill out vscode user snippets. Particularly this occurs when the placeholders are placed adjacent to each other (not seperated by a character, see Issue #9236), or when the snippet starts from a placeholder (see Issues #5969, #7068).

Note on the fix:

I noticed that in each of these placeholder cases where the cursor goes into visual mode, `isSnippetSelectionChange()` inside `ModeHandler.handleSelectionChange()`  returns `false` even though it is a snippet selection. 

I modified the `isSnippetSelectionChange()` to only consider comparing the previous cursors that represent a selection (`c.start` and `c.stop` are not equal) with `e.selections`. 

Fixes #9236, #7068, #6465, #5969 

### DIFF
--- a/src/mode/modeHandler.ts
+++ b/src/mode/modeHandler.ts
@@ -214,7 +214,11 @@ export class ModeHandler implements vscode.Disposable, IModeHandler {
     // must have been inserted.
     const isSnippetSelectionChange = () => {
       return e.selections.every((s) => {
-        return this.vimState.cursors.every((c) => !s.contains(new vscode.Range(c.start, c.stop)));
+        return this.vimState.cursors
+          .filter((c) => !c.start.isEqual(c.stop))
+          .every((c) => {
+            return !s.contains(new vscode.Range(c.start, c.stop));
+          });
       });
     };
 


HenryTSZ:
add test case for this

https://github.com/VSCodeVim/Vim/blob/2cb95520f10f0b7fe439f4f3d97a9c7f214b3416/test/mode/modeInsert.test.ts#L712-L724

append after this

```ts
    test('handles snippet when filling out snippets with adjacent placeholder', async () => {
      await modeHandler.handleKeyEvent('i');
      await vscode.commands.executeCommand('editor.action.insertSnippet', {
        snippet: '${3:foo} ${1:bar}${2:baz}',
      });
      await modeHandler.handleMultipleKeyEvents(['o', 'n', 'e']);
      await vscode.commands.executeCommand('jumpToNextSnippetPlaceholder');
      assertEqualLines(['foo onebaz']);
      assert.strictEqual(modeHandler.currentMode, Mode.Insert);
    });

    test('handles snippet with multiple cursors and start with a placeholder', async () => {
      await modeHandler.handleKeyEvent('i');
      await vscode.commands.executeCommand('editor.action.insertSnippet', {
        snippet: '${1:bar} ${2:baz} $1',
      });
      assertEqualLines(['bar baz bar']);
      assert.strictEqual(modeHandler.currentMode, Mode.Insert);
    });
```

Esonhugh:
two years bro. Any progress ?

PR #9242, opened 2024-09-04
Name VSCodium

buhtz:
IMHO the applications real name "VSCodium" should be named and not Microsofts re-branding "Visual Studio Code".


J-Fields:
Sorry, I don't agree

Issue #9195, opened 2024-08-10
Jumping to Previous Position After Pressing a Key


Bear-Frost:
**Describe the bug**
When searching for a symbol in VS Code, after finding the symbol and navigating to it, pressing any key on the keyboard causes the cursor to jump back to its previous position.

**To Reproduce**

1.  Toggle the search symbol in VS code ( Ctrl + Shift + o ) or (Ctrl + p ) and then type "@"
2.  start searching for a symbol that you want to go to
3.  click the symbol or press enter
4.  start pressing any key

**Expected behavior**
After navigating to the symbol I want to go to, the cursor should not jump back to its previous position after pressing any key.

**Environment**

- Extension (VsCodeVim) version: v1.27.3
- VSCode version: 1.92.1
- OS: Windows 10 ( Windows_NT x64 10.0.19045)

**Additional context**
This issue is not limited to searching for a symbol. If I also search for a word by pressing (Ctrl + F) and then navigate to that word, pressing any key causes the cursor to jump back to its previous position. This bug does not always occur, sometimes it doesn’t happen, and I’m still unsure what triggers it.

xiangfu1003:
I have the same issue, 
performance varies in different files, some files must appear, while others do not have this issue。
hope this problem can be resolved soon

zxch3n:
`gd` is used to go to the definition in my config. After press `i` to enter the insert mode, it jumps back to the previous position.

https://github.com/user-attachments/assets/3deba5bf-2d0a-4698-8832-52f528837fa3

sguillia:
This bug exists for more than a year and was reported here: https://github.com/microsoft/vscode/issues/177598

But it was closed after being identified as related to VSCodeVim.

If there are any pointers or guidance on how to address this, please let me know—I'm happy to contribute if I'm able.

This bug annoys me every single day.

qazxcdswe123:
This is because `<C-o>` uses custom jump list and track jump(goto) command that is issued by vim only.
To fix this simply add this to settings.json

```json
    {
      "before": ["<C-o>"],
      "commands": [
        "workbench.action.navigateBack"
      ]
    },
```

wolf0403:
Hi @qazxcdswe123 I'm slightly confused about where <C-o> some to play. 
I'm having this issue with "Go to Definition" (F12) followed by vim cursor moving command (h/j/k/l)

[...]

PR #9286, opened 2024-09-30
feat: use vscode native nav back to avoid surpurise and inconsistency

qazxcdswe123:
Notice: This is more or less a BREAKING CHANGE!

- Currently `<C-o>` use hand rolled track list to go back, and it breaks after mouse navigation (eg. `Cmd+Click`). This surprises lots of people, like https://github.com/microsoft/vscode/issues/177598
- This PR changed the default behaviour of `<C-o>` to let it use `workbench.action.navigateBack`

Changes:
- `<C-o>` now use workbench.action.navigateBack instead of hand rolled jump list, so that `gd` and mouse goto state are consistent.
- `<C-t>` is unchanged and still use hand rolled jump list.

- close: https://github.com/VSCodeVim/Vim/issues/9195

This will be a breaking change but IMO it's worth it since `workbench.action.navigateBack` is used much more often then Vim's jump list

### DIFF
--- a/src/actions/commands/actions.ts
+++ b/src/actions/commands/actions.ts
@@ -1286,9 +1286,9 @@ export class CommandInsertNewLineBefore extends BaseCommand {
 }
 
 @RegisterAction
-class CommandNavigateBack extends BaseCommand {
+class CommandJumpBack extends BaseCommand {
   modes = [Mode.Normal];
-  keys = [['<C-o>'], ['<C-t>']];
+  keys = [['<C-t>']];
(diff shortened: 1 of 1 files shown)


J-Fields:
I don't like the idea of breaking the jump list, especially for behavior that can be achieved by configuration.

Maybe it's possible to execute `workbench.action.navigateBack`, then check the cursor position against what `JumpTracker` would've done... if it's different, create a `Jump` on the fly?

PR #9301, opened 2024-10-09
Simple :grep using vscode search

jlahtinen:
image
image

I haven't yet polished this PR because I am not sure is this feature wanted or is this :grep correct command to be used like that.

It implements simple :grep

e.g. :grep foo opens vscode native search panel and runs search

I think there is no issue yet?

Is this kind of :grep what we/you need or should this be :vscodegrep or similar? Should we/you add a configuration parameter what user expects from :grep?

How about grep's argument parser? Should it be implemented differently?


J-Fields:
Thanks for the PR; sorry for not getting to it sooner!

Superseded by https://github.com/VSCodeVim/Vim/pull/9301

jlahtinen:
No problem.

> Superseded by #9301

Seems to be a typo?

Issue #651, opened 2016-08-25
bug: press % for matching does not work or goes to wrong locations


Viperform:
Please *thumbs-up* 👍 this issue if it personally affects you! You can do this by clicking on the emoji-face on the top right of this post. Issues with more thumbs-up will be prioritized.

-----

### What did you do?

Open this file https://github.com/telerik/kendo-ui-core/blob/master/src/kendo.color.js  in visual studio code and switch to NORMAL MODE, go to last line which begins with "}, typeof define == 'function' && define.amd", move cursor to col 1, that's should be under '}', now press %.

### What did you expect to happen?
The cursor should go to '{' in line 3.

### What happened instead?
Nothing happens.
Many times, press % will cause the cursor go to wrong places.

### Technical details:

* VSCode Version: 1.3 /1.4
* VsCodeVim Version: *[please ensure you are on the latest]* 0.1.7
* OS: win7, 64bit

johnfn:
Yeah, this is a known issue unfortunately, We don't correctly handle cases like `( "(" )`. Feel free to add it if you want 😉

xconverge:
@johnfn This seems to be fixed?

jpoon:
Indeed, this looks to be working on 0.2.0.

johnfn:
That's super weird, because our matching algorithm has never changed!

https://github.com/VSCodeVim/Vim/blob/master/src/matching/matcher.ts#L21

Something like `( ")" )` continues to trip us up, so I'm gonna keep this one open for now. 😉

Chillee:
Is there any reason we don't just use`editor.action.jumpToBracket`?

Chillee:
Oh, here's one... It doesn't support multiple cursors >:(

johnfn:
Ah multi cursors, the cause of, and solution to, all of VSCodeVim's problems.

antonburger-advanced:
Also affects motions like <kbd>[</kbd><kbd>{</kbd> and <kbd>]</kbd><kbd>}</kbd>. Would guess the same code is involved in finding pairs here?

bijoutrouvaille:
I don't use multiple cursors, so I made this mapping and it has worked so far:

```
{
      "before": ["%"],
      "commands": [
        {
          "command": "editor.action.jumpToBracket",
          "args": []
        },
        {
          "command": "editor.action.moveCarretLeftAction",
          "args": []
        },
        {
          "command": "editor.action.moveCarretRightAction",
[...]

PR #9378, opened 2024-11-27
[Draft] Improve % motion by using vscode editor.jumpToBracket

tim-tim707:
Using % motion on brackets doesn't work properly if there are unmatched brackets in comments 
```js
function test() {
    // if {
    return 1;
}
```

The % motion with `editor.jumpToBracket` was also discussed in (#651)[https://github.com/VSCodeVim/Vim/issues/651]. The motion didn't work properly with multicursor before, and this PR doesn't address this particular issue.
I also remember an issue talking specifically about the unmatched bracket but I cannot find it.

This PR calls `editor.jumpToBracket` in the case of bracket matching without selection. I tried to make it work with selection, but the command doesn't move the cursor and clears the selection if it exists. Trying to call `jumpToBracket` twice in a row when there is a selection, and saving + trying to restore the selection afterward creates a conflict between vscode cursor and vim cursor where they are not at the same location.

I would like to make it work with selection in the future but I don't really understand how to fix this yet.

Maybe I should call the editor commands elsewhere than PairMatcher like in the '%' operator somehow ?

### DIFF
--- a/src/actions/motion.ts
+++ b/src/actions/motion.ts
@@ -1798,12 +1798,12 @@ class MoveToMatchingBracket extends BaseMovement {
       if (pairing) {
         // We found an opening char, now move to the matching closing char
         return (
-          PairMatcher.nextPairedChar(
+          (await PairMatcher.nextPairedChar(
             new Position(position.line, col),
             lineText[col],
             vimState,
             false,
-          ) || failure
+          )) || failure
         );
       }
     }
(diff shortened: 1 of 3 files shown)


tim-tim707:
Doesn't work properly with Visual Block. Selection is empty in this case, vscode cursor moves, but vim one doesn't

PR #9430, opened 2025-01-11
Fix normal command double quote bug

s-kai273:
- [x] Fix normal command double quote bug

https://github.com/VSCodeVim/Vim/issues/9431


src/vimscript/parserUtils.ts
```diff
@@ -113,7 +113,7 @@ const escapedParser = string('\\')
 export const keystrokesExpressionParser: Parser<string[]> = alt(
   escapedParser,
   specialCharacterParser,
-  noneOf('"'),
+  regexp(/./),
```
s-kai273, line 116: Due to the use of `noneOf('"')`, parsing parameters that contain double quotes fails. 
However, I believe that including double quotes is not an invalid case in normal command.

If replacing the original parser with `regexp(/./)` introduces other issues, please let me know!

J-Fields, line 116: This should work, right?

```suggestion
  regexp(any),
```

s-kai273: Yes, that should work!
The main difference is that `any` can parse raw special characters like `\n`, whereas `/./` cannot.
However, in this plugin, the input keystrokes are not expected to include raw special characters such as `\t` or `\n`. These should always be escaped as `\\t` or `\\n`, in which case they are handled by escapedParser.

So as long as the plugin is working as intended and only receives properly escaped input, keystrokesExpressionParser should work correctly with either option.
However, if raw special characters are accidentally included, the behavior would differ depending on which parser is used.

Personally, I think using `/./` is safer, since if any raw special character slips in, it simply won't be parsed or included in the result, making it easier to notice unexpected input.

J-Fields: Got it, thanks!

J-Fields (approved):
Thanks!

PR #9441, opened 2025-01-15
Update README.md: Fix Quick Example Config

carloluis:
## Fix Quick Example

> Make JSON Configuration valid

Correct the JSON format in the README example so that it can be copied & pasted as a valid initial config. for the plugin.

https://github.com/VSCodeVim/Vim/blob/02cf1f3b2c1df1bb53996064c7ad65315b2e8884/README.md?plain=1#L118-L123


J-Fields (approved):
Thanks!

PR #9483, opened 2025-02-13
Add MarkCommand to handle mark creation in command line

arunchaganty:
**What this PR does / why we need it**: This PR implements the `mark` Ex command which allows users to set marks. While the `m` key also achieves this goal, implementing it as an ex command (1) is a missing feature for which PRs were invited and (2) allows other VSCode extensions to create marks by calling `vim.remap` with the appropriate `commands` argument. I have a separate use-case where I'd like to make sure a position is added to the jumplist (i.e. do `:mark '`).

**Which issue(s) this PR fixes**: N/A

**Special notes for your reviewer**: This is my first PR to the vscode vim extension, so I'd appreciate any input on how to contribute better. I'm a long time user and fan, so super excited to be submitting a PR!

Also, in full disclosure, this PR was partially written with AI using AugmentCode.


src/cmd_line/commands/marks.ts
```diff
@@ -133,3 +133,22 @@ export class DeleteMarksCommand extends ExCommand {
...
+
+  private markName: string;
+  constructor(markName: string) {
+    super();
+    this.markName = markName;
+  }
+
+  async execute(vimState: VimState): Promise<void> {
+    const position = vimState.cursorStopPosition;
+    vimState.historyTracker.addMark(vimState.document, position, this.markName);
```
J-Fields, line 152: Thanks for the PR!

From `:help :mark`:
```
:[range]ma[rk] {a-zA-Z'}
			Set mark {a-zA-Z'} at last line number in [range],
			column 0.  Default is cursor line.
```

You can make use of the range by implementing `executeWithRange` (yes - it's dumb that there are two different methods).

J-Fields (changes requested):

src/cmd_line/commands/marks.ts
arunchaganty: Alright, good point and done below!

J-Fields (approved):

PR #9511, opened 2025-03-03
Fix Python async function motions with `async def`

### LINKED ISSUE #9510: `async def` breaks Python function motions
**Describe the bug**
`async def` is not recognized by Python function motions

**To Reproduce**
Use `[m` or `]m` in a Python language file with `async def` functions

**Expected behavior**
`[m` should recognize `async def`

- Extension (VsCodeVim) version: 1.29.0
- VSCode version: unknown -- using cursor 0.45.10
- OS: macos

**Additional context**
is replicable in vscode too. i already wrote PR. this is just paperwork

nathan-gage:
fixes Python function motions `[m` and `]m` to recognize `async def`

fixes #9510

**Additional notes**
i opted to keep old tests and introduce new.

in this process, i found a bug related to function boundaries, captured in "valid findPrevFunctionStart, from inner regular to outer async" and "valid findPrevFunctionStart, from inner async to inner regular".

The bug was that when searching backwards for function boundaries, the original code would include the current function in the results because isBefore() returns true for equal positions, causing the motion to incorrectly stay at the current function instead of moving to the previous one. the original tests didn't have any test cases for moving backwards from within a function definition line itself.


J-Fields:
Thanks!

PR #9521, opened 2025-03-11
fix: Copy to clipboard on yank

asynchroza:
As per the current moment, yank commands do not copy content to external clipboard.
This PR introduces functionality for transferring yanked text to external clipboard.


asynchroza:
This is actually redundant and would enforce the copying behaviour. Apparently, there's an option to explicitly enable this behaviour.

PR #9525, opened 2025-03-17
Add `$` as a surround character (#8895)

### LINKED ISSUE #8895: Surround Input mode (`ys` in Normal or `S` in Visual) doesn't recognize dollar sign
## Description

When trying to surround a piece of text in dollar signs (such as starting a math environment for a markdown file) the dollar sign isn't recognized and instead snaps me back into Normal mode. The Surround plugin is enabled, and this works for other characters (`"`, `'`, `()`, etc. This is supposed to work in Vim (as suggested by StackOverflow posts)

## To Reproduce

Option A (normal mode)

* 1.  Enter an editor in normal mode
* 2.  Type `ys[motion]$` with any motion (for example, `ysiw$`)
* 3.  Watch as nothing happens (except maybe the cursor moving)

Option B (visual mode)

* 1. Enter an editor in visual mode (using `v`, `V` or `<C-v>` from normal mode)
* 2. Make a selection (for example `iw`)
* 3. Type `S` to enter Surround Input mode
* 4. Type `$`
* 5. Watch as you are put in normal mode with no dollar signs anywhere you expected

## Expected behavior
The dollar signs should appear around the selected text thanks to the Surround plugin, as it happens with other characters.

## Environment

- VsCodeVim version: 1.27.2
- VSCode version: 1.86.1
- OS: Windows 11 Home x64 23H2, build 22631.3155

cjohnson19:

This PR adds `$` as a "quote"-like character to support surround and bracket-like actions (e.g. `yi$`, `ysi$}`). These actions are supported by vim surround, so users expect this to line up when interacting  with VSCodeVim.

Fixes #8895 

This creates a weird mismatch between the names for `Quote` behavior, since it now supports a `$`. The naming for "quotes" makes sense given things like `diq`, but I think in essense the quote matching really lines up with a "directionless" behavior.

### DIFF
--- a/src/actions/motion.ts
+++ b/src/actions/motion.ts
@@ -2090,7 +2090,7 @@ export class MoveAroundSquareBracket extends MoveInsideCharacter {
 export abstract class MoveQuoteMatch extends BaseMovement {
   override modes = [Mode.Normal, Mode.Visual, Mode.VisualBlock];
   protected readonly anyQuote: boolean = false;
-  protected abstract readonly charToMatch: '"' | "'" | '`';
+  protected abstract readonly charToMatch: '"' | "'" | '`' | '$';
   protected includeQuotes = false;
   override isJump = true;
   readonly which: WhichQuotes = 'current';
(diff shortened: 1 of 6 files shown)


J-Fields:
Is `$` a special character treated like a quote, or does this work for any character in Vim surround?

cjohnson19:
That's a great question, and one I didn't consider previously. Looks like `vim-surround` will treat any non-alphabetic character as a valid surrounding target. From [the documentation](https://github.com/tpope/vim-surround/blob/3d188ed2113431cf8dac77be61b842acb64433d9/doc/surround.txt#L118):

> Undefined replacement characters (with the exception of alphabetic characters) default to placing themselves at the beginning and end of the destination, which can be useful for characters like / and |.

cjohnson19:
Sorry to revive this PR.

Would we be interested in merging a version which extends the surround behavior to work for all undefined and non-alphabetic characters as opposed to just $? If so, I would be happy to work on that.

PR #9548, opened 2025-03-29
Fix global marks jumping to the wrong position (#8662)

### LINKED ISSUE #8662: Jump To Mark With Single Quote Broken Since Latest Update (1.26.1)
**Describe the bug**

I noticed that the extension's jump-to-mark functionality has been broken since the latest version upgrade (v1.26.1). TLDR, you can only jump to a mark via backtick, but no longer single-quote, which seems to execute the mark character's raw command instead.

**To Reproduce**

1.  Set a mark, for example, `ma`
2. Go to another line
3. Press `'a` to jump to that mark --> **No jump occurs, instead this is treated like a raw Vim command** (for example, `a` for append)
4. Press `` `a `` to try again jumping to that mark --> **Jump successful**

- Extension (VsCodeVim) version: 1.26.1
- VSCode version: 1.83.1 (also reproduced on a previous version)
- OS: Mac OS 14.0 (also reproduced via a remote IDE session on a Linux host)

**Additional context**

I don't have any extension config, so there shouldn't be any remapping of the single-quote character (I am aware that the most recent extension patch included a vix to re-mapping).

This functionality was working just fine before the latest extension update.

NeedsSoySauce:

Fixes an issue with global marks not placing the cursor in the correct position when the mark is not in the currently active document

Fixes #8662


src/history/historyTracker.ts
```diff
@@ -96,13 +96,22 @@ class DocumentChange {
...
-  document?: vscode.TextDocument; // only required when using global marks (isUppercaseMark is true)
 }
 
+export interface ILocalMark extends IMarkBase {
+  isUppercaseMark: false;
+}
+
+export interface IGlobalMark extends IMarkBase {
+  isUppercaseMark: true;
+  document: vscode.TextDocument; // Required for global marks
```
J-Fields, line 110:
```suggestion
  document: vscode.TextDocument;
```

src/history/historyTracker.ts
```diff
@@ -96,13 +96,22 @@ class DocumentChange {
   }
 }
 
-export interface IMark {
+export interface IMarkBase {
```
J-Fields, line 99:
```suggestion
interface IMarkBase {
```

J-Fields (approved):
Very nice PR; thanks a lot!

PR #9584, opened 2025-04-20
Fix: keep the selection of the space after the last character

kajikentaro:

After executing o twice, keep the selection of the space after the last character

AS-IS
as-is

TO-BE
to-be

not yet


J-Fields (approved):
Thanks!

PR #9595, opened 2025-04-27
fix: allow to disable ctrl+<up|down|left|right> by vim.handleKeys

maksimr:
Now these mappings prevent to use macOS default tiling shortcuts


maksimr:
I can remove it by "Preference: Open Keyboard Shortcuts"

PR #9615, opened 2025-05-12
Update README.md

cjromb:
Changed System Preferences to System Settings, which is what it's called on a Mac.

**What this PR does / why we need it**: It corrects where a setting is located on a mac so others can find it.

**Which issue(s) this PR fixes**:  https://github.com/VSCodeVim/Vim/issues/9616

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,7 @@ defaults write com.microsoft.VSCodeExploration ApplePressAndHoldEnabled -bool fa
 defaults delete -g ApplePressAndHoldEnabled                                           # If necessary, reset global default
 ```
 
-We also recommend increasing Key Repeat and Delay Until Repeat settings in _System Preferences -> Keyboard_.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -67,7 +67,7 @@ defaults write com.microsoft.VSCodeExploration ApplePressAndHoldEnabled -bool fa
 defaults delete -g ApplePressAndHoldEnabled                                           # If necessary, reset global default
 ```
 
-We also recommend increasing Key Repeat and Delay Until Repeat settings in _System Preferences -> Keyboard_.
+We also recommend increasing Key Repeat and Delay Until Repeat settings in _System Settings -> Keyboard_.
```
rebornix, line 70: to support OSes before Ventura, we might want to mention both: _System Settings/Preference -> Keyboard_.

rebornix (approved):

Issue #5991, opened 2021-01-13
Implementation of vimgrep


TryerGit:
Hello,

I have a project structure thus:

```
/myproject/
    ----src/
        ----file.cpp
    ----include/
        ----file.h
```

I open up `code .` from `/myproject/`

With `file.cpp` open in the editor window, I am trying to run a vimgrep command thus:

`vimgrep search_item **/*`

On hitting enter, no search is returned, no quick fix list seems to be populated. This happens even though `search_item` features quite prominently in my code. Also, running the above command seems to make the file into a changed state (with no visible change to the file, yet the file "changes" / becomes dirty) requiring me to either save it explicitly or else undo multiple times until it gets back to the original state.

On the bottom status line, I obtain

"E37: No write since last change (add ! to override)"

Instead of VSCode, if I open up `vim` in the same folder, and run the same command, the quick fix list is populated and I am able to traverse this using `:cnext`, `:cprev`, etc.

I searched this forum for previous issues related to `vimgrep` but I could not find any. Hence this OP.

Is there something I am doing wrong? The above is after enabling `NeoVim`.

Thanks.

sql-koala:
> Is there something I am doing wrong?

Not really. But vimgrep is not implemented and doing so would be  a lot of "reinventing the wheel". Because vscode has a reasonably powerful project-wide search feature.
My personal advice is to learn and use vscodes ide features and add vim modal editing on top.

J-Fields:
Re-opening because this is a reasonable request, though I agree VSCode's implementation is probably the way to go

PR #9630, opened 2025-05-17
Implements vimgrep, #5991

AzimovParviz:
Adds a basic implementation of vimgrep, which doesn't reinvent the wheel and uses native VSCode search function
https://github.com/VSCodeVim/Vim/issues/5991

I still haven't thought of how to implement the E480 error for "no match".  
Also flags like `g` and `j` are still missing. while `j` is implementable, I am not sure how to limit VS Code's built-in search results by a number


AzimovParviz:
https://github.com/user-attachments/assets/0d8bc072-fc1d-4421-8213-c87d567a9296

this is how it looks like as of now

AzimovParviz:
@J-Fields I apologize for pinging. What is the bare minimum that would be expected from a vimgrep implementation? 

And an off-topic question, does the VSCodeVim Slack channel still exist? I would love to ask more questions from contributors and maybe get some advice as well. 

Thank you in advance

J-Fields:
> What is the bare minimum that would be expected from a vimgrep implementation?

I think what you've got here is merge-able, in terms of features.

> And an off-topic question, does the VSCodeVim Slack channel still exist? I would love to ask more questions from contributors and maybe get some advice as well.

Not in any meaningful way unfortunately. At the moment I'm the only active maintainer (and that's being very loose with the term "active"). Are there are any particular questions you have?

src/vimscript/exCommandParser.ts
```diff
@@ -577,7 +578,10 @@ export const builtinExCommands: ReadonlyArray<[[string, string], ArgParser | und
   [['vert', 'ical'], undefined],
   [['vi', 'sual'], undefined],
   [['vie', 'w'], undefined],
-  [['vim', 'grep'], undefined],
+  [['vim', 'grep'], GrepCommand.argParser],
+  [['vimg', 'rep'], GrepCommand.argParser],
+  [['vimgrep', ''], GrepCommand.argParser],
```
J-Fields, line 583:
```suggestion
  [['vimg', 'rep'], GrepCommand.argParser],
  [['vimgrep', ''], GrepCommand.argParser],
```

test/cmd_line/grep.test.ts
```diff
@@ -0,0 +1,84 @@
...
+    const pattern = Pattern.parser({ direction: SearchDirection.Backward, delimiter: '/' });
+    const command = grep(pattern.tryParse('t*st'), ['Untitled-1']);
+    await command.execute();
+    await vscode.commands.executeCommand('search.action.focusNextSearchResult');
+    // Assert that the active editor is Untitled-1
+    const activeEditor = vscode.window.activeTextEditor;
+    assert.ok(activeEditor, 'There should be an active editor');
+    assert.strictEqual(
+      activeEditor?.document.fileName.endsWith('Untitled-1'),
+      true,
```
J-Fields, line 74:
```suggestion
    assert.ok(
      activeEditor?.document.fileName.endsWith('Untitled-1'),
```

src/cmd_line/commands/grep.ts
```diff
@@ -0,0 +1,68 @@
+/* eslint-disable @typescript-eslint/no-unsafe-member-access */
+/* eslint-disable @typescript-eslint/no-unsafe-call */
```
J-Fields, line 2: Why are these disabled?

AzimovParviz: Oh thank you for pointing out. I have no idea, I most likely fat fingered these when calling quick actions on some missing imports when I just started writing the file. I'll remove these

AzimovParviz:
>  Are there are any particular questions you have?

I had some questions about writing tests (I was confused on how to write them at all. I was wondering, for example, if it was required to use the test simplifier. I wasn't able to figure out on how to include it in my tests properly) and I was also wondering about some minor things like the command strings in the command parser. 

All of these got resolved now, although it's a bit saddening that the Slack is not active. Thank you for maintaining the repo, it's my favorite Vim mode implementation across different text editors and IDEs

AzimovParviz:
> Can you clarify what you mean?

I was a little confused on why the commands were parsed like 'gr' 'ep', 'fu' 'nction' etc. I tried reading the vim manual and the commands didn't look like that in there.

mikaelmedina:
> I was a little confused on why the commands were parsed like 'gr' 'ep', 'fu' 'nction' etc. I tried reading the vim manual and the commands didn't look like that in there.

My guess is that the first part in the tuple is the shorthand alias that works, i.e. 'gr' is unique so we can shorthand it and use `:gr pattern file`, then the second part expands it to what the full command would be. Similar to how in .vimrc you can use `set nu` instead of `set number` as nu is unique already. 

If this is correct the I guess the entry for `[['grep', ''], GrepCommand.argParser],` is strictly unnecessary. (?)

Also this looks like a nice way to get this feature going, nice to embrace the native VScode features sometimes. 👍

J-Fields:
> My guess is that the first part in the tuple is the shorthand alias that works

That's correct - this allows us to specify all the equivalent commands in one line.

> If this is correct the I guess the entry for `[['grep', ''], GrepCommand.argParser],` is strictly unnecessary. (?)

Yep - I must've missed that in my first read.

src/vimscript/exCommandParser.ts
```diff
@@ -577,7 +578,8 @@ export const builtinExCommands: ReadonlyArray<[[string, string], ArgParser | und
   [['vert', 'ical'], undefined],
   [['vi', 'sual'], undefined],
   [['vie', 'w'], undefined],
-  [['vim', 'grep'], undefined],
+  [['vim', 'grep'], GrepCommand.argParser],
+  [['grep', ''], GrepCommand.argParser],
```
J-Fields, line 582:
```suggestion
```

test/cmd_line/grep.test.ts
```diff
@@ -0,0 +1,63 @@
...
+      fileExtension: '.txt',
+      contents: 'test, pattern nnnn, t*st, ttst',
+    });
+    // second file without a match
+    let file2 = await createFile({
+      fileExtension: '.txt',
+      contents: 'no pattern match here ',
+    });
+    // We open the second file where we know there is no match
+    const document1 = await vscode.workspace.openTextDocument(vscode.Uri.file(file1));
```
AzimovParviz, line 34: Not sure if there's a better way to open the files. I just copied the way you open them in the testUtils in the setupWorkspace function (sorry, not sure how to link to functions on GitHub)

Issue #8903, opened 2024-02-26
Vim cursor does not move with "Go to Next Problem" - Reopening: #5746


ls1xt:
Hi, I want to re-open a bug report: Vim cursor does not move with "Go to Next Problem"  #5746 

The problem is that after "Go to Next Problem" the vim cursor is not updated accordingly. Thus any hjkl movement will take place at the old cursor position.

This was fixed several years ago but the problem seems to come back consistently.

Would be great to have this resolved!

lucaSartore:
A possible, even not ideal solution, could be found here: https://github.com/VSCodeVim/Vim/issues/5746#issuecomment-2205307612

erquhart:
Yep, definitely a regression. Pretty high on the frustration scale.

erquhart:
Current workaround: This pops up every hour or so, so I hit `Developer: Restart Extension Host` when it does to fix. Makes it livable, although still pretty frustrating, until a fix is released.

erquhart:
I switched from Copilot to Supermaven and this stopped happening. Not certain that was what did it, but that's the only big change I'm aware of. Curious if anyone else would try turning off Copilot or using something else for a bit to see if it fixes it for them as well.

loc:
Confirming that disabling the Copilot extension fixed it for me. Also fixes a similar symptom where `Ctrl-f` search would drop the cursor back to where it started after `<Esc>`ing out of the search.

AndruC:
> ...I hit `Developer: Restart Extension Host` in the command palette when it does to fix...

Confirmed this resolved the issue for me.

erquhart:
Feels like a memory leak, keeps getting worse until you restart the server.

alythobani:
> Current workaround: This pops up every hour or so, so I hit Developer: Restart Extension Host in the command palette when it does to fix. Makes it livable, although still pretty frustrating, until a fix is released.

Personally I just toggle Vim off and on (run `Vim: Toggle Vim Mode` twice, keybound to `ctrl+cmd+V`). I find that a little smoother than restarting the whole extension host (which restarts all your extensions)

And yeah I experience this issue at least a few times a day these days

Issue #5746, opened 2020-11-24
Vim cursor does not move with "Go to Next Problem"


nzig:
**Describe the bug**
When using the keyboard shortcut for "Go to Next Problem", the focus and cursor move to the next problem in the file. However, using any vim action starts from where the cursor was last, before the keyboard action.

**To Reproduce**

1.  Create a file with linting enabled (I used Python with Pylance language server).
1. Make a syntax error somewhere down the file
1. Position cursor in beginning of file
1. Trigger "Go to Next Problem" keyboard action (by default `F8` on my machine).
1. press `l`.
1. The cursor will move to one character after it was before the keyboard action.

**Expected behavior**
The cursor should be one character after the problem that "Go to Next Problem" jumped to.

**Screenshots**
None

**Environment (please complete the following information):**

- Extension (VsCodeVim) version: 1.17.1
- VSCode version: 1.51.1
- OS: Ubuntu 20.04

**Additional context**
None

As an aside, it would also be nice to have a vim action for "next problem".

JW9506:
Hi, check out the latest commits and build it manually, this problem has been fixed a while ago via[#5250]

berknam:
Closing since this should be fixed on the new version "1.18" out now. If you still experience this feel free to reopen.

jeffomatic:
I'm using v1.18.9 of the extension (VSC Insiders, commit `93f705ab`) and I still observe this problem. The cursor visually appears to move when calling "Go to next problem", but performing a motion (`hjkl`) updates the _previous_ location of the cursor.

A similar thing occurs if you choose a problem in the "Problems" tab of the lower panel.

i-am-the-slime:
This regressed in 1.18.9. I just tried with 1.18.8 and it works there.

jeffomatic:
Confirmed that installing 1.18.8 fixes.

@berknam Would it be possible to re-open this issue?

dmbfm:
This issue is still happening in the latest release...

JW9506:
> This issue is still happening in the latest release...

Closing and reopening can clear the issue, I can confirm that the issue does actually surface from time to time after a long usage. I don't know exactly how to reproduce the problem at my will though.  And plus it is not a huge deal because closing and reopening do not take much time.

dmbfm:
[...]

PR #9644, opened 2025-05-22
fix: Clean up stale `ourSelections` to not indefinitely ignore external selections

alythobani:
## What this PR does / why we need it

This PR fixes the issue of external selections changes (e.g. from commands like `Go to Next/Previous Change`) often not being properly tracked in VSCodeVim's internal selections state.

For instance, here's a screen recording where running `Go to Previous Change` and then pressing `Escape` takes the cursor back to where it was, as if `Go to Previous Change` was never run: 

https://github.com/user-attachments/assets/6c420fda-633a-43e0-a470-d0f39b523078

Fixes: #8903 (assuming this is indeed the same root issue)
Fixes: #5746 (although it's currently marked as closed)

## Special notes for your reviewer

I've been encountering this issue pretty regularly for a good while. It would happen at varying frequencies for me (sometimes every couple days, sometimes many times a day), so I figured it'd be a pretty obscure problem to fix (e.g. a memory leak issue like one commenter suggested). I dug into it today though and think I've found the cause and a reasonable fix (although not perfect).

I first turned on VSCodeVim's debug logs (helpfully [described](https://github.com/VSCodeVim/Vim/?tab=readme-ov-file#debugging-remappings) in the README, thanks) and waited for the issue to crop up again. When it did, I saw some useful warning logs:

Screen Shot 2025-05-21 at 9 37 02 AM

Then I toggled vim off/on as usual for a workaround (`Vim: Toggle Vim Mode`) and contrasted those with normal behavior logs:

Screen Shot 2025-05-21 at 9 37 28 AM

I eventually found a reliable repro that can be done in seconds instead of waiting hours or days:

### Repro steps

1. Have a keymapping like `J` mapped to `5+`
2. Hold down `J` for a few seconds until it's run at least about 5-7 times
3. Try `Go to Next/Previous Change`
4. Press `Escape` and observe the cursor jumps back to before step 3

### Root cause

There are a few complicated things going on under the hood, so I'll try to be detailed for the sake of anyone reviewing and future maintainers.

After digging into the relevant code and past PRs (#5015, #5250), and lots of debugging, here's what I concluded:

The [`ModeHandler.selectionsChanged.ourSelections`](https://github.com/VSCodeVim/Vim/blob/fa5461b056dff9dedad2502141d2ffdfc275cd41/src/mode/modeHandler.ts#L105) array of hashed selections is used to track 'internal' selection changes (performed by VSCodeVim itself), so that when a [`vscode.TextEditorSelectionChangeEvent` fires](https://github.com/VSCodeVim/Vim/blob/fa5461b056dff9dedad2502141d2ffdfc275cd41/extensionBase.ts#L264-L267) due to an internal selection update, we can ignore it as already accounted for. We only want to update internal selection state because of externally-triggered SelectionChange events (e.g. by mouse clicks or builtin / extension commands).

Since **VSCode only triggers SelectionChange events if the selection actually changes in value**, we determine whether or not to add a given internal selection update to `ourSelections` by comparing it to the current `editor.selections` value.

However, VSCode unreliably fires those SelectionChange events. In this case, after key-repeating something like `J` (mapped to `5+`) around 5-7 times, `editor.selections` can lag behind `updateView`. Worse yet, after we [set `editor.selections`](https://github.com/VSCodeVim/Vim/blob/fa5461b056dff9dedad2502141d2ffdfc275cd41/src/mode/modeHandler.ts#L1473), VSCode seems to quickly reset `editor.selections` to its previous value before properly registering the change and firing a SelectionChange event for it. **By the time we calculate `willTriggerChange` again, `editor.selections` can have lagged enough to still have the old value, and we can add `selectionsHash` to `ourSelections` even if this selection update won't actually trigger a VSCode SelectionChange event**.

For example, with cursor starting on line 10 and `J` mapped to `5+`:
1. Key repeat `JJ` (usually takes 5-7 to repro but we'll assume 2 for simplicity)
2. `handleKeyAsAnAction('5')` => `updateView` =>  set `editor.selections` to line 10 => `editor.selections` was already line 10 => `willTriggerChange = false` and we do nothing else
3. `handleKeyAsAnAction('+')` => `updateView` => change `editor.selections` from line 10 to line 15 => `willTriggerChange = true`, push line 15's hash  to `ourSelections`
4. Under the hood, VSCode seems to reset `editor.selections` to 10
5. `handleKeyAsAnAction('5')` => `updateView` => we'll set `editor.selections` to line 15 => even though we already previously set `editor.selections` to 15, when calculating `willTriggerChange` the value of `editor.selections` is 10 => `willTriggerChange = true` and we push line 15's hash to `ourSelections` *again*
6. `vscode.TextEditorSelectionChangeEvent(15)` fires for step 3 => we pop off only one of the two hashes for line 15 in `ourSelections`, and the other remains
7. `handleKeyAsAnAction('+')` => `updateView` => set `editor.selections` to line 15 => `editor.selections` finally reads as 15, so `willTriggerChange = false` and we do nothing
8. End result: `ourSelections = [hash(15)]`

**Once we have a `selectionsHash` in `ourSelections` that won't ever trigger a SelectionChange event, it'll stay there indefinitely.**

Almost all subsequent SelectionChange events will hit [this block of code](https://github.com/VSCodeVim/Vim/blob/fa5461b056dff9dedad2502141d2ffdfc275cd41/extensionBase.ts#L303-L310) and result in those `Ignoring slipped selection` warnings and being ignored by VSCodeVim's internal selections tracking state. If they're external events, they won't be accounted for at all, leading to the undesired behavior seen in the filed issues.

### Fix

The fix here is quite simple compared to the more complicated explanation above: we timestamp each `ourSelections` entry so that we can remove each one when it gets stale.

- `ourSelections` is now renamed to `ourSelectionsUpdates` and is an array of `{selectionHash, updatedAt}` objects
- In the `TextEditorSelectionChangeEvent` handler, before checking against `ourSelectionUpdates`, we clean out any stale entries older than 1 second (hardcoded constant)
    - We want enough time to receive change events for our selection updates so we don't treat them as external changes, but we also don't want to take an annoyingly long time to clean up stale items when this bug occurs due to key repeats
    - Not sure @berknam is still around, but looks like they were considering a timeout approach of [500ms to 1-2s](https://github.com/VSCodeVim/Vim/pull/5250#issuecomment-700367535)

Detailed docstrings have been added to the logic in the new `src/mode/ourSelectionsUpdates.ts` file to help explain the reasoning behind the cleanup and the various pitfalls and tradeoffs involved with all this stuff.

---

More than happy to hear any alternate approaches. And thanks for making VSCodeVim :) it's been awesome on the whole.

### DIFF
--- a/src/mode/modeHandler.ts
+++ b/src/mode/modeHandler.ts
@@ -11,6 +11,7 @@ import { IBaseAction } from '../actions/types';
 import { Cursor } from '../common/motion/cursor';
 import { configuration } from '../configuration/configuration';
 import { decoration } from '../configuration/decoration';
+import { isLiteralMode, remapKey } from '../configuration/langmap';
 import { Notation } from '../configuration/notation';
 import { Remappers } from '../configuration/remapper';
 import { Jump } from '../jumps/jump';
@@ -59,7 +60,7 @@ import {
   isStatusBarMode,
   isVisualMode,
 } from './mode';
-import { isLiteralMode, remapKey } from '../configuration/langmap';
+import { OurSelectionsUpdate, hashSelections } from './ourSelectionsUpdates';
 
 interface IModeHandlerMap {
   get(editorId: Uri): ModeHandler | undefined;
@@ -89,7 +90,7 @@ export class ModeHandler implements vscode.Disposable, IModeHandler {
    * Used internally to ignore selection changes that were performed by us.
    * 'ignoreIntermediateSelections': set to true when running an action, during this time
    * all selections change events will be ignored.
-   * 'ourSelections': keeps track of our selections that will trigger a selection change event
+   * 'ourSelectionsUpdates': keeps track of our selections that will trigger a selection change event
    * so that we can ignore them.
    */
   public selectionsChanged = {
@@ -102,7 +103,7 @@ export class ModeHandler implements vscode.Disposable, IModeHandler {
      * keeps track of our selections that will trigger a selection change event
      * so that we can ignore them.
      */
-    ourSelections: Array<string>(),
+    ourSelectionsUpdates: Array<OurSelectionsUpdate>(),
   };
 
   /**
@@ -1456,16 +1457,11 @@ export class ModeHandler implements vscode.Disposable, IModeHandler {
         );
 
       if (willTriggerChange) {
-        const selectionsHash = selections.reduce(
-          (hash, s) =>
-            hash +
-            `[${s.anchor.line}, ${s.anchor.character}; ${s.active.line}, ${s.active.character}]`,
-          '',
-        );
-        this.selectionsChanged.ourSelections.push(selectionsHash);
+        const selectionsHash = hashSelections(selections);
+        this.selectionsChanged.ourSelectionsUpdates.push({ selectionsHash, updatedAt: Date.now() });
         Logger.trace(
           `Adding selection change to be ignored! (total: ${
-            this.selectionsChanged.ourSelections.length
+            this.selectionsChanged.ourSelectionsUpdates.length
           }) Hash: ${selectionsHash}, Selections: ${selections[0].anchor.toString()}, ${selections[0].active.toString()}`,
         );
       }
--- a/extensionBase.ts
+++ b/extensionBase.ts
@@ -8,6 +8,7 @@ import { Jump } from './src/jumps/jump';
 import { Mode } from './src/mode/mode';
 import { ModeHandler } from './src/mode/modeHandler';
 import { ModeHandlerMap } from './src/mode/modeHandlerMap';
+import { cleanupOurSelectionsUpdates, hashSelections } from './src/mode/ourSelectionsUpdates';
 import { Register } from './src/register/register';
 import { CompositionState } from './src/state/compositionState';
 import { globalState } from './src/state/globalState';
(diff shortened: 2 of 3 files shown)


src/mode/modeHandler.ts
```diff
@@ -102,7 +103,7 @@ export class ModeHandler implements vscode.Disposable, IModeHandler {
      * keeps track of our selections that will trigger a selection change event
      * so that we can ignore them.
      */
-    ourSelections: Array<string>(),
+    ourSelectionsUpdates: Array<OurSelectionsUpdate>(),
```
alythobani, line 106: I've kept these changes fairly minimal to make them easier to review in the context of the existing code. But we could consider further renaming things for clarity, e.g. `ourSelectionsUpdates` => `selectionsUpdatesToIgnore` and the `selectionsChanged` parent object => `selectionsToIgnoreState`.

Or, if we want to go a bit further with modularizing code / keeping related logic closer together, I'd be happy to refactor the whole `selectionsChanged` object into a `internalSelectionsTracker: InternalSelectionsTracker` class with:

Fields
- `shouldIgnoreIntermediateSelections: boolean`
- `selectionsUpdatesToIgnore: Array<InternalSelectionsUpdate>`

Methods
- `startIgnoringIntermediateSelections(): void`
- `stopIgnoringIntermediateSelections(): void`
- `trackSelectionsUpdateToIgnore(updatedSelections: vscode.Selection[]): void` (would hash and timestamp internally)
- `removeSelectionsUpdateToIgnore(updatedSelectionsHash: string): boolean` (returns `true` if found and removed)
- `cleanupStaleSelectionsUpdatesToIgnore(): void`
- `get numSelectionsUpdatesToIgnore(): number`

J-Fields (approved):
Very nice PR; thanks a lot!

The weirdness here makes me feel like either VS Code has a bug around this stuff or we're not using its API correctly. Either way, this is an excellent band-aid.

src/mode/modeHandler.ts
J-Fields: I'm going to merge this as-is, but I'd accept another PR with this refactor

alythobani:
@austin-beer  Thanks for the appreciation! And yup I'll be one of those people :) thanks for releasing v1.30.1 @J-Fields !

I've updated and things seem to be working much better now. I can still repro the bug on demand now that I know how to, but it now only happens if I run `Go to Next Change` within 1s of releasing the held-down remapped key, and then the bug auto-resolves itself after 1s has passed since the key release.

alythobani:
For a followup it might be nice to update the `EXPIRY_TIME_MS` default to be smaller e.g. 250-500ms, and/or add a config option for users to customize the expiry time. My local debugging saw that in general `vscode.TextEditorSelectionChangeEvent` events got fired around 20ms after a selections update, so 1s might have been over-generous.

PR #9655, opened 2025-05-30
Refactor: add InternalSelectionsTracker class (followup to #9644)

alythobani:
Followup to #9644

## What this PR does / why we need it

This PR replaces the `ModeHandler.selectionsChanged` object with an instance of a new class `InternalSelectionsTracker`, which encapsulates logic that was previously scattered across multiple files:

  - Controlling the ignoring of intermediate selections during an action
    - Previously: `selectionsChanged.ignoreIntermediateSelections`
    - Now: `InternalSelectionsTracker.shouldIgnoreIntermediateSelections` and helper methods `InternalSelectionsTracker.startIgnoringIntermediateSelections` and `InternalSelectionsTracker.stopIgnoringIntermediateSelections`
  - Determining whether an internal selection will trigger a selection change event, and tracking it if so
    - Previously: `willTriggerChange` logic at the bottom of `updateView`
    - Now: `InternalSelectionsTracker.maybeTrackSelectionsUpdateToIgnore` method
  - Determining whether to treat a selection change event as an internal change we can ignore
    - Previously: logic placed under `if (e.kind !== vscode.TextEditorSelectionChangeKind.Mouse)` branch in `vscode.window.onDidChangeTextEditorSelection` listener registration
    - Now: `InternalSelectionsTracker.maybeIgnoreInternalSelectionChangeEvent` method

The benefit is reduced mental overhead, and hopefully better maintainability, now that the various pieces of logic related to handing internally-triggered selection change events are encapsulated together in a single class. As well as some reduced clutter in functions like `updateView` and the `vscode.window.onDidChangeTextEditorSelection` listener registration.

Not sure it's worth filing an issue for this but happy to if you'd prefer that!

## Special notes for your reviewer

This is a pure refactor with no functionality change. I also added a good amount more explanation around details, pitfalls, and tradeoffs through class/method/property docstrings. Maybe even *too much* explanation, although with behavior this complex and connected, and knowing I had to figure out certain things by diving deep into old PRs and debugging with logs, I did try to err on more details than fewer.

My biggest uncertainty was in replacing and extending @berknam 's earlier ([2020](https://github.com/VSCodeVim/Vim/pull/5015/files#diff-bab551caa561c01d6cc4d9fdec6924895f48e9fc8666ea9c3d2ad4d856654411R298-R301)) comments and logs about what he called "slipped selections", i.e. the fallback case when the event's selections aren't found in our tracked internal updates, and we're not currently ignoring intermediate selections, but we still have some internal selections updates tracked. I did my best to understand and document the logic there, but I'm not currently sure about any concrete examples where an intermediate selection has actually slipped through past the end of an action and would corrupt state if we didn't ignore it. So I've mentioned in `InternalSelectionsTracker.shouldIgnoreAsIntermediateSelection` that we may want to revisit the tradeoff in the future, to help the documentation not come across as more certain than we (or at least I) actually am about the assumption we're operating on there. But if you have a better grasp of things there let me know and I can adjust the docstring as needed.


src/util/selections.ts
```diff
@@ -0,0 +1,40 @@
...
+  return (
+    selectionsA.length === selectionsB.length &&
+    selectionsA.every((sA, i) => areSelectionsEqual(sA, selectionsB[i]))
+  );
+}
+
+/**
+ * Returns whether two selections are equal (i.e. have the same anchor and active positions).
+ */
+function areSelectionsEqual(selectionA: vscode.Selection, selectionB: vscode.Selection): boolean {
```
J-Fields, line 36: `Selection` has an `isEqual` method that should be equivalent - let's just use that

alythobani: Pushed a change right now to clarify the distinction here in the docstring

J-Fields: Ah, nice catch

J-Fields (approved):
Thanks!

PR #9671, opened 2025-06-14
Document enable key-repeating for Windsurf

madflow:

- Document how to enable key-repeating for Windsurf

- None - but maybe some will be prevented in the future :)

- https://windsurf.com/editor


J-Fields (approved):
Thanks!

Issue #9337, opened 2024-10-30
`Vi{` incorrectly selects the line with the closing bracket


KnorpelSenf:
## Description

In Vim, you can do `Vi{` (visual line inside block) to select all lines inside the block but not the lines that contain `{` and `}`.

This plugin incorrectly selects the line with the closing bracket.

## Reproduction

Write the following text.

```txt
Begin
{
  Block A
  Block B

  Block C
}
End
```

Move the cursor to any line inside the block, such as the empty line, and press `Vi{`.

### Expected

Only
```txt
  Block A
  Block B

  Block C
```
are selected.

### Actual

The closing bracket is highlighted, too, so the following text ends up in the selection.
```txt
  Block A
  Block B

  Block C
}
```

![grafik](https://github.com/user-attachments/assets/6c9945af-b109-4be9-bab2-d58d00e7c7e1)

**Environment (please complete the following information):**

- Extension (VsCodeVim) version: 1.28.1
- VSCode version: 1.95.0
- OS: Debian 12 (bookworm)

karshPrime:
Yup error's persistent on my system as well

**Environment**
* Extension (VsCodeVim) version: 1.28.1
* VSCode version: 1.95.2
* OS: MacOS 15.1

karshPrime:
I don't think this is an error. `V` is to select the whole line. If you want to select code inside the braces `{}`, then use `vi{` - that'd give you the functionality you're after.

KnorpelSenf:
I'm not after selecting code inside the braces `{}`.

KnorpelSenf:
> `vi{` and `Vi{` are two different things.

I know. The former works correctly, the latter doesn't.

`vi{` will take your cursor just before the final closing brace, giving you the behavior you are expecting. `Vi{` takes you cursor to the same position, but since `V` means it will select the entire line the cursor is on, it includes the final closing brace as well.

Can you tell me what happens when you try it in vim? Because on my machine, that is not what is happening.

I expect this extension to work like vim. Right now, it doesn't.

Whiskas101:
You are correct, the selection terminates just before the newline character in vim or neovim, but it doesn't seem to do so here in vscodevim.

I was mistaken. I have only ever used `vi{` so I assumed the difference was intended behaviour.

Whiskas101:
Do you have neovim path set up in the json config?

KnorpelSenf:
[...]

PR #9685, opened 2025-06-25
Fixing Vi{ incorrectly grabbing the ending braces

Whiskas101:
Fixes the weird behaviour when using `VisualLine` mode with the `Vi{` or `Vi(` or `Vi[` when the matching braces are on different lines. The issue is well explained in #9337 (relevant images explaining the problem are present there so I have omitted them here).

As a fix for issue #9337 

Only affects `VisualLine` mode when using `Vi{`. Everything else is left untouched, all tests pass locally.

### DIFF
--- a/src/actions/motion.ts
+++ b/src/actions/motion.ts
@@ -1901,6 +1901,7 @@ export abstract class MoveInsideCharacter extends ExpandingSelection {
     firstIteration: boolean,
     lastIteration: boolean,
   ): Promise<IMovement> {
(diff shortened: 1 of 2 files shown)


J-Fields:
Thanks; mind adding a test case?

Whiskas101:
Sure, will do.

Whiskas101:
I've added a test case, let me know if im supposed to add more, or anything else I might have missed.

J-Fields (approved):
Thanks!

PR #9688, opened 2025-06-26
gI goToImplementation

klaskow:

`gI` for goToImplementation


Whiskas101:
is it not just `gd` for going to the implementation of some class or function?

Issue #9712, opened 2025-07-21
IME input causes significant delay or inconsistent behavior in insert mode


s-kai273:
**Describe the bug**
When using an IME (e.g. for Japanese input) in insert mode, there is a noticeable delay between confirming a composition (e.g. pressing Enter to finalize a phrase) and the text actually appearing in the editor.
The confirmed input does appear correctly, but there is a perceptible lag that makes the experience feel slow and unnatural.

**To Reproduce**

1.  Open a file in VS Code with the VSCodeVim extension enabled
2. Switch to insert mode (`i`)
3. Use a Japanese IME (e.g. Google Japanese Input, Mozc, Microsoft IME)
4. Type a word like `konnichiwa` to start composition (`こんにちは` appears as candidate)
5. Press Enter to confirm the composition
6. Observe that the confirmed text appears with a noticeable delay, rather than instantly

**Expected behavior**
The confirmed text (e.g. `こんにちは`) should appear in the editor immediately after pressing Enter to finalize the IME composition, without any perceptible delay. 
Input via IME should feel as responsive as in the default VS Code editor.

**Screenshots**

![Image](https://github.com/user-attachments/assets/98fb1663-353b-4ba0-8296-02aecac7f533)

**Environment (please complete the following information):**

- Extension (VsCodeVim) version: 1.30.1
- VSCode version: 1.102.1
- OS: Linux Mint
- IME: Mozc (Japanese)

s-kai273:
I’ve identified the root cause of the issue.

When users input text using an IME, the `compositionEnd` event is triggered.
At that point, the `handleMultipleKeyEvents` method of `ModeHandler` is called with the composed text split into individual characters.
As a result, `await vscode.commands.executeCommand('default:type', { text })` is executed once for each character, inside `handleMultipleKeyEvents`.
This behavior causes a noticeable delay, since `default:type` is called multiple times unnecessarily.

However, VSCode natively handles IME input by calling `default:type` once with the full composition text, which is much more efficient.
I’ve proposed a fix and submitted a PR to address this:
https://github.com/VSCodeVim/Vim/pull/9713

Please take a look. Thanks in advance!

PR #9713, opened 2025-07-21
Fix significant delay of IME input

s-kai273:
- [x] Changes `handleMultipleKeyEvents` so that it receives **the entire IME composition text as a single keystroke sequence** instead of passing each character one by one.  
  This significantly reduces the lag experienced after confirming IME input.

https://github.com/VSCodeVim/Vim/issues/9712

**Screenshots**

| Before | After |
| ------ | ----- |
| Before | After |


J-Fields:
Thank you! Sorry for the slow review

s-kai273:
Thanks for merging this — I really appreciate your time! @J-Fields

I’d love to continue contributing more actively to VSCodeVim, not only with PRs but also by helping with reviews and issue triage.  
If it would help reduce the review load, would it make sense to grant me collaborator/triage rights?  
I’m happy to start small and follow whatever guidelines you prefer. Thanks again!

PR #9730, opened 2025-08-19
Allow pasting registers from QuickPick menu

MGessinger:

This PR adds extends the `:display` command, so that the QuickPick which opens to show the registers also offers action buttons to paste the register content. This is equivalent to running `:di`, followed by `:put` with the selected register. While the functionality is not new per se, it makes the QuickPick of the registers slightly more useful.

In order to do this, most of the changes are within `registers.ts` but some small changes were made to the `Register` class as well. These are convenience functions, and they do not alter any existing functionality.

I couldn't find an issue which talks about this feature. If a linked issue is required, I can open one afterwards and discussion can happen there, but at least for me this feature would improve my productivity greatly.

I know it's a mouse action rather than a keyboard action. The Vim gods shall have mercy on my soul.

### DIFF
--- a/src/cmd_line/commands/register.ts
+++ b/src/cmd_line/commands/register.ts
@@ -6,7 +6,7 @@ import { Register, RegisterContent } from '../../register/register';
 import { RecordedState } from '../../state/recordedState';
 import { VimState } from '../../state/vimState';
 import { ExCommand } from '../../vimscript/exCommand';
-import { IPutCommandArguments, PutExCommand } from './put';
+import { PutExCommand } from './put';
 
 class RegisterDisplayItem implements vscode.QuickPickItem {
   public readonly label: string;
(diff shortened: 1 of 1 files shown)


src/cmd_line/commands/register.ts
```diff
@@ -2,90 +2,133 @@ import * as vscode from 'vscode';
...
+        if (!action || action !== paste) {
+          return;
         }
+
+        void RegisterCommand.paste(vimState, item);
       });
+  }
+
+  private static async paste(vimState: VimState, item: RegisterDisplayItem) {
+    // TODO: Can I reuse PutCommand here?
```
MGessinger, line 116: This `paste` method feels wrong. I don't want to have to reinvent the wheel just to actually paste register content. I would like to reuse existing functionality, but I couldn't figure out the correct way to do that.

What I tried was to create an instance of `PutCommand` here and pass it the register key as part of its arguments, but somehow that didn't give me the desired result (I had to press ESC after clicking the paste button, for the text to show up). 

Perhaps someone can give me some pointers for how to implement this properly.

Issue #9742, opened 2025-09-03
Add support to configure custom text objects


gayanper:
Today we lack support for treesitter text objects like 
- Inside Function
- Around Function

Rather than this plugin try to build these text objects, another suggestion is open up for registering text object commands via settings into this plugin

This way other plugins can provide support to implement new text-object implementations and configure them

I would like to explore the possibility to 

- Create a new type of TextObject which is configured for each configuration found in the settings. This new type will take the TextObject.keys from a provided key array in the setting object and a vscode command id. The command represented by the id must have a return type of `Omit<IMovement, "removed">`.

- The settings will follow a schema as  follows 

```json

"vim.textobjects": [
     {
        "keys": ["a", "f"]
        "command" : ["x.arround.function"]
     }
]
```

- When we load settings, we will load these text objects into actions map.

If the idea sounds good let me know, I can provide a PR for this suggestion.

gayanper:
I made a PR which I would like to get  review feedback https://github.com/VSCodeVim/Vim/pull/9781

gayanper:
https://github.com/user-attachments/assets/9271f04f-bc29-40db-b9bb-a32693cc4792

The above screencast shows this feature together with a custom private extension I implemented treesitter base textobjects implementation works.

The screencast demos 

- select inside method
- select around method
- delete around method
- delete inside method

The configuration for this looks as follows with this feature

```
"vim.textObjects": [        
        {
            "keys": [
                "i",
                "m"
            ],
            "command": "gap-vscode-vim-keys.text-obj-i-method"
        },
        {
            "keys": [
                "a",
                "m"
            ],
            "command": "gap-vscode-vim-keys.text-obj-a-method"
        }
]
```

PR #9781, opened 2025-10-05
#9742 : Add support for externally provided text objects

gayanper:
## Background

This PR implements support for providing new text objects from other extensions. For example an extension which has support for parting source code using treesitter library can provide more text objects based on treesitter queries. 

## What has implemented

The implementation is based on binding extension commands to vim keybindings using the same approach that has taken for in built text objects.

## What has not implemented

I have not considered validating against conflicting text objects which another extension might have provided. This is because the implementation depends on the user to configure such external text objects in settings rather than they become available automatically. Only exception for this assumption is default setting value an extension might provide, But it is unlikely an external extension might provide a default value using Vim extension configuration keys.

## Testing

I have my own private extension which implements treesitter support for few languages I work with daily. But as automated tests I'm open to help out implementing such tests here in this project if this feature is something the project owners consider including.

### DIFF
--- a/README.md
+++ b/README.md
@@ -802,13 +802,10 @@ The command should return a result object:
 {
   start: Position;
   stop: Position;
-  failed?: boolean;
 }
 ```
 
 or `undefined` if the text object cannot be determined.
-
-If `failed` is true, the text object operation will be cancelled.
 
 ## 🎩 VSCodeVim tricks!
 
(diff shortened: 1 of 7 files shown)


gayanper:
@J-Fields I would love to hear your thoughts on this feature improvement.

gayanper:
@J-Fields  A demo of this feature can be found in https://github.com/VSCodeVim/Vim/issues/9742#issuecomment-3567735121

PR #9840, opened 2025-11-22
fix: fix text decoration change when on visual mode

iabhiyaan:
When selecting a word like `getData` using `viw` -> `visual mode`, the last character becomes hard to see in dark mode due to the current text-decoration styling. This PR fixes that visibility issue so the full selection is clearly readable.

This PR solve the last character visibility issue shows in https://cloud.githubusercontent.com/assets/2787987/24952577/7ab97fee-1faa-11e7-81aa-80bfc0da7bbc.png

Open to any feedback or improvements 😊


J-Fields:
Could you please provide screenshots before & after?

Issue #9851, opened 2025-12-04
The effect of {backtick}< and {backtick}> are opposite to those in Vim, when the cursor is on the left of selected text


paakmau:
The effect of `` `< `` and `` `> `` are opposite to those in Vim, when the cursor is on the left of selected text

**To Reproduce**

1.  Select some text in view mode, make sure the cursor is on the left
2. Trigger `esc`, transition to normal mode
3.  Trigger `` `< `` or `` `> `` command
4. The problem happens:
    - `` `< `` will go right
    - `` `> `` will go left

**Expected behavior**

- `` `< `` should always go left
- `` `> `` should always go right

**Screenshots**

<img width="480" height="222" alt="Image" src="https://github.com/user-attachments/assets/991a10de-acd6-41e1-b6bf-d7eef8a58f9c" />

<img width="463" height="199" alt="Image" src="https://github.com/user-attachments/assets/96900c20-2fe9-423d-b7fb-f615b4244b05" />

**Environment (please complete the following information):**

- Extension (VsCodeVim) version: 1.32.1
- VSCode version: 1.105.1
- OS: win11 26200.7171

paakmau:
Doc's here: <https://vimdoc.sourceforge.net/htmldoc/motion.html#%60%3C>

PR #9853, opened 2025-12-04
fix: make {backtick}<  and {backtick}> move to correct position

paakmau:

Fix the behaviour of `` `< `` and `` `> ``:

- make `` `< `` always go left
- make `` `> `` always go right

According to the doc: <https://vimdoc.sourceforge.net/htmldoc/motion.html#%60%3C>

#9851


J-Fields (approved):
Thanks!

Issue #9887, opened 2026-01-01
Left-clicking the end of a line in Normal mode switches to Visual mode


sho16jp:
**Describe the bug**
When left-clicking the end of a line in Normal mode, the editor sometimes switches to Visual mode unexpectedly.

**To Reproduce**
1. Open any file and confirm you are in Normal mode.
2. Left-click the end of a line.
3. Move to another line and left-click the end of that line as well.
4. Visual mode is activated unexpectedly.

**Expected behavior**
A left-click in Normal mode should only move the cursor.
This action should not cause the editor to enter Visual mode.

**Environment (please complete the following information):**
- Extension (VsCodeVim) version:1.32.4
- VSCode version:1.107.1
- OS:Linux Mint 22.2 Cinnamon
- (Same behavior also confirmed on Windows 11)

**Additional context**
A reproduction video is attached below.

https://github.com/user-attachments/assets/b2f143a2-4b6f-4921-b873-8c43b6cf38ad

sho16jp:
Thanks for looking into this. Much appreciated.

PR #9889, opened 2026-01-03
added revert to normal mode in case of bad sequence of selectionchang…

teslae1:
Fixes mouseclick at eol when in normal mode sometimes triggers visual mode.

I was able to reproduce the problem - the core issue is that vscode
sometimes issues a specific sequence of selection change events on a single mouse click.
When the error occurs the sequence of events for the single mouse click is:
1. Selection change event that selects the last character of the line
- This causes the change to visual mode
2. Selection change event that selects no characters and is at end of line
- This moves the cursor to the end of line having no characters selected

These two events leave the editor in the state which was also shown in
the issue: visual mode and no characters selected.

#9887

### DIFF
--- a/src/mode/modeHandler.ts
+++ b/src/mode/modeHandler.ts
@@ -253,8 +253,7 @@ export class ModeHandler implements vscode.Disposable, IModeHandler {
             // like 'editor.action.smartSelect.grow' are handled.
             if (this.vimState.currentMode === Mode.Visual) {
               const isEol =
-                e.textEditor.document &&
-                e.textEditor.document.lineAt(selection.active.line).text.length ===
+                e.textEditor.document?.lineAt(selection.active.line).text.length ===
                   selection.active.character;
               const lastSelection = this.vimState.lastVisualSelection;
               const previousVisualSelectionIsFromLastCharToEol =


J-Fields:
>    1. Selection change event that selects the last character of the line
>
>    - This causes the change to visual mode

Is there a way to detect this, and prevent it from changing the mode, rather than reverting it after the fact?

teslae1:
> > 1. Selection change event that selects the last character of the line
> > 
> > 
> > * This causes the change to visual mode
> 
> Is there a way to detect this, and prevent it from changing the mode, rather than reverting it after the fact?

I tried researching this (I agree it would be best to never end up in the bad state),
but the fact that the first event in isolation is a completely viable event
that is also dispatched when a user left clicks and drags to select the last char of a line
lead me to the conclusion that we can only detect this after the fact.

teslae1:
@J-Fields 
Hey Jason, 
just wanted to know if there is anything I myself can do to move my pr's along
(including the one belonging to this conversation)
https://github.com/VSCodeVim/Vim/pull/9874
https://github.com/VSCodeVim/Vim/pull/9872
https://github.com/VSCodeVim/Vim/pull/9894
https://github.com/VSCodeVim/Vim/pull/9895
https://github.com/VSCodeVim/Vim/pull/9896

teslae1:
Hey again @J-Fields ,
let me know if I can help in any way.
Just wanted to add to the list of pending pr's I have: 
https://github.com/VSCodeVim/Vim/pull/9911
https://github.com/VSCodeVim/Vim/pull/9910
https://github.com/VSCodeVim/Vim/pull/9901

Issue #9902, opened 2026-01-17
`gf` command fails to open files in WSL environment


NeyasG:
Type: <b>Bug</b>

## Environment
- **VSCode Version**: 1.108.1
- **VSCodeVim Version**: 1.32.4
- **OS**: Windows 11 with WSL 2 (Ubuntu)
- **VSCode Remote**: WSL extension enabled

## Description
The `gf` (go to file) command fails to open files in a WSL environment, producing the error:
```
E447: Can't find file "test.txt" in path
```

This occurs even when the file exists in the same directory as the currently open file.

## Steps to Reproduce
1. Open VSCode connected to WSL (Remote - WSL extension)
2. Create a workspace with two files in the same directory:
   - `main.py`
   - `test.txt`
3. In `main.py`, add a line with just the filename: `test.txt`
4. Place cursor on `test.txt`
5. Press `gf` in normal mode

**Expected behavior**: Opens `test.txt` in a new editor tab

**Actual behavior**: Error message "E447: Can't find file "test.txt" in path"

## What I've Tried
- ✗ Bare filename: `test.txt`
- ✗ Relative path with `./`: `./test.txt`
- ✗ Absolute path: `/home/my_user/repos/VSCodeSettings/test.txt` (opened wrong file at duplicate nested path)
   - File path that was opened was `~/repos/VSCodeSettings/home/my_user/repos/VSCodeSettings/test.txt`
- ✗ Setting `editor.links: true` in VSCode settings (doesn't seem to affect vim commands)

## Working scenario
✓ **The `gf` command works correctly in a non-WSL environment** (native Linux/Windows)
- Created `test.txt` and `main.py` in the same directory
- Used bare filename `test.txt` in `main.py`
- `gf` command successfully opened the file

## Analysis

I'm not too familiar with the internal of VSCodeVim this so forgive me if my analysis is a bit rough.

VSCodeVim source code ([src/actions/commands/file.ts](https://github.com/VSCodeVim/Vim/blob/master/src/actions/commands/file.ts#L86-L109)), the `gf` implementation:

1. First tries to resolve paths relative to the current file's directory
2. Falls back to the workspace root
3. Uses VSCode's URI and file system APIs

The issue appears to be that VSCodeVim's path resolution doesn't properly handle VSCode Remote URI schemes (`vscode-remote://wsl+ubuntu/...`), treating them as local `file://` URIs instead?

## What I think is happening
[...]

PR #9910, opened 2026-01-23
fixed gf when wsl

teslae1:
Added file path resolution that tries to use "path" property of Uri to support filepaths gotten when workspace is remote WSL. Including a small refactor to a seperate method for clarity
https://github.com/VSCodeVim/Vim/issues/9902

### DIFF
--- a/src/actions/commands/file.ts
+++ b/src/actions/commands/file.ts
@@ -79,7 +79,7 @@ class OpenFile extends BaseCommand {
 
     const fileInfo = fullFilePath.match(/(.*?(?=:[0-9]+)|.*):?([0-9]*)$/);
     if (fileInfo) {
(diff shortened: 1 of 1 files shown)


src/actions/commands/file.ts
```diff
@@ -79,35 +79,44 @@ class OpenFile extends BaseCommand {
...
-        file: fileUri.fsPath,
+        file: filePath,
         cmd: isNaN(line) ? undefined : { type: 'line_number', line: line - 1 },
         createFileIfNotExists: false,
       });
       void fileCommand.execute(vimState);
     }
   }
+
+  async GetExistingFilePath(fileInfo: string[], vimState: VimState): Promise<string> {
```
J-Fields:
```suggestion
  async getExistingFilePath(fileInfo: string[], vimState: VimState): Promise<string> {
```

Issue #9645, opened 2025-05-22
Repeat of Visual Block edit is directional


dexusnl:
**Describe the bug**
The repeat of an edit from visual block is not repeated on the same lines if the selection is made from the bottom upwards.

**To Reproduce**

- Open a document with multiple lines
- Press CTRL+Q to make a visual block slection.
- Press up or K a couple of times to select multiple lines.
- Press SHIFT+A to append the selection and type something.
- Go back to normal mode by pressing escape.
- Repeat the last action by pressing "."
- The lines will be added from the cursor upwards instead of on the same lines.

**Expected behavior**
I expect the repeat to be executed on the same lines as the previous action regardless of the selection was done upwards or downwards, this is the behavior Vim has.

**Screenshots**
![Image](https://github.com/user-attachments/assets/542f4451-67ce-4066-8c71-b530d0e7c9c5)

**Environment (please complete the following information):**
- Extension (VsCodeVim) version: 1.29.2
- VSCode Version: 1.100.23258 (system setup)
  Commit: 557d9224936804f4aeb93625d7a45c97113627e4
  Date: 2025-05-15T18:24:23.128Z
  Electron: 34.5.4
  ElectronBuildId: undefined
  Chromium: 132.0.6834.210
  Node.js: 20.19.0
  V8: 13.2.152.41-electron.0
  OS: Windows_NT x64 10.0.19045

prakhargupta-jan:
https://github.com/user-attachments/assets/523f31d1-2e8c-43ac-9801-b9de2f36507b

Hi, I am unable to reproduce the bug

PFA video

PR #9912, opened 2026-01-23
fix: re-calculate starting point for repeated action on block

fuulish:
**What this PR does / why we need it**: repeating a visual block/line selection and modification is directional and does not emulate VIM behavior properly

**Which issue(s) this PR fixes**: This fixes issue #9645 

**Special notes for your reviewer**: Claude Code was used in identifying the issue and generating the fix.


J-Fields:
Mind writing a test case?

Issue #8321, opened 2023-03-19
unicode


aloispichler:
Some languages (as Julia) allow unicode characters as Ø, Đ, Λ, 𝒟 and many others.
If the document contains unicode characters, then VSCodeVim does strange things:

- wrong hjkl movements, 
- wrong easymotion jumps, 
- ‘x’ or ‘r’ will delete wrong, i.e., offset (displaced) characters.

Would be nice to fix that.

k1832:
The `r` and other character-level command issues with surrogate pair characters (like `𝒟`, `𝐄`, and other mathematical symbols from U+1D400+) are addressed in #9929.

Specifically, `a`, `r`, `s`, `~`, and `ga` now correctly handle characters encoded as UTF-16 surrogate pairs. The `hjkl` and `x`/`X` commands were already fixed in earlier changes.

k1832:
Note: easymotion still has surrogate pair issues — it uses raw character arithmetic (e.g. `character + 1`, `character + match.text.length`) without surrogate pair checks in match positioning and marker decoration. This would need a separate fix.

PR #9929, opened 2026-02-10
fix: surrogate pair (emoji) handling for a, r, s, ~, ga commands

### LINKED ISSUE #9931: Surrogate pair characters (emoji, rare CJK, etc.) break a, r, s, ~, ga commands
**Describe the bug**

When the cursor is on a character encoded as a UTF-16 surrogate pair — emojis (`😄`), rare CJK characters (`𩸽`), musical symbols (`𝄞`), etc. — several character-level commands produce incorrect results.

| Command | Broken behavior |
|---------|----------------|
| `a` (Append) | Cursor lands before the character instead of after |
| `r` (Replace) | Only replaces half the character, corrupting the text |
| `s` (Change char) | Only deletes half the character before entering Insert mode |
| `~` (Toggle case) | Corrupts the character into a lone surrogate |
| `ga` (Unicode info) | Shows the half-surrogate value instead of the full codepoint |

**To Reproduce**

1. Open a file containing `😄text`
2. Place cursor on `😄`
3. Press `a` (append), type `!`, press `Esc`
4. See `!😄text` — `!` inserted before the emoji instead of after

**Expected behavior**

`😄!text` — `!` should be inserted after the emoji.

- Extension (VsCodeVim) version: 1.32.4
- VSCode version: 1.109.0
- OS: Ubuntu 24.04

**Additional context**

`position.getRight()` increments by 1 UTF-16 code unit, but these characters are 2 code units (a surrogate pair). Moving by 1 lands between the pair, and VSCode's `validatePosition` clamps it back to the start.

`x`/`X`, `l`/`h`, and `y` already have surrogate boundary correction and work correctly.

k1832:

When the cursor is on a character encoded as a UTF-16 surrogate pair — emojis (`😄`), rare CJK characters (`𩸽`), mathematical symbols (`𝒟`, `𝔸`), musical symbols (`𝄞`), etc. — several character-level commands break.

`position.getRight()` increments by 1 UTF-16 code unit, but these characters are 2 code units. Moving by 1 lands between the pair, and VSCode's `validatePosition` clamps it back to the start — so the cursor effectively goes backward.

| Command | Broken behavior |
|---------|----------------|
| `a` (Append) | Cursor lands before the character instead of after |
| `r` (Replace) | Only replaces half the character, corrupting the text |
| `s` (Change char) | Only deletes half the character before entering Insert mode |
| `~` (Toggle case) | Corrupts the character into a lone surrogate |
| `ga` (Unicode info) | Shows the half-surrogate value instead of the full codepoint |

Previous fixes addressed insert mode (PR #7977 / #6046) and motions (`l`/`h`) and operators (`x`/`X`/`y`), but these 5 commands were missed. This PR completes the surrogate pair handling for character-level commands and adds 12 regression tests to prevent future breakage.

Adds `getSurrogateAwareRight()`/`getSurrogateAwareLeft()` helpers on the Position prototype that skip past surrogate pairs, following the same pattern already used by `MoveRight` (`l`) / `MoveLeft` (`h`). Uses these helpers in the 5 affected commands. Also switches `ga` from `charCodeAt()` to `codePointAt()` to report the full Unicode codepoint.

Fixes #9931
Partially addresses #8321 — this PR fixes `r` and other character-level commands reported there. Remaining issues not addressed: easymotion uses raw character arithmetic for match positioning and marker decorations; `xp` (transpose) on surrogate pairs is also still broken (TODO added in `put.ts`).

- [x] 19/19 surrogate pair tests passing (7 existing + 12 new)
- [x] Full test suite passes (3157 passing, 0 regressions)
- Tests cover emojis, rare CJK (𩸽), and mathematical symbols (𝒟, 𝔸) from U+1D400 block
- Tests require #9928 to build when re-resolving dependencies


src/common/motion/position.ts
```diff
@@ -151,6 +151,17 @@ declare module 'vscode' {
      */
     getUp(count?: number): Position;
 
+    /**
+     * Like getRight(), but skips past surrogate pairs so the cursor never
+     * lands between the high and low surrogate of an emoji.
+     */
+    getSurrogateAwareRight(line: string, count?: number): Position;
```
J-Fields:
```suggestion
    getSurrogateAwareRight(doc: TextDocument, count?: number): Position;
```
This API is easier to use and less error prone

k1832: Done — applied in 963c1b24. Both helpers now take `TextDocument` and look up the line internally.

k1832:
Forgot to re-request a review...

NanoUser3000:
note about `p`: `p` working depends one additional factor
when using `j` to move cursor over 2-wide / surrogate pair symbol, we can go from two columns (assuming they are normal characters). the cursor over emoji may be in 1 of 2 states: "in column more to the left" or "more to the right".

Screen capture showing the difference of going from space versus t (I `x` the a, then do `jjpukljp…`). Despite using `p` with cursor showing over the same character, the results are different)
[Screencast from 2026-05-22 18-45-15.webm](https://github.com/user-attachments/assets/4598f695-ca33-46cc-8169-ccd2722a2f22)

Same thing is happening for a yanked word: the word is pasted before or after the emoji under cursor depending on this internal state.

shift+p / P seems to be working fine

src/actions/commands/replace.ts
```diff
@@ -33,17 +33,7 @@ export class ReplaceCharacter extends BaseCommand {
...
-    // Return if tried to repeat longer than linelength
-    if (endPos.character > vimState.document.lineAt(endPos).text.length) {
-      return;
-    }
-
-    // If last char (not EOL char), add 1 so that replace selection is complete
-    if (endPos.character > vimState.document.lineAt(endPos).text.length) {
-      endPos = new Position(endPos.line, endPos.character + 1);
-    }
+    const endPos = position.getSurrogateAwareRight(vimState.document, timesToRepeat);
```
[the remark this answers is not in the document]
k1832: Good catch — fixed in 00f986bb. The old guard counted UTF-16 code units while `endPos` is codepoint-aware, so `2r` on a single emoji slipped past it and inserted an extra character. The check now runs after computing `endPos` and counts codepoints in the spanned range (`[...line.slice(...)].length`), so it refuses the operation when fewer than `count` characters remain — matching Vim. Added two regression tests (`2rx` spanning two surrogate pairs, and `2rx` on a single emoji as a no-op).

k1832:
Thanks for the detailed report and the screen recording, @NanoUser3000! That `p` behavior is a separate issue from this PR's scope — this PR only covers the character-level commands `a`, `r`, `s`, `~`, and `ga`. The paste positioning you're seeing (text landing before/after the emoji depending on the cursor's internal column state) shares the same surrogate-pair root cause but lives in different code paths (`put.ts` / cursor column tracking). I'd rather not expand this PR's surface, so I'll open a follow-up issue to track `p`/`xp` on surrogate pairs so it doesn't get lost here.

J-Fields (approved):
Thank you!

PR #9959, opened 2026-03-12
Fix dot repeat for tab-only insertions

### LINKED ISSUE #2337: using the repeat command (.) after inserting a tab, does nothing
BUG REPORT

**What happened**:
after inserting a tab, the _._ key cannot be used to repeat the last command (but repeat works for any other character, it seems to be only tab which does not work)

I expected the _._ to insert a tab for me

position to start of a line of text
_i_ _<tab>_ _ESC_ 
position to start of next line
_._ (to repeat last command)
nothing happens.

**Environment**:
- Extension Name: vim
- Extension Version: 0.10.13
- OS Version: Linux x64 4.4.0-38-generic
- VSCode version: 1.19.1

cdthompson:

Fixes `.` repeat for insert sessions that only change the document through VS Code-managed edits, such as a plain `<Tab>` in Insert mode.

Before this change, a tab-only insert like:

`I<Tab><Esc>j.j.`

would not replay with `.` because no `DocumentContentChangeAction` was ever recorded for that insert session. Mixed insertions like `iabc<Tab><Esc>.` worked because normal typing had already created a content-change action.

Fixes #2337

Added regression coverage for:
- `I<Tab><Esc>j.j.`
- `5i<Tab><Esc>`

I also ran the repo's Docker-based test flow from the contributing guide:

- `npx gulp test`

Full Docker suite result on this branch:
- `3150 passing`
- `19 pending`
- `0 failing`

For comparison, clean `master` in the same Docker path produced:
- `3148 passing`
- `19 pending`
- `0 failing`

So this branch adds the 2 new regression tests without introducing new failures.

### DIFF
--- a/src/actions/commands/insert.ts
+++ b/src/actions/commands/insert.ts
@@ -224,11 +224,39 @@ export class ExitInsertMode extends BaseCommand {
       vimState.cursors = vimState.cursors.map((x) => x.withNewStop(x.stop.getRight()));
     }
 
+    const lastActionBeforeEsc =
+      vimState.recordedState.actionsRun[vimState.recordedState.actionsRun.length - 2];
+    const hasPendingContentChanges = vimState.historyTracker.currentContentChanges.length > 0;
+    const hasRecordedContentChange = vimState.recordedState.actionsRun.some(
+      (action) => action instanceof DocumentContentChangeAction,
+    );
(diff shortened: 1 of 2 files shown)


J-Fields (approved):
Thanks!

PR #10012, opened 2026-05-23
Feat: Support localleader (#3382)

### LINKED ISSUE #3382: Support localleader
In addition to the `leader`,  could `localleader` also be supported? This would be benefit to me as I want to map my existing vim keymap to VSCode.

dheerusri324:
This PR adds support for the `<localleader>` keymap, acting analogously to the existing `<leader>` configuration. 

In Vim, `<localleader>` provides a prefix key for buffer-local or filetype-local mappings (often `,` or `\`), separate from the global `<leader>`. Users often rely heavily on these mappings in their existing `.vimrc` configurations, and previously, they could not easily port those configurations into VSCodeVim without manual translations.

This PR implements `<localleader>` across the entire codebase to mirror the precise functionality of `<leader>`, including:
- Exposing the `vim.localleader` extension setting (`\` by default, same as Vim).
- Resolving `<localleader>` securely through key normalization (`Notation.NormalizeKey`) and execution action loops (`CompareKeypressSequence`).
- Integrating with the `.vimrc` Vimscript parser so `<localleader>` is fully supported in user map configurations.
- Accurately processing `<localleader>` replacement during UI rendering (`showcmd` status).
- Enabling TextMate syntax highlighting for `<LocalLeader>` within `.vim` files.

fixes #3382

**Special notes for reviewer**:
* All changes strictly adhere to the pre-existing `<leader>` implementation structure. `NormalizeKey` and `printableKey` have updated signatures to accept an optional third argument for `localLeaderKey`.
* Inside `recordedState.ts` getters, `localleader` display replacements explicitly run *before* `leader` replacements conditionally to prevent collision clobbering when both variables default to the identical `\` key sequence.
* Tested the integration locally in both normal user configuration bounds and custom remapping contexts to verify functionality.


src/vimscript/parserUtils.ts
```diff
@@ -93,9 +93,10 @@ export const fileCmdParser: Parser<FileCmd | undefined> = string('+')
 
 // TODO: re-create parser when leader changes
 const leaderParser = regexp(/<leader>/).map(() => configuration.leader); // lazy evaluation of configuration.leader
+const localleaderParser = regexp(/<localleader>/).map(() => configuration.localleader);
```
J-Fields, line 96: Shouldn't need to use `regexp` for either of these - just `parsimmon.string`

src/configuration/notation.ts
```diff
@@ -40,7 +40,7 @@ export class Notation {
    * and converts the characters to their literals
    * (e.g. <space>, <cr>, <leader>)
    */
-  public static NormalizeKey(key: string, leaderKey: string): string {
+  public static NormalizeKey(key: string, leaderKey: string, localLeaderKey?: string): string {
```
J-Fields, line 43: Why is `localLeaderKey` optional?

src/vimscript/parserUtils.ts
J-Fields: Except, maybe they're meant to be case insensitive? In that case they need the `i` flag

src/configuration/notation.ts
J-Fields: If it shouldn't be, we can do something like this so callers can just pass the config
```suggestion
  public static NormalizeKey(key: string, leaderKeys: {leader: string, localleader: string}): string {
```

src/vimscript/parserUtils.ts
dheerusri324: Good catch! Vim key notations are indeed case-insensitive. I've updated both leader and localleader to use the /i regex flag so it properly handles <Leader>, <LEADER>, etc.

src/configuration/notation.ts
dheerusri324: Great suggestion, this is much cleaner. I've updated the signature to accept the object leaderKeys: { leader: string, localleader: string } and updated the call sites across the codebase to pass the configuration object. I also applied this same pattern to printableKey to keep things consistent.

dheerusri324:
Hi @J-Fields, it should be good to go now!
I pulled down the recent merges from master and fixed the resulting build errors. I added the missing localleader property to the testConfiguration.ts mock so the interface matches, and I also went ahead and patched the glob v11 API update in test/index.ts that was causing the CI to fail on the master merge.
The local build, lint, and tests are all passing!

dheerusri324:
Hi @J-Fields,
First, I want to sincerely apologize for all the back-and-forth on this, and thank you again for your patience and guidance!
I finally tracked down the CI issue. It turns out the test environment actually pulls in glob v7 transitively, rather than v11. 
Because v7 doesn't return a Promise, my async approach was failing the updated, stricter ESLint rules. I've reverted the test runner to use the v7 callback API, which fully resolved the type errors.
To be absolutely sure before pinging you again, I set up a test PR on my own fork and confirmed that the ubuntu-latest build now passes completely green. This is ready for another look whenever you have the time. Thanks again!

